Issue #39, opened 2017-01-13
Remove hadoop 1 support


jerryjch:
Do we still need to keep the hadoop 1 profile/module?

sjudeng:
Just noticed the TinkerPop 3.1.0 release notes for Hadoop Gremlin state "Hadoop1 is no longer supported". Based on this I'd agree the hadoop1 profile/module should just be removed from JanusGraph.

http://tinkerpop.apache.org/docs/3.1.1-incubating/upgrade/#_tinkerpop_3_1_0_2

jerryjch:
Thanks for chiming in. Good info.

jerryjch:
Resolved by Pull request #65

PR #65, opened 2017-01-25
Remove hadoop 1 support

jerryjch:
Remove-hadoop1-support #39


jerryjch:
Cleaned up hadoop 1 related code. 
Kept the hadoop 2 profile and dist structure. It is possible we will add hadoop 3 support in the near future.

PR #75, opened 2017-02-01
Automatically run tests on Travis CI

mbrukman:
Factored out the build and test steps to be able to easily parameterize them
using environment variables in both Linux and OS X.

Starting by testing just the `janusgraph-solr` and `janusgraph-test` modules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,49 @@
 language: java
 sudo: false
 
+jdk:
+  - oraclejdk8
+
 cache:
   directories:
     - ${HOME}/.m2
 
 env:
   global:
-   # This is the encrypted COVERITY_SCAN_TOKEN, created via the
-   # `travis encrypt` command using the project repo's public key.
-   - secure: "v5ixqTeb74y0vRuPcDbe3C28GDDYvqyEXA2dt+9UVU6GG7WpnmpkBf05gI1dIhp51lBhwx9WSlFBtzho+KdCBmNY/CzBRhVHe/lCQYK9Hb6uGPvuwBvC0WjJgJXsVrLFjppeRhcf+OAweVQ3uw2RPMDRvKIVMUcO1BTFjjJl6REJXNUdzGS57MtH2mmRyOEz250EwgqUELZvcOytG7fNrjMJKVK2nSsoxi0BqZIpItTWPWWeQ1wi1FplJ18A2qtD+MPfAGNSB+/a+r0Av+VCT2eGl06ZyZAzP3q/vG5IYjQ3AJsSPqcZUt4ms+2us1+kwuzXIILjzZmcfImu29+y/thndU5E5b2v+nZ4H69CUCc5OmKW2RwozLNmBIUhO0n+35va/J7FiPIqm3pwxCz5vWA3YTHDADxnIYe7+9uY/+dOK/AvP5fyu7u07vuF3liKNBdrX7ylP3kYc7FXGmYl8wCZv31iy1yTtndQ9qKef7bo8lM9Cdh39KyowrygH+Um7pr9gqf2S9jn99nQ3bib32fBWgBkLpJRwhZYHPUupZjZfgu/9woby0DuriuHZKMqZd7QUawYz6wXGlhzu78x5Tohlj1pGBwHYdcJ/Tm3PiEpyH4aYQLffkjGHJAcCW5tO8QbB0qrLYWC8xVMWuFz1TpSBRXOqVYdBfIa2UZDtOU="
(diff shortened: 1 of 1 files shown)


mbrukman:
Sadly, I'm not sure it's quite working well: https://travis-ci.org/JanusGraph/janusgraph/builds/197378987 — I think it's re-running the tests in both modules due to module dependencies.

@sjudeng – do you have any suggestions for how to configure the Travis CI build to just build one module at-a-time or fix the duplicate env vars?

sjudeng:
Try nesting matrix under env as shown below.

https://github.com/ngageoint/titan/blob/eb12c9090b1f2fc4333e90b5718bd41206ce3e45/.travis.yml

sjudeng:
Ah sorry you're already doing that. Let me look again.

sjudeng:
Maybe you have to single quote the values (e.g. `MODULE='solr'` instead of `MODULE=solr`).

hsaputra:
Hmm, kinda misleading since the build seems to pass

sjudeng:
Also try doing an install with dependencies on the first command and then don't build dependencies in the second call to avoid running tests on dependencies.

```bash
mvn install -DskipTests=true -B -pl janusgraph-${MODULE} -am
mvn verify -pl janusgraph-${MODULE}
```

mbrukman:
@sjudeng – thanks, trying now with single quotes and the commands you suggested.

mbrukman:
Nope, still insists on running with each `MODULE` env var as well as the combination: https://travis-ci.org/JanusGraph/janusgraph/builds/197416780; might be a Travis bug.

sjudeng:
It looks like it's being caused by some unexpected interaction between the env matrix and the top level (os) matrix. I reproduced the issue by adding the below to my .travis file. Try removing it from yours as a test.

```
matrix:
  include:
    - os: linux
      dist: trusty
    - os: osx
```

mbrukman:
@sjudeng – you're right, that looks like a Travis issue, removing the top-level `matrix` fixes it: https://travis-ci.org/JanusGraph/janusgraph/builds/197427608; let's see if it passes now.

sjudeng:
Yeah there's the first issue (log limit). I have a fix for that in the below branch. I'll hopefully be able to submit a PR shortly.

https://github.com/ngageoint/janusgraph/tree/test-logging-cleanup

mbrukman:
@sjudeng – Thanks! Disabled the Solr module, enabled all others, let's see what happens. May need to disable the Cassandra and HBase ones as you mentioned in the other issue; at least we'll have a record of what was tried, and we'll know where to go from here.

I'd like to close this issue with enabling at least a few of the modules going forward so that we have some notion of ongoing testing for each PR being submitted.

We still have to figure out how to get our main backends tested (Cassandra and HBase); do you think my proposal for grouping tests into categories and running by category is OK, or do you have any other suggestions?

mbrukman:
Looks like the `es` module has the same log length problem: https://travis-ci.org/JanusGraph/janusgraph/jobs/197472453

On the other hand, HBase module passed: https://travis-ci.org/JanusGraph/janusgraph/jobs/197472455 but too quickly, as if no tests were run.

sjudeng:
Yeah it's excellent progress already. That's great berkeleydb and lucene tests are passing, hopefully the sporadic errors I was seeing with these don't resurface. It looks like the builds for hadoop-parent and hbase-parent didn't descend into the child modules. You'd need to explicitly include hadoop-2, hbase-10, etc. in your build matrix.

sjudeng:
I think you can remove the below rows from your matrix as they don't run any tests.

```
    - MODULE='hadoop-parent'
    - MODULE='hadoop-parent/janusgraph-hadoop-core'
    - MODULE='hadoop-parent/janusgraph-hadoop'
    - MODULE='hbase-parent'
    - MODULE='hbase-parent/janusgraph-hbase-core'
    - MODULE='hbase-parent/janusgraph-hbase'
```

mbrukman:
@sjudeng – Thanks! Removed, re-running tests on Travis now. It's now a much shorter set of modules that will run tests, since most of them fail either with timeout or too many logs. Looking a bit more like your `.travis.yml` config now.

jerryjch:
Trying to follow here.  Do we have a build that includes a full run of hbase tests yet?

mbrukman:
@jerryjch — no, I had to disable HBase tests because they time out. See the [diff of the commit in this PR](https://github.com/JanusGraph/janusgraph/pull/75/files) for all the modules I'm commenting out due to too much space used by log files or timeouts.

jerryjch:
ok. @mbrukman

PR #76, opened 2017-02-03
Incorrect Cassandra keyspace regex for non-alphanumeric-underscore characters

sjudeng:
Regex to detect if Cassandra keyspace name contains non-alphanumeric-underscore characters is incorrect. This causes test failures in `janusgraph-solr` tests with the following error.

```
InvalidRequestException(why:Keyspace name must not be empty, more than 48 characters long, or contain non-alphanumeric-underscore characters (got "org.janusgraph.diskstorage.solr.ThriftSolrTest"))
```

This wasn't a problem previously because `com.thinkaurelius.titan`-based keyspace names had lengths longer than max length of 48, so they'd be updated based on that constraint, but `org.janusgraph`-based names can have lengths less than 48.

### DIFF
--- a/janusgraph-cassandra/src/test/java/org/janusgraph/CassandraStorageSetup.java
+++ b/janusgraph-cassandra/src/test/java/org/janusgraph/CassandraStorageSetup.java
@@ -137,7 +137,7 @@ public class CassandraStorageSetup {
         Preconditions.checkNotNull(raw);
         Preconditions.checkArgument(0 < raw.length());
 
-        if (48 < raw.length() || raw.matches("[^a-zA-Z_]")) {
+        if (48 < raw.length() || raw.matches("^.*[^a-zA-Z0-9_].*$")) {
             return "strhash" + String.valueOf(Math.abs(raw.hashCode()));
         } else {
             return raw;


sjudeng:
This fixes tests that are currently broken in `janusgraph-solr`. Reproduce as shown below. The result is a build failure with 46 errors.

```
git clone https://github.com/JanusGraph/janusgraph.git
cd janusgraph
mvn clean install -DskipTests
mvn verify -pl janusgraph-solr
```

Confirm the fix by running the tests on this branch. The result is a build success.

```
git remote add ngageoint https://github.com/ngageoint/janusgraph.git
git fetch ngageoint
git checkout cassandra-keyspacename-fix
mvn clean install -DskipTests
mvn verify -pl janusgraph-solr
```

janusgraph-cassandra/src/test/java/org/janusgraph/CassandraStorageSetup.java
```diff
@@ -137,7 +137,7 @@ public static String cleanKeyspaceName(String raw) {
         Preconditions.checkNotNull(raw);
         Preconditions.checkArgument(0 < raw.length());
 
-        if (48 < raw.length() || raw.matches("[^a-zA-Z_]")) {
+        if (48 < raw.length() || raw.matches("^.*[^a-zA-Z0-9_].*$")) {
```
mbrukman: Do we actually need the additional complexity of `^.*[^a-zA-Z0-9_].*$` or can we just use `[^a-zA-Z0-9_]` here? Seems like the test is to find whether or not there exists an invalid character, so we don't need to match the full string.

sjudeng: With the matches method the regex does need to match the full string. The original doesn't and that causes the issue.

mbrukman: Makes sense, thanks!

mbrukman:
@sjudeng – please rebase this PR on the `master` branch and `git push -f` so we can see what Travis CI says about it before merging.

sjudeng:
Done but tests in janusgraph-solr which this affects are not currently passing under Travis. Recommend merging #77 which should at least resolve the log limit-based issues.

Issue #67, opened 2017-01-27
Update ElasticSearch dependency


sharpau:
JanusGraph officially is only compatible with ElasticSearch 1.5.x, a release from late 2014 that is no longer supported (EOL was September 2016 https://www.elastic.co/support/eol).

http://docs.janusgraph.org/0.1.0-SNAPSHOT/version-compat.html

I am running Titan in production with ES 1.7.5 with no issues, but that just hit EOL as well. The 2.0 branch is still maintained, and ES recently jumped to 5.x releases. There are security concerns with the older, unmaintained branches. 

Related: Titan requires the dynamic scripting feature of ES, which is off by default in newer versions. I do not know how strict of a requirement this is, but it would be nice to not rely on it if possible.
https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-security.html

sjudeng:
The branch below updates JanusGraph to Elasticsearch 2.4.4. I hope to be able to sumit a PR off this shortly. But note this breaks support for Elasticsearch 1.x. Not sure if that'll be acceptable to the community.

https://github.com/ngageoint/janusgraph/tree/elasticsearch-geoshape-upgrade

I do plan to update to support Elasticsearch 5.x unless someone else is already on it. Question is whether we'd want to maintain backwards compatiblity with 2.x.

MichaelMarkieta:
I am interested in the support for ElasticSearch 5.x. Let me know if there is anything I can help with.

sjudeng:
@MichaelMarkieta There's related discussion on the list below. It would be useful if you're interested in chiming in there on your opinion on whether to maintain support for ES 1.x, or just go forward with 2.x/5.x compatibility, and whether to support node/transport clients or just the HTTP client (ES RestClient or Jest client).

https://groups.google.com/forum/#!topic/janusgraph-dev/F-zfIsuiIoQ

Current status is #79 pending merge adds support for 2.x and there are at least a couple of efforts underway to get to ES 5.x support. I have updates to add support for 5.x with backwards compatibility with 2.x using ES 5.x REST client. @krlohnes has updates to use the Jest high level HTTP ES client. Hopefully some combination of these will get JanusGraph to full 2.x/5.x compatibility in the near term.

sjudeng:
Closing after merge of #79 which adds support for ES 2.x and 5.x.

PR #78, opened 2017-02-03
Update to TinkerPop 3.2.3

sjudeng:
This PR includes commits by @dylanht and myself. All default and TinkerPop tests are passing (`mvn clean install -Dtest.skip.tp=false`). Several tests do require https://github.com/apache/tinkerpop/pull/458 but in my opinion this fix is not critical to a release (e.g. it's likely just relevant to tests).

## Update from 3.1.1-incubating to 3.2.0-incubating

The following is an excerpt of @dylanht's notes from his [original PR (May 2016)](https://github.com/thinkaurelius/titan/pull/1312):

> Missing from this PR are many practical things such as a CassandraOutputFormat, CassandraInput/OutputRDD, GraphFilter support for CassandraInputFormat, and updates to examples / docs that would really allow people to take advantage of the new features and understand how they can work with JanusGraph. Also missing from this PR is compiling g.V(x) and g.V().hasId(x) to graph.vertices(x) in JanusGraphStepStrategy for JanusGraph as I believe @okram has previously recommended. I attempted to modify the HasStepFolder.foldInHasContainers() method which seemed to match the code example in the TinkerPop upgrade docs about the new GraphStep.processHasContainerIds() helper method but couldn't get it to go.
>
> I commented out occurrences of JanusGraphTest.verifyMetrics() because I was getting a null pointer from trying to access the TraversalMetrics stored via ...profile("metrics") and accessed via t.getSideEffects().get("metrics") after changing the test traversals to match the new syntax implemented the profiling overhaul @rjbriody did. I am hopeful I just missed what was really happening under the hood and changes to TP3ProfileWrapper will be sufficient down the line - I failed to grasp how QueryProfiler works to let JanusGraph interface with profile() step, and surprisingly in the console on an inmemory JanusGraph the same traversals seemed to work and I could access the metrics and nested metrics from getSideEffects().get("the-metrics-key-or-~metrics-whichever") as the tests attempt to do. I tried other ways of getting the metrics out of the traversal in the tests and couldn't get it done.

## Update from 3.2.0-incubating to 3.2.3

The following notes are from the [update PR (November 2016)](https://github.com/dylanht/titan/pull/1):

* Explicitly attaching ReferenceElements before persisting in FulgoraGraphComputer to resolve numerous tests failures associated with missing properties
* Opting out of `GraphComputerTest#shouldSupportGraphFilter` since FulgoraGraphComputer doesn't support graph filters but also can't throw the expected exception without breaking a number of tests that otherwise pass
[...]

PR #77, opened 2017-02-03
Ensure Log4j is used for testing

sjudeng:
Log4j should be used for testing but Logback is being used in some modules (janusgraph-es, janusgraph-hadoop, janusgraph-hbase and janusgraph-solr). In these cases the log4j.properties test resources are ignored and console logging is too verbose. In a standard test run (`mvn verify`) console logging exceeds 75M. With the fixes in this PR the logging output for the same run is about 3M.

Note the solution was already present in janusgraph-cassandra (see logging comment in [janusgraph-cassandra/pom.xml](https://github.com/JanusGraph/janusgraph/blob/fc48e9e680273ee62292126f55585b4f1a6d8294/janusgraph-cassandra/pom.xml#L44-L62)) and this PR just propagates that into the offending modules. More minor updates were to clean up some of the test log4j.properties files and ensure that these were used in relevant tests (e.g. `log4j.configuration` file).

PR #79, opened 2017-02-03
Issue 67: Update Elasticsearch and improve geoshape indexing

sjudeng:
## Features

- Updates core Elasticsearch version used for settings/query builders to 2.4.4 (#67)
- Adds support for the Elasticsearch 5 HTTP REST client, which supports both 2.x and 5.x ES servers (#67, #92)
- Uses Painless instead of Groovy scripts for inline updates in ES 5.x, which eliminates the dynamic scripting requirement when using ES 5.x
- Limit use of scripts to deletion and updates of List/Set fields (#154)
- Adds support for indexing and querying by line and polygon geometries.

## Compatibility

<table>
  <tr>
    <td rowspan="2">ES Client</td>
    <td colspan="2" align="center">ES Server Version</td>
  </tr>
  <tr>
    <td>Supported</td>
    <td>Tested</td>
  </tr>
  <tr>
    <td>HTTP REST</td>
    <td>2.x,5.x</td>
    <td>5.3.0,5.2.2,5.2.1,5.1.2,5.0.2,2.4.4,2.3.5,2.2.2*,2.1.2*,2.0.2*</td>
  </tr>
  <tr>
    <td>Transport</td>
    <td>2.x</td>
    <td>2.4.4,2.3.5,2.2.2*,2.1.2*,2.0.2*</td>
  </tr>
</table>

\* Pre-2.3.0 versions do not support indexing non-point geometries or contains operations and require additional dependency exclusions to satisfy dependency convergence checks

## Breaking Changes

- **Elasticsearch 1.x is no longer supported.** Justification is that currently Elasticsearch is on version 5.2 and 1.7.x was end of life as of 2017-01-16 (1.5.x EOL date was 2016-09-23). See https://www.elastic.co/support/eol for more information.
- **Node client is no longer supported**. As a result support for starting embedded Elasticsearch instances is limited. Embedded instances are managed automatically in testing and when using JanusGraph server (`janusgraph.sh`). In all other cases Elasticsearch must already be running to use with JanusGraph. As a mitigation and convenience, an Elasticsearch distribution is packaged with JanusGraph. See the updated JanusGraph Elasticsearch documentation page for more information.
- **TTL is no longer supported**. TTL was deprecated in 2.0 and removed in 5.0 with guidance to migrate to an index-per-timeframe approach (see https://www.elastic.co/guide/en/elasticsearch/reference/2.4/mapping-ttl-field.html)
- The default client/port was changed from Transport/9300 to REST/9200
- Geoshape Kryo serialization now requires a custom serializer. The serializer, `org.janusgraph.hadoop.serialize.JanusGraphKryoRegistrator` is defined in the janusgraph-hadoop-core module to avoid adding an (unshaded) Kryo dependency to janusgraph-core. This serializer must be registered with Spark when indexing Geoshape properties.
```
spark.kryo.registrator=org.janusgraph.hadoop.serialize.JanusGraphKryoRegistrator
```

## Dependency Updates

Dependencies were updated as shown in the following table.

| Dependency | Previous Version | New Version |
| --- | --- | --- |
| Elasticsearch | 1.5.1 | 2.4.4 |
| Lucene | 4.10.4 | 5.5.2 |
| Spatial4j | 0.4.1 | 0.5 |
| Jackson2 | 2.4.4 | 2.6.6 |
| Netty | 3.6.6 | 3.10.5 |
| Joda | 1.6.2 | 2.8.2 |
| commons-cli | 1.2 | 1.3.1 |

Additional Elasticsearch version properties were added to allow changing the version separately for different parts of the system.

| Property | Scope | Default |
| ---- | ---- | ---- |
| elasticsearch.version | Core version used for constructing settings, query objects, etc. | 2.4.4 |
| elasticsearch.rest.version | REST client version | 5.3.0 |
| elasticsearch.dist.version | Default distribution version used for running tests (see below) and in release | 5.3.0 |
| elasticsearch.groovy.inline | Flag for whether to enable Groovy dynamic scripting. Required only for ES 2.x. | false | 

## Testing

Updated project configuration to download and run a local Elasticsearch instance during testing and updated all janusgraph-es test suites to use same. A separate property, `elasticsearch.dist.version`, can be used to run the janusgraph-es test suites against both 2.x and 5.x ES distributions. Tests were updated to use REST client by default but separate Transport client subclasses were also added. This allows transport client testing when using ES 2.x but not 5.x (since transport client connections to ES 5.x are not currently supported).

The following entries in the Travis build matrix show the cases required to test all supported ES major distribution and client variants. Note that `elasticsearch.groovy.inline` must be set to `true` when testing against ES 2.x distributions.

```
# REST client tests against default ES 5.x distribution version
- MODULE='es' ARGS='-DthreadCount=1'
# REST client tests against specific ES 5.x distribution version
- MODULE='es' ARGS='-DthreadCount=1' -Delasticsearch.dist.version=5.2.2
# REST and Transport client tests against default ES 2.x distribution (v2.4.4)
- MODULE='es' ARGS='-DthreadCount=1 -Pelasticsearch2 -Delasticsearch.dist.version=2.3.5
```

All tests are passing.

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 4:10:55.873s
[INFO] Finished at: Sat Mar 11 06:13:06 UTC 2017
[INFO] Final Memory: 141M/1459M
[INFO] ------------------------------------------------------------------------
```


amcp:
How did you test these changes? In other words, do the tests perform better now or at least the same that they were before changing the version?

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geo.java
```diff
@@ -108,6 +108,34 @@ public boolean hasNegation() {
...
+        }
+
+        @Override
+        public String toString() {
+            return "contains";
+        }
+
+        @Override
+        public boolean hasNegation() {
+            return false;
```
amcp: comment on why this is not supported. I think it is because when shapes intersect at their edges, part of the shapes are on the outside.

janusgraph-core/src/main/java/org/janusgraph/graphdb/types/ParameterType.java
```diff
@@ -23,7 +23,7 @@
  */
 public enum ParameterType {
 
-    MAPPING("mapping"), INDEX_POSITION("index-pos"), MAPPED_NAME("mapped-name"), STATUS("status");
+    MAPPING("mapping"), INDEX_POSITION("index-pos"), MAPPED_NAME("mapped-name"), STATUS("status"), INDEX_GEO_MAX_LEVELS("index-geo-max-levels"), INDEX_GEO_DIST_ERROR_PCT("index-geo-dist-error-pct");
```
amcp: one line for each enum and comment on each?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java
```diff
@@ -198,21 +208,21 @@ public Connection connect(Configuration config) throws IOException {
...
-        String disableScriptsVal = settings.get(disableScriptsKey);
-        if (null != disableScriptsVal && !"false".equals(disableScriptsVal)) {
-            log.warn("JanusGraph requires Elasticsearch dynamic scripting.  Setting {} to false.  " +
-                    "Dynamic scripting must be allowed in the Elasticsearch cluster configuration.",
-                    disableScriptsKey);
+        // Force-enable inline scripting.  This is probably only useful in Node mode.
+        String inlineScriptsKey = "script.inline";
+        String inlineScriptsVal = settings.get(inlineScriptsKey);
+        if (null != inlineScriptsVal && !"true".equals(inlineScriptsVal)) {
+            log.warn("JanusGraph requires Elasticsearch inline scripting.  Setting {} to true.  " +
```
amcp: requirement sounds like a user error. loglevel=error and fail fast?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -387,12 +414,22 @@ private static String getDualMappingName(String key) {
         return key + STRING_MAPPING_SUFFIX;
     }
 
+    private static Map<Geo, ShapeRelation> spatialPredicates() {
+        return Collections.unmodifiableMap(Stream.of(
```
amcp: return ImmutableMap.of(...)

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -782,91 +860,141 @@ public FilterBuilder getFilter(Condition<?> condition, KeyInformation.StoreRetri
...
+                    queryBuilder = QueryBuilders.geoDistanceQuery(key).lat(center.getLatitude()).lon(center.getLongitude()).distance(shape.getRadius(), DistanceUnit.KILOMETERS);
                 } else if (shape.getType() == Geoshape.Type.BOX) {
                     Geoshape.Point southwest = shape.getPoint(0);
                     Geoshape.Point northeast = shape.getPoint(1);
-                    return FilterBuilders.geoBoundingBoxFilter(key).bottomRight(southwest.getLatitude(), northeast.getLongitude()).topLeft(northeast.getLatitude(), southwest.getLongitude());
-                } else
+                    queryBuilder = QueryBuilders.geoBoundingBoxQuery(key).bottomRight(southwest.getLatitude(), northeast.getLongitude()).topLeft(northeast.getLatitude(), southwest.getLongitude());
+                } else if (shape.getType() == Geoshape.Type.POLYGON) {
+                    queryBuilder = QueryBuilders.geoPolygonQuery(key);
+                    for (int i=0; i<shape.size(); i++) {
```
amcp: stream.map.forEach

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -782,91 +860,141 @@ public FilterBuilder getFilter(Condition<?> condition, KeyInformation.StoreRetri
...
+                    sb = ShapeBuilder.newCircleBuilder().center(center.getLongitude(), center.getLatitude()).radius(shape.getRadius(), DistanceUnit.KILOMETERS);
+                    break;
+                case BOX:
+                    Geoshape.Point southwest = shape.getPoint(0);
+                    Geoshape.Point northeast = shape.getPoint(1);
+                    sb = ShapeBuilder.newEnvelope().bottomRight(northeast.getLongitude(),southwest.getLatitude()).topLeft(southwest.getLongitude(),northeast.getLatitude());
+                    break;
+                case LINE:
+                    sb = ShapeBuilder.newLineString();
+                    for (int i = 0; i < shape.size(); i++) {
```
amcp: stream

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -782,91 +860,141 @@ public FilterBuilder getFilter(Condition<?> condition, KeyInformation.StoreRetri
...
+                        Geoshape.Point point = shape.getPoint(i);
+                        ((LineStringBuilder) sb).point(point.getLongitude(), point.getLatitude());
+                    }
+                    break;
+                case POLYGON:
+                    sb = ShapeBuilder.newPolygon();
+                    for (int i = 0; i < shape.size(); i++) {
```
amcp: stream

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java
```diff
@@ -262,25 +272,17 @@ static void applySettingsFromJanusGraphConf(ImmutableSettings.Builder settings,
...
-                String subdir = dataDirectory + File.separator + sub;
-                f = new File(subdir);
-                if (!f.exists()) {
-                    log.info("Creating ES {} directory: {}", sub, f);
-                    f.mkdirs();
-                }
-                settingsBuilder.put("path." + sub, subdir);
-                log.debug("Set ES {} directory: {}", sub, f);
-            }
+            settingsBuilder.put("path.home", dataDirectory);
```
amcp: ive seen this property key before. can we collect all ES property keys someplace and use string constants?

janusgraph-es/src/test/resources/es_jvmlocal.yml
```diff
@@ -1,7 +1,4 @@
 node.data: true
 node.client: false
 node.local: true
-path.data: ${project.build.directory}/es/jvmlocal_yml/data
-path.work: ${project.build.directory}/es/jvmlocal_yml/work
-path.logs: ${project.build.directory}/es/jvmlocal_yml/logs
-cluster.name: jvmlocal
\ No newline at end of file
+path.home: ${project.build.directory}/es
```
amcp: newline at the end?

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/serialize/JanusGraphKryoRegistrator.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Geoshape serializer for Kryo.
+     */
+    public static class GeoShapeKryoSerializer extends Serializer<Geoshape> {
+        @Override
+        public void write(Kryo kryo, Output output, Geoshape geoshape) {
+            try {
+                ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
amcp: try with resources? it implements AutoCloseable

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/serialize/JanusGraphKryoRegistrator.java
```diff
@@ -0,0 +1,71 @@
...
+            } catch (IOException e) {
+                throw new RuntimeException("I/O exception writing geoshape");
+            }
+        }
+
+        @Override
+        public Geoshape read(Kryo kryo, Input input, Class<Geoshape> aClass) {
+            int length = input.read();
+            assert length>0;
+            InputStream inputStream = new ByteArrayInputStream(input.readBytes(length));
```
amcp: try with resources?

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/serialize/JanusGraphKryoRegistrator.java
```diff
@@ -0,0 +1,71 @@
...
+                output.write(bytes);
+            } catch (IOException e) {
+                throw new RuntimeException("I/O exception writing geoshape");
+            }
+        }
+
+        @Override
+        public Geoshape read(Kryo kryo, Input input, Class<Geoshape> aClass) {
+            int length = input.read();
+            assert length>0;
```
amcp: why assert and not Precondition.checkArgument?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -60,5 +111,62 @@ public void testGeoShapeSerializationReadWriteAsGryo() throws Exception {
...
+
+    private Function<Geoshape,Geoshape> makeMultiLine = place -> {
+        double x = Math.floor(place.getPoint().getLongitude());
+        double y = Math.floor(place.getPoint().getLatitude());
+        return Geoshape.geoshape(gf.createMultiLineString(new LineString[] {
+                gf.createLineString(new Coordinate[] {new Coordinate(x,y), new Coordinate(x+1,y+1)}),
+                gf.createLineString(new Coordinate[] {new Coordinate(x-1,y-1), new Coordinate(x,y)})}));
+    };
+
+    private Function<Geoshape,Geoshape> makeMultiPolygon = place -> {
```
amcp: static final?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -60,5 +111,62 @@ public void testGeoShapeSerializationReadWriteAsGryo() throws Exception {
...
+        return Geoshape.line(Arrays.asList(new double[][] {{x,y},{x,y+1},{x+1,y+1},{x+1,y}}));
+    };
+
+    private Function<Geoshape,Geoshape> makeMultiPoint = place -> {
+        double x = Math.floor(place.getPoint().getLongitude());
+        double y = Math.floor(place.getPoint().getLatitude());
+        return Geoshape.geoshape(gf.createMultiPoint(new Coordinate[] {new Coordinate(x,y), new Coordinate(x+1,y+1)}));
+    };
+
+    private Function<Geoshape,Geoshape> makeMultiLine = place -> {
```
amcp: static final?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -60,5 +111,62 @@ public void testGeoShapeSerializationReadWriteAsGryo() throws Exception {
...
+        return Geoshape.polygon(Arrays.asList(new double[][] {{x,y},{x,y+1},{x+1,y+1},{x+1,y},{x,y},{x,y}}));
+    };
+
+    private Function<Geoshape,Geoshape> makeLine = place -> {
+        double x = Math.floor(place.getPoint().getLongitude());
+        double y = Math.floor(place.getPoint().getLatitude());
+        return Geoshape.line(Arrays.asList(new double[][] {{x,y},{x,y+1},{x+1,y+1},{x+1,y}}));
+    };
+
+    private Function<Geoshape,Geoshape> makeMultiPoint = place -> {
```
amcp: static final?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -60,5 +111,62 @@ public void testGeoShapeSerializationReadWriteAsGryo() throws Exception {
...
+        });
     }
+
+    private Function<Geoshape,Geoshape> makePoly = place -> {
+        double x = Math.floor(place.getPoint().getLongitude());
+        double y = Math.floor(place.getPoint().getLatitude());
+        return Geoshape.polygon(Arrays.asList(new double[][] {{x,y},{x,y+1},{x+1,y+1},{x+1,y},{x,y},{x,y}}));
+    };
+
+    private Function<Geoshape,Geoshape> makeLine = place -> {
```
amcp: static final?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -60,5 +111,62 @@ public void testGeoShapeSerializationReadWriteAsGryo() throws Exception {
...
+        graph.traversal().E().has("place").toList().stream().forEach(e-> {
+            assertTrue(e.property("shape").isPresent());
+            Geoshape place = (Geoshape) e.property("place").value();
+            Geoshape expected = makeGeoshape.apply(place);
+            Geoshape actual = (Geoshape) e.property("shape").value();
+            assertEquals(expected, actual);
+        });
     }
+
+    private Function<Geoshape,Geoshape> makePoly = place -> {
```
amcp: static final? it seems like these functions dont depend on member state, so lets make it clear

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geo.java
sjudeng: A JanusGraphPredicate can only have a negation if it can be defined directly as another JanusGraphPredicate. This isn't possible for Geo.Contains (e.g. negation is not Geo.WITHIN for example).

Note this is a separate question from whether the negation of the contains Contains spatial predicate exists mathematically, though I'm not sure it does because of subtleties in the definition.

https://en.wikipedia.org/wiki/DE-9IM#Spatial_predicates
http://lin-ear-th-inking.blogspot.com/2007/06/subtleties-of-ogc-covers-spatial.html

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java
sjudeng: I agree this would be an improvement but since this is consistent with how Elasticsearch properties are currently being handled in the code I'd prefer not to pull out just path.home here. I think the effort to improve Elasticsearch configuration property handling should be separate from this PR. Can it be backlogged separately?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sjudeng: Done but note while Geoshape has a get(int i) method it is not a Collection so doesn't support stream directly if that's what you were after.

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/serialize/JanusGraphKryoRegistrator.java
sjudeng: Done but close is a no-op on ByteArrayInput/OutputStream

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
sjudeng: Good idea

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/serialize/JanusGraphKryoRegistrator.java
sjudeng: Good catch

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sjudeng: Much better thanks

sjudeng:
@amcp Thanks for the review. All updates are there with a couple exceptions as mentioned above.

All tests pass in this branch with the additional application of the fix in #76. Attached is relevant output from a test run that completed two days ago but note this was on a combined master branch which also included other open PRs (#76, #77, #78 and #81).

[test.out.excerpt.txt](https://github.com/JanusGraph/janusgraph/files/752461/test.out.excerpt.txt)
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 04:49 h
[INFO] Finished at: 2017-02-02T00:46:25-06:00
[INFO] Final Memory: 146M/1538M
[INFO] ------------------------------------------------------------------------
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/types/ParameterType.java
sjudeng: Yeah good call. I documented the ones I added but wasn't sure about the others.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java
amcp: please create an issue

amcp:
Can we establish a 72 hour comment period on this PR? Would like to get these changes merged in.

janusgraph-core/pom.xml
```diff
@@ -64,6 +64,11 @@
             <artifactId>spatial4j</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.vividsolutions</groupId>
+            <artifactId>jts</artifactId>
```
pluradj: This is a [LGPL](http://search.maven.org/#artifactdetails%7Ccom.vividsolutions%7Cjts%7C1.13%7Cjar) library. Are there alternatives? It looks like the code is only used in test cases. Maybe we can make it an optional, off-by-default, inclusion via a Maven build flag or profile. TinkerPop had to do something similar for Neo4j.

amcp: @pluradj Good catch.
@mbrukman We should have some license automation checks too.
@sjudeng Please let us know if there are any alternatives.

amcp: @mbrukman I created #128 to track automating dependency license check as part of CI

sjudeng: JTS is currently a non-optional dependency in janusgraph-solr and is distributed with JanusGraph (and previously Titan) releases.

[janusgraph-solr/pom.xml](https://github.com/JanusGraph/janusgraph/blob/236dd930a7af35061e393ea8bb1ee6eb65f924b2/janusgraph-solr/pom.xml#L90-L94)

What could be done is to mark it as optional in the poms so it's available for tests and users with requirements for relevant features (Solr, Geoshapes, etc.) can manually add it to their lib directory.

Given it's already a dependency can this be a separate issue or is it a blocker here?

sjudeng: Looks like the longer term solution might be a new Eclipse Foundation project aiming to replace JTS.

https://github.com/locationtech/jts
https://www.locationtech.org/projects/technology.jts
https://www.locationtech.org/proposals/jts-topology-suite

It's being developed by LocationTech, who also develop Spatial4j, through-which the JTS dependency comes in to Solr/JanusGraph. I'd guess the roadmap is for a stable new-JTS release and then a Spatial4j update/release migrating to same, at which point JanusGraph could update as well (likely involving a Solr version upgrade).

pluradj: ha, you're right @sjudeng - had no idea that dependency was around for so long.
Good find on that new project. any indications that they are close to a release? It seems like they've been at it for a while.

sjudeng: I don't see anything on a release timeline. There's recent mailing list activity but there was a gap for a few months before that. Also I don't think a full migration in JanusGraph would be possible until Spatial4j and Solr were updated to use JTS 2. The good news is the core Spatial4j developer, who I think also is a core contributor on the Solr spatial connectors, has been active on the JTS 2 mailing list so he is at least aware of the effort.

amcp:
@pluradj I created #133 to track the issue of marking JTS optional in janusgraph-solr. I think it is a high priority item, and we should try to fix it before our initial release. Can @sjudeng move on with his PR?
@sjudeng please rebase.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -16,22 +16,38 @@
...
 import org.janusgraph.diskstorage.ScanBuffer;
 import org.janusgraph.diskstorage.WriteBuffer;
 import org.janusgraph.graphdb.database.idhandling.VariableLong;
-import org.janusgraph.graphdb.relations.RelationIdentifier;
-import org.apache.commons.lang.builder.HashCodeBuilder;
+
+import com.spatial4j.core.shape.jts.JtsGeometry;
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.GeometryFactory;
```
pluradj: This usage is directly in `janusgraph-core` not in a test flow. @amcp I think we should hold off on including any of the LGPL code until we figure out an optional and clean way to do so. @sjudeng is it possible to separate out the ES 2.4.4 update from the Geoshape indexing improvements?

pluradj:
@ptgoetz @hsaputra any other ideas on how to mitigate LGPL usage?

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
amcp: Of the little I saw in janusgraph-es, it will require some refactoring, and probably we need to add new StoreFeatures like supportsGeo. Interesting.

amcp: @pluradj Put another way, I think removing JTS dependency in this PR is way out of scope for this PR. Titan/JanusGraph depended on JTS up until today, so I think it is OK to upgrade these versions and then figure out a way off of current JTS. Sound OK?

sjudeng: There are a few specialty functions in there that use JTS. I'll need to verify/test but if it's made an optional dependency it would still compile and be used in tests but wouldn't be included in distributions (at least not through janusgraph-core). Is this acceptable? In this way users of relevant (e.g. polygon) features could manually add JTS to their JanusGraph classpath but other users would be unaffected.

pluradj: @amcp OK, I think we're in agreement. In the current code, you can use Solr without JTS as long as you don't define a Solr schema with it, although the example [schema.xml](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-dist/src/assembly/static/conf/solr/schema.xml#L534-L540) includes it. Updating the docs and making that dependency optional will be handled with https://github.com/JanusGraph/janusgraph/issues/133. 

@sjudeng JTS wouldn't be included in distributions -- that's what we're shooting for. Users would need to add the jar themselves to take advantage of those features, and the docs alerts them of the licensing restrictions they're agreeing to, similar to what is in the [Solr wiki](https://cwiki.apache.org/confluence/display/solr/Spatial+Search).

sjudeng:
Updated to make JTS optional in janusgraph-core as discussed. When JTS is not on the path geoshape functions involving point, circle and rectangle shapes will still work as expected. User must add JTS to their classpath when use of line and polygon shapes is required, as indicated in documentation.

All tests are passing.

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 3:26:18.819s
[INFO] Finished at: Wed Feb 22 09:29:09 UTC 2017
[INFO] Final Memory: 139M/1481M
[INFO] ------------------------------------------------------------------------
```

Note during testing JTS was marked optional in janusgraph-solr to ensure proper behavior building/running distribution as well as unit tests, but this PR does not include that update which will be handled under #133.

pluradj:
@sjudeng after building the distribution zip, Elasticsearch isn't starting with `bin/janusgraph.sh start`. I think some of the files under `janusgraph-dist/src/assembly/static` might need to be updated.

sjudeng:
@pluradj Just pushed updates to resolve this issue.

Are there automated integration tests that catch this or do you just test manually? When building a release [this execution](https://github.com/JanusGraph/janusgraph/blob/af3fa20ecd362a694f685f92c218b092a273b89d/janusgraph-dist/janusgraph-dist-hadoop-2/pom.xml#L136-L142) runs, which seems to be at least a pseudo integration test on the distribution. Maybe something similar could be added to run `janusgraph.sh start/stop` and/or other relevant scripts.

pluradj:
@sjudeng I tested it manually, but automation sounds like a good idea.

The fix worked fine on Ubuntu, and I was able to connect through the Gremlin Server remotely with the Gremlin Console.

I'm getting this error in the `log/elasticsearch.log` when run the `bin/janusgraph.sh start` on Mac:

```
[2017-02-23 11:09:19,991][ERROR][bootstrap                ] Exception
java.lang.IllegalStateException: jar hell!
class: groovy.beans.Bindable
jar1: /Users/pluradj/src/github/JanusGraph/janusgraph/pull/janusgraph-0.1.0-SNAPSHOT-hadoop2/bin/../lib/groovy-2.4.7-indy.jar
jar2: /Users/pluradj/src/github/JanusGraph/janusgraph/pull/janusgraph-0.1.0-SNAPSHOT-hadoop2/bin/../lib/groovy-2.4.7.jar
	at org.elasticsearch.bootstrap.JarHell.checkClass(JarHell.java:280)
	at org.elasticsearch.bootstrap.JarHell.checkJarHell(JarHell.java:186)
	at org.elasticsearch.bootstrap.JarHell.checkJarHell(JarHell.java:87)
	at org.elasticsearch.bootstrap.Bootstrap.setup(Bootstrap.java:180)
	at org.elasticsearch.bootstrap.Bootstrap.init(Bootstrap.java:286)
	at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:45)
```

It's odd that the stack trace is getting chopped. I tried manually cleaning up some of the jar conflicts, `groovy-2.4.7-indy.jar` then `high-scale-lib-1.0.6.jar`, but then ended up with a clash on `elasticsearch-2.4.4.jar` with `janusgraph-es-0.1.0-SNAPSHOT.jar`.

sjudeng:
@pluradj The workaround for the JarHell issue was to override it in JanusGraph. I think the original JarHell class is getting picked up first under OSX for whatever reason.

To test would you change bin/elasticsearch.in.sh and replace the `ES_CLASSPATH` definition on L13 with the following?

```bash
ES_CLASSPATH="`dirname $0`/../janusgraph-es-0.1.0-SNAPSHOT.jar:`dirname $0`/../lib/*"
```

If this works I'll look at getting that updated (obviously without the hard coded version).

pluradj:
worked great @sjudeng

sjudeng:
@pluradj Great, thanks. I'll try to push an update for this later today.

sjudeng:
@pluradj That update is pushed. Also created #138 for looking at automating validation of at least the start/stop operations on `janusgraph.sh` as part of the release build.

amcp:
@hsaputra @jerryjch @twilmes @dylanht Shall we call a moratorium on this PR? seems like the goal of this PR, ES 2.4.4 support and improved geoshape support is met, and other issues like licensing #132 and build quality #138 are being tracked elsewhere.

aphelps:
I know I'm a bit late to this PR, but I would really like to strongly discourage completely removing support for ES 1.X.  While 1.7 is technically end-of-lifed, any of us that are currently running production Titan systems are using 1.X (we use 1.7 with no problems despite Titan 1.0 only officially supporting 1.5).

If the goal of JanusGraph is to be a step forward for Titan users as well as pulling in new users, having a relatively seamless transition is going to be critical.  While I can update the libraries our code uses easily enough, I cannot just decide to update our ElasticSearch infrastructure.  Things like this should be taken one step at a time.

sjudeng:
@aphelps This is definitely a tough one. Currently JanusGraph doesn't do compatibility shims for the indexing backends like it does for the storage backends. I think this is what would be required in this case and I didn't purse it at the time. It's additionally complicated by the tight connection between Elasticsearch and Lucene as well as the support for embedded Elasticsearch instances.

Since submitting this PR I've gone more in the other direction, in particular looking at updating to support Elasticsearch 5.x while understanding that unlike here maintaining backwards compatibility with LTR Elasticsearch 2.4 is a must. Luckily in this case I think the REST client will allow us to build with Elasticsearch 5.x but support both 2.x and 5.x servers. It's conceivable that this might also work to communicate with 1.x servers, but I wouldn't be very optimistic on this. In particular I think I read somewhere that whereas there's good REST compatibility between 2.x and 5.x the same is not true with 1.x.

krlohnes:
Not sure this is the right place to start this discussion, but I have some code in the works that would add Http support (that would address #92 ) which could support ES 1.x and 2.x, with 5.x  support coming in the near future. It's a Jest (Apache Licensed high level http client for ES) based implementation. Users could change the Jest jar version and use either 1.x or 2.x and use the same code. The Jest folks have open PRs for adding 5.x support, so it would be a pretty easy story adding support 5.x once those are merged. I think that would help address @aphelps and @sjudeng concerns. With the lucene version coupling, maybe that could get configured along with the Jest jar version?

sjudeng:
@krlohnes This sounds great to me. Where I'm at with my own follow-up work is I've replaced node and transport clients with just the ES 5.x RestClient and reworked tests to unpack and run org.elasticsearch.distribution artifacts as part of janusgraph-es test suites to support testing against ES 2.x and 5.x in the Travis build matrix. But while this works when keeping search builders tied to ES 2.x, I was having trouble when trying to upgrade to ES 5.x builders (so I could update to Lucene 6), because of some new features added in ES 5.x search objects. But it sounds like your updates to use Jest might solve this very nicely.

I'm hoping JanusGraph is willing to drop support for 1.x (at least temporarily given you might be able to resurrect it with your work). If so this PR can get merged and then I'll push an updated branch (based on this) with the above refactoring, which I think might be a better starting place for your updates depending on how far along you are, especially regarding supporting swapping ES distribution for tests/embedded instances. In addition (and admittedly biased) this would help me avoid what I'm suspecting would be a painful process to reimplement some of the other work here (e.g. geoshape and testing updates) on top of your updates. But are you already pretty far along on your implementation off of master?

Out of curiosity are you dropping node and transport clients or supporting all where applicable? In my updates I dropped node and transport and was just going to go with HTTP/REST ... though I was curious what the push back would be on this from the community.

krlohnes:
@sjudeng 

>But it sounds like your updates to use Jest might solve this very nicely

That's what I'm hoping 😄 

>...which I think might be a better starting place for your updates depending on how far along you are

It's not a _whole_ lot of change, just a couple hours of work likely (I already did it once in Titan)  so I can go either way on this. Merging your changes would likely make supporting the node and transport clients a bit easier on my end. Which leads me to...

> are you dropping node and transport clients or supporting all where applicable?

I was going to support them, thinking that marking them as deprecated in JanusGraph docs might make sense. So far, I split the http client and the node/transport client out into two separate classes, and I'm working on conditionally supporting them based on the configuration in JanusGraph, that way it's a simple delete when support gets dropped for the other clients. At least that's where my heads at, I'd love to get more opinions on that.

pluradj:
@krlohnes can you start a top level [DISCUSS] thread on https://groups.google.com/forum/#!forum/janusgraph-dev? It will encourage more open discussion. Key decision points we need to flush out for ES:
1. Versions to support (1.x, 2.x, 5.x)
2. HTTP vs Transport/Node

krlohnes:
Sure thing @pluradj Created https://groups.google.com/forum/#!topic/janusgraph-dev/F-zfIsuiIoQ

aphelps:
@sjudeng I completely agree that moving to a more modern ES is critical, however from a project acceptance standpoint I feel it would be a really good idea to release an initial milestone release of JanusGraph that can be a mostly drop-in replacement for Titan production systems.  I know in the case of our system needing to first upgrade ES would likely result in many more months before a switch would be considered, and its not even immediately clear to me how one would switch an existing Titan DB from 1.X to 2.X without fully rebuilding the system (much less sufficiently seamlessly to avoid downtime).

sjudeng:
@aphelps We're discussing this over on the mailing list where the current proposal is to merge this PR, remove node/transport client support in favor of HTTP, and update to use Jest. This would provide full ES 2.x/5.x support including test suites by default. But thanks to @krlohnes users would still be able to use with ES 1.x by making a minor classpath update to swap out the default Jest jar with one that's compatible with 1.x. The most controversial thing here I think would be dropping support for node/transport clients. But HTTP is the way forward for ES anyway and node client is already dropped in ES 5. Here I'm hoping JanusGraph can play the "we're not Titan" card and allow us to reshape the code base a bit more abruptly than would normally be done leading up to the first initial release.

sjudeng:
@amcp, @pluradj, @jerryjch While we discuss the way forward on this on the dev list I did want to update this PR since it's been awhile and I've since added support for ES 5.x via REST. But note this last commit would need separate review before this PR could get merged.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -0,0 +1,221 @@
...
+            outputStream.write(mapper.writeValueAsBytes(actionData));
+            outputStream.write("\n".getBytes());
+            if (request.getSource() != null) {
+                outputStream.write(mapper.writeValueAsBytes(request.getSource()));
+                outputStream.write("\n".getBytes());
+            }
+        }
+
+        // TODO: Refresh needed for tests but would prefer this in test classes/configuration instead
+        final Response response = performRequest("POST", "/_bulk?refresh=wait_for", outputStream.toByteArray());
```
sjudeng: @krlohnes, @MichaelMarkieta Any thoughts on this refresh? It's necessary for tests to pass but could this be a performance hit in production?

docs/elasticsearch.txt
```diff
@@ -18,101 +18,86 @@ JanusGraph supports http://elasticsearch.org[Elasticsearch] as an index backend.
...
+$ elasticsearch/bin/elasticsearch
+----
 
-. The new <<es-cfg-interface-track, `interface` track>>
-. The <<es-cfg-legacy-track, legacy track>>
+Elasticsearch Configuration Overview
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-These tracks are mutually exclusive.  A configuration uses one track or the other, but not both simultaneously.  The `interface` track is recommended over the legacy track.  The `interface` track, introduced in 0.5.1, offers a superset of the legacy track's functionality.  The legacy track will be maintained through at least the end of the 0.5.x patch series.
+JanusGraph supports HTTP and Transport client connections to a running Elasticsearch cluster.
```
sjudeng: Yep, thanks for the review. I'll get that updated and rebase later today.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
krlohnes: Would using the Elasticsearch integration test stuff help with this?

krlohnes: https://www.elastic.co/guide/en/elasticsearch/reference/5.x/integration-tests.html

sjudeng: It might be doable with transport client but I haven't had success getting the REST client through ESIntegTestCase to work. This would also complicate current support for swapping out the distribution used for testing.

Maybe this could at least be moved to a property. I was hoping there was something I could put in the test instance elasticsearch.yml that would resolve it cleaner.

krlohnes: @sjudeng  Jest has their tests running with it [example](https://github.com/searchbox-io/Jest/blob/master/jest/src/test/java/io/searchbox/client/JestClientFactoryIntegrationTest.java). Might help?

sjudeng: Thanks @krlohnes that's helpful. I'd prefer not to move to that just for this specific issue though I could see needing to revisit if there are broader issues with the current full ES test instance approach (e.g. if the underlying Maven artifacts become unreliable or there are port conflict issues running tests in broader dev/CI environments).

sjudeng:
Added one more commit with minor updates to run scripts as [Painless](https://www.elastic.co/guide/en/elasticsearch/reference/master/modules-scripting-painless.html) instead of Groovy under ES 5.x. This removes the JanusGraph requirement for enabling Groovy dynamic scripting on ES 5.x clusters (relates to https://github.com/JanusGraph/janusgraph/issues/67 and https://github.com/JanusGraph/janusgraph/issues/154).

tippytto:
Hi,
I think there is may be a problem with the compatibaility with ES5, when Janus will try to register an index with a String.
In ES5, there is no more String type (https://www.elastic.co/guide/en/elasticsearch/reference/5.2/string.html). String type was replaced by Text and Keyword.
Text is the same as analyzed String but keyword is more powerful because you can apply a normalizer to it.
I think with this code, when Janus will try to push a mapping with String type, ES will refuse to register it.

According to the Doc of ES (https://www.elastic.co/guide/en/elasticsearch/reference/5.2/keyword.html and https://www.elastic.co/guide/en/elasticsearch/reference/5.2 /text.html).
"Indexes imported from 2.x do not support text or keyword.Instead they will attempt to downgrade text into string." This allows you to merge modern mappings with legacy mappings. "
I did not try, but if you make the change from String to Text and Keyword, it should be working for ES5 and ES2

pluradj:
@sjudeng likely has more input on this, but in ES 2.x, there is no text or keyword datatype. There is only String datatype https://www.elastic.co/guide/en/elasticsearch/reference/2.4/string.html

ES 5.x will upgrade the String datatype into text or keyword, but log a warning.

tippytto:
Ok, it's maybe just an improvement to do when this big PR will be merged.
I think it's better to use text and keyword and let ES2 downgrade it into String than the inverse.

sjudeng:
@tippytto I took a look at this but ES 2.x does not support mappings with `keyword` or `text` types. 

```
Caused by: org.elasticsearch.client.ResponseException: PUT http://127.0.0.1:9200/janusgraph/_mapping/vertex: HTTP/1.1 400 Bad Request
{"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [keyword] declared on field [name]"}],"type":"mapper_parsing_exception","reason":"No handler for type [keyword] declared on field [name]"},"status":400}
...
Caused by: org.elasticsearch.client.ResponseException: PUT http://127.0.0.1:9200/janusgraph/_mapping/vertex: HTTP/1.1 400 Bad Request
{"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [text] declared on field [name]"}],"type":"mapper_parsing_exception","reason":"No handler for type [text] declared on field [name]"},"status":400}
```

But as @pluradj said ES 5.x will accept mappings with `string` types and properly upgrade to text/keyword. There is a deprecation warning but I don't think support will be dropped until ES 6.x.

sjudeng:
@amcp, @pluradj, @jerryjch - Because commit [b34be73](https://github.com/JanusGraph/janusgraph/pull/79/commits/b34be738ac76a14d25d85302696aa7924badf718) is pretty large and was added (together with the smaller [7a2c590](https://github.com/JanusGraph/janusgraph/pull/79/commits/7a2c590627f961bfbfc9dda1b70745d61b4d7ac2) and [2660ed5](https://github.com/JanusGraph/janusgraph/pull/79/commits/2660ed5cbe505961afd8c125422d53f0579c33ef)) after your initial approvals do you want to [dismiss your reviews](https://help.github.com/articles/dismissing-a-pull-request-review/) until you (or @hsaputra, @twilmes, @dylanht) have time to review the new commits?

.travis.yml
```diff
@@ -13,11 +13,15 @@ env:
...
     # `travis encrypt` command using the project repo's public key.
     - secure: "v5ixqTeb74y0vRuPcDbe3C28GDDYvqyEXA2dt+9UVU6GG7WpnmpkBf05gI1dIhp51lBhwx9WSlFBtzho+KdCBmNY/CzBRhVHe/lCQYK9Hb6uGPvuwBvC0WjJgJXsVrLFjppeRhcf+OAweVQ3uw2RPMDRvKIVMUcO1BTFjjJl6REJXNUdzGS57MtH2mmRyOEz250EwgqUELZvcOytG7fNrjMJKVK2nSsoxi0BqZIpItTWPWWeQ1wi1FplJ18A2qtD+MPfAGNSB+/a+r0Av+VCT2eGl06ZyZAzP3q/vG5IYjQ3AJsSPqcZUt4ms+2us1+kwuzXIILjzZmcfImu29+y/thndU5E5b2v+nZ4H69CUCc5OmKW2RwozLNmBIUhO0n+35va/J7FiPIqm3pwxCz5vWA3YTHDADxnIYe7+9uY/+dOK/AvP5fyu7u07vuF3liKNBdrX7ylP3kYc7FXGmYl8wCZv31iy1yTtndQ9qKef7bo8lM9Cdh39KyowrygH+Um7pr9gqf2S9jn99nQ3bib32fBWgBkLpJRwhZYHPUupZjZfgu/9woby0DuriuHZKMqZd7QUawYz6wXGlhzu78x5Tohlj1pGBwHYdcJ/Tm3PiEpyH4aYQLffkjGHJAcCW5tO8QbB0qrLYWC8xVMWuFz1TpSBRXOqVYdBfIa2UZDtOU="
+    # Default Elasticsearch heap size can be too large for Travis
+    - ES_JAVA_OPTS="-Xms256m -Xmx512m"
 
   matrix:
     - MODULE='berkeleyje'
     - MODULE='cassandra'
     - MODULE='es' ARGS='-DthreadCount=1'
+    - MODULE='es' ARGS='-DthreadCount=1 -Delasticsearch.dist.version=2.4.4'
```
amcp: use a variable for es version here and below?

docs/versions.txt
```diff
@@ -13,5 +13,7 @@ JanusGraph.
 [options="header"]
 |==========================
 | JanusGraph | Cassandra | HBase | Elasticsearch | Solr | TinkerPop
-| 0.1.0 | 1.2.z, 2.0.z, 2.1.z | 0.94.z, 0.96.z, 0.98.z, 1.0.z | 1.5.z |  5.2.z | 3.0.z
+| 0.1.0 | 1.2.z, 2.0.z, 2.1.z | 0.98.z, 1.0.z, 1.2.z | 2.z, 5.z* |  5.2.z | 3.2.z
```
amcp: if this PR is about ES why are we changing HBase versions? if we forgot to do it then its ok

docs/elasticsearch.txt
```diff
@@ -17,7 +17,11 @@ JanusGraph supports https://www.elastic.co/[Elasticsearch] as an index backend.
 Please see <<version-compat>> for details on what versions of ES will work with JanusGraph.
 
 [IMPORTANT]
-JanusGraph currently requires https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-security.html#enable-dynamic-scripting[Elasticsearch's dynamic scripting feature].  The `script.engine.groovy.inline.update` setting must be set to `true` on the Elasticsearch cluster.  This configuration requirement may be removed in future JanusGraph versions.
+===============================
+Beginning with Elasticsearch 5.0 JanusGraph uses sandboxed https://www.elastic.co/guide/en/elasticsearch/reference/master/modules-scripting-painless.html[Painless scripts] for inline updates, which are enabled by default in Elasticsearch 5.x.
+
+Using JanusGraph with Elasticsearch 2.x requires enabling Groovy inline scripting by setting `script.engine.groovy.inline.update` to `true` on the Elasticsearch cluster (see https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-security.html#enable-dynamic-scripting[dynamic scripting documentation] for more information).
```
amcp: seems like we need 2.4+, update?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -190,6 +189,9 @@ public ElasticSearchIndex(Configuration config) throws BackendException {
         maxResultsSize = config.get(INDEX_MAX_RESULT_SET_SIZE);
         log.debug("Configured ES query result set max size to {}", maxResultsSize);
 
+        scriptLang = client.getMajorVersion() < 5 ? "groovy" : "painless";
```
amcp: pull 5 out into constant

amcp: or even better into enum ESMajorVersion where it would contain TWO and FIVE

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/TransportElasticSearchClient.java
```diff
@@ -194,6 +194,11 @@ public void close() throws IOException {
         client.close();
     }
 
+    @Override
+    public int getMajorVersion() {
+        return 2;
```
amcp: maybe use the MajorVersion enum described above

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -76,6 +82,29 @@ public void close() throws IOException {
     }
 
     @Override
+    public int getMajorVersion() {
+        if (majorVersion == null) {
```
amcp: to un-nest code lets check majorVersion != null first and return, and then do the logic to compute it

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -223,4 +252,19 @@ private Response performRequest(String method, String path, byte[] requestData)
         return response;
     }
 
+    @JsonIgnoreProperties(ignoreUnknown=true)
+    private static final class ClusterInfo {
```
amcp: maybe add a comment as to why we have a map to represent versions? is this class consistent with a remote ES deployment model? thinking along the lines of AWS ES where JanusGraph may not be in control of the cluster topology with its configuration.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -563,6 +569,22 @@ private String getAdditionScript(KeyInformation.IndexRetriever informations, Str
         return script.toString();
     }
 
+    private Map<String,Object> getAdditionDoc(KeyInformation.IndexRetriever informations, String storename, IndexMutation mutation) throws PermanentBackendException {
+        Map<String,Object> doc = new HashMap<>();
+        for (IndexEntry e : mutation.getAdditions()) {
+            KeyInformation keyInformation = informations.get(storename).get(e.field);
+            switch (keyInformation.getCardinality()) {
```
amcp: if you only care about one value then maybe say if keyInformation.getCardinality != SINGLE continue and get rid of the switch-case-break

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -563,6 +569,22 @@ private String getAdditionScript(KeyInformation.IndexRetriever informations, Str
         return script.toString();
     }
 
+    private Map<String,Object> getAdditionDoc(KeyInformation.IndexRetriever informations, String storename, IndexMutation mutation) throws PermanentBackendException {
+        Map<String,Object> doc = new HashMap<>();
+        for (IndexEntry e : mutation.getAdditions()) {
```
amcp: maybe switch to a .stream().map().filter().collect()

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/TransportElasticSearchClient.java
```diff
@@ -119,9 +119,16 @@ public void bulkRequest(List<ElasticSearchMutation> requests) throws IOException
...
-                    Script script = new Script(inline, ScriptService.ScriptType.INLINE, null, null);
-                    UpdateRequestBuilder update = client.prepareUpdate(indexName, type, id).setScript(script);
+                    UpdateRequestBuilder update = client.prepareUpdate(indexName, type, id);
+                    if (request.getSource().containsKey("script")) {
+                        Map<String,String> script = ((Map<String, String>) request.getSource().get("script"));
+                        String inline = script.get("inline");
+                        String lang = script.get("lang");
+                        update.setScript(new Script(inline, ScriptService.ScriptType.INLINE, lang, null));
+                    }
+                    if (request.getSource().containsKey("doc")) {
```
amcp: externalize doc to constant

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/TransportElasticSearchClient.java
```diff
@@ -119,9 +119,16 @@ public void bulkRequest(List<ElasticSearchMutation> requests) throws IOException
                     brb.add(new IndexRequest(indexName, type, id).source(request.getSource()));
                     break;
                 } case UPDATE: {
-                    String inline = ((Map<String, String>) request.getSource().get("script")).get("inline");
-                    Script script = new Script(inline, ScriptService.ScriptType.INLINE, null, null);
-                    UpdateRequestBuilder update = client.prepareUpdate(indexName, type, id).setScript(script);
+                    UpdateRequestBuilder update = client.prepareUpdate(indexName, type, id);
+                    if (request.getSource().containsKey("script")) {
+                        Map<String,String> script = ((Map<String, String>) request.getSource().get("script"));
+                        String inline = script.get("inline");
```
amcp: externalize inline to constant

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/TransportElasticSearchClient.java
```diff
@@ -119,9 +119,16 @@ public void bulkRequest(List<ElasticSearchMutation> requests) throws IOException
                     brb.add(new IndexRequest(indexName, type, id).source(request.getSource()));
                     break;
                 } case UPDATE: {
-                    String inline = ((Map<String, String>) request.getSource().get("script")).get("inline");
-                    Script script = new Script(inline, ScriptService.ScriptType.INLINE, null, null);
-                    UpdateRequestBuilder update = client.prepareUpdate(indexName, type, id).setScript(script);
+                    UpdateRequestBuilder update = client.prepareUpdate(indexName, type, id);
+                    if (request.getSource().containsKey("script")) {
```
amcp: externalize script to constant

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/TransportElasticSearchClient.java
```diff
@@ -119,9 +119,16 @@ public void bulkRequest(List<ElasticSearchMutation> requests) throws IOException
...
+                        Map<String,String> script = ((Map<String, String>) request.getSource().get("script"));
+                        String inline = script.get("inline");
+                        String lang = script.get("lang");
```
amcp: externalize lang to constant

amcp:
@sjudeng I reviewed everything once, i will take a look at the enormous https://github.com/JanusGraph/janusgraph/pull/79/commits/b34be738ac76a14d25d85302696aa7924badf718 again later today

docs/versions.txt
hsaputra: Has this concern being addressed? Always nice to move away from 0.94 but should it be addressed by this PR

docs/elasticsearch.txt
sjudeng: You mean in terms of compatibility? All 2.x releases are supported. Test suite has been run against 2.4.4, 2.3.5, 2.2.2, 2.1.2, 2.0.2 (though see note at the top regarding limitations in versions before 2.3.0).

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
sjudeng: The version is obtained from a ES request with default path (e.g. `http://localhost:9200/`). The response is expected to be of the form below and this class is used for unmarshalling to avoid nested `Map` casts. But would AWS ES (or other deployment models) not return a response of this form to this request? I can look into this more.

```javascript
{
  "name" : "X29X896",
  "cluster_name" : "elasticsearch",
  "cluster_uuid" : "ROrFjDOqSIGLBDsZhxFOcA",
  "version" : {
    "number" : "5.2.1",
    "build_hash" : "db0d481",
    "build_date" : "2017-02-09T22:05:32.386Z",
    "build_snapshot" : false,
    "lucene_version" : "6.4.1"
  },
  "tagline" : "You Know, for Search"
}
```

.travis.yml
sjudeng: I moved this all out into an `elasticsearch2` profile, which I think resolves this.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sjudeng: Done (good idea)

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sjudeng: I did take out the switch but I wasn't able to get a clean implementation using stream API in this case. Let me know if you want me to take another crack at it.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/TransportElasticSearchClient.java
sjudeng: Yep, externalized "script", "inline" and "doc.

sjudeng:
@amcp and @hsaputra - Thanks for the feedback. Everything should be addressed with a couple exceptions as mentioned above. If you haven't already please also review the breaking changes section in the PR notes above.

docs/versions.txt
sjudeng: Rebased to pick up #176 and now just updating the ES version here.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
sjudeng: @amcp I confirmed that AWS ES does support requests with default path (e.g. https://search-estest-*.*.es.amazonaws.com/) and the response is formatted as expected and shown above.

amcp: Thanks for checking.

sjudeng:
@amcp, @jerryjch, @pluradj, @hsaputra, @twilmes, @dylanht - I think this PR has been sufficiently reviewed at this point. There were two early commits that were reviewed/approved by three other committers and the more recent commits have been approved by two other committers. All issues have been mitigated and there has been a community discussion+vote on the dev list. In particular on the issue of ES 1.x support, the 0.1 release branch will maintain the original ES v1.5.1 support.

Unless there are any objections I would like to merge this PR later this week.

hsaputra:
+1 @sjudeng

PR #82, opened 2017-02-03
Added a Travis CI badge.

mbrukman:
Click on the "View" button in the diffs to see the rendered view on my branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,11 @@ env:
 
     - MODULE='test'
 
+matrix:
+  # https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail
+  allow_failures:
+    - env: MODULE='berkeleyje'
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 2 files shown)


mbrukman:
@sjudeng – looks like `berkeleyje` is now failing: https://travis-ci.org/JanusGraph/janusgraph/jobs/198165185

Should we switch it to a single-thread test suite, or is this a deeper problem?

sjudeng:
Unfortunately I think it's a deeper problem. I know I saw sporadic failures running tests in Travis in all but the janusgraph-test module. You can try going to single thread but that was a shot in the dark on my part and it didn't consistently resolve the issue for me. For now I'd recommend dropping back to just running tests for janusgraph-test in master. Otherwise I suspect you'll see this (e.g. build failures not related to PR) more often then not.

Meanwhile you might keep one of your existing automated test issues open or create a new one to track the effort to try and understand what the issue is and if there are any workarounds.

mbrukman:
@sjudeng – turns out, it's possible to mark a specific job as `allowed_failure`, so I've just done this for `berkeleyje`, and let's see what happens. If needed, we can mark the `lucene` similarly.

Unfortunately, this requires a top-level `matrix` section, which previously conflicted with our nested

```
env:
  matrix:
    ...
```

section, so we may need to remove it if this doesn't quite work.

mbrukman:
Looks like it nicely separates required from allowed failures: https://travis-ci.org/JanusGraph/janusgraph/builds/198198304 — this is awesome!

sjudeng:
Yeah that's perfect. For diagnosing the sporadic fail issue it will be great to have history on the failures. Nice find.

Issue #40, opened 2017-01-13
Make snapshot builds available


pluradj:
https://twitter.com/funkatron/status/819972558871293952

Making snapshot distribution zips available before the first official release will lower the bar for people wanting to try out JanusGraph.

funkatron:
Thanks! That will help a lot, esp for folks who are not Java-savvy.

rafatuita:
can you guys prioritize this issue @pluradj ? quite a few of us can't migrate from titan without snapshots available. Since this is a fork from latest version, don't see the issue to release it as is...

looking forward to move our code to Janus...

twilmes:
Ok, made some progress but still had to deploy the packaged zip manually via `mvn deploy:deploy-file` command line.  I'm continuing to work that so a `mvn deploy` will suffice.

In the meantime though:
* https://oss.sonatype.org/content/repositories/snapshots/org/janusgraph/
* installation zip can be found [here](https://oss.sonatype.org/content/repositories/snapshots/org/janusgraph/janusgraph-dist/0.1.0-SNAPSHOT/)

mbrukman:
@twilmes — thanks for building and deploying the snapshot!

Can you please create a tag at the same Git hash that you built the snapshot from? E.g., `v0.1.0-SNAPSHOT-2017-xx-yy` (with `xx-yy` being the date of the hash)? That would make it possible to create a source release mark on GitHub, and we can attach the binary artifacts on GitHub to host it there for ease of downloading.

Our docs currently send folks to the GitHub release page, so this will make it very easy for users to discover the downloadable artifacts, and get either the source at the same version or the built JARs.

twilmes:
Good call @mbrukman.  I'll take care of that.  That naming scheme sounds good to me.

twilmes:
I got things setup and uploaded the zip to https://github.com/JanusGraph/janusgraph/releases/tag/0.1.0-SNAPSHOT-2017-03-18.  The verbiage is pretty minimal at this point but can be easily updated if we so desire.

mbrukman:
Per discussion on the mailing list, I created a tag [`v0.1.0-rc1`](https://github.com/JanusGraph/janusgraph/releases/tag/v0.1.0-rc1) and removed the earlier tag `0.1.0-SNAPSHOT-2017-03-18` from the repo and releases.

[...]

PR #98, opened 2017-02-09
Support for deploying to GitHub Releases from Travis CI

sjudeng:
@mbrukman This PR is just for consideration in case there's interest. It adds support for deploying JanusGraph distributions to [GitHub Releases](https://help.github.com/articles/about-releases/) as part of Travis CI builds. For example result see [this project](https://github.com/ngageoint/titan/releases). Note it only works for tags, so for example it wouldn't immediately solve #40 unless you'd want to create periodic tags (not sure if it would work to overwrite files).

If you want to try it out you'll need to follow [Travis CI instructions](https://docs.travis-ci.com/user/deployment/releases/) and update .travis.yml to change `CHANGE_ME` to the relevant encrypted api key.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ before_deploy:
 deploy:
   provider: releases
   api_key:
-    secure: CHANGE_ME
+    secure: NpXWvr1kuKs+nv5xsQWQsEC1ulhLE5Cf2siO4NXHgOwRiPqKePx5WPH1EloBtf+dopq7sogMuQQf+5t3/KGhRUFvbEm8oEMylzfuz7L615i53GoQcaDhDUDd5maF+bYDKh7y9dyfSgRHxuVMItFQAhJpUtBFGBwO9lsqJQaIk4ar8RUW0vSlYriigZJdewapH3GoawfXXrbbo+nFjwadLTpBBaD4EMBcvPEPhpX8wMnjySAI/GXK3YlWEK4b8+Uh4cEcwByrzKxPIJaz4gcQTzfvA/9n6sh/39i/X+w4+ouG/ZeteSmFmVfxh1yY5CiNKpdC/b5c96iRHg81Umo7craT9D4ve6i1AF7KbWZOnRjvKg36lIIu1tPOwPstgymzbivU5NUFc/EkW0h3sHCHWHdl8mASNfagh723/qWgMM0KvMPxjg2uf2LtdYUoQ3hmTpZyVqYaCiLJxp5pehYEWfg3ax1GewXTPFTzqHWuohFO60fGf7Cfy9oygN3sF0dC+N0aCk9TOHlPEhmKFylZ3f2W7QkuGMNeFSdme9g5BUgBpPlWpCTKX9rfror8VEAr143RlOGCneR43flMUYpjLWbUAKEY+I5jQnuSiSOlmXLnlMR7DSqVZudjzvV+64lMiHtUKhR2PwofNQ2/B/GvbSYQKi1NVn/n0pa3yRgPZQ4=
   file_glob: true
   skip_cleanup: true
   file: "$JANUSGRAPH_DIST_ZIP"
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,21 @@ script:
   - mvn install -DskipTests=true -B -pl janusgraph-${MODULE} -am
   - mvn verify -pl janusgraph-${MODULE} ${ARGS}
 
+before_deploy:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -86,6 +86,21 @@ script:
   - mvn install -DskipTests=true -B -pl janusgraph-${MODULE} -am
   - mvn verify -pl janusgraph-${MODULE} ${ARGS}
 
+before_deploy:
+  - mvn package -DskipTests -Dgpg.skip=true -Pjanusgraph-release
```
amcp: why skip gpg?

sjudeng: I never tried to work through getting GPG signing to work as part of a Traivs CI job. Not sure if it's possible. If not maybe deploying in this way is a non-starter? I do like the general idea of handling release deployment as part of the CI process though.

amcp: lets make an issue to track GPG signing in order to unblock this PR.

amcp:
@mbrukman I am all for automatic SNAPSHOT releases. Can you take a look and if it make sense approve this?

amcp:
@sjudeng Are these __all__ the changes necessary for GitHub Releases deployments? Do we also need to set a CHANGE_ME?

amcp:
@mbrukman Marko Rodriguez says as soon as we make some kind of release (even if it is a snapshot) we can get listed on the Tinkerpop landing page (#107). Seems like a very good avenue for visibility so lets prioritize this action item?

sjudeng:
@amcp Yes someone with write permissions on the repo would need to setup the encrypted API key and update .travis.yml in this PR to change `CHANGE_ME` to that key. See [Travis releases deployment instructions](https://docs.travis-ci.com/user/deployment/releases/) for more info.

But note that this only works if a tag (e.g. 0.1.0) is pushed to the repository, so it's really more for the standard release case. A snapshot tag (e.g.  or 0.1.0-SNAPSHOT-201702121349) would need to be pushed to get this to work as-is for the snapshot release case .. so it's not automatic. It looks like it is possible to create and push tags as part of the Travis build, so a fully automatic solution might be possible but isn't currently implemented in this PR.

amcp:
@sjudeng I've got a key. Ill file a PR against this one and you can merge it into yours

amcp:
@sjudeng now you can add the line to the build script to create the tag.

.travis.yml
```diff
@@ -86,6 +86,21 @@ script:
...
+  - export JANUSGRAPH_DIST_ZIP="$(ls janusgraph-dist/janusgraph-dist-hadoop-2/target/janusgraph-*-hadoop2.zip)"
+  - echo "Deploying $JANUSGRAPH_DIST_ZIP to GitHub releases"
```
amcp: create and push the tag here

sjudeng:
I don't think it would go in `before_deploy` as this would only be reached if deploying (e.g. tag was already created). It could go in `after_success` (see below). But I think pushing will fail because Travis only has read access to the repo. It looks like there are examples of accomplishing write/push from Travis build, but I haven't looked into this. Until this is resolved or maybe a better alternative for snapshot releases is found, I'd recommend going with the manual tag creation at this point, either for stable release or an initial alpha "snapshot" release.

https://github.com/travis-ci/travis-ci/issues/1476
http://stackoverflow.com/questions/28217556/travis-ci-auto-tag-build-for-github-release

amcp:
Can you configure Travis to run once a day for the purpose of creating snapshot releases?

sjudeng:
It looks like there's support for [Cron Jobs in Travis](https://docs.travis-ci.com/user/cron-jobs). To use with the deploy capabilities I'm guessing there would need to be a conditional for `TRAVIS_EVENT_TYPE="cron"` or something like that. Regarding maybe the harder problem of enabling write/push from Travis to GitHub, it seems like using the [GitHub API](https://developer.github.com/v3/git/tags/) would be the best. One would setup a [GitHub API token](https://github.com/blog/1509-personal-api-tokens) with write access, [encrypt it with Travis](https://docs.travis-ci.com/user/encryption-keys/) and then create a script (probably parametrized by build number/etc. to be passed from Travis) to perform the tag using the API.

But again this might not be necessary for this project if standard releases, with maybe some intermediate manually-created snapshot releases, would be sufficient.

amcp:
@sjudeng lets close out the remaining items on this PR?

sjudeng:
Closing this for now. We're releasing infrequently and handling release artifact deployment manually at this point. Also GPG signing which isn't included here seems like it would be required for formal release artifacts.

PR #99, opened 2017-02-09
Cleanup some dependencies and implicit casts to make things cleaner in eclipse 

lindsayismith:
This CL should change nothing, I've set up JanusGraph in eclipse and made some changes to make things saner.  For reference I'm using M2E to configure the project, no special settings.

* Switch to org.codehaus.gmavenplus version 1.5 for all gmaven plugins, there used to be a mix of gmaven and gmavenplus.  The latest version works nicely with M2E.
* Remove the janusgraph-hadoop project from the dependencies section of janusgraph-all/pom.xml as it's not a java project.
* Do explicit casts wherever eclipse was generating errors.
* A -Xlint:unchecked to compiler options so we get more warnings for this kind of thing.

Eclipse builds still aren't clean, theres errors from with the assembly plugin and something to do with packing output locations I haven't been able to track down yet.

### DIFF
--- a/janusgraph-all/pom.xml
+++ b/janusgraph-all/pom.xml
@@ -49,6 +49,11 @@
             <version>${hbase098.core.version}-hadoop2</version>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>janusgraph-hadoop</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
             <version>${hadoop2.version}</version>
--- a/janusgraph-all/pom.xml
+++ b/janusgraph-all/pom.xml
@@ -49,11 +49,6 @@
             <version>${hbase098.core.version}-hadoop2</version>
         </dependency>
         <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>janusgraph-hadoop</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
             <version>${hadoop2.version}</version>
--- a/janusgraph-core/pom.xml
+++ b/janusgraph-core/pom.xml
@@ -116,15 +116,15 @@
         </testResources>
         <plugins>
             <plugin>
-                <groupId>org.codehaus.gmaven</groupId>
-                <artifactId>gmaven-plugin</artifactId>
+                <groupId>org.codehaus.gmavenplus</groupId>
+                <artifactId>gmavenplus-plugin</artifactId>
                 <version>1.5</version>
                 <executions>
                     <execution>
                         <goals>
                             <goal>generateStubs</goal>
                             <goal>compile</goal>
-                            <goal>generateTestStubs</goal>
+                            <goal>testGenerateStubs</goal>
                             <goal>testCompile</goal>
                         </goals>
                         <configuration>
--- a/janusgraph-core/pom.xml
+++ b/janusgraph-core/pom.xml
@@ -118,7 +118,7 @@
             <plugin>
                 <groupId>org.codehaus.gmavenplus</groupId>
                 <artifactId>gmavenplus-plugin</artifactId>
-                <version>1.5</version>
+                <version>${gmavenplus.version}</version>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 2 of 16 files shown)


janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPartitionGraphTest.java
```diff
@@ -83,7 +83,8 @@ public WriteConfiguration getConfiguration() {
 
     private IDManager idManager;
 
-    @Before
+    @Override
+  @Before
```
mbrukman: Fix indent.

lindsayismith: Done.

mbrukman:
LGTM, but other folks should also review + approve.

janusgraph-all/pom.xml
```diff
@@ -49,11 +49,6 @@
             <version>${hbase098.core.version}-hadoop2</version>
         </dependency>
         <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>janusgraph-hadoop</artifactId>
```
jerryjch: I am a little confused here.
If we remove janusgraph-hadoop from the janusgraph-all pom, does it mean that an application that depends on janusgraph-all in its pom will not have janusgraph-hadoop artifact pulled in, thus missing this dependency?

amcp: is it possible this got pulled in by another dependency? `mvn dependency:tree` should tell us exactly what is getting pulled in.

janusgraph-core/pom.xml
```diff
@@ -116,15 +116,15 @@
...
+                <groupId>org.codehaus.gmavenplus</groupId>
+                <artifactId>gmavenplus-plugin</artifactId>
                 <version>1.5</version>
                 <executions>
                     <execution>
                         <goals>
                             <goal>generateStubs</goal>
                             <goal>compile</goal>
-                            <goal>generateTestStubs</goal>
+                            <goal>testGenerateStubs</goal>
```
amcp: how did you test the changed goal?

janusgraph-test/pom.xml
```diff
@@ -100,13 +100,13 @@
...
                 <artifactId>gmavenplus-plugin</artifactId>
-                <version>1.2</version>
+                <version>1.5</version>
                 <executions>
                     <execution>
                         <goals>
                             <goal>addSources</goal>
                             <goal>addTestSources</goal>
-                            <goal>generateStubs</goal>
+                            <goal>testGenerateStubs</goal>
```
amcp: again, how did you test this name change?

janusgraph-test/pom.xml
```diff
@@ -100,13 +100,13 @@
             <plugin>
                 <groupId>org.codehaus.gmavenplus</groupId>
                 <artifactId>gmavenplus-plugin</artifactId>
-                <version>1.2</version>
+                <version>1.5</version>
```
amcp: can we pull this version number up to parent pom

amcp:
any updates?

janusgraph-all/pom.xml
lindsayismith: You're quite right.  This dependency is used to output janusgraph-hadoop-0.1-SNAPSHOT.jar in the final lib folder and without it the jar is dropped, my bad.   

The issue is that this is a dependency on a module that doesn't have any classes, it's just an assembly project.  So it's not a classpath dependency, it's a packaging dependency.  I'm not sure how to handle this nicely in Maven, suggestions are welcome.  I've added the dependency back int.

janusgraph-core/pom.xml
lindsayismith: This goal was renamed in this plugin.  But, it turns out there's no groovy sources at all in the janusgraph-core project, so I'm not sure what this is supposed to be doing.

janusgraph-test/pom.xml
lindsayismith: This one I messed up, and shouldn't have renamed that goal, as the testGenerateStubs one is already mentioned lower down, Thanks for checking.

janusgraph-test/pom.xml
lindsayismith: Done.

janusgraph-core/pom.xml
lindsayismith: As far as testing goes, I was wary of the changes meaning that some files weren't generated correctly.  I tested by comparing the directory structures, including all target/generated folders, of a master repo vs one with these changes.

lindsayismith:
Thanks for close reading, I reinstated the hadoop dependency as it's used for packaging, fixed one of the edits I made to the groovy plugin definition, move version number up.

amcp:
fixes part of #113

PR #96, opened 2017-02-08
Including Eclipse IDE building

blacknred0:

PR #106, opened 2017-02-10
Issue 95: Adding first version of development process documentation.

twilmes:
Here is a first cut at developer process docs.  The release policy section is pretty sparse but I think we can expand this when we go through our first release process.  I imagine we'll also add explicit release steps and directions at that point also.  I'm not wed to any of these particular policies but I think they're a good start.  Specifically I'm curious if folks think we should allow CTR in limited cases like I've defined here.

### DIFF
--- a/docs/development.txt
+++ b/docs/development.txt
@@ -1,0 +1,82 @@
+[[development-process]]
+= JanusGraph Development Process
+
+The following sections describe the JanusGraph development process.
+
+
+[[development-decisions]]
+== Development Decisions
+
+Many development decisions will be made during the day-to-day work of JanusGraph contributors without any extra process overhead.
+However, for larger bodies of work and significant updates and additions, the following process shall be followed.
+Significant work may include, but is not limited to:
+
+* A major new feature or subproject, e.g., a new storage adapter
+* Incrementing the version of a core dependency such as a Apache TinkerPop
+* Addition or deprecation of a public API
+* Internal shared data structure or API change
+* Addition of a new major dependency
+
+For these sorts of changes, contributors will:
+
+* Create one or more issues in the GitHub issue tracker
+* Start a DISCUSS thread on the https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] list where the proposed change may be discussed by committers
+and other community members
+* When the proposer feels it appropriate, a VOTE shall be called
+* Two +1 votes are required for the change to be accepted
+
+[[branching]]
+== Branching
+
+For features that involve only one developer, developers will work in their own JanusGraph forks, managing their own branches, and submitting pull requests when ready. If multiple developers wish to collaborate on a feature,
+they may request that a feature branch be created in JanusGraph repository.  If the developers are not committers, a JanusGraph committer will be assigned as the shepherd for that branch.  The shepherd will be responsible for merging pull requests to that branch.
+
+=== Branch Naming Conventions
+
+All branch names hosted in the JanusGraph repository shall be prepended with `Issue_#_`.
+
+[[pull-requests]]
+== Pull Requests
+
+Most changes to JanusGraph must be reviewed before they are committed (RTC - review-then-commit).
+This workflow is performed using GitHub's pull request process. In instances where the committer deems the full process unnecessary, commit-then-review (CTR) may be employed.
+This should be used infrequently and only for extremely trivial changes.
+Changes of this sort may include:
+
+* Documentation typo or small documentation addition
+* Addition of a new test case
+
+Any commit that is made following CTR shall include the fact that it is a CTR in the commit comments.
+Community members who wish to review CTRs may subscribe to the https://groups.google.com/forum/#!forum/janusgraph-commits[JanusGraph commits] list so that they will see CTRs as they come through.
+If another committer responds with a -1, the commit should be rolled back and the formal RTC process should be followed.
+
+Users wishing to contribute to JanusGraph should fork the https://github.com/janusgraph/janusgraph[JanusGraph] repository and then submit pull requests using the GitHub pull request process.
+Every pull request must be associated with an existing issue.
+Be sure to include the relevant issue number in the title of your pull request and a description of what test suites were run to validate the pull request.
+
+Pull requests commit messages should clearly describe what work was accomplished.
+Intermediate work-in-progress commits should be squashed prior to pull request submittal.
+
+Pull requests must be reviewed before being merged.
+The JanusGraph project uses the https://help.github.com/articles/about-pull-request-reviews/[GitHub review process] to review pull requests.
+Non-committers are free, and encouraged to review pull requests.
+Their review will be non-binding, but can be taken into consideration and are still very valuable community input.
+The following rules apply to the voting process.
+
+* Two committer approvals are required to merge a pull request
+* One or more -1 votes within a change request will veto the pull request until the noted issue(s) are addressed and the -1 vote is withdrawn
+* Change requests will not be considered valid without an explanation
+
+[[release-policy]]
+== Release Policy
+
+Any JanusGraph committer may propose a release.
+To propose a release, simple start a new RELEASE thread on https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] proposing the new release and requesting feedback on what should be included in the release.
+After consensus is reached the release manager will perform the following tasks:
+
+* Create a release branch so that work may continue on `master`
+* Prepare the release artifacts
+* Call a vote to approve the release on https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev]
+* Committers will be given 72 hours to review and vote on the release artifacts
+* Three +1 votes are required for a release to be approved
+* One or more -1 votes with explanation will veto the release until the noted issues are addressed and the -1 votes are withdrawn
--- a/docs/index.txt
+++ b/docs/index.txt
@@ -15,4 +15,6 @@ include::advanced.txt[]
 
 include::internals.txt[]
 
+include::development.txt[]
+
 include::appendices.txt[]


mbrukman:
@twilmes — you seem to have the same issue as @blacknred0 did in PR #96, namely that GitHub is missing the metadata for your username in your commit, so the automated tool cannot tell who submitted this PR. @blacknred0 submitted a ticket to GitHub support and they fixed it; please do the same.

Once GitHub fixes it, @janusgraph-bot will flip the CLA bit for you automatically.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+[[release-policy]]
+Release Policy
+--------------
+
+Any JanusGraph committer may propose a release. To propose a release, simple start a new [RELEASE] thread on
+https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] proposing the new release and requesting feedback on what should be included in the release. After
+consensus is reached, decide on a date to start the release code freeze. Allow for one week between the start of the
+code freeze and the beginning of the release vote. During that week, a release manager shall create release
+artifacts. Committers will be given seven days to review and vote on the release artifacts. Three +1 votes are required for a release to happen.
+If one or more committers casts a -1 vote, the release will be halted.
```
ptgoetz: In Apache projects, release votes can't be vetoed, they require at least 3 +1 votes and more +1s than -1s. Just pointing that out in case we want to follow suit.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+[[release-policy]]
+Release Policy
+--------------
+
+Any JanusGraph committer may propose a release. To propose a release, simple start a new [RELEASE] thread on
+https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] proposing the new release and requesting feedback on what should be included in the release. After
+consensus is reached, decide on a date to start the release code freeze. Allow for one week between the start of the
+code freeze and the beginning of the release vote. During that week, a release manager shall create release
+artifacts. Committers will be given seven days to review and vote on the release artifacts. Three +1 votes are required for a release to happen.
```
ptgoetz: Is a code freeze necessary? Couldn't we just create a release branch that forms the base for the release (i.e. so work on master can continue)?

We may also want to shorten the vote time (Apache uses min. 72hrs.). As currently worded it looks like there would be a minimum turnaround of two weeks for a release. What if, for example we need to rush out a security patch?

amcp: I agree with the release branch approach, lets unblock ourselves.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+perform before their change is accepted. In the instance where the change is not approved or significant updates to
+the proposal are made, we wish to minimize the amount of rework required.
+
+[[branching]]
+Branching
+---------
+For features that involve only one developer, developers will work in their own JanusGraph forks, managing their
+own branches, and submitting pull requests when ready. Feature branches in the main JanusGraph repository will be
+created when more than one developer collaborates on a feature.
+
```
amcp: Perhaps this means that the owner of a feature (=feature branch) should be a committer?

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+Branch Naming Conventions
+~~~~~~~~~~~~~~~~~~~~~~~~~
+All branch names shall be prepended with `Issue_#_`.
+
+[[pull-requests]]
+Pull Requests
+-------------
+
+Most changes to JanusGraph must be reviewed before they are committed (RTC). This workflow is performed using Github's
```
amcp: s/(RTC)/(RTC - review-then-commit)/

docs/development.txt
amcp: We need a break-glass mechanism for releases. If we are affected by a CVE, we would like to act even sooner. Usually CVE can be mitigated by updating dependency versions, so in that case the mechanics of creating a release would only be limited by TinkerPop test time? Currently I understand this to be 24 hours.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+* Create one or more issues in the Github issue tracker
+* Start a [DISCUSS] thread on the https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] list where the proposed change may be discussed by committers
+and other community members
+* When the proposer feels it appropriate, a VOTE shall be called. As with pull requests three +1 votes, including
+one from the relevant Technical Steering Committee (TSC) member if the change is limited to a single project, are required for the change to be accepted.
+
+Though not required, it is strongly recommended that contributors limit the amount of development work they
+perform before their change is accepted. In the instance where the change is not approved or significant updates to
+the proposal are made, we wish to minimize the amount of rework required.
```
twilmes: I think this is superfluous upon rereading so I'm taking it out.  My intent was to discourage folks from spending a ton of time on development before seeking consensus on whether or not a change is a good idea or not but I think the process makes that clear already.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+Non-committers are free, and encouraged to vote. Their votes will be non-binding, but can be taken into consideration
+and are still very valuable community input. The following rules apply to the voting process.
+
+* Three +1 votes are required merge a pull request
+* One or more -1 votes will veto the pull request until the noted issue(s) are addressed and the pull request
+is resubmitted
+* Committers may vote on their own pull requests
+* Pull requests that are limited to one component require at least one of the +1 votes to come from the relevant
+TSC member. If TSC members are unable to review after 2 weeks, the pull request may be merged
+after three +1 votes
```
twilmes: I'll remove the TSC requirement.  I think that's too stringent to start and we always add it in later if need be.

docs/development.txt
```diff
@@ -0,0 +1,84 @@
...
+* When the proposer feels it appropriate, a VOTE shall be called
+* Three +1 votes are required for the change to be accepted.
+
+[[branching]]
+Branching
+---------
+For features that involve only one developer, developers will work in their own JanusGraph forks, managing their
+own branches, and submitting pull requests when ready. If multiple developers wish to collaborate on a feature,
+they may request that a feature branch be created in JanusGraph repository.  If the developers are not committers, a
+JanusGraph committer will be assigned as the shepard for that branch.  The shepard will be responsible for merging
```
amcp: s/shepard/shepherd/

docs/development.txt
```diff
@@ -0,0 +1,84 @@
+[[development-process]]
+JanusGraph Development Process
+==============================
```
mbrukman: Please use Asciidoc best practices for section titles, as described in [their docs](http://asciidoctor.org/docs/asciidoc-recommended-practices/#section-titles).

Specifically, this line should be:

> = JanusGraph Development Process

and other section headings should use the number of leading `=` equal to their header level. 

The current approach of using different types of characters on a line is complex (need to remember which character means which heading level) and error-prone (must equal in length to the heading, or else it breaks. This is the type of thing I had to deal with during the global `s/Titan/Janusgraph/` cleanup; see [this commit](https://github.com/JanusGraph/janusgraph/commit/e21734fd1ffef90ef44a0ad7d2da0b2f5859baa4) for more details.

I am working on a change to convert all existing headers to the Asciidoc-recommended approach, so it would be nice to not add any more of these anachronistic section titles.

docs/development.txt
```diff
@@ -0,0 +1,84 @@
...
+* Addition of a new test case
+
+Any commit that is made following CTR shall include the fact that it is a CTR in the commit comments.  The committer
+should also send a message to https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] to notify other
+project members of the CTR.  If another committer responds with a -1, the commit should be rolled back and the formal
+RTC process should be followed.
+
+Users wishing to contribute to JanusGraph should fork the https://github.com/janusgraph/janusgraph[JanusGraph]
+repository and then submit pull requests using the Github pull request process. Please include the relevant issue
+number in the title of your pull request. A vote (-1, 0, +1) is required before a pull request may be merged.
```
mbrukman: Instead of +1 / 0 / -1, can we just use the GitHub approval flow? This makes the approvals stand out, and it's easy to count them: they show up as checkmarks in the right-hand side vs. having to scan for +1 in the message stream.

Only committers / maintainers will have the ability to approve a PR, which makes it easy to tell which ones are binding vs. non-binding +1s.

docs/development.txt
```diff
@@ -0,0 +1,84 @@
...
+Non-committers are free, and encouraged to vote. Their votes will be non-binding, but can be taken into consideration
+and are still very valuable community input. The following rules apply to the voting process.
+
+* Three +1 votes are required merge a pull request
```
mbrukman: Is this standard practice for ASF projects to require three +1s for every commit? Is this not too many? I ask because a requirement of 3 reviewers with +1 may slow down the review / submit process for PRs. Thoughts?

jerryjch: Have the same feeling that 3 sounds a little too much.
How about this?  Two +1 or approvals from other than the author.
In this way, committer author does not needs to explicitly cast +1.

docs/development.txt
```diff
@@ -0,0 +1,84 @@
...
+* Documentation typo or small documentation addition
+* Addition of a new test case
+
+Any commit that is made following CTR shall include the fact that it is a CTR in the commit comments.  The committer
+should also send a message to https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] to notify other
+project members of the CTR.  If another committer responds with a -1, the commit should be rolled back and the formal
+RTC process should be followed.
+
+Users wishing to contribute to JanusGraph should fork the https://github.com/janusgraph/janusgraph[JanusGraph]
+repository and then submit pull requests using the Github pull request process. Please include the relevant issue
```
jerryjch: Could you clarify the relationship between Issue and PR? Every PR needs to be associated with an Issue? I am for it.
Could we include some recommendations for the commits going into the pull request? i.e. the commit message, organize the commits.

jerryjch: In the PR, include a statement of how the Pull Request has been tested. e.g. unit test, new test, or other testing.

docs/development.txt
```diff
@@ -0,0 +1,84 @@
...
+https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] proposing the new release and requesting
+feedback on what should be included in the release. After consensus is reached the release manager will perform
+the following tasks:
+
+* Create a release branch so that work may continue on `master`
+* Prepare the release artifacts
+* Call a vote to approve the release on https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev]
+
+Committers will be given 72 hours to review and vote on the release artifacts. A minimum of three +1 votes and more
++1's than -1's are required for a release to be approved.
```
jerryjch: Could you clarify this?  Is this more relaxed than the rule for a Pull Request?  Should we just do the same -1 policy as for PR?

docs/development.txt
twilmes: There isn't a standard across ASF projects so I was just going off what we do on TinkerPop in this case.  Two +1 approvals from other than the author sounds good to me.

docs/development.txt
twilmes: Good call.  I'll also move back to sentence per line.

docs/development.txt
twilmes: @ptgoetz I know this was just a suggestion, you okay with me switching this back to a -1 vetoing a release?  I'm on the fence with this one because in many cases, I would imagine if anyone felt there was a serious enough issue to warrant a veto, we'd probably be in agreement, but maybe not...

ptgoetz: @twilmes Yes that's fine. The way it usually works on Apache projects is that the release manager takes a -1 seriously and usually cancels the release vote to address it. In my experience as a release manager I've never let a release proceed if there were binding -1 votes.

twilmes:
Pushed an update, I'll squash commits after we reach consensus and are ready to merge.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+
+[[development-decisions]]
+== Development Decisions
+
+Many development decisions will be made during the day-to-day work of JanusGraph contributors without any extra process overhead.
+However, for larger bodies of work and significant updates and additions, the following process shall be followed.
+Significant work may include, but is not limited to:
+
+* A major new feature or subproject eg. a new storage adapter
```
mbrukman: s/eg./, e.g.,/

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+* Addition or deprecation of a public API
+* Internal shared data structure or API change
+
+For these sorts of changes, contributors will:
+
+* Create one or more issues in the Github issue tracker
+* Start a DISCUSS thread on the https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] list where the proposed change may be discussed by committers
+and other community members
+* When the proposer feels it appropriate, a VOTE shall be called
+* Three +1 votes are required for the change to be accepted.
```
mbrukman: Did you want to change this to be "Two +1 votes are required..."?

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+[[branching]]
+== Branching
+
+For features that involve only one developer, developers will work in their own JanusGraph forks, managing their own branches, and submitting pull requests when ready. If multiple developers wish to collaborate on a feature,
+they may request that a feature branch be created in JanusGraph repository.  If the developers are not committers, a JanusGraph committer will be assigned as the sheperd for that branch.  The sheperd will be responsible for merging pull requests to that branch.
```
mbrukman: s/sheperd/shepherd/g :-)

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+=== Branch Naming Conventions
+
+All branch names shall be prepended with `Issue_#_`.
+
+[[pull-requests]]
+== Pull Requests
+
+Most changes to JanusGraph must be reviewed before they are committed (RTC - review-then-commit).
+This workflow is performed using Github's pull request process. In instances where the committer deems the full process unnecessary, commit-then-review (CTR) may be employed.
```
mbrukman: s/Github/GitHub/g

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+Most changes to JanusGraph must be reviewed before they are committed (RTC). This workflow is performed using Github's
+pull request process. In instances where the committer deems the full process
+unnecessary, commit-then-review (CTR) may be employed. This should be used infrequently and only for extremely trivial
+changes. Changes of this sort may include:
+
+* Documentation typo or small documentation addition
+* Addition of a new test case
+
+Any commit that is made following CTR shall include the fact that it is a CTR in the commit comments.
```
mbrukman: GitHub provides automated notifications: https://help.github.com/articles/managing-notifications-for-pushes-to-a-repository/

For example, we can create a list with the firehose of all commits / merges, e.g., janusgraph-commits@, and folks who want the firehose can subscribe to this. I've seen this done with the commit-then-review workflow in the LLVM project, it worked well.

I can set this up.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+Any JanusGraph committer may propose a release.
+To propose a release, simple start a new RELEASE thread on https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] proposing the new release and requesting feedback on what should be included in the release.
+After consensus is reached the release manager will perform the following tasks:
+
+* Create a release branch so that work may continue on `master`
+* Prepare the release artifacts
+* Call a vote to approve the release on https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev]
+
+Committers will be given 72 hours to review and vote on the release artifacts.
```
mbrukman: This paragraph is written as a bulleted list (each sentence starts on a new line), but that's not how it will be rendered in HTML. If you'd like it to become a bulleted list, please make it so. Otherwise, please reflow the paragraph so that it reads as it will be rendered.

twilmes: Hm, not sure if I understand.  When I build the docs this renders as a bulleted as I expected.

mbrukman: I'm not referring to "Prepare" and "Call", I'm referring to "Committers will be given ..." and following sentences. That also renders as a bulleted list?

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+[[development-decisions]]
+== Development Decisions
+
+Many development decisions will be made during the day-to-day work of JanusGraph contributors without any extra process overhead.
+However, for larger bodies of work and significant updates and additions, the following process shall be followed.
+Significant work may include, but is not limited to:
+
+* A major new feature or subproject eg. a new storage adapter
+* Incrementing the version of a core dependency such as a Apache TinkerPop
+* Addition or deprecation of a public API
```
jerryjch: Bringing in a major dependency.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+Pull requests commit messages should clearly describe what work was accomplished.
+Intermediate work-in-progress commits should be squashed prior to pull request submittal.
+
+Pull requests must be reviewed before being merged.
+The JanusGraph project uses the https://help.github.com/articles/about-pull-request-reviews/[GitHub review process] to review pull requests.
+Non-committers are free, and encouraged to review pull requests.
+Their review will be non-binding, but can be taken into consideration and are still very valuable community input.
+The following rules apply to the voting process.
+
+* Two committer approvals are required merge a pull request
```
amcp: s/required merge/required to merge/

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+
+[[branching]]
+== Branching
+
+For features that involve only one developer, developers will work in their own JanusGraph forks, managing their own branches, and submitting pull requests when ready. If multiple developers wish to collaborate on a feature,
+they may request that a feature branch be created in JanusGraph repository.  If the developers are not committers, a JanusGraph committer will be assigned as the shepherd for that branch.  The shepherd will be responsible for merging pull requests to that branch.
+
+=== Branch Naming Conventions
+
+All branch names shall be prepended with `Issue_#_`.
```
amcp: For example, if there was an issue 114 to add Tupl support to JanusGraph, and people needed to collaborate on this branch, then it could be named `Issue_114_Tupl`. This naming convention only applies to hosted branches created for collaboration, not to branches in forks created for normal pull requests.

docs/development.txt
```diff
@@ -0,0 +1,82 @@
...
+* Documentation typo or small documentation addition
+* Addition of a new test case
+
+Any commit that is made following CTR shall include the fact that it is a CTR in the commit comments.
+The committer should also send a message to https://groups.google.com/forum/#!forum/janusgraph-dev[janusgraph-dev] to notify other project members of the CTR.
+If another committer responds with a -1, the commit should be rolled back and the formal RTC process should be followed.
+
+Users wishing to contribute to JanusGraph should fork the https://github.com/janusgraph/janusgraph[JanusGraph] repository and then submit pull requests using the GitHub pull request process.
+Every pull request must be associated with an existing issue.
+Be sure to include the relevant issue number in the title of your pull request and a description of what test suites were run to validate the pull request.
```
amcp: CI will run tests for you (ideally). Eventually, we should probably ask for contributors to note any special testing steps they took.

docs/development.txt
twilmes: Updated to note that this only applies to the main JanusGraph repo.

docs/development.txt
twilmes: Agreed.  I think we should leave this in for now and then when CI is in good shape, we can update.

docs/development.txt
twilmes: Good catch, fixed.

twilmes:
I made a few updates based upon @amcp feedback.  At this point there has been good review of this so I cleaned up the commits.  Are we ready to merge?

amcp:
I'm ok with the changes, how about @mbrukman ?

twilmes:
@mbrukman you okay with leaving the CTR text as is and then when we get the janusgraph-commits firehose setup, I'll make an update to remove the requirement to notify the dev list?  If so, I think we can go ahead and merge and close this issue out.

mbrukman:
@twilmes — https://groups.google.com/forum/#!forum/janusgraph-commits is now being populated.

docs/development.txt
mbrukman: This is now done; please see https://groups.google.com/forum/#!forum/janusgraph-commits

twilmes:
Thanks for creating the list @mbrukman.  I'll add the link in and get this merged.

PR #145, opened 2017-03-01
Make optimisticLocking a StoreFeature. Silence noisy kcvslog test.

amcp:
These changes are a rework of something I tried to do two years ago but did not make it in Titan before the end of 2015. Contains an extremely minor refactoring of tests so that the DynamoDB Storage Backend can change some parameters. Also promotes optimisticLocking to a StoreFeature, as that is where it belongs.


amcp:
These changes are in support of DynamoDB compatibility [here](https://github.com/amcp/dynamodb-titan-storage-backend/tree/janusgraph).

PR #148, opened 2017-03-03
Enable setting custom vertex ids in TP3 interface #147 

### LINKED ISSUE #147: Enable custom id setting in JanusGraphBlueprintsTransaction
JanusGraphFeatures suggests that the use of custom vertex ids through TinkerPop Structure [depends on the allowVertexIdSetting() value](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphFeatures.java#L179). However the TitanBlueprintsTransaction implementation [forbids custom ids regardless of the allowVertexIdSetting](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphBlueprintsTransaction.java#L102). I then unfortunately had to drill down to [vendor-specific implementation to use custom vertex ids](https://github.com/classmethod/tupl-titan-storage-backend/blob/titan10/src/test/java/jp/classmethod/titan/TuplStorageSetup.java#L100) even though they should be supported, despite [having enabled setting vertex IDs](https://github.com/classmethod/tupl-titan-storage-backend/blob/titan10/src/test/java/jp/classmethod/titan/TuplStorageSetup.java#L124).

amcp:
fixes #147, another bug fix that didn't make it into Titan by the end of 2015

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/idmanagement/IDManager.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/idmanagement/IDManager.java
@@ -567,6 +567,33 @@ public class IDManager {
         return ids;
     }
 
+    /**
+     * Converts a user provided long id into a JanusGraph vertex id. The id must be positive and less than {@link #getVertexCountBound()}.
+     * This method is useful when providing ids during vertex creation via {@link org.apache.tinkerpop.gremlin.structure.Graph#addVertex(Object...)}.
+     *
+     * @param id long id
+     * @return a corresponding JanusGraph vertex id
+     * @see #fromVertexId(long)
+     */
+    public long toVertexId(long id) {
+        Preconditions.checkArgument(id > 0, "Vertex id must be positive: %s", id);
+        Preconditions.checkArgument(vertexCountBound > id, "Vertex id is too large: %s", id);
+        return id<<(partitionBits+USERVERTEX_PADDING_BITWIDTH);
+    }
+
+    /**
+     * Converts a JanusGraph vertex id to the user provided id as the inverse mapping of {@link #toVertexId(long)}.
+     *
+     * @param id JanusGraph vertex id (must be positive)
+     * @return original user provided id
+     * @see #toVertexId(long)
+     */
+    public long fromVertexId(long id) {
+        Preconditions.checkArgument(id >>> USERVERTEX_PADDING_BITWIDTH+partitionBits > 0
+            && id <= (vertexCountBound-1)<<USERVERTEX_PADDING_BITWIDTH+partitionBits, "Invalid vertex id provided: %s", id);
+        return id>>USERVERTEX_PADDING_BITWIDTH+partitionBits;
+    }
+
     public boolean isPartitionedVertex(long id) {
         return isUserVertexId(id) && VertexIDType.PartitionedVertex.is(id);
     }
--- a/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphTransaction.java
@@ -49,9 +49,13 @@ public interface JanusGraphTransaction extends Transaction {
      * Note, that an exception is thrown if the vertex id is not a valid JanusGraph vertex id or if a vertex with the given
      * id already exists.
      * <p/>
-     * Custom id setting must be enabled via the configuration option {@link org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration#ALLOW_SETTING_VERTEX_ID}.
-     * <p/>
-     * Use {@link org.janusgraph.core.util.JanusGraphId#toVertexId(long)} to construct a valid JanusGraph vertex id from a user id.
+     * Custom id setting must be enabled via the configuration option {@link org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration#ALLOW_SETTING_VERTEX_ID}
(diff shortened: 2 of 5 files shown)


sjudeng:
I wonder if this might solve the issue I saw with custom vertex ids in upgrading to TinkerPop 3.2.3 (see notes in #78). To resolve there I updated test GraphComputerImplementations to include configuration around id generation but I wasn't happy with that approach.

amcp:
@sjudeng I've got to put in for the night it being 2AM here, but if this interests you and given the extremely small size of the fix I encourage you to pick up from here and rip out the config workaround you put in GraphComputerImplementations to see if it fixes the issue you encountered.

sjudeng:
I took a look at this branch. Initially I was having test failures of the form `The provided key/value array must have a String or T on even array indices`, which I resolved with the below update.

```patch
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphBlueprintsTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphBlueprintsTransaction.java
@@ -115,7 +115,8 @@ public abstract class JanusGraphBlueprintsTransaction implements JanusGraphTrans
             label = (labelValue instanceof VertexLabel)?(VertexLabel)labelValue:getOrCreateVertexLabel((String) labelValue);
         }
 
-        final JanusGraphVertex vertex = addVertex(ElementHelper.getIdValue(keyValues).orElse(null), label);
+        final Number id = (Number) ElementHelper.getIdValue(keyValues).orElse(null);
+        JanusGraphVertex vertex = addVertex(id != null ? id.longValue() : null, label);
```

I then tried running through tests in the PeerPressureTest suite via InMemoryJanusGraphComputerTest, which use custom vertex ids. But these gave errors at [StandardJanusGraphTx.java#L514](https://github.com/amcp/janusgraph/blob/ac5172d6ace652854263f068105b387eefaf3cad/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java#L514). I experimented with an additional conditional there for `graph.getConfiguration().allowVertexIdSetting() && UserVertex.is(vertexId)` but that also failed (`UserVertex.is(vertexId)` was false). I also tried forcing it past that error but then ended up with an error lower down at [IDManager.java#L474](https://github.com/amcp/janusgraph/blob/ac5172d6ace652854263f068105b387eefaf3cad/janusgraph-core/src/main/java/org/janusgraph/graphdb/idmanagement/IDManager.java#L474).

sjudeng:
@amcp I rebased and added a commit to this PR. The commit includes a fix to JanusGraphBlueprintsTransaction to resolve test failures as described in the comment above. Also during testing I found that the existing documented method for converting user ids to vertex ids, `JanusGraphId#toVertexId(long)`, is no longer valid as it doesn't take into account partitioning bits used in JanusGraph vertex assignment. Attempts to flush vertices with custom ids created using this method can lead to assert/IllegalArgumentException errors. To resolve this issue I added methods to IDManager for converting user vertex ids to/from JanusGraph vertex ids and deprecated JanusGraphId.

All default and TinkerPop tests are passing with these updates. Given my involvement in the implementation I think it would be best to have another independent review.

janusgraph-core/src/main/java/org/janusgraph/graphdb/idmanagement/IDManager.java
```diff
@@ -567,6 +567,33 @@ public long getPartitionedVertexId(long partitionedVertexId, long otherPartition
...
 
+    /**
+     * Converts a user provided long id into a JanusGraph vertex id. The id must be positive and less than {@link #getVertexCountBound()}.
+     * This method is useful when providing ids during vertex creation via {@link org.apache.tinkerpop.gremlin.structure.Graph#addVertex(Object...)}.
+     *
+     * @param id long id
+     * @return a corresponding JanusGraph vertex id
+     * @see #fromVertexId(long)
+     */
+    public long toVertexId(long id) {
```
amcp: can you test this method?

janusgraph-core/src/main/java/org/janusgraph/graphdb/idmanagement/IDManager.java
```diff
@@ -567,6 +567,33 @@ public long getPartitionedVertexId(long partitionedVertexId, long otherPartition
...
+    }
+
+    /**
+     * Converts a JanusGraph vertex id to the user provided id as the inverse mapping of {@link #toVertexId(long)}.
+     *
+     * @param id JanusGraph vertex id (must be positive)
+     * @return original user provided id
+     * @see #toVertexId(long)
+     */
+    public long fromVertexId(long id) {
```
amcp: can you test this method?

amcp:
@sjudeng I agree we need another review.
@analytically the content of this PR changed slightly so can you re-review please?

janusgraph-core/src/main/java/org/janusgraph/graphdb/idmanagement/IDManager.java
sjudeng: @amcp There's a test added as part of the commit (see [VertexIDAssignerTest#testCustomIdAssignment](https://github.com/amcp/janusgraph/blob/fixBlueprintsCustomIds/janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java#L186-L208)).

PR #164, opened 2017-03-15
Fix capitalization of Bigtable in the docs

mbrukman:
Essentially: `s/BigTable/Bigtable/`


mbrukman:
@jerryjch – it's only BigQuery that's CamelCased, everything else is not: Bigtable, Dataflow, Spanner, Datastore, etc. CamelCase is BigQuery's (and MapReduce's, I suppose) fault. ;-)

Issue #35, opened 2017-01-12
New Cassandra backend using CQL w/ prepared statements


rustyrazorblade:
Thrift is deprecated and will be removed in Cassandra 4.0.

For perf reasons, avoid IN() clauses for selects, prefer concurrent async queries.

sharpau:
A request for any new Cassandra backend work: please consider better handling for vertices with large numbers of edges. In particular, although one can partition adjacency lists in Titan (http://s3.thinkaurelius.com/docs/titan/1.0.0/graph-partitioning.html), the results of traversals are still subject to hitting the Cassandra frame size limit, even though each individual partition should not.

Given how well the Tinkerpop API adheres to a lazy, streaming data model, it's frustrating when Titan's backend implementation does not.

dylanht:
@sharpau thank you for the advice. I'm going to self-assign this one, and intend to deprecate thrift support.

sjudeng:
The thread below mentions using thrift to enable support for ScyllaDB. Would this capability be affected by the proposed updates?

https://github.com/scylladb/scylla/issues/693
https://github.com/JanusGraph/janusgraph/issues/53

rustyrazorblade:
@dylanht I can help with the data modeling on this.  I'm one of the first people to put titan + cassandra in production, and seen where it fails (miserably).

dylanht:
@sjudeng I just had a look over ScyllaDB's docs on [Cassandra compatibility](http://docs.scylladb.com/cassandra-compatibility/) and it's my understanding that we can use [CQL Binary Protocol](https://git-wip-us.apache.org/repos/asf?p=cassandra.git;a=blob_plain;f=doc/native_protocol_v3.spec) with ScyllaDB just as well as Thrift (the REGISTER and AUTH_RESPONSE requests, and the AUTHENTICATE response are not supported by Scylla in case that alarms anyone), and reading through the comments in the issue you linked it seems Thrift support in ScyllaDB was actually the initial barrier to integration. Now that they have implemented support for Thrift https://github.com/scylladb/scylla/issues/399 and given the patches @duartan linked recently that apparently fix the OLAP queries @pluradj was running, my impression is that things are "working" on ScyllaDB master via Thrift. Correct me if I'm wrong, but it looks like support for CQL in Titan especially with @rustyrazorblade offering to help us out (thank you Jon!) might immediately provide a more stable/performant integration with ScyllaDB. 

[...]

PR #166, opened 2017-03-15
CQL storage backend

pkendall64:
Adds a CQL stotage backend that is compatible with the table structure of the Astynanx code.
This PR addresses issue #35.

A new library (javaslang) is introduced which provides very nice functional library which makes the code a lot more concise and understandable.


pkendall64:
We'll get onto those changes and update the PR, thanks for the feedback.

mbrukman:
@pkendall64 -- why did you delete the branch and close the PR? You can just update the code on your side and `git push -f` to update the branch and we'll continue the review. No need to open a new PR each time we have a round of reviews.

You can probably re-open this PR (since you created it), recreate the branch and force-push to it via `git push -f` and we can continue where we left off.

pkendall64:
Yeah, thats what I did and it just disappeared!

mbrukman:
@pkendall64 – you pushed "delete branch" button on GitHub, so it deleted the branch in `orionhealth:feature/cql-backend` but presumably you have a local copy of it on your workstation? In which case, you can just push to the same repo / branch to continue.

You can also [fetch PR](http://stackoverflow.com/a/15397064) on the CLI using `git`.

In the worst case scenario, if you've deleted the branch in the `orionhealth` repo as well as locally, you can still attach [`.patch`](https://patch-diff.githubusercontent.com/raw/JanusGraph/janusgraph/pull/166.patch) or [`.diff`](https://patch-diff.githubusercontent.com/raw/JanusGraph/janusgraph/pull/166.diff) to the PR URL to get the full contents — click on either of those links to see what I mean.

pkendall64:
Sorry about that, all fixed now. I had to checkout the old hash and force push that, then force push the new stuff over the top and now it's ok again.
Don't really know what happened the first time!

mbrukman:
Great to hear your PR is back, @pkendall64! :-)

janusgraph-cql/.classpath
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
amcp: is this an eclipse specific file? should we be committing this to the repo, or should people be able to determine their own IDE configurations?

janusgraph-cql/.project
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
amcp: same as above, eclipse specific

janusgraph-cql/pom.xml
```diff
@@ -0,0 +1,342 @@
...
+                                    </action>
+                                </pluginExecution>
+                            </pluginExecutions>
+                        </lifecycleMappingMetadata>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
+</project>
```
amcp: add newline to the end of file

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.cql;
+
+import static com.datastax.driver.core.querybuilder.QueryBuilder.*;
```
amcp: please import specific things and not wildcard

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+    private final PreparedStatement insertColumnWithTTL;
+
+    public CQLKeyColumnValueStore(final CQLStoreManager storeManager, final String tableName, final Configuration configuration, final Runnable closer) {
+        this.storeManager = storeManager;
+        this.closer = closer;
+        this.session = this.storeManager.getSession();
+        this.tableName = tableName;
+        this.getter = new Getter(storeManager.getMetaDataSchema(this.tableName));
+
+        initialiseTable(this.session, this.storeManager.getKeyspaceName(), tableName, configuration);
```
amcp: spelled initialized elsewhere in this project so please match existing spelling

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+    private final Runnable closer;
+
+    private final PreparedStatement getSlice;
+    private final PreparedStatement getKeysAll;
+    private final PreparedStatement getKeysRanged;
+    private final PreparedStatement deleteColumn;
+    private final PreparedStatement insertColumn;
+    private final PreparedStatement insertColumnWithTTL;
+
+    public CQLKeyColumnValueStore(final CQLStoreManager storeManager, final String tableName, final Configuration configuration, final Runnable closer) {
```
amcp: add javadoc

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+        this.getter = new Getter(storeManager.getMetaDataSchema(this.tableName));
+
+        initialiseTable(this.session, this.storeManager.getKeyspaceName(), tableName, configuration);
+
+        // @formatter:off
+        this.getSlice = this.session.prepare(select()
+                .column(COLUMN_COLUMN_NAME)
+                .column(VALUE_COLUMN_NAME)
+                .fcall("writetime", column(VALUE_COLUMN_NAME)).as(WRITETIME_COLUMN_NAME)
+                .fcall("ttl", column(VALUE_COLUMN_NAME)).as(TTL_COLUMN_NAME)
```
amcp: use one of the string constants above

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+        this.tableName = tableName;
+        this.getter = new Getter(storeManager.getMetaDataSchema(this.tableName));
+
+        initialiseTable(this.session, this.storeManager.getKeyspaceName(), tableName, configuration);
+
+        // @formatter:off
+        this.getSlice = this.session.prepare(select()
+                .column(COLUMN_COLUMN_NAME)
+                .column(VALUE_COLUMN_NAME)
+                .fcall("writetime", column(VALUE_COLUMN_NAME)).as(WRITETIME_COLUMN_NAME)
```
amcp: externalize write time constant

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+                .where(eq(KEY_COLUMN_NAME, bindMarker(KEY_BINDING)))
+                .and(gte(COLUMN_COLUMN_NAME, bindMarker(SLICE_START_BINDING)))
+                .and(lt(COLUMN_COLUMN_NAME, bindMarker(SLICE_END_BINDING)))
+                .limit(bindMarker(LIMIT_BINDING)));
+
+        this.getKeysRanged = this.session.prepare(select()
+                .column(KEY_COLUMN_NAME)
+                .column(COLUMN_COLUMN_NAME)
+                .column(VALUE_COLUMN_NAME)
+                .fcall("writetime", column(VALUE_COLUMN_NAME)).as(WRITETIME_COLUMN_NAME)
```
amcp: externalize writetime

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+                .fcall("ttl", column(VALUE_COLUMN_NAME)).as(TTL_COLUMN_NAME)
```
amcp: use string constant above for ttl

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+                .where(gte(token(KEY_COLUMN_NAME), bindMarker(KEY_START_BINDING)))
+                .and(lt(token(KEY_COLUMN_NAME), bindMarker(KEY_END_BINDING)))
+                .and(gte(COLUMN_COLUMN_NAME, bindMarker(SLICE_START_BINDING)))
+                .and(lte(COLUMN_COLUMN_NAME, bindMarker(SLICE_END_BINDING))));
+
+        this.getKeysAll = this.session.prepare(select()
+                .column(KEY_COLUMN_NAME)
+                .column(COLUMN_COLUMN_NAME)
+                .column(VALUE_COLUMN_NAME)
+                .fcall("writetime", column(VALUE_COLUMN_NAME)).as(WRITETIME_COLUMN_NAME)
```
amcp: externalize writetime

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+                .fcall("ttl", column(VALUE_COLUMN_NAME)).as(TTL_COLUMN_NAME)
```
amcp: use string constant above

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+        // the same iterator would be reused and would be exhausted.
+        return StaticArrayEntryList.ofStaticBuffer(() -> Iterator.ofAll(lazyList.get())
+                .<Tuple3<StaticBuffer, StaticBuffer, Row>> map(row -> Tuple.of(
+                        StaticArrayBuffer.of(row.getBytes(COLUMN_COLUMN_NAME)),
+                        StaticArrayBuffer.of(row.getBytes(VALUE_COLUMN_NAME)),
+                        row)),
+                getter);
+    }
+
+    Statement deleteColumn(final StaticBuffer key, final StaticBuffer column, final long timestamp) {
```
amcp: why package private? add a comment, annotate or make private.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+        return this.deleteColumn.bind()
+                .setBytes(KEY_BINDING, key.asByteBuffer())
+                .setBytes(COLUMN_BINDING, column.asByteBuffer())
+                .setLong(TIMESTAMP_BINDING, timestamp);
+    }
+
+    Statement insertColumn(final StaticBuffer key, final Entry entry, final long timestamp) {
```
amcp: why package private? add a comment, annotate or make private.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+                        .setFetchSize(this.storeManager.getPageSize())
+                        .setConsistencyLevel(getTransaction(txh).getReadConsistencyLevel()))))
+                .getOrElseThrow(EXCEPTION_MAPPER);
+    }
+
+    // ------------------------------------------------------------------------
+    // Inner classes
+    // ------------------------------------------------------------------------
+
+    static class ResultSetKeyIterator extends AbstractIterator<StaticBuffer> implements KeyIterator {
```
amcp: externalize this class

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+        public RecordIterator<Entry> getEntries() {
+            return new EntryRecordIterator(this.sliceQuery, this.getter, Iterator.of(this.currentRow).concat(this.iterator), this.currentKey);
+        }
+
+        @Override
+        public void close() throws IOException {
+            // NOP
+        }
+
+        static class EntryRecordIterator extends AbstractIterator<Entry> implements RecordIterator<Entry> {
```
amcp: externalize this class

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+            }
+
+            @Override
+            public void close() throws IOException {
+                // NOP
+            }
+        }
+    }
+
+    static class Getter implements GetColVal<Tuple3<StaticBuffer, StaticBuffer, Row>, StaticBuffer> {
```
amcp: externalize

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.cql;
+
+import static com.datastax.driver.core.querybuilder.QueryBuilder.truncate;
+import static com.datastax.driver.core.schemabuilder.SchemaBuilder.createKeyspace;
+import static javaslang.API.*;
```
amcp: don't use wildcards

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+
+import javaslang.Tuple;
+import javaslang.collection.Array;
+import javaslang.collection.HashMap;
+import javaslang.collection.Iterator;
+import javaslang.collection.Seq;
+import javaslang.concurrent.Future;
+import javaslang.control.Option;
+
+public class CQLStoreManager extends DistributedStoreManager implements KeyColumnValueStoreManager {
```
amcp: add classdoc

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,460 @@
...
+import javaslang.Lazy;
+import javaslang.Tuple;
+import javaslang.Tuple2;
+import javaslang.Tuple3;
+import javaslang.collection.Array;
+import javaslang.collection.Iterator;
+import javaslang.concurrent.Future;
+import javaslang.control.Try;
+
+public class CQLKeyColumnValueStore implements KeyColumnValueStore {
```
amcp: add classdoc

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -0,0 +1,177 @@
...
+package org.janusgraph.diskstorage.cql;
+
+import org.janusgraph.diskstorage.configuration.ConfigElement;
+import org.janusgraph.diskstorage.configuration.ConfigNamespace;
+import org.janusgraph.diskstorage.configuration.ConfigOption;
+import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
+import org.janusgraph.graphdb.configuration.PreInitializeConfigOptions;
+
+@PreInitializeConfigOptions
+public interface CQLConfigOptions {
```
amcp: add classdoc

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+    private static final int DEFAULT_PORT = 9042;
+
+    private final String keyspace;
+    private final boolean atomicBatch;
+    private final Cluster cluster;
+    private final Session session;
+    private final StoreFeatures storeFeatures;
+    private final Map<String, CQLKeyColumnValueStore> openStores;
+
+    public CQLStoreManager(final Configuration configuration) throws BackendException {
```
amcp: add javadoc

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+import javaslang.collection.Array;
+import javaslang.collection.HashMap;
+import javaslang.collection.Iterator;
+import javaslang.collection.Seq;
+import javaslang.concurrent.Future;
+import javaslang.control.Option;
+
+public class CQLStoreManager extends DistributedStoreManager implements KeyColumnValueStoreManager {
+
+    private static final int DEFAULT_PORT = 9042;
```
amcp: doesn't the configuration class above also contain a default port? don't define twice

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+    public CQLStoreManager(final Configuration configuration) throws BackendException {
+        super(configuration, DEFAULT_PORT);
+        this.keyspace = configuration.get(KEYSPACE);
+        this.atomicBatch = configuration.get(ATOMIC_BATCH_MUTATE);
+
+        this.cluster = initialiseCluster();
+        this.session = initialiseSession(this.cluster, this.keyspace);
+
+        final Configuration global = buildGraphConfiguration()
+                .set(READ_CONSISTENCY, "QUORUM")
```
amcp: externalize QUORUM constant

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+                .set(WRITE_CONSISTENCY, "QUORUM")
+                .set(METRICS_PREFIX, METRICS_SYSTEM_PREFIX_DEFAULT);
+
+        final Configuration local = buildGraphConfiguration()
+                .set(READ_CONSISTENCY, "LOCAL_QUORUM")
```
amcp: externalize LOCAL_QUORUM constant

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+            default: {
+                throw new IllegalArgumentException("Unrecognized partitioner: " + partitioner);
+            }
+        }
+        this.storeFeatures = fb.build();
+        this.openStores = new ConcurrentHashMap<>();
+    }
+
+    private Cluster initialiseCluster() throws PermanentBackendException {
+        // XXX: Deal with retry policy and potentially others
```
amcp: XXX?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+        fb.multiQuery(true);
+
+        final String partitioner = this.cluster.getMetadata().getPartitioner();
+        switch (partitioner) {
+            case "org.apache.cassandra.dht.RandomPartitioner":
+            case "org.apache.cassandra.dht.Murmur3Partitioner": {
+                fb.keyOrdered(false).orderedScan(false).unorderedScan(true);
+                break;
+            }
+            case "org.apache.cassandra.dht.ByteOrderedPartitioner": {
```
amcp: create string constants that get the canonical name from the actual classes. that way if version update changes the package of ByteOrderedPartitioner, you can catch it as compile time error here

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+
+            } catch (NoSuchAlgorithmException | CertificateException | IOException | KeyStoreException | KeyManagementException e) {
+                throw new PermanentBackendException("Error initialising SSL connection properties", e);
+            }
+        }
+
+        return builder.build();
+    }
+
+    private Session initialiseSession(@SuppressWarnings("hiding") final Cluster cluster, final String keyspaceName) {
```
amcp: why is hiding the member cluster ok?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,349 @@
...
+                final StaticBuffer key = keyAndMutations.getKey();
+                final KCVMutation keyMutations = keyAndMutations.getValue();
+
+                final Iterator<Statement> deletions = Iterator.of(commitTime.getDeletionTime(this.times))
+                        .flatMap(deleteTime -> Iterator.ofAll(keyMutations.getDeletions()).map(deletion -> columnValueStore.deleteColumn(key, deletion, deleteTime)));
+                final Iterator<Statement> additions = Iterator.of(commitTime.getAdditionTime(this.times))
+                        .flatMap(addTime -> Iterator.ofAll(keyMutations.getAdditions()).map(addition -> columnValueStore.insertColumn(key, addition, addTime)));
+
+                return Iterator.concat(deletions, additions)
+                        .grouped(20) // XXX Pull this out as a configuration option
```
amcp: please pull out the batch size into configuration

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.janusgraph.diskstorage.cql.CQLConfigOptions.*;
+
+import org.janusgraph.diskstorage.BaseTransactionConfig;
+import org.janusgraph.diskstorage.common.AbstractStoreTransaction;
+import org.janusgraph.diskstorage.keycolumnvalue.StoreTransaction;
+
+import com.datastax.driver.core.ConsistencyLevel;
+import com.google.common.base.Preconditions;
+
+public class CQLTransaction extends AbstractStoreTransaction {
```
amcp: add classdoc

janusgraph-cql/.classpath
pkendall64: Sure, I'll remove the eclipse files

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
pkendall64: I'd like to leave this one as an inner class of the Itertor and make it private as it's really an internal implementation detail.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
pkendall64: No, the default port for CQL is not defined anywhere else that I can find.
There is one in the AbstractCassandraStoreManager which defines the Thrift port.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
pkendall64: I don't think you want to have compile time dependencies on these classes as they are internals to cassandra and we really only want this depend on the driver jar and not the cassandra internals.
Perhaps it would be better to get the simplename of the class and switch on that accordingly?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
pkendall64: This is just our eclispe settings. I'll remove the annotation.

boliza:
should the CQL storage Backend just a part of the module `janusgraph-cassandra`, not a single module?

pkendall64:
We have seperated the CQL backend because we don't want to bring in cassandra-all, we only want the driver.

We are also working on a PR that creates a janusgraph-cassandra parent that has embedded, thrift, astyanax and CQL as child projects along with a core project which will contain common configuration and other cassandra helper and abstract classes. There will also be a common test project which will contain common test code.

This should make the while cassandra section a lot cleaner. Our ultimate goal is that you should be able to have maven dependencies on the backends you require (e.g. cassandra-cql and elasticsearch) and have a minimal set of transitive dependencies pulled in to create a runnning instance if JanusGraph.

janusgraph-cql/pom.xml
```diff
@@ -0,0 +1,342 @@
...
+                                <property>
+                                    <name>test.cassandra.datadir</name>
+                                    <value>${project.build.directory}/cassandra/${test.murmur}/data</value>
+                                </property>
+                            </systemProperties>
+                            <skip>${test.skip.murmur-serial}</skip>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>${test.murmur-ssl}-test</id>
```
sjudeng: Same as above ... this test execution is run by default in janusgraph-cql but not under janusgraph-cassandra

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLColValGetter.java
```diff
@@ -0,0 +1,45 @@
+package org.janusgraph.diskstorage.cql;
```
sjudeng: Add license header

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -0,0 +1,190 @@
...
+            "True to use Cassandra atomic batch mutation, false to use non-atomic batches",
+            ConfigOption.Type.MASKABLE,
+            false);
+
+    // Replication
+    public static final ConfigOption<Integer> REPLICATION_FACTOR = new ConfigOption<>(
+            CQL_NS,
+            "replication-factor",
+            "The number of data replicas (including the original copy) that should be kept. " +
+                    "This is only meaningful for storage backends that natively support data replication.",
```
sjudeng: The second sentence doesn't seem to make sense in context of specific backend configuration item

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,385 @@
...
+import javaslang.Tuple;
+import javaslang.Tuple2;
+import javaslang.Tuple3;
+import javaslang.collection.Array;
+import javaslang.collection.Iterator;
+import javaslang.concurrent.Future;
+import javaslang.control.Try;
+
+/**
+ * An implementation of {see KeyColumnValueStore} which stores the data in a CQL
```
sjudeng: Change `see` to `@link` (or `@see`)

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -0,0 +1,190 @@
...
+ */
+@PreInitializeConfigOptions
+public interface CQLConfigOptions {
+
+    public static final ConfigNamespace CQL_NS = new ConfigNamespace(
+            GraphDatabaseConfiguration.STORAGE_NS,
+            "cql",
+            "CQL storage backend options");
+
+    public static final ConfigOption<String> KEYSPACE = new ConfigOption<>(
```
sjudeng: Generated configuration reference documentation page does not appear to include a section for `storage.cql` with these options. It should be automatically generated I think (see [configref.txt#L6](https://github.com/JanusGraph/janusgraph/blob/8914caed850b9d641c8e5b4e15c20f839f6d076b/docs/configref.txt#L6)).

sjudeng:
Great contribution. Some general questions to go with the more minor feedback above:

Documentation: Can you look at whether documentation (specifically [docs/cassandra.txt](https://github.com/JanusGraph/janusgraph/blob/8914caed850b9d641c8e5b4e15c20f839f6d076b/docs/cassandra.txt)) can be updated to include information on the new cql implementation? It might also make sense to include an example config in [docs/basics.txt](https://github.com/JanusGraph/janusgraph/blob/8914caed850b9d641c8e5b4e15c20f839f6d076b/docs/basics.txt#L14).

Testing: I was expecting to find cql test suite variants corresponding to all thrift test suites currently in janusgraph-cassandra. But only 6/20 of these are implemented. The following 14 thrift test suites don't have a corresponding cql test suite.

```java
org.janusgraph.blueprints.thrift.process.ThriftComputerTest
org.janusgraph.blueprints.thrift.process.ThriftProcessTest
org.janusgraph.blueprints.thrift.structure.ThriftStructureTest
org.janusgraph.diskstorage.cassandra.thrift.ThriftDistributedStoreManagerTest-
org.janusgraph.graphdb.thrift.ThriftEventualGraphTest
org.janusgraph.graphdb.thrift.ThriftGraphCacheTest
org.janusgraph.graphdb.thrift.ThriftGraphConcurrentTest
org.janusgraph.graphdb.thrift.ThriftGraphIterativeTest
org.janusgraph.graphdb.thrift.ThriftGraphPerformanceMemoryTest
org.janusgraph.graphdb.thrift.ThriftGraphSpeedTest
org.janusgraph.graphdb.thrift.ThriftGraphTest
org.janusgraph.graphdb.thrift.ThriftOLAPTest
org.janusgraph.graphdb.thrift.ThriftOperationCountingTest
org.janusgraph.graphdb.thrift.ThriftPartitionGraphTest
```

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
samant-orionhealth: These methods are used by CQLStoreManager in mutateMany. Added comments to document that.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLTransaction.java
samant-orionhealth: Added classdoc.

janusgraph-all/pom.xml
```diff
@@ -62,6 +67,16 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>org.ow2.asm</groupId>
+                    <artifactId>asm</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.ow2.asm</groupId>
+                    <artifactId>asm-commons</artifactId>
+                </exclusion>
+            </exclusions>
```
samant-orionhealth: This is to resolve a dependency conflict between `org.elasticsearch:elasticsearch` and ``com.datastax.cassandra:cassandra-driver`. Both of these depend on `org.ow2.asm:asm`.
Elasticsearch depends on version 4.1 whereas the Cassandra driver depends on 5.0.3.
I didn't feel that dependency managing it in the parent pom was the right solution as this issue only comes up when both janusgraph-es and janusgraph-cql are included as dependencies.
The cleanest resolution in my opinion is to select the more recent version by excluding the version from elasticsearch.
Another alternative would be to explicitly depend in `asm` at the correct version but we'd have to manage that version manually.

janusgraph-cql/pom.xml
```diff
@@ -0,0 +1,342 @@
...
+            <groupId>org.janusgraph</groupId>
+            <artifactId>janusgraph-test</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
```
samant-orionhealth: This is was an oversight. Updated the pom.

janusgraph-cql/pom.xml
```diff
@@ -0,0 +1,342 @@
...
+                                    <action>
+                                        <ignore></ignore>
+                                    </action>
+                                </pluginExecution>
+                            </pluginExecutions>
+                        </lifecycleMappingMetadata>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
```
samant-orionhealth: Removed

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLColValGetter.java
samant-orionhealth: Added

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
samant-orionhealth: Added `org.janusgraph.diskstorage.cql.CQLConfigOptions` to `org.janusgraph.core.util.ReflectiveConfigOptionLoader` so that it will find the CQL namespaced config options.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
samant-orionhealth: Removed second sentence.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
samant-orionhealth: Changed to `@link`

janusgraph-cql/pom.xml
samant-orionhealth: Regarding the serial and SSL tests, it appears to me that they both run by default in janusgraph-cassandra unless the appropriate skip property is defined.

I believe the janusgraph-cql module has the same behaviour but happy to change if this isn't correct.

janusgraph-all/pom.xml
sjudeng: Thanks for the explanation. It looks like asm is no longer a dependency in ES so we'll be able to remove this when that's updated.

janusgraph-cql/pom.xml
sjudeng: Ah sorry about that. I was comparing to a cassandra log I had on hand but now I see in that case there was a failure in the bop-test execution so the others didn't run.

samant-orionhealth:
I've added tests as per @sjudeng's suggestion.

Would appreciate some input in terms of what would be useful documentation for the CQL backend. I believe the configuration options are generally the same as the existing C* ones and have the same semantics.

sjudeng:
Thanks for getting those tests in! Would you run the full TinkerPop suite on the module (`mvn clean install -Dtest.skip.tp=false -pl janusgraph-cql`) and post the resulting build summary (e.g. "BUILD SUCCESS") block?

Regarding what (if any) documentation updates make sense for the [Cassandra page](http://docs.janusgraph.org/0.1.0-SNAPSHOT/cassandra.html), I didn't have anything specific in mind. I just wanted to make sure the page has been reviewed in the context of the new CQL support in case any changes were needed.

samant-orionhealth:
I ran the tests for 22hours but unfortunately they crashed when my machine ran out of file handles.
I've attached the test output as well as the test log and the final console output.

[MavenOutput.txt](https://github.com/JanusGraph/janusgraph/files/874301/MavenOutput.txt)
[CQL-Test-Output.zip](https://github.com/JanusGraph/janusgraph/files/874300/CQL-Test-Output.zip)

sjudeng:
The 20+ hour runtime is not surprising as it's similar for existing HBase and Cassandra modules. But I think the test errors need to be resolved.

samant-orionhealth:
@sjudeng We're running a suite again on a test server but in the meantime we decided to debug some of the failures and discovered that at least in one case the test doesn't appear to pass for the `CassandraThriftStoreManager` implementation either.

The test in question is `BulkDumperVertexProgramTest.shouldDumpWholeGraph`. We ran it against both the CQL and Thrift implementations and it failed at line 51 in both cases.

It appears that the calls to `StandardJanusGraphTx.addVertex(Long vertexId, VertexLabel label)` always pass null as the vertexId argument.

Does this test pass successfully for anyone?

sjudeng:
Sorry I forgot to point this out earlier ... BulkDumperVertexProgramTest and a couple other test classes require a TinkerPop commit that didn't make it into v3.2.3. You'll need to do a local TinkerPop build with that commit manually applied in order for those tests to pass (see [this post](https://gitter.im/janusgraph/janusgraph?at=58cf09c4ac314c8a1a68024b)).

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
amcp: yes switch on simplename please

janusgraph-cql/src/test/java/org/janusgraph/blueprints/cql/structure/CQLStructureTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.tinkerpop.gremlin.structure.StructureStandardSuite;
+import org.janusgraph.blueprints.cql.CQLGraphProvider;
+import org.janusgraph.core.JanusGraph;
+import org.junit.runner.RunWith;
+
+@RunWith(StructureStandardSuite.class)
+@GraphProviderClass(provider = CQLGraphProvider.class, graph = JanusGraph.class)
+public class CQLStructureTest {
+
+//    TP3 ignores @BeforeClass -- the following method is never executed
```
amcp: maybe leave just the comment that TP3 ignores @BeforeClass and get rid of the commented out method

janusgraph-cql/src/test/java/org/janusgraph/blueprints/cql/process/CQLComputerTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.janusgraph.blueprints.cql.CQLGraphComputerProvider;
+import org.janusgraph.core.JanusGraph;
+import org.junit.runner.RunWith;
+
+@RunWith(ProcessComputerSuite.class)
+@GraphProviderClass(provider = CQLGraphComputerProvider.class, graph = JanusGraph.class)
+public class CQLComputerTest {
+
+//    TP3 ignores @BeforeClass -- the following method is never executed
+//    @BeforeClass
```
amcp: delete 28-31

janusgraph-cql/src/test/java/org/janusgraph/blueprints/cql/process/CQLProcessTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.janusgraph.blueprints.cql.CQLGraphProvider;
+import org.janusgraph.core.JanusGraph;
+import org.junit.runner.RunWith;
+
+@RunWith(ProcessStandardSuite.class)
+@GraphProviderClass(provider = CQLGraphProvider.class, graph = JanusGraph.class)
+public class CQLProcessTest {
+
+//    TP3 ignores @BeforeClass -- the following method is never executed
+//    @BeforeClass
```
amcp: delete 28-31

LeeWong:
Where this pull request stands now? Thanks.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -0,0 +1,411 @@
...
+                        .setBytes(SLICE_END_BINDING, query.getSliceEnd().asByteBuffer())
+                        .setInt(LIMIT_BINDING, query.getLimit())
+                        .setConsistencyLevel(getTransaction(txh).getReadConsistencyLevel())))
+                .map(resultSet -> fromResultSet(resultSet, this.getter));
+        awaitInterruptibly(result);
+        return result.getValue().get().getOrElseThrow(EXCEPTION_MAPPER);
+    }
+
+    @Override
+    public Map<StaticBuffer, EntryList> getSlice(final List<StaticBuffer> keys, final SliceQuery query, final StoreTransaction txh) throws BackendException {
```
countrodrigo: BackendTransaction.java already knows how to implement a parallelized multi query inside the method [BackendTransaction::edgeStoreMultiQuery()](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java#L296). The effective implementation is the same as that of this method. From what I can see, a store should only advertise that it supports the feature `multiQuery` if the underlying backend truly supports multi queries, and the store invokes that functionality. In this case you are just implementing the parallelization manually at the CQLKeyColumnValueStore level.

It seems like we should just advertise that the CQL store does not support multiQuery and let BackendTransaction handle the multi query parallelization.

sjudeng:
@samant-orionhealth, @pkendall64 - I'm seeing a number of failures in the added TinkerPop tests. Is this something you're still looking at?

pkendall64:
@sjudeng we are still maintaining the branch and only get one failure when running the tests and it intermittent and seems related to the way that the cassandra store is created for each test run. We are looking at an alternative way of starting cassandra for the tests so that it is created one for the entire set.

BTW, did you know that in AbstractJanusGraphComputerProvider#48 causes each id generation to take 300ms minimum and thats why the tests take 22hours to run. We removed this line (perhaps that entire overridden method could be removed) the test time went down to 7.5 hours, and the tests passed as before.

amcp:
@pkendall64 wow great find. Do you mind if I pull the id generation time fix into a different PR?

sjudeng:
@pkendall64 That's fantastic regarding the runtime reduction! And I have to admit it's totally my fault ... I added that override in early in the TinkerPop 3.2.3 update work on Titan to resolve failures in `PeerPressureTest#g_V_peerPressure_byXclusterX_byXoutEXknowsXX_pageRankX1X_byXrankX_byXoutEXknowsXX_timesX2X_group_byXclusterX_byXrank_sumX` and `UnionTest#g_VX1_2X_localXunionXoutE_count__inE_count__outE_weight_sumX` (see [PR notes](UnionTest#g_VX1_2X_localXunionXoutE_count__inE_count__outE_weight_sumX) for more info). At the time I hadn't run TinkerPop tests before so I thought the timing was normal. If these tests are still a problem with this update I think there should be good agreement we should instead just `OptOpt`.

Regarding janusgraph-cql tests here, when I run the suite (`mvn clean install -DskipTests && mvn clean install -pl janusgraph-cql -D-Dtest.skip.tp=false`) I see a large number of errors that suggest tables/keyspace haven't been cleared (truncated) properly. What architecture are you testing under? I run tests in a CentOS7 Docker container using OpenJDK8.

pkendall64:
@amcp thats not part of our PR, it's just something we noted while we were
testing. We did some performance testing of the unit tests and that was one
of the things we noticed. So if you want to make that change as a seperate
PR then it's not going to affect our one :-)

sjudeng:
A little more info on the issue running TinkerPop tests on the cql module ... I think the root cause might be resource/thread cleanup as it's hitting OS limits on max number of processes casuing `java.lang.OutOfMemoryError: unable to create new native thread`. For example when running  CQLComputerTest the thread count (`ps huH -p pid | wc -l`) increases steadily to over 10000 in under and hour. By comparison thread count is steady around 100 when running ThriftComputerTest.

I tested under several x64 Linux distributions (CentOS, Ubuntu, Arch) in virtualized (EC2, VBox) and containerized environments with both OpenJDK and Oracle JDK.

pluradj:
Good find @sjudeng. It would be great to add some of the production settings for Cassandra (and the other backends) to get JanusGraph users started, then refer back to the docs of those packages for more details. That gives our users some upfront guidance without having to leave our docs. Also protects our users if those docs suddenly disappear.

amcp:
You could also rebase.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,395 @@
...
+                .map(tableMetadata -> tableMetadata.getOptions().getCompression())
+                .getOrElseThrow(() -> new PermanentBackendException(String.format("Unknown table '%s'", name)));
+    }
+
+    @Override
+    public void close() throws BackendException {
+        try {
+            this.session.close();
+        } finally {
+            this.cluster.close();
```
sjudeng: Add `this.executorService.shutdownNow();`

sjudeng:
I think the threading issue can be resolved by shutting down the thread pool (`this.executorService.shutdownNow()`) as part of `CQLStoreManager.close()`. With this update the thread count now appears to be stable when running the tests.

janusgraph-all/pom.xml
amcp: Please put a TODO comment in the POM mentioning that these deps will change in the future and may require deletion.

amcp:
Implemented block size fix here: https://github.com/JanusGraph/janusgraph/pull/244

janusgraph-all/pom.xml
sjudeng: Once this is rebased you might be able to remove this as ES has been updated on master.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,395 @@
...
+    @Override
+    public void clearStorage() throws BackendException {
+        final Future<Seq<ResultSet>> result = Future.sequence(
+                this.executorService,
+                Iterator.ofAll(this.cluster.getMetadata().getKeyspace(this.keyspace).getTables())
+                        .map(table -> Future.fromJavaFuture(this.executorService, this.session.executeAsync(truncate(this.keyspace, table.getName())))));
+        result.await();
+        if (result.isFailure()) {
+            throw EXCEPTION_MAPPER.apply(result.getCause().get());
+        }
```
sjudeng: Drop keyspace(`this.session.execute(dropKeyspace(this.keyspace));`)?

sjudeng:
As mentioned earlier I was seeing some test errors that looked to be associated with stale keyspaces. Digging in a little more on this I found the issue was when running multiple test suites (e.g. `CQLComputerTest` and `CQLProcessTest`), in which case the first suite runs without issue (with the above thread cleanup fix) but the second suite throws errors of the following form.

```
2017-04-28 13:13:37.801 [    MigrationStage:1] ERROR sandra.service.CassandraDaemon: Exception in thread Thread[MigrationStage:1,5,main]
java.lang.AssertionError
        at org.apache.cassandra.db.DefsTables.updateKeyspace(DefsTables.java:406)
        at org.apache.cassandra.db.DefsTables.mergeKeyspaces(DefsTables.java:242)
        at org.apache.cassandra.db.DefsTables.mergeSchemaInternal(DefsTables.java:193)
        at org.apache.cassandra.db.DefsTables.mergeSchema(DefsTables.java:166)
 ```

I found that this could be resolved by dropping the keyspace (`this.session.execute(dropKeyspace(this.keyspace))`) as part of `CQLStoreManager#clearStorage()`, as shown above. Not sure if this is recommended or there's a better way to address the issue.

**Comment by sjudeng** [comment]
With the above threading and keyspace updates all TinkerPop test suites in the cql module (CQLComputerTest, CQLProcessTest, CQLStructureTest) are passing for me. I am getting 3-4 errors in default tests. I thought the default tests were all passing at one point so maybe something picked up in recent rebasing is related? Three errors can be reproduced in isolation as follows.

```bash
mvn clean install -pl janusgraph-cql -Dtest=CQLDistributedStoreManagerTest#testGetDeployment
mvn clean install -pl janusgraph-cql -Dtest=CQLGraphTest#testTitanGraphBackwardCompatibility
mvn clean install -pl janusgraph-cql -Dtest=CQLGraphTest#testConsistencyEnforcement
```

The corresponding errors are summarized below.

```
testGetDeployment(org.janusgraph.diskstorage.cql.CQLDistributedStoreManagerTest)  Time elapsed: 2.985 sec  <<< FAILURE!
java.lang.AssertionError: expected:<LOCAL> but was:<REMOTE>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at org.janusgraph.diskstorage.DistributedStoreManagerTest.testGetDeployment(DistributedStoreManagerTest.java:39)
        ...

testTitanGraphBackwardCompatibility(org.janusgraph.graphdb.cql.CQLGraphTest)  Time elapsed: 11.377 sec  <<< ERROR!
java.lang.IllegalStateException: Need to set configuration value: root.graph.titan-version
        at com.google.common.base.Preconditions.checkState(Preconditions.java:197)
        at org.janusgraph.diskstorage.configuration.ConfigOption.get(ConfigOption.java:230)
        at org.janusgraph.diskstorage.configuration.BasicConfiguration.get(BasicConfiguration.java:69)
        at org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.<init>(GraphDatabaseConfiguration.java:1417)
        at org.janusgraph.core.JanusGraphFactory.open(JanusGraphFactory.java:107)
        ...

testConsistencyEnforcement(org.janusgraph.graphdb.cql.CQLGraphTest)  Time elapsed: 17.012 sec  <<< FAILURE!
java.lang.AssertionError: Storage backend does not abort conflicting transactions
        at org.junit.Assert.fail(Assert.java:88)
        at org.janusgraph.graphdb.JanusGraphTest.executeLockConflictingTransactionJobs(JanusGraphTest.java:2415)
        at org.janusgraph.graphdb.JanusGraphTest.testConsistencyEnforcement(JanusGraphTest.java:2326)
        ...
```

I also observed the below error in a full run but have not been able to reproduce it in isolation.
```
testVertexTTLImplicitKey(org.janusgraph.graphdb.cql.CQLGraphCacheTest)  Time elapsed: 15.987 sec  <<< FAILURE!
java.lang.AssertionError: expected:<PT1S> but was:<PT0S>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at org.janusgraph.graphdb.JanusGraphTest.testVertexTTLImplicitKey(JanusGraphTest.java:5137)
```

pkendall64:
We will get onto these latest comments this week and update the PR after
rebasing. Thanks for the review and comments.

sjudeng:
Thanks to you both for all your work on the PR in the first place! Looking forward to seeing it merged.

countrodrigo:
I second @sjudeng 's comment. This is a great change that's been long time coming.

pkendall64:
We've fix those failing tests. Also, when looking if we needed to implement `CQLStoreManager#getLocalKeyPartition`, we found that the CQL driver returns tokens of zero bytes long, which does not meet the requirements for this method, so we don't set the localKeyPartition flag in the features. This then meant that to fix that test we had to modify the `DistributedStoreManagerTest` to be ignored if that feature is not set.

Doing part of that investigation, we found a bug in the logic of the `CassandraHelper#transformRange` method where it adds one to the byte arrays. We have opened PR #257 for this one.

janusgraph-all/pom.xml
sjudeng: @samant-orionhealth Now that you've rebased and picked up the more recent ES I think you can remove these exclusions.

samant-orionhealth: @sjudeng I've removed the exclusions now as ES no longer has ASM dependencies.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -0,0 +1,421 @@
...
```
sjudeng: @pkendall64 `dropKeyspace` was present as part of commit 7fa2d8e but it looks like it was taken out again in the last round of commits. This causes errors as mentioned previously.

sjudeng:
The good news is using earlier commit 7fa2d8e I ran through all TinkerPop and default tests twice and everything passed both times (#258 did come up in one run but that's a separate issue). But in the last round of commits since then I'm seeing failures again I think because `dropKeyspace` was removed from clearStorage.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
pkendall64: Yeah, I've done that because dropKeyspace can hang according to reports around the cassandra forums. It's also more inline with what the other cassandra backends do as well. We are trying to fix the intermittent failures due to connection startup in startEmbedded and this is one one of those things that makes it go slow and cause some of the failures.

sjudeng: Since with dropKeyspace all tests pass consistently what about adding it back for now and creating a separate issue for updating to start a clean embedded instance for each test class and/or finding a dropKeyspace alternative so we can break that work off from this PR and move forward here?

The dropKeyspace usage doesn't seem to have a problem with hanging and it's called multiple times for each of over 2000 tests in the TinkerPop suite alone, so it seems like if it was going to hang it would at some point there. Also I haven't observed the intermittent failures due to connection startup (beyond the keyspace issue) that you mention. Can you elaborate on what you're seeing there?

pkendall64:
I'll change it back to dropKeyspace, but I've run the full test suite a couple of times without issue here as well at commit 6b71057, so it seems odd to me that you have failures the way it is now!

sjudeng:
Thanks for adding back `dropKeyspace`. I ran through default and TinkerPop tests again with this and everything passed.

The test issues when `truncate` is used instead of `dropKeyspace` are reproducible for me in the below test environments when running all tests (`mvn clean install -DskipTests && mvn clean install -Dtest.skip.tp=false -pl janusgraph-cql -fn`).
* Docker CentOS 7: [Dockerfile](https://gist.github.com/sjudeng/0c6be52c4a6be5dade6c403a40537e2a)
* EC2 CentOS 7: [CloudFormation template](https://gist.github.com/sjudeng/32821da2ec85fd36de51bf237b5a258a)

I suspect a timing issue associated with the truncate operations. If necessary please create a separate issue to track migration to an alternative approach (clean embedded instance for each test suite, etc.).

pkendall64:
Ok, thanks for the thorough review! We'll look into squashing where
appropriate to minimise it to the actual functional commits.

pkendall64:
I have squashed a lot of the commits so it shows the initial implementation included a lot of the fixes we applied along the way, and then the reviews and extra functionality. If there's any other squashing or rearranging, please let us know and we'll get them done.

sjudeng:
Can you squash 4c5672f, 3b00163, 7558097 and 2c11799 into a single commit, and 8415550, fc6fbf4 and 4850899 into a single commit. I think those go together well and the end result would be 7 commits, which seems reasonable to me.

pkendall64:
Squashed commits as suggested

sjudeng:
@mbrukman Has all your feedback been resolved here?

@hsaputra,@fppt,@dylanht Are there any objections to merging this PR (pending thumbs up from @mbrukman)? I'm not a Cassandra user myself (@amcp not sure if you are) so current reviews might not have covered all aspects of the contribution. But that said there has been a lot of community interest on this and I think once merged these users will be able to try it out and provide additional feedback that can be tracked in subsequent issues/PRs.

hsaputra:
@sjudeng The code in general looks good, I didn't have check for correctness on runtime since I am not Cassandra user. But, one main comment is the dependency on `io.javaslang` which now changed to `io.vavr`. I am ok to change it later but better be the next PR submission bc there is a good reason why they change from javaslang to vavr =P

sjudeng:
@hsaputra Good call. Looks like the first version under io.vavr just same out yesterday so ... way to be on top of it!

@pkendall64, @samant-orionhealth I added #271 to track work to make this dependency update. Hopefully you'll be able to help out with this work after this PR is merged.

pkendall64:
Sure, we can start on that now in preparation for the merge.

sjudeng:
@pkendall64 Thanks for jumping on that. Do hold it for another PR though as I think we're about ready to go on this and I wouldn't want to hold it up to wait for another set of reviews on the new update.

sjudeng:
@mbrukman It looks like your feedback has been addressed here. Can you confirm?

janusgraph-cql/pom.xml
```diff
@@ -0,0 +1,314 @@
...
+            <groupId>com.datastax.cassandra</groupId>
+            <artifactId>cassandra-driver-core</artifactId>
+            <version>${cassandra-driver.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.javaslang</groupId>
```
pkendall64: I've updated ticked #271 with a link to a branch in our repo with this migration. I'll create a PR once this is merged.

sjudeng:
@pkendall64 A minor conflict in .travis.yml was created with merge of separate PR. I resolved it but it created another commit. I think it's fine but you're also welcome to rebase/squash to remove the merge commit if you'd like. Separately we'll need to look at getting the module covered better under Travis.

@mbrukman Unless you have any objections I'd like to move forward with the merge here. We have four approvals and I've confirmed that your feedback has been addressed.

pkendall64:
@sjudeng I'm ok with the merge commit. Thanks for getting this merged and the thorough reviews from everyone involved.

PR #168, opened 2017-03-21
Fixed @return comments to remove javadoc warnings

dbolcioni:
Signed-off-by: Davide Bolcioni <dbolcion@uber.com>

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/MetaAnnotated.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/MetaAnnotated.java
@@ -22,14 +22,12 @@ import java.util.Map;
 public interface MetaAnnotated {
 
     /**
-     * Returns true if this entry has associated meta data
-     * @return
+     * @return true if this entry has associated meta data
      */
     public boolean hasMetaData();
 
     /**
-     * Returns all meta data associated with this entry
-     * @return
+     * @return all meta data associated with this entry
      */
     public Map<EntryMetaData,Object> getMetaData();
 


mbrukman:
@dbolcioni -- the change looks good, but could you please change your commit description to be more meaningful? E.g., "fixed `@return` comments to remove warnings" -- exercising GitHub workflow is nice, but won't be meaningful in history when someone is reviewing past commits to JanusGraph.

mbrukman:
You can do this easily via:

```bash
$ git commit --amend

  # change the description, save and exit your editor

$ git push -f
```

sjudeng:
I would have thought for validation you need to have both the method description and the description of return argument. Doesn't this change cause a new warning about a missing method description?

dbolcioni:
@sjudeng I tried building both with and without this diff, without the diff I get:
[WARNING] /home/uber/usecurity/janusgraph/janusgraph-core/src/main/java/org/janusgraph/diskstorage/MetaAnnotated.java:28: warning - @return tag has no arguments.
[WARNING] /home/uber/usecurity/janusgraph/janusgraph-core/src/main/java/org/janusgraph/diskstorage/MetaAnnotated.java:34: warning - @return tag has no arguments.

while with the diff I get no warning about MetaAnnotated.java.

mbrukman:
@sjudeng – will wait for your review/approval before merging.

sjudeng:
@dbolcioni, @mbrukman Thanks I'm good with this. I had to look it up: From http://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html#JSSOR645, "comments can have a tag section and no main description".

mbrukman:
@sjudeng – thanks for checking.

@dbolcioni – thanks again for the contribution!

PR #170, opened 2017-03-24
Add signatures for value-based type defaults

krlohnes:
The previous implementation only provided property key names to
DefaultSchemaMaker instances, which limits users' ability to customize
behavior. This changeset also passes the property key value down to the
DefaultSchemaMaker, allowing more fine-grained analysis and potentially
better decisions around default types. A base implementation has been
added to the JanusGraphDefaultSchemaMaker.

Signed-off-by: Keith Lohnes <krlohnes@us.ibm.com>

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/schema/DefaultSchemaMaker.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/schema/DefaultSchemaMaker.java
@@ -62,6 +62,18 @@ public interface DefaultSchemaMaker {
     }
 
     /**
+     * Creates a new property key with default settings against the provided {@link PropertyKeyMaker} and value.
+     *
+     * @param factory PropertyKeyMaker through which the property key is created
+     * @param value the value of the property. The default implementation does not use this parameter.
+     * @return A new property key
+     * @throws IllegalArgumentException if the name is already in use or if other configured values are invalid.
+     */
+    public default PropertyKey makePropertyKey(PropertyKeyMaker factory, Object value) {
+         return makePropertyKey(factory);
+    }
+
+    /**
      * Creates a new vertex label with the default settings against the provided {@link VertexLabelMaker}.
      *
      * @param factory VertexLabelMaker through which the vertex label is created
--- a/janusgraph-core/src/main/java/org/janusgraph/core/schema/SchemaInspector.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/schema/SchemaInspector.java
@@ -67,6 +67,25 @@ public interface SchemaInspector {
      */
     public PropertyKey getOrCreatePropertyKey(String name);
 
+
+    /**
+     * Returns the property key with the given name. If automatic type making is enabled, it will make the property key
+     * using the configured default type maker if a key with the given name does not exist.
+     *
+     * The default implementation simply calls the {@link #getOrCreatePropertyKey(String name) getOrCreatePropertyKey} method
+     *
+     * @param name name of the property key to return
+     * @param value the value of the property key. This param is not used by the default
+     * implementaion
+     * @return the property key with the given name
+     * @throws IllegalArgumentException if a property key with the given name does not exist or if the
+     *                                  type with the given name is not a property key
+     * @see PropertyKey
+     */
+    public default PropertyKey getOrCreatePropertyKey(String name, Object value) {
+       return getOrCreatePropertyKey(name);
+    }
+
     /**
      * Returns the property key with the given name. If it does not exist, NULL is returned
      *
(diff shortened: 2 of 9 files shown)


janusgraph-core/src/main/java/org/janusgraph/core/schema/DefaultSchemaMaker.java
```diff
@@ -62,6 +62,18 @@ public default PropertyKey makePropertyKey(PropertyKeyMaker factory) {
     }
 
     /**
+     * Creates a new property key with default settings against the provided {@link PropertyKeyMaker} and value.
```
pluradj: Should mention here that the default implementation does not utilize the `Object value`.

janusgraph-core/src/main/java/org/janusgraph/core/schema/SchemaInspector.java
```diff
@@ -67,6 +67,23 @@
...
 
+
+    /**
+     * Returns the property key with the given name. If automatic type making is enabled, it will make the property key
+     * using the configured default type maker if a key with the given name does not exist.
+     *
+     * The default implementation simply calls the {@link #getOrCreatePropertyKey(String name) getOrCreatePropertyKey} method
+     *
+     * @param name name of the property key to return
+     * @return the property key with the given name
```
pluradj: Missing `@param value` description

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphDefaultSchemaMaker.java
```diff
@@ -34,6 +40,39 @@ public Cardinality defaultPropertyCardinality(String key) {
...
+        } else if (value instanceof Double) {
+            klass = Double.class;
+        } else if (value instanceof Date) {
+            klass = Date.class;
+        } else if (value instanceof Geoshape) {
+            klass = Geoshape.class;
+        } else if (value instanceof UUID) {
+            klass = UUID.class;
+        } else {
+            klass = Object.class;
```
pluradj: Wondering if it makes sense to extract this class if/else block into a separate protected method. It would be a lot of logic to copy/paste if the implementation wants to extend with a new `instanceof` check.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphDefaultSchemaMaker.java
```diff
@@ -34,6 +40,39 @@ public Cardinality defaultPropertyCardinality(String key) {
...
+        } else if (value instanceof Short) {
+            klass = Short.class;
+        } else if (value instanceof Integer) {
+            klass = Integer.class;
+        } else if (value instanceof Long) {
+            klass = Long.class;
+        } else if (value instanceof Float) {
+            klass = Float.class;
+        } else if (value instanceof Double) {
+            klass = Double.class;
```
pluradj: The supported data types in [Table 5.1](http://docs.janusgraph.org/0.1.0-SNAPSHOT/schema.html#_defining_property_keys) lists also `Decimal` and `Precision`. Here's the set from [`StandardSerializer.java`](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java#L76-L92)

It doesn't look like those 2 are valid anymore, so please remove those from the doc.

pluradj:
@sjudeng @twilmes what do you think about including this for `jg01`?

sjudeng:
@pluradj I think it could be merged into `jg01` but I'd recommend waiting until after the 0.1.0 release. So it would be in 0.1.1 (and master) but not 0.1.0. At this point `jg01` has been fully tested including core and TinkerPop tests so I'm hoping we can have a code freeze on it leading up to the 0.1.0 release.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphDefaultSchemaMaker.java
krlohnes: I'll fix this in another commit. There's _quite_ a few places that mention `Precision` and `Decimal`. They've been out of Titan for a while though.

janusgraph-core/src/main/java/org/janusgraph/core/schema/DefaultSchemaMaker.java
krlohnes: fixed

janusgraph-core/src/main/java/org/janusgraph/core/schema/SchemaInspector.java
krlohnes: fixed

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphDefaultSchemaMaker.java
krlohnes: Fixed

krlohnes:
@pluradj @sjudeng I added the changes you both requested.

sjudeng:
Can you squash this into a single commit for merge?

krlohnes:
@sjudeng Squashed.

Issue #102, opened 2017-02-10
Update the Version Compatibility Matrix


jerryjch:
http://docs.janusgraph.org/0.1.0-SNAPSHOT/version-compat.html

Update the support matrix after the recent upgrades of the components.
We can wait until the major upgrades are all in before updating this doc for the first release.

jerryjch:
More information can be added there.  For example, hadoop version, Java version.

mbrukman:
Fixed by PR #176.

PR #176, opened 2017-03-31
Update version compatibility table for 0.1.0 release

mbrukman:
Partially addresses issue #102 (for HBase only) as part of the upcoming 0.1.0
release.

### DIFF
--- a/docs/versions.txt
+++ b/docs/versions.txt
@@ -12,6 +12,6 @@ JanusGraph.
 .Version Compatibility Matrix
 [options="header"]
 |==========================
-| JanusGraph | Cassandra | HBase | Elasticsearch | Solr | TinkerPop
-| 0.1.0 | 1.2.z, 2.0.z, 2.1.z | 0.94.z, 0.96.z, 0.98.z, 1.0.z | 1.5.z |  5.2.z | 3.0.z
+| JanusGraph | Cassandra | HBase | Bigtable | Elasticsearch | Solr | TinkerPop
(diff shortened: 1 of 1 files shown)


mbrukman:
@jerryjch — done for both. Bigtable has only a single version ("live"); noted as such in the docs.

sjudeng:
Can you also change the TinkerPop version to 3.2.x?

jerryjch:
Good point.  Should be in both jg01 and master.

mbrukman:
My recommendation would be to commit to `master` and then to cherry-pick / merge manually into the release branch.

I don't think we can just commit to a release branch and merge it into `master` (or _vice versa_) as `master` will quickly diverge from a release branch, at which point it won't cleanly merge in either direction.

sjudeng:
I think the release branch can always be merged into master but not the other way around. Can we work through this on the dev list?

Also regarding Bigtable version @jerryjch made a comment on this in #184 and here I see you're using version "live" but on [Maven central](https://search.maven.org/#search%7Cga%7C1%7Cbigtable) and in [bigtable.txt](https://github.com/JanusGraph/janusgraph/blob/e41daefe3eca0f4de951c9468c1e78d1862e71ce/docs/bigtable.txt#L40) I see versions of the form 0.9.x.

mbrukman:
@sjudeng — sorry for the delay, I was out for a few days. Will update this PR as follows:

* rebase on the `0.1` branch
* change Bigtable to 0.9.x (good point)
* change TinkerPop to 3.2.x

mbrukman:
@sjudeng — do you want to re-review this PR given the changes I made after your approval?

sjudeng:
LGTM!

PR #177, opened 2017-03-31
Fixed ETL wiki page url

tedhtchang:
Signed-off-by: Ted Chang <htchang@us.ibm.com>

### DIFF
--- a/docs/bulkloading.txt
+++ b/docs/bulkloading.txt
@@ -6,7 +6,7 @@ There are a number of configuration options and tools that make ingesting large 
 There are a number of use cases for bulk loading data into JanusGraph, including:
 
 * Introducing JanusGraph into an existing environment with existing data and migrating or duplicating this data into a new JanusGraph cluster.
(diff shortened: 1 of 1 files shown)


mbrukman:
@tedhtchang — thank you for the contribution! It's clearly broken on the docs page: http://docs.janusgraph.org/latest/bulk-loading.html; that said, this format is using Asciidoc, not Markdown, so the syntax is close to correct, and in particular, the anchor text needs to go **after** the URL, but it looks like the URL is not formatted properly.

To verify that the output is what we expect, please build the docs — you can find how to do that by looking at the `docs/build-and-copy-docs.sh` which is in the same directory as the file you modified.

@pluradj — could you please add Ted Chang to the IBM CCLA?

tedhtchang:
@mbrukman Thanks! I corrected. Also I created a PR for the generated html https://github.com/JanusGraph/docs.janusgraph.org/pull/10

sjudeng:
@mbrukman Since this is for the 0.1.0 release should the PR be retargeted to the `jg01` branch? The release branch can be merged into master at any time but in general I don't think it should go the other way (e.g. merging master into a release branch).

PR #179, opened 2017-04-01
Issue 171: Exclude TraversalInterruptionTest in BerkeleyJanusGraphProcessTest

sjudeng:
This PR resolves #171 by using a custom `ProcessStandardSuite` implementation to skip the `TraversalInterruptionTest` in that module.

All remaining tests are passing with this update.

```
$ mvn clean install -DskipTests=true -Dtest.skip.tp=false -pl janusgraph-berkeleyje
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 3:18:13.432s
[INFO] Finished at: Sat Apr 01 00:57:43 UTC 2017
[INFO] Final Memory: 26M/510M
[INFO] ------------------------------------------------------------------------

$ mvn clean install -pl janusgraph-berkeleyje
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 9:02.101s
[INFO] Finished at: Sat Apr 01 01:48:08 UTC 2017
[INFO] Final Memory: 33M/534M
[INFO] ------------------------------------------------------------------------
```

PR #188, opened 2017-04-05
Merge 0.1 into master

sjudeng:
Merge recent commits from 0.1 into master.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -15,3 +15,4 @@ DataStax
 Dylan Bethune-Waddell <dylan.bethune.waddell@mail.utoronto.ca>
 Expero
 Google
+Rafael Fernandes <luizrafael@gmail.com>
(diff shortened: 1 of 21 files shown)


mbrukman:
Please don't merge this PR just yet, let me figure out the issue with the CLA first.

sjudeng:
@mbrukman If it's easier and you're amenable I think you can just merge and push to master directly yourself (as a repo admin) without going through the PR/review process.

mbrukman:
Looks like this is marked as `[cla: no]` because @twilmes merged PR #179 (via web UI?), thus creating a commit using his default email address on GitHub, which is not the email address listed on the CLA.

I'll fix it by adding his other email address to the CLA signers group and we should be good to go here.

mbrukman:
@sjudeng, @twilmes -- CLA issues resolved, so this PR is good to go.

Should we wait for Travis to turn green, or assume it's fine since it's just a merge from one branch to another?

amcp:
Travis is OK now so merging

twilmes:
Good catch @mbrukman.  It was the web UI, didn't think about the fact it was associated with the other email.  Thanks for adding that email to CLA group.

PR #189, opened 2017-04-05
Issue 133: Mark JTS as optional dependency in janusgraph-solr

### LINKED ISSUE #133: Mark JTS as optional in janusgraph-solr and update docs
JTS is currently a non-optional dependency in janusgraph-solr and is distributed with JanusGraph (and previously Titan) releases. What could be done is to mark it as optional in the pom.xml files so it's available for tests and users with requirements for relevant features (Solr, Geoshapes, etc.) can manually add it to their lib directory. Docs also need to be updated to instruct users to manually add the dependency.

From @sjudeng - good catch:
Looks like the longer term solution might be a new Eclipse Foundation [project aiming to replace JTS](https://www.locationtech.org/projects/technology.jts), ([repo](https://github.com/locationtech/jts), [proposal](https://www.locationtech.org/proposals/jts-topology-suite)). It's being developed by LocationTech, who also develop Spatial4j, through-which the JTS dependency comes in to Solr/JanusGraph. I'd guess the roadmap is for a stable new-JTS release and then a Spatial4j update/release migrating to same, at which point JanusGraph could update as well (likely involving a Solr version upgrade). I don't think a full migration in JanusGraph would be possible until Spatial4j and Solr were updated to use JTS 2. The good news is the core Spatial4j developer, who I think also is a core contributor on the Solr spatial connectors, has been active on the JTS 2 mailing list so he is at least aware of the effort.

sjudeng:
Resolves #133

### DIFF
--- a/docs/solr.txt
+++ b/docs/solr.txt
@@ -223,10 +223,10 @@ org.apache.solr.common.SolrException: com/vividsolutions/jts/geom/Geometry
         ...
 ----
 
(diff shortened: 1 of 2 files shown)


twilmes:
@sjudeng just want to double check, I'm good to merge this one, right?

sjudeng:
I'm good with it but I thought we needed one more review.

twilmes:
As we have the dev process currently, I think we're good to go: "Two committer approvals are required to merge a pull request". It's not explicitly stated, but I think the intent was that if the submitter is a  committer, their vote on the PR counts towards the two approvals.  I'll go ahead and merge.

PR #192, opened 2017-04-05
Add ability to await on multiple statuses

dpitera:
This fixes the following bug for example:
If a user wants to update an index by adding a propertyKey to it, then
the existing propertyKeys will have status ENABLED, but the
IndexStatusWatcher will be looking for REGISTERED only, so the index update
is never completed.

Issues: #196

Signed-off-by: David Pitera <dpitera@us.ibm.com>

PR #193, opened 2017-04-05
UpdateTriggerStatus from Set -> List

dpitera:
Previously, we were seeing new index creations on the same graph act as
if they were equal to previous triggers. Thus, the previous but
not finished updateStatusTriggers were being removed from the SET and
getting stuck in INSTALLED. In this change, we have done two things:
1. Change the datastructure housing the updateStatusTriggers from a Set
to a List.
2. Remove the code that removes any updateStatusTriggers from the data
structure.

These changes ensure all index creations on a given graph fulfill their
INSTALLing.

Issues: #195

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
@@ -106,6 +106,8 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -140,7 +142,7 @@ public class ManagementSystem implements JanusGraphManagement {
     private final StandardJanusGraphTx transaction;
 
     private final Set<JanusGraphSchemaVertex> updatedTypes;
-    private final Set<Callable<Boolean>> updatedTypeTriggers;
+    private final List<Callable<Boolean>> updatedTypeTriggers;
 
     private final Instant txStartTime;
     private boolean graphShutdownRequired;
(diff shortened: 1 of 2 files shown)


dpitera:
Hello @mbrukman my PR is not being labelled as `cla:yes` even though I updated my signature on all commits to match my [CLA signature](https://github.com/JanusGraph/legal/blob/master/CLA_SIGNERS.yaml#L178)

dpitera:
Given that the unit tests already test things like creating, deleting, reindexing indexes, I don't think there's really a need for an additional unit test to be included in this PR.

This change ensures a production grade Janus cluster does not evict incomplete update index triggers. I think the change is self contained/ self explanatory without an additional test. LMK if YMMV? 

Having said that, I am adding a unit test for #192

dpitera:
Issues created and linked in the PR and commits

mbrukman:
@dpitera -- the CLA checker verifies everything:

* GitHub id
* full name
* email

The name on the commits is not the name on the CLA, it's your GitHub user id.

Please fix it and re-push this branch (same for the other PR), using [these instructions](https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla) and configure your repo to avoid this issue in the future.

dpitera:
Thank you. I fixed. I see now there is a difference between the commit author and the signature listed in the commit message.

amcp:
@twilmes Given the presence of new tests in #192 are we OK to merge this after #192 is merged?

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -870,8 +872,6 @@ public boolean equals(Object oth) {
     }
 
     private void setUpdateTrigger(Callable<Boolean> trigger) {
-        //Make sure the most current is the one set
-        if (updatedTypeTriggers.contains(trigger)) updatedTypeTriggers.remove(trigger);
```
twilmes: Should the trigger still be removed at some point, just not prematurely?

dpitera: I'm not too worried about it; the `ManagementSystem` is something used by users to do ~case-by-case schema mutations, as in the List in question probably won't get that long (i.e. users are not going to be doing enough schema mutations to blow out the stack--especially not in one transaction (and every commit requires a new instantiation of the ManagementSystem)) and the GC should do just fine collecting the entire ManagementSystem object once no longer used. And also note this behavior is exactly the same as before.

mbrukman:
@dpitera – thank you for the contribution! In a future PR, please also add your name and email to `CONTRIBUTORS.txt`; IBM, as copyright holder, is already listed in `AUTHORS.txt`, so that's all set.

PR #202, opened 2017-04-07
Add instructions for non-Google account holders

mbrukman:
Add directions for subscribing and unsubscribing using an arbitrary email
address, not just those associated with a Google account (whether Gmail or
corporate email).


mbrukman:
I've tested it, and yes, it does work.

PR #210, opened 2017-04-11
0.1 release prep

twilmes:
* Release date set in change logs
* mvn release:prepare run which bumped version 0.1.0, committed, tagged, and then bumped to the new dev version 0.1.1-SNAPSHOT and committed

### DIFF
--- a/CHANGELOG.asc
+++ b/CHANGELOG.asc
@@ -3,8 +3,8 @@
 Release Notes
 -------------
 
-Version 0.1.0 (TBD)
-~~~~~~~~~~~~~~~~~~~
+Version 0.1.0 (Release Date: April 11, 2017)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 [source, m2]
 <dependency>
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -3,8 +3,8 @@
 Release Notes
 -------------
 
-Version 0.1.0 (TBD)
-~~~~~~~~~~~~~~~~~~~
+Version 0.1.0 (Release Date: April 11, 2017) 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 [source, m2]
 <dependency>
--- a/janusgraph-all/pom.xml
+++ b/janusgraph-all/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.janusgraph</groupId>
         <artifactId>janusgraph</artifactId>
-        <version>0.1.0</version>
+        <version>0.1.1-SNAPSHOT</version>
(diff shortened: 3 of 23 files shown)


sjudeng:
What would be the process if the 0.1.0 release vote fails and changes are required? Would that work happen on top of these commits (which would then have project version of 0.1.1-SNAPSHOT)?

twilmes:
The maven release plugins bump from 0.1.0 to 0.1.1-SNAPSHOT is a bit problematic in the case where further changes are required. We could rollback to the RC2 tag, force a push to 0.1 on that and do any further work after that point.  In the past, I've used the versions plugin to do the bump to the release version, and then later on, the bump to the new SNAPSHOT after the release is squared away.

CHANGELOG.asc
```diff
@@ -3,8 +3,8 @@
 Release Notes
 -------------
 
-Version 0.1.0 (TBD)
-~~~~~~~~~~~~~~~~~~~
+Version 0.1.0 (Release Date: April 11, 2017)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
mbrukman: Can you please file an issue to cleanup the headers here to use the new headings format? Feel free to assign it to me.

docs/changelog.txt
```diff
@@ -3,8 +3,8 @@
 Release Notes
 -------------
 
-Version 0.1.0 (TBD)
-~~~~~~~~~~~~~~~~~~~
+Version 0.1.0 (Release Date: April 11, 2017) 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
mbrukman: Same comment here as in `CHANGELOG.asc`.

ptgoetz:
@mbrukman That's kind of non-standard practice. What I usually do when an RC vote fails is (assuming maven release plugin):

1. Delete the tag in the local and remote repos.
2. `git revert` the two commits made by the maven release plugin.

CHANGELOG.asc
twilmes: Good catch, created #211.

twilmes:
Yes, agreed, a revert is the way to go so we don't lose the history.

mbrukman:
OK, so we have the following options in case a vote fails:

* force-push to the branch and rewrite history
* `git revert`
* work in a local branch and push only when vote passes

I'm in favor of the second or third, I prefer not to set a precedent that we rewrite history, it should be moving forward. The second option has the benefit of recording what happened, including the reverts, the third one seems like it's inviting things to be done outside of the repo, so I agree that `git revert` is a good style to follow.

twilmes:
Thanks for the quick feedback guys.  I'll take note of the `git revert` in case of vote failure and capture that in the updated instructions after we complete the release.

PR #217, opened 2017-04-15
Merge 0.1 into master

sjudeng:
Resolves merge conflicts in all POMs from project version updates.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -1,10 +1,8 @@
 [[changelog]]
 [appendix]
-Release Notes
--------------
+== Release Notes
 
-Version 0.1.0 (Release Date: April 11, 2017) 
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+=== Version 0.1.0 (Release Date: April 11, 2017) 
 
 [source, m2]
 <dependency>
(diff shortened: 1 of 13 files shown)


docs/changelog.txt
```diff
@@ -1,7 +1,41 @@
 [[changelog]]
 [appendix]
-== Release Notes
+Release Notes
```
mbrukman: This is undoing the change that I made to use the recommended style of header. The correct version here is the current version on `master`:

> == Release Notes

Similarly for the other header below.

Should I submit a change to the `0.1` branch to fix this and we can then merge with that fix, or do you want to merge this PR, and I'll submit separate fixes to `master` and `0.1` branches?

LGTM outside of these two concerns.

sjudeng: Sorry I didn't catch this. Either way if you don't mind fixing it on 0.1 then I think we can merge it in with this PR or after. Otherwise I can take a stab at fixing (is it just changelog.txt?).

sjudeng: @mbrukman I submitted #218 to fix this. I was going to do it as CTR so I could include it here but it doesn't look like repo permissions currently allow this (e.g. can't push directly to protected branch, can't merge PR without review and can't self-approve PR).

mbrukman: Now that #218 has been merged, please update this PR from the `0.1` branch and let me know when it's ready to re-review.

sjudeng:
@mbrukman I merged again and picked up the commit in #218.

mbrukman:
Thanks again, @sjudeng!

PR #240, opened 2017-04-28
#228 [WIP] Make ID store name configurable

### LINKED ISSUE #228: ID_STORE_NAME table renamed from Titan
In migrating from Titan, I found (by way of Cassandra table metrics reporting) that the `titan_ids` table has changed to `janusgraph_ids`. I think if this table were used it would be a breaking change.

However, it appears to me that this table isn't actually used much if at all? It's defined under the constant ID_STORE_NAME, which seems unused here:
https://github.com/JanusGraph/janusgraph/search?utf8=%E2%9C%93&q=ID_STORE_NAME&type=

Either there should be a migration path for this table from Titan to JanusGraph, or it should not exist at all. So far migrating to JanusGraph from Titan seems to work but things like this make me worry.

amcp:
Fixes #228

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java
@@ -81,11 +81,12 @@ public class Backend implements LockerProvider, AutoCloseable {
      * <p/>
      * These names are fixed and should NEVER be changed. Changing these strings can
      * disrupt storage adapters that rely on these names for specific configurations.
+     * In the past, the store name for the ID table, janusgraph_ids, was also marked here,
+     * but to clear the upgrade path from Titan to JanusGraph, we had to pull it into
+     * configuration.
      */
     public static final String EDGESTORE_NAME = "edgestore";
     public static final String INDEXSTORE_NAME = "graphindex";
-
-    public static final String ID_STORE_NAME = "janusgraph_ids";
 
     public static final String METRICS_STOREMANAGER_NAME = "storeManager";
     public static final String METRICS_MERGED_STORE = "stores";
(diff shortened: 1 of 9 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -718,6 +718,14 @@ public boolean apply(@Nullable Integer integer) {
 //    public static final int IDS_BLOCK_SIZE_DEFAULT = 10000;
 
     /**
+     * The name of the ID store. Currently this defaults to janusgraph_ids. You can override the ID store to
+     * facilitate migration from JanusGraph's predecessor, Titan. Previously, this KCVStore was named titan_ids.
+     */
+    public static final ConfigOption<String> IDS_STORE_NAME = new ConfigOption<>(IDS_NS,"store-name",
```
jerryjch: Nit. Space
`IDS_NS,"store-name"`

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java
```diff
@@ -85,8 +85,6 @@
     public static final String EDGESTORE_NAME = "edgestore";
     public static final String INDEXSTORE_NAME = "graphindex";
 
-    public static final String ID_STORE_NAME = "janusgraph_ids";
```
jerryjch: Can we add a comment telling why ID_STORE_NAME is missing here and moved to other place? e..g To make Make ID store name configurable

jerryjch:
Let's say in the documentation that this IDS_STORE_NAME is meant to be used only for backward compatibility with Titan, and warn against using this property explicitly in normal operations.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java
amcp: done

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
amcp: done

srosenthal:
It looks like this commit never made it to the master branch, only the branch for the 0.1.1. Am I right in assuming master is meant to be the latest work-in-progress for 0.2.0?

Issue #243, opened 2017-04-28
Reduce Cassandra TP test time


amcp:

amcp:
@sjudeng What is the new Cassandra TP test time?

sjudeng:
4.2 hours! Compare this to 14.9 hours from the last run prior to the fix.

PR #244, opened 2017-04-28
Use larger id block size and avoid deprecated methods.

amcp:
I now use the default block size here. I think we need to opt-out of some tests still


amcp:
Fixes #243

analytically:
I think this can be merged now.

sjudeng:
@analytically I thought there were a number of test failures in the TinkerPop suite with this update as is. @amcp Can you confirm?

sjudeng:
@amcp @analytically I (finally) ran through the whole suite myself with this commit (rebased on master) and all tests passed. Thanks for fixing this (and sorry that I caused the issue)!

PR #278, opened 2017-05-21
Add Seeq to the list of JanusGraph users.

mbrukman:
Thanks to @sharpau for confirmation that Seeq has deployed JanusGraph and permission to use the name here.

See also https://github.com/JanusGraph/janusgraph.org/pull/11 for the website update.


mbrukman:
@sjudeng — yep, I keep forgetting to add those tags to the commits, so I manually skip Travis (as I did here) to avoid wasting CPU cycles.

PR #286, opened 2017-05-28
Update README badges and community pointers

mbrukman:
* Renamed "Contact" to "Community", reorganized order of sections, and added Stack Overflow
* Added Maven and Stack Overflow badges to the top of the page; changed Markdown formatting

Note: I added `[skip ci]` to each of the commits so Travis shouldn't run build/test for this PR.


sjudeng:
@mbrukman I noticed Travis did run on the merge commits for this, I think because `[skip ci]` didn't make it into the merge commit message since it wasn't on the first line of the original commit message.

Going forward the easiest solution might just be to put `[skip ci]` in the first line of the commit message so then it should be in the merge commit message by default. Otherwise those performing merge would need to remember to make sure to add it manually before merging.

mbrukman:
@sjudeng – thank for the tip! Seems like putting it on the PR title is sufficient rather than the commit message, because the merge commit appears to use the PR title as its entire message.

Issue #92, opened 2017-02-06
migrate es to HTTP client and not TransportClient (on deprecation path)


amcp:
At some point in the future, support for the [Transport Client (native Java client) will be removed](https://www.elastic.co/blog/elasticsearch-the-server).
To mitigate, we need to use the HTTP API.

sjudeng:
The current Java REST client in Elasticsearch 5.x is very low level. I'd recommend waiting until they release a higher level Java REST client before working on migration to avoid unnecessary work.

amcp:
I agree it is extremely low level and would require major refactoring. We should wait for a higher level REST client. I just wanted to get ES lack of commitment to the transport client on our radar (so that we know to plan for it).

pluradj:
Related dev mailing list discussion https://groups.google.com/d/msg/janusgraph-dev/F-zfIsuiIoQ/q_R-kkhHDgAJ

sjudeng:
With merge of #79 JanusGraph now supports the low-level ES REST HTTP client. But keeping this issue open to track work on migrating to higher level HTTP client for ES (e.g. Jest or maybe eventually ES 6.x REST).

sjudeng:
With merge of #289 JanusGraph now only supports the ES REST HTTP client and the Transport client is no longer support.

PR #233, opened 2017-04-26
Can use external ES mapping instead of push mapping

davidclement90:
This adds the possibility to use external mapping in order to use custom ES mapping.
If you set USE_EXTERNAL_MAPPINGS to true, JanusGraph will check if the mapping is good instead of push it.
Whitout this adds, even if I push a mapping JanusGraph will update it during register phase. 
This feature will allows to manage mapping directly in Kibana and not with ES_CREATE_EXTRAS_NS properties or to set for example a property with are declare in JanusGraph as String to IP, or use copy-to ES feature.

Issue : https://github.com/JanusGraph/janusgraph/issues/222
Signed-off-by: David Clement <david.clement90@laposte.net>

PR #289, opened 2017-05-29
Remove core Elasticsearch dependency

sjudeng:
This PR removes support for the Elasticsearch Transport client and updates to manually build request objects instead of using the ES Java API. This allows for the removal of the `org.elasicsearch:elasticsearch` dependency and the addition of compatibility shims that resolve deprecation warnings under Elasticsearch 5.x and also restores support for Elasticsearch 1.x without requiring custom builds.

Related issue is #92.


janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/ES2Compat.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final IndexFeatures FEATURES = coreFeatures().supportsGeoContains().build();
+
+    @Override
+    public String scriptLang() {
+        return "groovy";
+    }
+
+    @Override
+    public Map<String,Object> createKeywordMapping() {
+        return ImmutableMap.of(ES_TYPE_KEY, "string", ES_INDEX_KEY, "not_analyzed");
```
amcp: externalize string constant

docs/versions.txt
```diff
@@ -13,7 +13,7 @@ JanusGraph.
 [options="header"]
 |==========================
 | JanusGraph | Cassandra | HBase | Bigtable | Elasticsearch | Solr | TinkerPop
-| 0.1.0 | 1.2.z, 2.0.z, 2.1.z | 0.98.z, 1.0.z, 1.1.z, 1.2.z | 0.9.z | 2.z,5.z* |  5.2.z | 3.2.z |
+| 0.1.0 | 1.2.z, 2.0.z, 2.1.z | 0.98.z, 1.0.z, 1.1.z, 1.2.z | 0.9.z | 1.z*,2.z,5.z |  5.2.z | 3.2.z |
```
sjudeng: No we dropped support for 1.z with previous update (#79) and this PR restores it. But unit tests require manually starting a 1.z ES instance so it'll be tough to maintain the support for new features/updates. Should we take 1.z back out of the table?

amcp: Version compatibility is different from ease of testing, so I think 1.z can stay in this list.

sjudeng:
Please hold off on merging this until #233 is merged.

amcp:
@sjudeng ok to rebase and resolve conflicts here.

Issue #74, opened 2017-01-31
Enable automated testing on Travis CI


mbrukman:
Right now, our Travis config only runs a build, but no tests, because the full suite times out: Travis CI build/test jobs time out after 50 minutes, and the test suite requires far longer than that.

Note that a Travis CI has a concept of a _Build_ which consists of one or more _Jobs_: each _Job_ has a 50 minute timeout, but there's no limit (to my knowledge) on the number of _Jobs_ that can be included in a single _Build,_ so in theory, we can split up the test suite as needed to fit into the 50 minute timeout.

The goal of this issue is to enable automated test runs for each PR and on the `master` branch on Travis CI. To accomplish this, we may also need to address one or more of the following issues:

* [#5: Categorize tests by runtime](https://github.com/JanusGraph/janusgraph/issues/5)
* [#18: Add timeouts to test classes and methods](https://github.com/JanusGraph/janusgraph/issues/18)

sjudeng:
@mbrukman See comment at end of #5. I think there might be a non-timeout issue running tests under Travis. But on the timeout issue note if you break tests out by module all except hbase and cassandra pass under 50 minutes.

sjudeng:
Regarding running hbase and cassandra tests under Travis, again assuming no non-timeout issues crop up, how about grouping tests as part of Travis build matrix?

```
env:
  matrix:
    - MODULE='cassandra' ARGS='-Dtest=ThriftPartitionGraphTest'
    - MODULE='cassandra' ARGS='-Dtest=ThriftDistributedStoreManagerTest'
```

The largest run time for a test class I see is about 1500 seconds, well under 50 minute timeout. Could group tests as well based on timing (e.g. `-Dtest=ThriftPartitionGraphTest,ThriftDistributedStoreManagerTest`). I'd write a script to generate the matrix based on test output from a successful run. It would be a huge ugly test matrix but given the choice I'd prefer this in Travis configuration rather than modifying JanusGraph test classes with category annotations. For one it is really all just to accomodate a (not unreasonable) Travis limitation. Also there would be only one file to maintain.

mbrukman:
Partial progress made: PR https://github.com/JanusGraph/janusgraph/pull/85 adds testing of all modules via Travis CI, even though some of them are marked as optional, because they are either known to fail, or fail non-deterministically. Changes are in flight to fix the tests, including:

[...]

PR #290, opened 2017-05-29
Further split Cassandra Thrift tests and make optional in Travis

sjudeng:
Cassandra Thrift test groups have been regularly timing out under Travis. This PR adds another test group set and also moves all Thrift test groups to optional in the Travis build matrix. Justification is that tests for the astyanax and embedded clients are still required.

Related issue is #74.


mbrukman:
Thanks, @sjudeng!

PR #292, opened 2017-05-31
Remove hardcoded slf4j version from cassandra.in.sh

### LINKED ISSUE #291: Cassandra doesn't log to its log file
When I start JanusGraph with the provided `janusgraph.sh` script without modifying anything I get the following warning and Cassandra doesn't log to the default log file `log/cassandra.log`:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/janusgraph/janusgraph-0.1.1-hadoop2/lib/slf4j-log4j12-1.7.12.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/janusgraph/janusgraph-0.1.1-hadoop2/lib/logback-classic-1.1.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
```

This seems to only affect Cassandra as Gremlin Server and Elasticsearch both log to their own log files.
I manually deleted the `logback-classic-1.1.2.jar` from the `lib` directory which fixed the issue for me. So it seems that Cassandra uses by default the logback logger which is a bit odd because `conf/cassandra` contains a `log4j-cassandra.properties` file but no `logback.xml` file.
[The `pom.xml` of `janusgraph-cassandra`](https://github.com/JanusGraph/janusgraph/blob/e1b087566bbc85ecfadaef9455c4f8358ae88947/janusgraph-cassandra/pom.xml) contains a comment about this issue, but it sounds like this should only affect testing.

Is this a known problem with JanusGraph and Cassandra? And isn't there a way this can be resolved in the pre-packaged zip archive? Maybe by adding a `logback.xml` file to the `conf/cassandra` directory?

sjudeng:
Fixes #291

### DIFF
--- a/janusgraph-dist/src/assembly/static/bin/cassandra.in.sh
+++ b/janusgraph-dist/src/assembly/static/bin/cassandra.in.sh
@@ -49,7 +49,7 @@ cassandra_storagedir="$CASSANDRA_HOME/data"
 # The java classpath (required)
 CLASSPATH="$CASSANDRA_CONF:$cassandra_bin"
 
-CLASSPATH="$CLASSPATH":"$CASSANDRA_HOME"/lib/slf4j-log4j12-1.7.5.jar
+CLASSPATH="$CLASSPATH":$(ls "$CASSANDRA_HOME"/lib/slf4j-log4j12-*.jar)
 
 for jar in "$CASSANDRA_HOME"/lib/*.jar; do
     if [ $jar != slf4j-log4j12* ] ; then


janusgraph-dist/src/assembly/static/bin/cassandra.in.sh
```diff
@@ -49,7 +49,7 @@ cassandra_storagedir="$CASSANDRA_HOME/data"
 # The java classpath (required)
 CLASSPATH="$CASSANDRA_CONF:$cassandra_bin"
 
-CLASSPATH="$CLASSPATH":"$CASSANDRA_HOME"/lib/slf4j-log4j12-1.7.5.jar
+CLASSPATH="$CLASSPATH":`ls "$CASSANDRA_HOME"/lib/slf4j-log4j12-*.jar`
```
mbrukman: `$(...)` appears to be preferred to <code>\`...\`</code>: http://mywiki.wooledge.org/BashFAQ/082

janusgraph-dist/src/assembly/static/bin/cassandra.in.sh
```diff
@@ -49,7 +49,7 @@ cassandra_storagedir="$CASSANDRA_HOME/data"
...
 
 for jar in "$CASSANDRA_HOME"/lib/*.jar; do
     if [ $jar != slf4j-log4j12* ] ; then
```
jerryjch: Do we need to clean these few lines that follow?

janusgraph-dist/src/assembly/static/bin/cassandra.in.sh
sjudeng: @mbrukman Yep I'll look at making that update later today.

janusgraph-dist/src/assembly/static/bin/cassandra.in.sh
sjudeng: @jerryjch Those lines seem pretty clean to me, what do you have in mind?

amcp:
@jerryjch are you suggesting that line 55 should add a dash after log4j2?

sjudeng:
@jerryjch Are you good with this or is there anything else you're looking for here?

jerryjch:
Sorry for the confusion and delay.
My question was why single out slf4j-log4j12-*.jar?  Can it be included in the loop?
Or I misread it?

```
CLASSPATH="$CLASSPATH":$(ls "$CASSANDRA_HOME"/lib/slf4j-log4j12-*.jar)
 
 for jar in "$CASSANDRA_HOME"/lib/*.jar; do
     if [ $jar != slf4j-log4j12* ] ; then
         CLASSPATH="$CLASSPATH:$jar"
     fi
 done
```

sjudeng:
@jerryjch I think this block is there to make sure the slf4j-log4j12 jar is first on the classpath so log4j is used as the logging backend for slf4j. This is similar to what's done in project configuration for tests (see [janusgraph-cassandra/pom.xml](https://github.com/JanusGraph/janusgraph/blob/e1b087566bbc85ecfadaef9455c4f8358ae88947/janusgraph-cassandra/pom.xml#L52-L55)).

Issue #293, opened 2017-05-31
Allows to index points, circles and boxes in PREFIX_TREE mapping in ES


davidclement90:
With the current code, JanusGraph can not index points, circle and boxes for PREFIX_TREE mappings in Elasticsearch, but Elasticsearch allow to index this.

PR #294, opened 2017-05-31
This feature allows to index points, circles and boxes in PREFIX_TREE mapping for ES.

davidclement90:
I also fix some indentation issue.
https://github.com/JanusGraph/janusgraph/issues/293

### DIFF
--- a/docs/searchpredicates.txt
+++ b/docs/searchpredicates.txt
@@ -92,7 +92,7 @@ Additional data types will be supported in the future.
 
 [[geoshape]]
 === Geoshape Data Type
-The Geoshape data type supports representing a point, circle, box, line, polygon, multi-point, multi-line and multi-polygon. Index backends currently support indexing points, lines and polygons. Indexing multi-point, multi-line and multi-polygon properties has not been tested.
+The Geoshape data type supports representing a point, circle, box, line, polygon, multi-point, multi-line and multi-polygon. Index backends currently support indexing points, circles, boxes, lines and polygons. Indexing multi-point, multi-line and multi-polygon properties has not been tested.
 Geospatial index lookups are only supported via mixed indexes.
 
 .Note:
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -19,6 +19,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
+import com.spatial4j.core.shape.Rectangle;
 import org.apache.commons.lang.StringUtils;
 import org.elasticsearch.Version;
 import org.elasticsearch.common.geo.ShapeRelation;
@@ -317,7 +318,7 @@ public class ElasticSearchIndex implements IndexProvider {
                         if (textAnalyzer != null) {
                             mapping.field(ANALYZER, textAnalyzer);
                         }
-                    	break;
+                        break;
                     case TEXTSTRING:
                         if (textAnalyzer != null) {
                             mapping.field(ANALYZER, textAnalyzer);
@@ -422,11 +423,11 @@ public class ElasticSearchIndex implements IndexProvider {
             Object value = null;
             switch (keyInformation.getCardinality()) {
                 case SINGLE:
-                    value = convertToEsType(Iterators.getLast(add.getValue().iterator()).value);
+                    value = convertToEsType(Iterators.getLast(add.getValue().iterator()).value, Mapping.getMapping(keyInformation));
                     break;
                 case SET:
                 case LIST:
-                    value = add.getValue().stream().map(v -> convertToEsType(v.value))
+                    value = add.getValue().stream().map(v -> convertToEsType(v.value, Mapping.getMapping(keyInformation)))
                         .filter(v -> {
                             Preconditions.checkArgument(!(v instanceof byte[]), "Collections not supported for " + add.getKey());
                             return true;
@@ -446,7 +447,7 @@ public class ElasticSearchIndex implements IndexProvider {
         return doc;
     }
 
-    private static Object convertToEsType(Object value) {
+    private static Object convertToEsType(Object value, Mapping mapping) {
         if (value instanceof Number) {
             if (AttributeUtil.isWholeNumber((Number) value)) {
                 return ((Number) value).longValue();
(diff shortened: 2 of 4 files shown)


janusgraph-test/src/main/java/org/janusgraph/diskstorage/indexing/IndexProviderTest.java
```diff
@@ -424,12 +424,12 @@ private void storeTest(String... stores) throws Exception {
 
 
             //Update some data
-            add(store, "doc4", getDocument("I'ts all a big Bob", -100, 11.2, Geoshape.point(48.0, 8.0), Geoshape.line(Arrays.asList(new double[][] {{7.5, 47.5}, {8.5, 48.5}})), Arrays.asList("10", "11", "12"), Sets.newHashSet("10", "11"), Instant.ofEpochSecond(4)), true);
+            add(store, "doc4", getDocument("I'ts all a big Bob", -100, 11.2, Geoshape.point(-48.0, 8.0), Geoshape.point(-48.0, 8.0), Arrays.asList("10", "11", "12"), Sets.newHashSet("10", "11"), Instant.ofEpochSecond(4)), true);
```
sjudeng: Is there a reason you changed the sign on the first point latitude?

davidclement90: Yes, I want to test the reindexing process for prefix_tree in "doc1" so I change the sign on latitude to be sure that the document has really changed.
I do not want to change all tests thoroughly so I change also the sign en "doc4".

sjudeng: Makes sense thanks

docs/searchpredicates.txt
```diff
@@ -92,7 +92,7 @@ Additional data types will be supported in the future.
 
 [[geoshape]]
 === Geoshape Data Type
-The Geoshape data type supports representing a point, circle, box, line, polygon, multi-point, multi-line and multi-polygon. Index backends currently support indexing points, lines and polygons. Indexing multi-point, multi-line and multi-polygon properties has not been tested.
+The Geoshape data type supports representing a point, circle, box, line, polygon, multi-point, multi-line and multi-polygon. Index backends currently support indexing points, circles, boxes, lines and polygons. Indexing multi-point, multi-line and multi-polygon properties has not been tested.
```
sjudeng: Can you add language to make clear that indexing circles and boxes is only supported under Elasticsearch?

davidclement90: Why only on Elasticsearch ?
Solr is also tested. And I think that the code allow it. https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java#L361

sjudeng: Right, perfect! That code was added in #79 I just hadn't tested circle/box indexing. Thanks!

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -468,18 +469,31 @@ private static Object convertToEsType(Object value) {
...
 
-    private static Object convertgeo(Geoshape geoshape) {
-        if (geoshape.getType() == Geoshape.Type.POINT) {
+    @SuppressWarnings("unchecked")
+    private static Object convertgeo(Geoshape geoshape, Mapping mapping) {
+        if (geoshape.getType() == Geoshape.Type.POINT && Mapping.PREFIX_TREE != mapping) {
             Geoshape.Point p = geoshape.getPoint();
             return new double[]{p.getLongitude(), p.getLatitude()};
-        } else if (geoshape.getType() != Geoshape.Type.BOX && geoshape.getType() != Geoshape.Type.CIRCLE) {
+        } else if (geoshape.getType() == Geoshape.Type.BOX) {
```
amcp: why is it OK not to catch IOException for BOX but required for CIRCLE?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -583,11 +597,13 @@ private String getAdditionScript(KeyInformation.IndexRetriever informations, Str
             switch (keyInformation.getCardinality()) {
                 case SET:
                 case LIST:
-                    script.append("ctx._source[\"" + e.field + "\"].add(" + convertToJsType(e.value, scriptLang) + ");");
+                    script.append("ctx._source[\"" + e.field + "\"].add(" + convertToJsType(e.value, scriptLang, Mapping.getMapping(keyInformation)) + ");");
```
amcp: use a String.format to reduce the number of frame switches

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -583,11 +597,13 @@ private String getAdditionScript(KeyInformation.IndexRetriever informations, Str
...
                     if (hasDualStringMapping(keyInformation)) {
-                        script.append("ctx._source[\"" + getDualMappingName(e.field) + "\"].add(" + convertToJsType(e.value, scriptLang) + ");");
+                        script.append("ctx._source[\"" + getDualMappingName(e.field) + "\"].add(" + convertToJsType(e.value, scriptLang, Mapping.getMapping(keyInformation)) + ");");
```
amcp: String.format

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sjudeng: Switching to `String.format` may be a performance hit. I was looking up whether it was preferred over StringBuilder and results seemed to say concatenation and StringBuilder end up being similar (though StringBuilder may be better style), but String.format is 5-20 time slower than both.

1. https://kylewbanks.com/blog/java-string-concatenation-vs-stringbuilder-vs-string-format-performance
2. http://jacksondunstan.com/articles/3015
3. https://stackoverflow.com/questions/513600/should-i-use-javas-string-format-if-performance-is-important/513705#513705

amcp: @sjudeng Wow. Thank you for this important learning. StringBuilder is the way to go from now on.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
amcp: not an issue!

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
amcp: bump

davidclement90: Yes it's OK.
IOException is throw by _geoshape.toMap()_

For Box, the map is constructed without using _geoshape.toMap()_.
For Circle, the map is constructed with _geoshape.toMap()_.

PR #300, opened 2017-06-03
Skip build and tests on coverity_scan branch

sjudeng:
This fixes a typo to ensure tests don't run, and also updates to skip the install stage, in the `coverity_scan` branch under Travis. Hopefully this helps get the Travis build passing on the branch (https://travis-ci.org/JanusGraph/janusgraph/builds/238948501).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ env:
     - MODULE='hbase-parent/janusgraph-hbase-098' ARGS='-Dtest=**/diskstorage/hbase/*'
     - MODULE='hbase-parent/janusgraph-hbase-098' ARGS='-Dtest=**/graphdb/hbase/*'
     - MODULE='cql'
+    - COVERITY_ONLY=true
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -67,22 +67,20 @@ addons:
...
-    # https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
     build_command: "mvn clean package -B -DskipTests=true"
     branch_pattern: coverity_scan
 
 install:
-  # Build and install current module and dependencies. Handle output timeouts and retry on error.
-  - travis_retry travis_wait mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -pl janusgraph-${MODULE} -am
+  - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then
+      # Build and install current module and dependencies. Handle output timeouts and retry on error.
+      travis_retry travis_wait mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -pl janusgraph-${MODULE} -am;
```
amcp: Can you split this mvn install command into multiple lines and comment on each command line argument?

.travis.yml
```diff
@@ -67,22 +67,20 @@ addons:
...
+    fi
 
 script:
-  - if [[ "${COVERITY_SCAN_BRANCH}}" == 1 ]]; then
-      echo "Skipping build/test on 'coverity_scan' branch.";
-      exit 0;
+  - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then
+      # Build and test module. Handle output timeouts and retry with a clean build on error.
+      travis_retry travis_wait 50 mvn clean verify -pl janusgraph-${MODULE} ${ARGS}
```
amcp: comment on the command line arguments here as well

.travis.yml
sjudeng: @amcp Thanks for the review. I split into a few lines and switched to long-form argument names for clarity. Good enough or would more comments still be helpful?

mbrukman:
So, ironically, the [failing message](https://travis-ci.org/JanusGraph/janusgraph/jobs/238948505) was:

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Which is to say, previously, it was having trouble finding `travis_wait` command but what it really needs is to use the `travis_wait` command to ensure that it doesn't kill the build when there's no output for 10min.

I've recently submitted a separate, manual build for Coverity, which they've accepted and analyzed, but it's waiting for them to approve the "open source"-ness of our project before they'll let us see the results.

I'm also thinking that it doesn't make sense to run Coverity separately for each module. What I was thinking is to simplify this entirely, drop the special Coverity integration from Travis CI config, and do the following instead:

* add a new module `"all"` which corresponds to `janusgraph-all`
* exit immediately on the `coverity_scan` branch if running for any module other than `all`
* conversely, exist immediately for the `all` branch for any branch besides `coverity_scan`
* create a manual script to download, run the build (skip the tests), and if successful, upload scan results to Coverity — I have a script for this already, just requires some more polish.

Thoughts?

sjudeng:
What I thought was most interesting about the timeout error is that it happened during the compile step of janusgraph-core, not during testing or dependency download. I assume this compile delay is because of coverity itself? In this case your approach of a separate script might be the only way to go so it can be wrapped by travis_wait (e.g. even if we turned up logging or built with debug on I don't think there'd be any output during the actual compile where it's hanging).

If you're going to add a new element to the build matrix for this there's no need to define `MODULE`, instead I'd define a single variable like `COVERITY_ONLY` that you'd then check as described.

```
matrix:
  - COVERITY_ONLY=1
  ...
```

mbrukman:
> What I thought was most interesting about the timeout error is that it happened during the compile step of janusgraph-core, not during testing or dependency download. I assume this compile delay is because of coverity itself?

I'm not sure if it's Coverity that's causing the delay, it's also possible that Travis CI free tier is using single-CPU VM or shared-CPU VM, which could be slowing down compilation, and we have `travis_wait` wrapping all other commands, possibly for the same reason. We used to have `travis_wait` for the `coverity_scan` branch, too and it worked when the command was found (but seems like it was flaky some of the time).

@sjudeng — do you want to adjust your PR to use `COVERITY_ONLY` and I'll add the runner script in a follow-up PR?

sjudeng:
@mbrukman I just pushed the update. There's now a `COVERITY_ONLY` job that will only run on the coverity_scan branch. Also Travis should exit prior to running Coverity scan for all other jobs, but I don't have a Coverity token to test (can you?). Hopefully exiting prior to scanning will keep the other jobs from counting against the project quota. Incidentally I'm not sure if it's a coincidence but in the [failing job](https://travis-ci.org/JanusGraph/janusgraph/builds/238948501) there were four successes, which is the maximum number of allowable Coverity scans per day according to the [FAQ](https://scan.coverity.com/faq#frequency). I'm curious if just limiting the scan to a single job as now done here will resolve the issue.

.travis.yml
```diff
@@ -64,25 +65,33 @@ addons:
     # https://scan.coverity.com/travis_ci
     project:
       name: "JanusGraph/janusgraph"
-      version: "0.2.0-SNAPSHOT"
```
mbrukman: Not sure why this line was removed; this identifies the build on the Coverity dashboard, so that we can see what version of JanusGraph the given Coverity scan refers to. It appears that it'll need to be updated manually, but it doesn't seem like there is any other way.

Alternatively, we can remove this entire stanza and use the script I wrote separately to submit the build for Coverity to scan, which also knows how to parse `pom.xml` to retrieve the current version automatically. I'll look into this later.

sjudeng: @mbrukman Sorry with that line in there I got a validation warning through https://lint.travis-ci.org/ but then forgot to call it out to you for review. I added #334 to restore it.

I think ideally it'd be great to get the scan to work through the Travis plugin as it seems like that would be the easiest to maintain. But if we can't resolve Coverity build failures then the script will probably be the only option left.

Issue #313, opened 2017-06-08
Fix JanusGraphIndexQuery indexQuery misleading javadoc


wojciechwojcik:
Configuration:
```
# index backend
index.search.backend=lucene
index.search.directory=build/searchindex
```

Mixed index creation:
```
JanusGraphManagement mgmt = = graph.openManagement();
mgmt.buildIndex("indexByName", Vertex.class).addKey("name").buildMixedIndex("search");
```

When you try to query the created index using:
```
org.janusgraph.core.Transaction

    /**
     * Returns a {@link org.janusgraph.core.JanusGraphIndexQuery} to query for vertices or edges against the specified indexing backend using
     * the given query string. The query string is analyzed and answered by the underlying storage backend.
     * <p/>
     * Note, that using indexQuery will may ignore modifications in the current transaction.
     *
     * @param indexName Name of the indexing backend to query as configured
     * @param query Query string
     * @return JanusGraphIndexQuery object to query the index directly
     */
    public JanusGraphIndexQuery indexQuery(String indexName, String query);
```

**@param indexName Name of the indexing backend to query as configured** is misleading.

The call does NOT expect indexing backend name (in our case **search**), but the actual indexName created by janus management (in our case **indexByName**)

robertdale:
Please review the fix and close this issue if you agree.

wojciechwojcik:
This issues can be closed now. Thank you.

PR #323, opened 2017-06-08
Fixed indexName param description in 0.1 - issue #313 [skip ci]

robertdale:
Fixed indexName param description - issue #313

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java
@@ -70,7 +70,7 @@ public interface Transaction extends Graph, SchemaManager {
      * <p/>
      * Note, that using indexQuery will may ignore modifications in the current transaction.
      *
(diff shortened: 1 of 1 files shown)


mbrukman:
FYI, comment changes like this a great time to add `[skip ci]` to commit title to save CPU on Travis CI, which lets us get more build time for the other changes which actually change the code.

The tests actually run twice for each PR:

1. when the PR is submitted for review
1. when the PR is merged to the base branch

Having `[skip ci]` in the commit skips the first one, but the merge commit also needs it, so having it in the title (first line of commit) helps it easily propagate to both places.

I've updated the title of this PR and [the other PR](https://github.com/JanusGraph/janusgraph/pull/324) to avoid the CPU hit on the merge in both cases.

robertdale:
@mbrukman Thanks for the info. Was not aware of that.

mbrukman:
Credit belongs to @sjudeng who pointed this out to me. :-)

Issue #335, opened 2017-06-12
Split index in Elasticsearch


davidclement90:
Currently, JanusGraph use one Elasticsearch index for all its indexes and use elasticsearch type for choose the right index.
But this leads to : 
1. I can not delete data in Elasticsearch when I delete an index in JanusGraph because it takes too much time.
2. "Sparcity :  Types almost always increase the sparsity of your data because different types have different fields."
3. "Scoring: Documents are scored by index and not by type so storing different entities in the same index can interfere with the relevance calculation for each entity type."
https://www.elastic.co/blog/index-type-parent-child-join-now-future-in-elasticsearch

So I think we should have one Elasticsearch index by JanusGraph index.

PR #336, opened 2017-06-12
Split Elasticsearch index into N indexes.

davidclement90:
With this pull request, JanusGraph will use one Elasticsearch index by JanusGraph index instead of use one Elasticsearch index for all JanusGraph indexes.
Issue : https://github.com/JanusGraph/janusgraph/issues/335

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -22,7 +22,6 @@ import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
 import org.locationtech.spatial4j.shape.Rectangle;
 import org.apache.commons.lang.StringUtils;
-import org.apache.tinkerpop.shaded.jackson.core.JsonProcessingException;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectWriter;
 import org.apache.tinkerpop.shaded.jackson.databind.SerializationFeature;
@@ -103,12 +102,12 @@ public class ElasticSearchIndex implements IndexProvider {
             new ConfigNamespace(INDEX_NS, "elasticsearch", "Elasticsearch index configuration");
 
     public static final ConfigOption<String> CLUSTER_NAME =
-            new ConfigOption<String>(ELASTICSEARCH_NS, "cluster-name",
+            new ConfigOption<>(ELASTICSEARCH_NS, "cluster-name",
             "The name of the Elasticsearch cluster.  This should match the \"cluster.name\" setting " +
             "in the Elasticsearch nodes' configuration.", ConfigOption.Type.GLOBAL_OFFLINE, "elasticsearch");
 
     public static final ConfigOption<Boolean> CLIENT_SNIFF =
-            new ConfigOption<Boolean>(ELASTICSEARCH_NS, "sniff",
+            new ConfigOption<>(ELASTICSEARCH_NS, "sniff",
             "Whether to enable cluster sniffing.  This option only applies to the TransportClient.  " +
             "Enabling this option makes the TransportClient attempt to discover other cluster nodes " +
             "besides those in the initial host list provided at startup.", ConfigOption.Type.MASKABLE, true);
@@ -122,20 +121,20 @@ public class ElasticSearchIndex implements IndexProvider {
             disallowEmpty(String.class));
 
     public static final ConfigOption<Boolean> IGNORE_CLUSTER_NAME =
-            new ConfigOption<Boolean>(ELASTICSEARCH_NS, "ignore-cluster-name",
+            new ConfigOption<>(ELASTICSEARCH_NS, "ignore-cluster-name",
             "Whether to bypass validation of the cluster name of connected nodes.  " +
             "This option is only used on the interface configuration track (see manual for " +
             "information about ES config tracks).", ConfigOption.Type.MASKABLE, true);
 
     public static final ConfigOption<String> HEALTH_REQUEST_TIMEOUT =
-            new ConfigOption<String>(ELASTICSEARCH_NS, "health-request-timeout",
+            new ConfigOption<>(ELASTICSEARCH_NS, "health-request-timeout",
             "When JanusGraph initializes its ES backend, JanusGraph waits up to this duration for the " +
             "ES cluster health to reach at least yellow status.  " +
             "This string should be formatted as a natural number followed by the lowercase letter " +
             "\"s\", e.g. 3s or 60s.", ConfigOption.Type.MASKABLE, "30s");
 
     public static final ConfigOption<String> BULK_REFRESH =
-            new ConfigOption<String>(ELASTICSEARCH_NS, "bulk-refresh",
+            new ConfigOption<>(ELASTICSEARCH_NS, "bulk-refresh",
             "Elasticsearch bulk API refresh setting used to control when changes made by this request are made " +
             "visible to search", ConfigOption.Type.MASKABLE, "false");
 
@@ -146,7 +145,7 @@ public class ElasticSearchIndex implements IndexProvider {
             new ConfigNamespace(ELASTICSEARCH_NS, "create", "Settings related to index creation");
 
     public static final ConfigOption<Long> CREATE_SLEEP =
-            new ConfigOption<Long>(ES_CREATE_NS, "sleep",
+            new ConfigOption<>(ES_CREATE_NS, "sleep",
             "How long to sleep, in milliseconds, between the successful completion of a (blocking) index " +
             "creation request and the first use of that index.  This only applies when creating an index in ES, " +
             "which typically only happens the first time JanusGraph is started on top of ES. If the index JanusGraph is " +
@@ -156,11 +155,12 @@ public class ElasticSearchIndex implements IndexProvider {
             new ConfigNamespace(ES_CREATE_NS, "ext", "Overrides for arbitrary settings applied at index creation", true);
 
     public static final ConfigOption<Boolean> USE_EXTERNAL_MAPPINGS =
-            new ConfigOption<Boolean>(ES_CREATE_NS, "use-external-mappings",
+            new ConfigOption<>(ES_CREATE_NS, "use-external-mappings",
             "Whether JanusGraph should make use of an external mapping when registering an index.", ConfigOption.Type.MASKABLE, false);
 
-    private static final IndexFeatures ES_FEATURES = new IndexFeatures.Builder()
-            .setDefaultStringMapping(Mapping.TEXT).supportedStringMappings(Mapping.TEXT, Mapping.TEXTSTRING, Mapping.STRING).setWildcardField("_all").supportsCardinality(Cardinality.SINGLE).supportsCardinality(Cardinality.LIST).supportsCardinality(Cardinality.SET).supportsNanoseconds().supportsCustomAnalyzer().build();
+    public static final ConfigOption<Boolean> USE_DEPRECATED_MULTITYPE_INDEX =
+            new ConfigOption<>(ELASTICSEARCH_NS, "use-deprecated-multitype-index",
+            "Whether JanusGraph should group these indices into a single Elasticsearch index.", ConfigOption.Type.GLOBAL_OFFLINE, false);
 
     public static final int HOST_PORT_DEFAULT = 9200;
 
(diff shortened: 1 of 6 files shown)


janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -133,7 +133,7 @@ public void clusterHealthRequest(String timeout) throws IOException {
     public boolean indexExists(String indexName) throws IOException {
         boolean exists = false;
         try {
-            delegate.performRequest("GET", "/" + indexName);
+            delegate.performRequest("GET", "/" + indexName.toLowerCase());
```
davidclement90: Ok,
I also protect indexName because it can have uppercase ?

sjudeng: It's a store parameter and is used directly without case check in the existing implementation. So I think it's okay not to protect it but if you want to improve what about checking it in the ElasticSearchIndex constructor when it's first read and throw an illegal argument exception if it's not lower case?

robertdale:
Does this change any of the documentation or configuration possibilities?

davidclement90:
I did not find any doc on how JanusGraph push its indexes in ES.
I can add a paragraph ?

I do not make any switch via the configuration to keep backward compatibility.
If 0.2.0 keep the compatibility with 0.1.x, I can make easily this switch.

sjudeng:
I hadn't tested ES 1.x support with this branch either. Running those tests it looks like they are throwing errors because [ElasticSearchIndex#clearStorage](https://github.com/davidclement90/janusgraph/blob/e1086f6423880159de3ea2dd6ba66239ba189f53/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java#L1019) deletes by indexName, which isn't found since it's an alias. If updates are made to instead get and delete the index by alias then the tests pass. So if you can make that code change no configuration is required (at least for backwards compatibility).

davidclement90:
How do you launch test over ES 1.x ?

sjudeng:
You can start a local ES 1.x instance either by [downloading binaries](https://www.elastic.co/downloads/past-releases) or using [legacy Docker image](https://hub.docker.com/_/elasticsearch/). Then run tests with the `is.external.es` flag set.

```bash
mvn clean install -pl janusgraph-es -Dis.external.es=true
```

Note you may need to enable dynamic scripting in the elasticsearch.yml.

robertdale:
Sorry, but can you help me understand this change, the use case, and benefit? Is it optional?  I read https://www.elastic.co/blog/index-vs-type and it seems "single large index is more efficient than several small indices" is contrary to this change. Of course, neither is one size fits all. So if the change is not optional, maybe it should be?   I'm not an expert and I assume you have some real-world experience to validate this change.  There are several reasons I'm probing.
- is this good for performance? for both JanusGraph and Elasticsearch
- does the management of Elasticsearch become easier, harder, or stay the same?
- how would one configure current behavior, i.e. put all indexes in one elasticsearch index ?
   - Perhaps an option like `index.[X].elasticsearch.split=false`
- should something similar be done in the solr backend? (not that you would have to do it, but more of a philosophical question for JanusGraph architecture.  Should index backends exhibit like-behavior where possible?)

sjudeng:
@robertdale There are some notes on this in #335, which include a link to a follow-up to the above article (https://www.elastic.co/blog/index-type-parent-child-join-now-future-in-elasticsearch) in which they recommend against multiple types in the same index and indicate ES 6.x will begin (and ES 7.x will finish) breaking changes to forbid the practice.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -228,29 +232,26 @@ public ElasticSearchIndex(Configuration config) throws BackendException {
...
-                log.debug("Sleeping {} ms after {} index creation returned from actionGet()", sleep, indexName);
-                Thread.sleep(sleep);
+                log.debug("Sleeping {} ms after {} index creation returned from actionGet()", createSleep, index);
+                Thread.sleep(createSleep);
             } catch (InterruptedException e) {
                 throw new JanusGraphException("Interrupted while waiting for index to settle in", e);
             }
         }
-        if (!client.indexExists(indexName)) throw new IllegalArgumentException("Could not create index: " + indexName);
+        if (!client.indexExists(index)) throw new IllegalArgumentException("Could not create index: " + index);
```
amcp: use Preconditions.checkState

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -213,11 +215,13 @@ public ElasticSearchIndex(Configuration config) throws BackendException {
 
         try {
             client.clusterHealthRequest(config.get(HEALTH_REQUEST_TIMEOUT));
-            checkForOrCreateIndex(config);
         } catch (IOException e) {
             throw new PermanentBackendException(e);
```
amcp: add a message to this PermanentBackendException

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -164,7 +170,7 @@ public void createMapping(String indexName, String typeName, Map<String,Object>
 
     @Override
     public Map getMapping(String indexName, String typeName) throws IOException{
-        Response response = performRequest("GET", "/" + indexName.toLowerCase() + "/_mapping/"+typeName, null);
+        Response response = performRequest("GET", "/" + indexName + "/_mapping/" + typeName, null);
```
amcp: I see GET appear a few times. Please create string constants for all the HTTP verbs and reuse them.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -186,9 +192,8 @@ public void deleteIndex(String indexName) throws IOException {
     public void bulkRequest(List<ElasticSearchMutation> requests) throws IOException {
         final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         for (final ElasticSearchMutation request : requests) {
-            Map actionData = ImmutableMap.of(request.getRequestType().name().toLowerCase(),
-                ImmutableMap.of("_index", request.getIndex(), "_type", request.getType(), "_id", request.getId()));
-            outputStream.write(mapWriter.writeValueAsBytes(actionData));
+            final StringBuilder actionData = new StringBuilder("{\"").append(request.getRequestType().name().toLowerCase()).append("\":{\"_index\":\"").append(request.getIndex()).append("\", \"_type\" : \"").append(request.getType()).append("\", \"_id\" : \"").append(request.getId()).append("\"}}");
```
amcp: are you building JSON by hand? can you use a library to do this to reduce the potential of syntax errors etc?

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchConfigTest.java
```diff
@@ -167,29 +170,34 @@ public void testIndexCreationOptions() throws InterruptedException, BackendExcep
...
             //Test create index KO mapping is not push
             try {
-                new ElasticSearchIndex(indexConfig);
+                final IndexProvider idx = new  ElasticSearchIndex(indexConfig);
+                final KeyInformation.IndexRetriever indexRetriever = IndexProviderTest
+                        .getIndexRetriever(IndexProviderTest.getMapping(idx.getFeatures(), ANALYZER_ENGLISH, ANALYZER_KEYWORD));
+                final BaseTransactionConfig txConfig = StandardBaseTransactionConfig.of(TimestampProviders.MILLI);
+                final IndexTransaction itx = new IndexTransaction(idx, indexRetriever, txConfig, maxWrite);
+                idx.register(storeName, "date", IndexProviderTest.getMapping(idx.getFeatures(), ANALYZER_ENGLISH, ANALYZER_KEYWORD).get("date"), itx);
                 fail("should failed");
```
amcp: minor: should fail

davidclement90:
@robertdale : Solr backend already split its index via collection.(https://wiki.apache.org/solr/SolrTerminology)
With this PR, in all three search backend we will can drop easily an index.

sjudeng:
@robertdale's question got me thinking ... is this compatible with existing Titan/JanusGraph installations (e.g. how will it behave if there's already a populated janusgraph index)? If that's a problem it might still be worth considering starting with a configuration item for this and looking at more gently deprecating multiple-type indices in JanusGraph.

davidclement90:
No is not compatible with existing Titan/JanusGraph installations but 0.2.0 already break backward compatibility with geoshape serialization? 
Otherwise: 
1. We create a script to migrate index.
2. We make the configuration switch.

I prefer 1 because I would like to add a delete method by index to solve https://github.com/JanusGraph/janusgraph/issues/354 and to remove this https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRemoveJob.java#L105

sjudeng:
Can you start a discussion on "potential breaking changes in 0.2.0" on the dev list (https://groups.google.com/forum/#!forum/janusgraph-dev) and see if we can get some other opinions/consensus on this? I'm okay with your proposal but let's make it more visible.

Regarding geoshape serialization being a breaking change in 0.2.0, could you elaborate? Pre-0.2.0 indices would only have serialized Geoshape.Point (e.g. no lines/polys) and I thought the geo-point serialization path is unchanged.

davidclement90:
For me, there is no problem in serialization on ES indices, but there is problem in serialization in primary backend (https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java#L542).
In 0.1.x Geoshape serialization was en tab of float, in 0.2.0 JanusGraph use JtsBinaryCodec. So a geoshape serializes in 0.1.x can not be read by 0.2.0. 

I add a discussion in janusgraph-dev : https://groups.google.com/forum/#!topic/janusgraph-dev/GRGWJ44VlOM.

davidclement90:
So now, if index.[X].elasticsearch.use-deprecated-multitype-index to true, so JanusGraph will query Elasticsearch in legacy mode.

sjudeng:
With this I assume someone with a legacy graph/index would have to add index.[X].elasticsearch.use-deprecated-multitype-index=true to their config, correct? If so is it possible to remove this requirement? Otherwise I wonder if it should be made required since it seems like it could be pretty damaging if it's not set correctly.

davidclement90:
I change the constructor to check if the config is good.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchMutliTypeIndexTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
+import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
+import static org.janusgraph.diskstorage.es.ElasticSearchIndex.BULK_REFRESH;
+import static org.janusgraph.diskstorage.es.ElasticSearchIndex.INTERFACE;
+import static org.janusgraph.diskstorage.es.ElasticSearchIndex.USE_DEPRECATED_MULTITYPE_INDEX;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INDEX_HOSTS;
+
+/**
+ * @author Matthias Broecheler (me@matthiasb.com)
+ */
```
davidclement90: It was a miss in my copy past.

sjudeng:
@amcp Can you check that your requested changes are all resolved here?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -159,8 +158,9 @@
             new ConfigOption<Boolean>(ES_CREATE_NS, "use-external-mappings",
             "Whether JanusGraph should make use of an external mapping when registering an index.", ConfigOption.Type.MASKABLE, false);
 
-    private static final IndexFeatures ES_FEATURES = new IndexFeatures.Builder()
-            .setDefaultStringMapping(Mapping.TEXT).supportedStringMappings(Mapping.TEXT, Mapping.TEXTSTRING, Mapping.STRING).setWildcardField("_all").supportsCardinality(Cardinality.SINGLE).supportsCardinality(Cardinality.LIST).supportsCardinality(Cardinality.SET).supportsNanoseconds().supportsCustomAnalyzer().build();
+    public static final ConfigOption<Boolean> USE_DEPRECATED_MULTITYPE_INDEX =
+            new ConfigOption<Boolean>(ELASTICSEARCH_NS, "use-deprecated-multitype-index",
```
amcp: could be new ConfigOption<>?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -47,11 +49,25 @@
 
 public class RestElasticSearchClient implements ElasticSearchClient {
 
+
     private static final Logger log = LoggerFactory.getLogger(RestElasticSearchClient.class);
 
+    private static final String REQUEST_TYPE_DELETE = "DELETE";
+
```
amcp: minor, remove extra lines between the constants, lets make these into a group of constants.

sjudeng:
@amcp Looks like this is ready for merge?

amcp:
Yes. Merging.

Issue #239, opened 2017-04-27
JanusGraph search predicates and data types should be supported in remote traversals


robertdale:
Console:
```
gremlin> graph = EmptyGraph.instance()
==>emptygraph[empty]
gremlin> g = graph.traversal().withRemote('conf/remote-graph.properties')
==>graphtraversalsource[emptygraph[empty], standard]
gremlin> g.V().has('name',textContains('marko')).valueMap(true)
17:39:00 ERROR org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler  - Could not process the response
java.lang.NullPointerException
	at java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:936)
	at org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler.channelRead0(Handler.java:196)
	at org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler.channelRead0(Handler.java:184)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:308)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:294)
	at org.apache.tinkerpop.gremlin.driver.Handler$GremlinSaslAuthenticationHandler.channelRead0(Handler.java:112)
	at org.apache.tinkerpop.gremlin.driver.Handler$GremlinSaslAuthenticationHandler.channelRead0(Handler.java:66)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:308)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:294)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:308)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:294)
[...]

PR #339, opened 2017-06-14
Issue #239 Serialize geo and text search predicates in gryo (0.1)

robertdale:
Issue #239 

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Text.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Text.java
@@ -14,17 +14,20 @@
 
 package org.janusgraph.core.attribute;
 
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Sets;
-import org.janusgraph.graphdb.query.JanusGraphPredicate;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
 import org.apache.commons.lang.StringUtils;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
+import org.janusgraph.graphdb.query.JanusGraphPredicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
 
 /**
  * Comparison relations for text objects. These comparisons are based on a tokenized representation
@@ -66,6 +69,12 @@ public enum Text implements JanusGraphPredicate {
             else if (condition instanceof String && StringUtils.isNotBlank((String) condition)) return true;
             else return false;
         }
+
+        @Override
+        public String toString() {
+            return "textContains";
+        }
+
     },
 
     /**
@@ -90,6 +99,11 @@ public enum Text implements JanusGraphPredicate {
         @Override
         public boolean isValidCondition(Object condition) {
             return condition != null && condition instanceof String;
+        }
+
+        @Override
+        public String toString() {
+            return "textContainsPrefix";
         }
 
     },
@@ -118,6 +132,11 @@ public enum Text implements JanusGraphPredicate {
             return condition != null && condition instanceof String && StringUtils.isNotBlank(condition.toString());
         }
 
+        @Override
+        public String toString() {
+            return "textContainsRegex";
+        }
+
     },
 
     /**
@@ -141,6 +160,11 @@ public enum Text implements JanusGraphPredicate {
             return condition != null && condition instanceof String;
         }
 
+        @Override
+        public String toString() {
+            return "textPrefix";
+        }
+
     },
 
     /**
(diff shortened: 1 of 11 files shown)


robertdale:
I have a separate commit for master that I'll submit after we work through this one.

sjudeng:
Looks like there are some test errors with this update. See example exception below and [Travis job output](https://travis-ci.org/JanusGraph/janusgraph/jobs/242790023).

```
Caused by: java.lang.IllegalArgumentException: Text mapped string values only support CONTAINS queries and not: textContains
```

robertdale:
Yeah, there's logic that switches between using enum comparison and toString comparison.  Will need to clean that up.

robertdale:
This should be good to go now.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Text.java
```diff
@@ -14,75 +14,95 @@
...
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
 
 /**
- * Comparison relations for text objects. These comparisons are based on a tokenized representation
- * of the text, i.e. the text is considered as a set of word tokens.
+ * Comparison relations for text objects. These comparisons are based on a
```
sjudeng: Is your IDE set to auto split long lines and what's the limit set at? This is another one for common style guide as mentioned above.

robertdale: Yes, some of the commits were formatting only.  I can go back through and remove those.

robertdale:
Pushed changes without formatting changes.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphPSerializer.java
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Serialize Janus Graph Predicates.
+ * 
+ * Since all predicates end up being an instance of P and the TinkerPop P
+ * serializer is not extensible, it is copied here and the JanusGraph predicates
+ * are added.
+ *
+ */
+public class JanusGraphPSerializer extends Serializer<P> {
```
sjudeng: Is it possible instead to implement SerializerShim here for consistency with PSerializer? If so maybe you can pass through to the PSerializer write method to reduce some of the code redundancy. But I'm not sure the same would be possible for read() so I'm not sure it's worth it. Your call.

```java
public class JanusGraphPSerializer implements SerializerShim<P> {

    private GryoSerializers.PSerializer delegate = new GryoSerializers.PSerializer();

    @Override
    public <O extends OutputShim> void write(final KryoShim<?, O> kryo, final O output, final P p) {
        delegate.write(kryo, output, p);
    }
    ...
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphPSerializer.java
```diff
@@ -0,0 +1,128 @@
...
+        } else {
+            value = kryo.readClassAndObject(input);
+        }
+
+        try {
+
+            if (value instanceof Collection) {
+                switch (predicate) {
+                case "and":
+                    return new AndP((List<P>) value);
```
sjudeng: Should the "and" and "or" cases be in the else (not collection) case as in [PSerializer.read](https://github.com/apache/tinkerpop/blob/dfba83f046cc1c54ddc5d2ee26a8228c7af6bf1b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoSerializers.java#L209-L210)?

robertdale: No, they are always a List hence the cast to List<P> thus they are safe to put in the Collection check.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphPSerializer.java
robertdale: I went the path of duplicating it entirely because JanusGraph 0.1 depends on TinkerPop 3.2.3 which has a `and, or` serialization bug. Less of an issue on master where it depends on TP 3.2.4 containing the fix.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphPSerializer.java
sjudeng: Now I see. Thanks!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphPSerializer.java
sjudeng: Works for me. Can you add a test? I didn't see one for PSerializer that you could model after so maybe just a new case in [JanusGraphIoTest](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java)?

robertdale: @sjudeng tests added and passed

Issue #306, opened 2017-06-06
AttributeSerializer conflict on HashMap data type for migrated Titan graph


pluradj:
JanusGraph now has a [native serializer](https://github.com/JanusGraph/janusgraph/blame/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java#L133) for the HashMap class and that's clashing with the custom HashMap serializer they had [added to Titan](http://s3.thinkaurelius.com/docs/titan/1.0.0/serializer.html) (even after the serializer is refactored to use the JanusGraph packages). So when starting up against the existing DB it fails with the following:

```
java.lang.IllegalArgumentException: DataType has already been registered: class java.util.HashMap
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:146)
	at org.janusgraph.graphdb.database.serialize.StandardSerializer.registerClassInternal(StandardSerializer.java:147)
	at org.janusgraph.graphdb.database.serialize.StandardSerializer.registerClass(StandardSerializer.java:141)
	at org.janusgraph.graphdb.configuration.RegisteredAttributeClass.registerWith(RegisteredAttributeClass.java:43)
	at org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.getSerializer(GraphDatabaseConfiguration.java:1869)
	at org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.getSerializer(GraphDatabaseConfiguration.java:1862)
	at org.janusgraph.graphdb.database.StandardJanusGraph.<init>(StandardJanusGraph.java:139)
	at org.janusgraph.core.JanusGraphFactory.open(JanusGraphFactory.java:107)
	at org.janusgraph.core.JanusGraphFactory.open(JanusGraphFactory.java:97)
	at org.janusgraph.core.JanusGraphFactory$Builder.open(JanusGraphFactory.java:152)
	...
```

Currently there is no way to unregister the default HashMap serializer from `StandardSerializer` or a way to override it with the user's preferred serializer.

sjudeng:
[...]

PR #342, opened 2017-06-15
AttributeSerializer conflict on HashMap data type for migrated Titan Graph

smcquillan:
This it to fix Issue #306 which prevents the JanusGraph from starting from an existing TitanGraph if that graph had had a custom serializer registered. 

The custom serializers are added  `StandardSerializer.registerClass()`. At that point we’ll not have used the default HashMap serializer yet. So if we spot an attempt to add a custom one in this method then de-register the default one and then register the custom serializer as normal.

For the unit test I extended the SerializableSerializer class to add an extra property into the HashMap when its deserialized, then check for this property which would only be present if the custom serialiser had been used.


smcquillan:
Closing until I figure out why it didn't pick up the fact that I should be on the CLA

PR #343, opened 2017-06-15
AttributeSerializer conflict on HashMap data type for migrated Titan

smcquillan:
This it to fix Issue #306 which prevents the JanusGraph from starting from an existing TitanGraph if that graph had had a custom serializer registered.

The custom serializers are added `StandardSerializer.registerClass()`. At that point we’ll not have used the default HashMap serializer yet. So if we spot an attempt to add a custom one in this method then de-register the default one and then register the custom serializer as normal.

For the unit test I extended the SerializableSerializer class to add an extra property into the HashMap when its deserialized, then check for this property which would only be present if the custom serialiser had been used.

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
@@ -138,6 +138,16 @@ public class StandardSerializer implements AttributeHandler, Serializer {
     public synchronized <V> void registerClass(int registrationNo, Class<V> datatype, AttributeSerializer<V> serializer) {
         Preconditions.checkArgument(registrationNo >= 0 && registrationNo < MAX_REGISTRATION_NO, "Registration number" +
                 " out of range [0,%s]: %s", MAX_REGISTRATION_NO, registrationNo);
+
+        if (datatype == HashMap.class) {
+            final Integer hashMapRegNo = registrations.inverse().get(normalizeDataType(HashMap.class));
+            if (hashMapRegNo != null) {
+            	// Remove the default HashMap serializer so we can replace it with the custom one
+            	registrations.remove(hashMapRegNo);
+            	handlers.remove(datatype);
+            }
+        }
+
         registerClassInternal(CLASS_REGISTRATION_OFFSET + registrationNo, datatype, serializer);
     }
 
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/attributes/THashMapSerializer.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/attributes/THashMapSerializer.java
@@ -1,0 +1,33 @@
+// Copyright 2017 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.serializer.attributes;
+
+import java.util.HashMap;
+
+import org.janusgraph.diskstorage.ScanBuffer;
+import org.janusgraph.graphdb.database.serialize.attribute.SerializableSerializer;
+
+/**
+ * @author Scott McQuillan (scott.mcquillan@uk.ibm.com)
+ */
+public class THashMapSerializer extends SerializableSerializer<HashMap> {
+
+    @Override
+    public HashMap read(ScanBuffer buffer) {
+        HashMap hashMap = super.read(buffer);
+        hashMap.put(THashMapSerializer.class.getName(), null); // Unit test will check this got added
+        return hashMap;
+    }
+}
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java
@@ -234,6 +234,27 @@ public class SerializerTest extends SerializerTestCommon {
         assertFalse(b.hasRemaining());
 
     }
+    
+    @Test
+    public void customHashMapSerializeTest() {
+        serialize.registerClass(1,HashMap.class, new THashMapSerializer());
+        DataOutput out = serialize.getDataOutput(128);
+
+        final String property1 = "property1";
+        final String value1 = "value1";
+        HashMap<String, Object> hashMapIn = new HashMap<String, Object>();
+        hashMapIn.put(property1, value1);
+        out.writeObjectNotNull(hashMapIn);
+
+        ReadBuffer b = out.getStaticBuffer().asReadBuffer();
+        if (printStats) log.debug(bufferStats(b));
+        
+        HashMap<String, Object> hashMapOut = serialize.readObjectNotNull(b, HashMap.class);
+        assertNotNull(hashMapOut);
+        assertEquals(2, hashMapOut.size());
+        assertEquals(value1, hashMapOut.get(property1));
+        assertTrue(hashMapOut.containsKey(THashMapSerializer.class.getName())); // THashMapSerializer adds this
+    }
 
     private StaticBuffer getStringBuffer(String value) {
         DataOutput o = serialize.getDataOutput(value.length()+10);


smcquillan:
Need to follow up on why I'm getting the "clan: no" tag as I should be cleared

sjudeng:
@mbrukman Is it obvious to you what's going on with the cla check here?

smcquillan:
Thanks to @pluradj for spotting the problem.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
```diff
@@ -138,6 +138,14 @@ public StandardSerializer() {
     public synchronized <V> void registerClass(int registrationNo, Class<V> datatype, AttributeSerializer<V> serializer) {
         Preconditions.checkArgument(registrationNo >= 0 && registrationNo < MAX_REGISTRATION_NO, "Registration number" +
                 " out of range [0,%s]: %s", MAX_REGISTRATION_NO, registrationNo);
+
+        if (datatype == HashMap.class)
+        {
```
sjudeng: Write on one line (e.g. `if (...) {`) for consistency with style used elsewhere in class.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
```diff
@@ -138,6 +138,14 @@ public StandardSerializer() {
...
+            Integer hashMapRegNo = registrations.inverse().get(normalizeDataType(HashMap.class));
+            registrations.remove(hashMapRegNo);
```
sjudeng: final int hashMapRegNo

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/attributes/THashMapSerializer.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.HashMap;
+
+import org.janusgraph.diskstorage.ScanBuffer;
+import org.janusgraph.graphdb.database.serialize.attribute.SerializableSerializer;
+
+/**
+ * @author Scott McQuillan (scott.mcquillan@uk.ibm.com)
+ */
+public class THashMapSerializer extends SerializableSerializer<HashMap<String, ?>> {
```
sjudeng: Drop `<String, ?>` (this allows `serialize.registerClass(1001, HashMap.class, new THashMapSerializer())`).

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/CustomHashMapSerializer.java
```diff
@@ -0,0 +1,82 @@
...
+
+import static org.junit.Assert.*;
+
+import java.util.HashMap;
+
+/**
+ *  @author Scott McQuillan (scott.mcquillan@uk.ibm.com)
+ */
+public class CustomHashMapSerializer {
+
```
sjudeng: This class is testing a lot more than just your feature. What about instead add a new test case to [SerializerTest](https://github.com/JanusGraph/janusgraph/blob/cb9c06aeb8ab95c3a4c4cc7575b7ad97ff79ddca/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java)?

```java
    @Test
    public void testCustomHashMapSerializer() {
        serialize.registerClass(1001, HashMap.class, new THashMapSerializer());
        ...
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
smcquillan: done

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
smcquillan: done

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/attributes/THashMapSerializer.java
smcquillan: done

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/CustomHashMapSerializer.java
smcquillan: Didn't seem appropriate to add as an extra test to [SerializerTest](https://github.com/JanusGraph/janusgraph/blob/cb9c06aeb8ab95c3a4c4cc7575b7ad97ff79ddca/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java) as that class wasn't creating any instances of `StandardJanusGraph`. But the [SerializerGraphConfiguration](https://github.com/JanusGraph/janusgraph/blob/cb9c06aeb8ab95c3a4c4cc7575b7ad97ff79ddca/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java) unit tests was and it seemed a more suitable place to add the new `testCustomHashMapSerializer` so I've moved it into there

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
```diff
@@ -138,6 +138,13 @@ public StandardSerializer() {
     public synchronized <V> void registerClass(int registrationNo, Class<V> datatype, AttributeSerializer<V> serializer) {
         Preconditions.checkArgument(registrationNo >= 0 && registrationNo < MAX_REGISTRATION_NO, "Registration number" +
                 " out of range [0,%s]: %s", MAX_REGISTRATION_NO, registrationNo);
+
+        if (datatype == HashMap.class) {
+            final Integer hashMapRegNo = registrations.inverse().get(normalizeDataType(HashMap.class));
```
sjudeng: Handle null?

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
```diff
@@ -89,5 +93,34 @@ public void testOnlyRegisteredSerialization() {
         }
     }
 
+    @Test
+    public void testCustomHashMapSerialization() {
+        JanusGraphManagement mgmt = graph.openManagement();
+        PropertyKey time = mgmt.makePropertyKey("time").dataType(Integer.class).make();
```
sjudeng: If you just want to test your custom attribute registration I think you could add a test to SerializerTest. In this case I agree your test wouldn't include a graph at all and would be more like [SerializerTest#enumSerializeTest](https://github.com/JanusGraph/janusgraph/blob/cb9c06aeb8ab95c3a4c4cc7575b7ad97ff79ddca/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java#L225-L236) (replacing enum test data/asserts with a HashMap). For the test as you have it written this looks like a good place for it but I'm not sure it runs by default under surefire (the class name doesn't end in "Test" and I don't see it in any test includes).

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
```diff
@@ -41,6 +43,8 @@ public void initialize() {
         config.set(GraphDatabaseConfiguration.CUSTOM_SERIALIZER_CLASS, TClass1Serializer.class.getName(), "attribute1");
         config.set(GraphDatabaseConfiguration.CUSTOM_ATTRIBUTE_CLASS, TEnum.class.getName(), "attribute4");
```
amcp: externalize attribute4

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
```diff
@@ -41,6 +43,8 @@ public void initialize() {
...
         config.set(GraphDatabaseConfiguration.CUSTOM_SERIALIZER_CLASS, TEnumSerializer.class.getName(), "attribute4");
+        config.set(GraphDatabaseConfiguration.CUSTOM_ATTRIBUTE_CLASS, HashMap.class.getName(), "attribute5");
```
amcp: externalize attribute5

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
amcp: externalize time

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
```diff
@@ -89,5 +93,34 @@ public void testOnlyRegisteredSerialization() {
         }
     }
 
+    @Test
+    public void testCustomHashMapSerialization() {
+        JanusGraphManagement mgmt = graph.openManagement();
+        PropertyKey time = mgmt.makePropertyKey("time").dataType(Integer.class).make();
+        mgmt.makePropertyKey("hashMap").cardinality(Cardinality.SINGLE).dataType(HashMap.class).make();
```
amcp: externalize hashMap

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
```diff
@@ -89,5 +93,34 @@ public void testOnlyRegisteredSerialization() {
...
+        mgmt.buildIndex("byTime",Vertex.class).addKey(time).buildCompositeIndex();
+        mgmt.makeVertexLabel("person").make();
```
amcp: externalize person

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java
smcquillan: done

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
smcquillan: @sjudeng good catch, that unit test would be no good if it wasn't getting run. I converted it into one similar to the [SerializerTest#enumSerializeTest ](https://github.com/JanusGraph/janusgraph/blob/cb9c06aeb8ab95c3a4c4cc7575b7ad97ff79ddca/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java#L225-L236) as suggested and removed the other one.

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
smcquillan: moot point now that the test has moved.

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
smcquillan: moot point now that the test has moved.

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
smcquillan: moot point now that the test has moved.

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerGraphConfiguration.java
smcquillan: moot point now that the test has moved.

amcp:
@smcquillian can you rebase on master and squash your commits?

smcquillan:
@amcp have pushed up a rebase on master with squashed commits

PR #345, opened 2017-06-15
withRemote() config [skip ci]

robertdale:
Added property file for use with withRemote()


pluradj:
LGTM, the configuration file works fine. It would be great to add some documentation on how to use it in the [JanusGraph Server](https://github.com/JanusGraph/janusgraph/blob/0.1/docs/basics.txt#L478) section.

amcp:
what issue does this PR address?

robertdale:
This is the default remote graph config for gremlin server. It is for submitting remote traversals to gremlin server. 
The issue it solves is convenience and providing complete configurations.

See also http://tinkerpop.apache.org/docs/current/reference/#connecting-via-remotegraph

PR #346, opened 2017-06-15
Added property file for use with withRemote() [skip ci]

robertdale:


amcp:
@robertdale what issue does this PR address?

amcp:
@robertdale do you want #345 or #346 merged? it seems like they have similar diff content.

robertdale:
@amcp This one is for master, #345 is on 0.1.  I would like to see both branches get the change. I don't have an opinion for how it gets there.

PR #360, opened 2017-06-24
Fix assertion error with list cardinality

sjudeng:
An assertion in the wrong spot causes failure of `GraphComputerTest.shouldSucceedWithProperTraverserRequirements`.


sjudeng:
Sorry I added the assert based on your request but I thought it was innocuous enough of a change that I didn't need to rerun the full TP test suite. I caught it when rerunning for a separate PR.

amcp:
@sjudeng no problem, thanks for catching.

PR #364, opened 2017-06-25
Add Uber as a production JanusGraph user [skip ci]

mbrukman:
Courtesy to use the company name courtesy of @tunaygur.

Also, note that we don't yet have permission to feature Uber's logo on http://janusgraph.org so there's no corresponding PR on that repo to add it.


mbrukman:
FWIW, this was previously announced publicly in:
https://twitter.com/JanusGraph/status/875806752767262720

PR #369, opened 2017-06-27
Support HBase cell level TTL

### LINKED ISSUE #356: Support HBase cell level TTL
JanusGraph supports vertex, edge and property level TTL. This requires backend storage to support cell level TTL.   
http://docs.janusgraph.org/latest/advanced-schema.html
The newer versions HBase support cell level TTL:
https://issues.apache.org/jira/browse/HBASE-10560
This is missing from the current JanusGraph HBase support.  Let's add it.

jerryjch:
Fix #356

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
@@ -659,11 +659,11 @@ public class GraphDatabaseConfiguration {
             ConfigOption.Type.GLOBAL, false);
 
     public static final ConfigOption<Boolean> STORE_META_TTL = new ConfigOption<>(STORE_META_NS,"ttl",
-            "Whether to include ttl in retrieved entries for storage backends that automatically annotated entries with timestamps",
+            "Whether to include ttl in retrieved entries for storage backends that support storage and retrieval of cell level TTL",
             ConfigOption.Type.GLOBAL, false);
 
     public static final ConfigOption<Boolean> STORE_META_VISIBILITY = new ConfigOption<>(STORE_META_NS,"visibility",
-            "Whether to include visibility in retrieved entries for storage backends that automatically annotated entries with timestamps",
+            "Whether to include visibility in retrieved entries for storage backends that support cell level visibility",
             ConfigOption.Type.GLOBAL, true);
 
 
--- a/janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
+++ b/janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
@@ -33,6 +33,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
@@ -55,6 +56,7 @@ import org.apache.hadoop.hbase.ZooKeeperConnectionException;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.HBaseAdmin;
 import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Row;
 import org.apache.hadoop.hbase.util.Pair;
@@ -63,6 +65,7 @@ import org.janusgraph.core.JanusGraphException;
 import org.janusgraph.diskstorage.BackendException;
 import org.janusgraph.diskstorage.BaseTransactionConfig;
 import org.janusgraph.diskstorage.Entry;
+import org.janusgraph.diskstorage.EntryMetaData;
 import org.janusgraph.diskstorage.PermanentBackendException;
 import org.janusgraph.diskstorage.StaticBuffer;
 import org.janusgraph.diskstorage.StoreMetaData;
@@ -395,7 +398,7 @@ public class HBaseStoreManager extends DistributedStoreManager implements KeyCol
         StandardStoreFeatures.Builder fb = new StandardStoreFeatures.Builder()
                 .orderedScan(true).unorderedScan(true).batchMutation(true)
                 .multiQuery(true).distributed(true).keyOrdered(true).storeTTL(true)
-                .timestamps(true).preferredTimestamps(PREFERRED_TIMESTAMPS)
+                .cellTTL(true).timestamps(true).preferredTimestamps(PREFERRED_TIMESTAMPS)
                 .optimisticLocking(true).keyConsistent(c);
 
         try {
@@ -413,18 +416,18 @@ public class HBaseStoreManager extends DistributedStoreManager implements KeyCol
         // In case of an addition and deletion with identical timestamps, the
         // deletion tombstone wins.
         // http://hbase.apache.org/book/versions.html#d244e4250
-        Map<StaticBuffer, Pair<Put, Delete>> commandsPerKey =
+        final Map<StaticBuffer, Pair<List<Put>, Delete>> commandsPerKey =
                 convertToCommands(
                         mutations,
                         commitTime.getAdditionTime(times),
                         commitTime.getDeletionTime(times));
 
-        List<Row> batch = new ArrayList<Row>(commandsPerKey.size()); // actual batch operation
+        final List<Row> batch = new ArrayList<>(commandsPerKey.size()); // actual batch operation
 
         // convert sorted commands into representation required for 'batch' operation
-        for (Pair<Put, Delete> commands : commandsPerKey.values()) {
-            if (commands.getFirst() != null)
-                batch.add(commands.getFirst());
+        for (Pair<List<Put>, Delete> commands : commandsPerKey.values()) {
+            if (commands.getFirst() != null && !commands.getFirst().isEmpty())
+                batch.addAll(commands.getFirst());
 
             if (commands.getSecond() != null)
                 batch.add(commands.getSecond());
@@ -450,6 +453,9 @@ public class HBaseStoreManager extends DistributedStoreManager implements KeyCol
 
     @Override
     public KeyColumnValueStore openDatabase(String longName, StoreMetaData.Container metaData) throws BackendException {
+        // HBase does not support retrieving cell-level TTL by the client.
+        Preconditions.checkArgument(!storageConfig.has(GraphDatabaseConfiguration.STORE_META_TTL, longName)
+            || !storageConfig.get(GraphDatabaseConfiguration.STORE_META_TTL, longName));
 
         HBaseKeyColumnValueStore store = openStores.get(longName);
 
@@ -699,14 +705,10 @@ public class HBaseStoreManager extends DistributedStoreManager implements KeyCol
              */
             if (adm.tableExists(tableName)) {
                 desc = adm.getTableDescriptor(tableName);
-                String fmt;
-                String msg;
                 // Check and warn if long and short cf names are mixedly used for the same table.
                 if (shortCfNames && initialCFName.equals(SHORT_CF_NAME_MAP.get(SYSTEM_PROPERTIES_STORE_NAME))) {
                     String longCFName = SHORT_CF_NAME_MAP.inverse().get(initialCFName);
                     if (desc.getFamily(longCFName.getBytes()) != null) {
-                        fmt = "Configuration {}=true, but the table \"{}\" already has column family with long name \"{}\".";
-                        msg = String.format(fmt, SHORT_CF_NAMES.getName(), tableName, longCFName);
                         logger.warn("Configuration {}=true, but the table \"{}\" already has column family with long name \"{}\".",
                             SHORT_CF_NAMES.getName(), tableName, longCFName);
                         logger.warn("Check {} configuration.", SHORT_CF_NAMES.getName());
@@ -715,8 +717,6 @@ public class HBaseStoreManager extends DistributedStoreManager implements KeyCol
                 else if (!shortCfNames && initialCFName.equals(SYSTEM_PROPERTIES_STORE_NAME)) {
                     String shortCFName = SHORT_CF_NAME_MAP.get(initialCFName);
                     if (desc.getFamily(shortCFName.getBytes()) != null) {
-                        fmt = "Configuration {}=false, but the table \"{}\" already has column family with short name \"{}\".";
-                        msg = String.format(fmt, SHORT_CF_NAMES.getName(), tableName, shortCFName);
                         logger.warn("Configuration {}=false, but the table \"{}\" already has column family with short name \"{}\".",
                             SHORT_CF_NAMES.getName(), tableName, shortCFName);
                         logger.warn("Check {} configuration.", SHORT_CF_NAMES.getName());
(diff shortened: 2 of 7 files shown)


jerryjch:
JanusGraphTest.testVertexTTLImplicitKey and JanusGraphTest.testEdgeTTLImplicitKey failed on my local machine too. They failed because HBase client does not support retrieving the cell TTLs from the server after they have been put.

jerryjch:
I didn't add any new test cases because JanusGraphTest has TTL tests, which all passed except the above two.

amcp:
@jerryjch is there a low level HBase API you can use to retrieve the cell TTLs from the server after they have been put?
Also, I agree with your test strategy. Relying on the GraphTests should be enough.

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
                 if (mutation.hasAdditions()) {
-                    if (commands.getFirst() == null) {
-                        Put p = new Put(key, putTimestamp);
-                        commands.setFirst(p);
-                    }
-
+                    // All the entries (column cells) with the rowkey use this one Put, except the ones with TTL.
+                    Put p = new Put(key, putTimestamp);
+                    
                     for (Entry e : mutation.getAdditions()) {
```
amcp: add a comment here: "At the end of this loop, there will be one Put entry in the commands.getFirst() list that contains all additions without TTL set, and one Put entry for each column that has TTL set."

amcp: also, I think you should refactor this loop into a method and test it.

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
                 }
 
                 if (mutation.hasAdditions()) {
-                    if (commands.getFirst() == null) {
-                        Put p = new Put(key, putTimestamp);
-                        commands.setFirst(p);
-                    }
-
+                    // All the entries (column cells) with the rowkey use this one Put, except the ones with TTL.
+                    Put p = new Put(key, putTimestamp);
```
amcp: probably better to rename p -> putColumnsWithoutTtl

amcp: and make final

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
+                        // HBase cell level TTL is actually set at the Mutation/Put level.
+                        // Therefore we need to construct a new Put for each entry (column cell) with TTL.
+                        // We can not combine them because column cells within the same rowkey may:
+                        // 1. have no TTL
+                        // 2. have TTL
+                        // 3. have different TTL
+                        Integer ttl = (Integer) e.getMetaData().get(EntryMetaData.TTL);
+                        if (null != ttl && ttl > 0) {
+                            // Create a new Put
+                            Put pTTL = new Put(key, putTimestamp);
```
amcp: final and rename pTTL -> putColumnWithTTL

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
-                        commands.getFirst().add(cfName,
+
+                        // Deal with TTL within the entry (column cell) first
+                        // HBase cell level TTL is actually set at the Mutation/Put level.
+                        // Therefore we need to construct a new Put for each entry (column cell) with TTL.
+                        // We can not combine them because column cells within the same rowkey may:
+                        // 1. have no TTL
+                        // 2. have TTL
+                        // 3. have different TTL
+                        Integer ttl = (Integer) e.getMetaData().get(EntryMetaData.TTL);
```
amcp: final

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
+                            // commands.getFirst() is the list of Puts for this rowkey.
+                            commands.getFirst().add(pTTL);
+                        } else {
+                            p.add(cfName,
+                                e.getColumnAs(StaticBuffer.ARRAY_FACTORY),
+                                putTimestamp,
+                                e.getValueAs(StaticBuffer.ARRAY_FACTORY));
+                        }
+                    }
+                    if (!p.isEmpty()) {
```
amcp: add a comment here: "If there were any mutations without TTL set, add them to commands.getFirst()".

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -876,10 +878,13 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
                 byte[] key = m.getKey().as(StaticBuffer.ARRAY_FACTORY);
                 KCVMutation mutation = m.getValue();
 
-                Pair<Put, Delete> commands = commandsPerKey.get(m.getKey());
+                Pair<List<Put>, Delete> commands = commandsPerKey.get(m.getKey());
 
                 if (commands == null) {
```
amcp: can you add a comment describing the situation where commands would be null?

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -422,9 +423,9 @@ public void mutateMany(Map<String, Map<StaticBuffer, KCVMutation>> mutations, St
         List<Row> batch = new ArrayList<Row>(commandsPerKey.size()); // actual batch operation
```
amcp: final

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
amcp: considering [other stores could implement TTL](https://aws.amazon.com/blogs/aws/new-manage-dynamodb-items-using-time-to-live-ttl/), should we add a method to Entry called getTTL()?

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
+                        // We can not combine them because column cells within the same rowkey may:
+                        // 1. have no TTL
+                        // 2. have TTL
+                        // 3. have different TTL
+                        Integer ttl = (Integer) e.getMetaData().get(EntryMetaData.TTL);
+                        if (null != ttl && ttl > 0) {
+                            // Create a new Put
+                            Put pTTL = new Put(key, putTimestamp);
+                            pTTL.add(cfName,
                                 e.getColumnAs(StaticBuffer.ARRAY_FACTORY),
```
amcp: Add a private static method that takes a Put and Entry as arguments and adds the column and value in entry to the put. Use that method here and in the ttl=null or ttl negative case.

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
                                 putTimestamp,
                                 e.getValueAs(StaticBuffer.ARRAY_FACTORY));
+                            // Convert ttl from second (JanusGraph TTL) to millisec (HBase TTL)
+                            pTTL.setTTL(ttl * 1000);
+                            // commands.getFirst() is the list of Puts for this rowkey.
+                            commands.getFirst().add(pTTL);
+                        } else {
+                            p.add(cfName,
```
amcp: use the method you created above to add cv to p

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
amcp: The [current implementation subclasses a map](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/EntryMetaData.java#L74). while providing the a similar interface, we could deprecate the raw .get() methods and add specific methods for getTTL, getVisibility and getTimestamp.

amcp: Also, instead of [subclassing ArrayList](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/EntryMetaData.java#L70) you could just use guava to create the IDENTIFYING_METADATA list.

amcp: Finally, dataType and isIdentifying return constant values for each of the three Enums. Can you create a package private constructor that just sets the values as part of the state of each enum instead of switch-casing?

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -891,21 +896,45 @@ private void setCFOptions(HColumnDescriptor cdesc, int ttlInSeconds) {
...
+                        // 3. have different TTL
+                        Integer ttl = (Integer) e.getMetaData().get(EntryMetaData.TTL);
+                        if (null != ttl && ttl > 0) {
+                            // Create a new Put
+                            Put pTTL = new Put(key, putTimestamp);
+                            pTTL.add(cfName,
                                 e.getColumnAs(StaticBuffer.ARRAY_FACTORY),
                                 putTimestamp,
                                 e.getValueAs(StaticBuffer.ARRAY_FACTORY));
+                            // Convert ttl from second (JanusGraph TTL) to millisec (HBase TTL)
```
amcp: Is JanusGraph TTL always seconds? Could you please add a comment pointing to the class/method where JanusGraph TTL is processed in seconds?

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
amcp: https://github.com/awslabs/dynamodb-janusgraph-storage-backend/issues/70

jerryjch:
Thanks for the comments. I will address them.

On the client API to retrieve column cell TTL, there is currently none. 
This JIRA gives some information: 
https://issues.apache.org/jira/browse/HBASE-12470
The column cell TTL are tags attached to the cell, along with visibility labels, etc.
They are striped off on server RPC before the results are returned to the server.

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
jerryjch: Can I open another ticket to add the getTTL API?

jerryjch:
@sjudeng  Yes, only these two test cases try to read TTL values.
In the HBaseKeyColumnValueStore:
```

        @Override
        public Object getMetaData(Map.Entry<byte[], NavigableMap<Long, byte[]>> element, EntryMetaData meta) {
            switch(meta) {
                case TIMESTAMP:
                    return element.getValue().lastEntry().getKey();
                default:
                    throw new UnsupportedOperationException("Unsupported meta data: " + meta);
            }
```

amcp:
@jerryjch Looks good. Is it possible to test the loop? Since we do have graph tests its ok not to but since the loop is complicated it may be worth it. Either way let me know. Then, please squash your commits and I will approve as soon as we have consensus on the ignore test question posed by @sjudeng .

jerryjch:
Thanks @amcp . I was trying to ask you about testing the loop. Do you mean to test convertToCommands() in a test case?

jerryjch:
There are test cases that do a lot of mutations and there are a few tests that do cell level TTL.
If I understand you correctly @amcp,  let me add testing to excise calls to convertToCommands() with multiple and mixed without TTL and withTTL in rows so that we are sure it works well.

amcp:
@jerryjch you got it right. Thank you for going forth with testing to exercise calls to convertToCommands()

sjudeng:
Does it make sense to check/assert that storage.meta.ttl=false in HBaseStoreManager?

```java
Preconditions.checkArgument(!config.has(STORE_META_TTL) || !config.get(STORE_META_TTL))
```

amcp:
@jerryjch #378 I did the cleanup we discussed above.

janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
amcp: yes, you can open a ticket please.

jerryjch:
@sjudeng It turns out the property STORE_META_TTL needs to go with a store name as a 'umbrella', so it will be something like this:
`        Preconditions.checkArgument(!storageConfig.has(GraphDatabaseConfiguration.STORE_META_TTL, longName)
            || !storageConfig.get(GraphDatabaseConfiguration.STORE_META_TTL, longName));`

jerryjch:
HBase cell-level TTL is available in versions 0.98.6 and above.
https://issues.apache.org/jira/browse/HBASE-10560
I will update the 0.98.x version in JanusGraph to 0.98.24 (the latest and last) as part of the changes.

jerryjch:
Ran hbase-098 tests as well.  All passed.

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -659,11 +659,11 @@ public boolean apply(@Nullable String s) {
             ConfigOption.Type.GLOBAL, false);
 
     public static final ConfigOption<Boolean> STORE_META_TTL = new ConfigOption<Boolean>(STORE_META_NS,"ttl",
-            "Whether to include ttl in retrieved entries for storage backends that automatically annotated entries with timestamps",
+            "Whether to include ttl in retrieved entries for storage backends that support cell level TTL",
```
sjudeng: "...for storage backends that support storage and retrieval of cell level TTL"

jerryjch: Ok

sjudeng:
Please squash commits for merge.

jerryjch:
Thanks for the reviews. 
I will wait a little for @mbrukman or @lindsayismith to chime in here for Bigtable since it shares the same client. I think the goals are:
1.  Does Bigtable support cell-level TTL.  I could not find much info, it does not seem to support.
2.  If it does not, the changes do not cause regression or surprise for Bigtable backend users.  If the user does not use or try to set edge, vertex or property level TTL with Bigtable backend, everything works as before.  If the user try to set cell level TTL, an exception should be thrown.  Or it is ignored with no surprise impact.
I think as long as the above is true, we are good to go.

jerryjch:
Rebased.

jerryjch:
If there is no more comment, I plan to commit it soon.

PR #370, opened 2017-06-27
Remove 3rd party stylesheet and font

### LINKED ISSUE #368: Eliminate 3rd party stylesheet and font dependencies
The Avada stylesheet is says its license is GPL. Avada also pulls in a jQuery plugin called Isotope, which has a several [license options](http://isotope.metafizzy.co/docs/license.html). Font Awesome uses [SIL OFL 1.1](http://fontawesome.io/license/).
- [avada-stub/style.css](https://github.com/JanusGraph/janusgraph/blob/v0.1.1/docs/static/avada-stub/style.css#L8)
- [avada-stub/js/main.js](https://github.com/JanusGraph/janusgraph/blob/v0.1.1/docs/static/avada-stub/js/main.js#L2167-L2170)
- [font-awesome-4.0.3/css/font-awesome.css](https://github.com/JanusGraph/janusgraph/blob/v0.1.1/docs/static/font-awesome-4.0.3/css/font-awesome.css#L3)

There's no reason other than good looks to use these artifacts. It would be cleaner for license scans of the JanusGraph source repository to remove these, and ultimately the docs won't look significantly different.

I have a pull request coming.

pluradj:
Remove 3rd party stylesheet and font. Fixes #368.
- Removed directories `avada-stub` and `font-awesome-4.0.3`
- Updated header and footer HTML/CSS
  - Added favicons
  - Replaced Titan red with JanusGraph green in some styles and architecture diagram
  - For hyperlinks, chose a blue instead for higher contrast with regular text

Documentation functionality removed/lost:
- Dynamically pinning the header at the top when scrolled
- Scroll to top indicator when scrolled

Signed-off-by: Jason Plurad <pluradj@us.ibm.com>


sjudeng:
@amcp You can view generated docs locally after building project in janusgraph-doc/target/docs/single/index.html.

docs/xsl/common.xsl
```diff
@@ -54,19 +54,10 @@
...
-    <!--[if IE]>
-    <link rel="stylesheet" href="avada-stub/css/ie.css" />
-    <![endif]-->
-
-    <link rel='stylesheet' id='style-css-css'  href='avada-stub/style.css' type='text/css' media='all' />
-    <link rel='stylesheet' id='media-css-css'  href='avada-stub/css/media.css' type='text/css' media='all' />
-    <link rel='stylesheet' id='animate-css-css'  href='avada-stub/css/animate-custom.css' type='text/css' media='all' />
-    <link rel='stylesheet' id='inline-blob-from-avada-prototype' href='avada-stub/css/inlineblob.css' type='text/css' media='all' />
-    <link rel='stylesheet' id='inline-blob-from-avada-prototype-janusgraph-docs-specific' href='css/docs.css' type='text/css' media='all' />
+    <link rel='icon' type='image/jpeg' href='images/janusgraph-logomark.png' />
```
mbrukman: This claims the icon is `image/jpeg` but the actual file is a PNG. On a related note, it would be just easier to put this file as `/favicon.ico` at the root of the directory, and we don't have to modify every single file or copy it into every version of every doc.

docs/xsl/common.xsl
```diff
@@ -54,19 +54,10 @@
...
+    <link rel='apple-touch-icon' type='image/jpeg' href='images/janusgraph-logomark.jpg' />
```
mbrukman: It would be better to use PNG than JPG for logos as PNGs are not lossy and don't have funny artifacts due to JPG compression. We have plenty of high-resolution PNGs — does this feature not accept PNGs?

docs/xsl/common.xsl
```diff
@@ -34,7 +34,7 @@
   <xsl:param name="table.cell.border.color" select="'#747474'"/>
   <xsl:param name="table.cell.border.style" select="'solid'"/>
   <xsl:param name="table.cell.border.thickness" select="'1px'"/>
-  <xsl:param name="table.frame.border.color" select="'#FF8080'"/>
+  <xsl:param name="table.frame.border.color" select="'#108768'"/>
```
mbrukman: Is the idea here to switch from red to green borders? If you want the JanusGraph green color, based on my research it's #1b8669; see [`style.css`](https://github.com/JanusGraph/janusgraph.org/blob/master/style.css#L18) in the janusgraph.org repo.

pluradj: Thanks, I'll use the same value.

docs/xsl/common.xsl
pluradj: I'll convert to the png to favicon.ico.

docs/xsl/common.xsl
pluradj: The PNG has a transparent background which looks great as the favicon. However, I found that saving the webpage to the home screen on iPhone caused the transparent background to be all black. The JPEG has a white background with no transparency, so it worked out better. I grabbed both images from the logos repo.

pluradj:
@mbrukman I've made all your suggested updates.

@amcp I tested the documentation like this
```
cd janusgraph-doc
mvn clean install
cd target/docs/chunk/
python3 -m http.server 8080
firefox http://localhost:8080
```

docs/xsl/common.xsl
mbrukman: Maybe create / use a PNG with a white background? Should be better than using JPGs.

mbrukman:
FWIW, I wrote up [these instructions](https://github.com/janusgraph/janusgraph.org#preview-via-github) for how I upload my in-progress changes to janusgraph.org so that everyone doesn't have to manually pull changes and build locally to preview the impact of a PR, maybe this could be used here as well? 

I can adapt those instructions for the docs.janusgraph.org repo, and we can add a README to the `docs/` folder of this repo for how to make changes in that repo with appropriate preview.

Thoughts?

pluradj:
OK, I'll give those instructions a shot.

mbrukman:
@pluradj — sorry for being unclear, but in this case, I was suggesting for you to create a preview of docs.janusgraph.org with the output HTML files produced by the script (since the files you're modifying drive the output of that site), not janusgraph.org which is what those instructions are talking about.

I just sent a PR to the docs.janusgraph.org repo (see link above) which clarifies the process; please try it out and let me know if anything is confusing or if you run into any issues.

PR #376, opened 2017-06-29
Support deserializing points stored in legacy graphs

### LINKED ISSUE #374: GeoshapeSerializer does not support legacy serialization method
Need to update GeoshapeSerializer to support legacy (coordinate float array) point serialization so pre-0.2.0 graphs with Geoshapes can be read in 0.2.0. Thanks to @davidclement90 for pointing this out in #336.

sjudeng:
These updates allow Geoshape points stored in legacy (pre-0.2) graphs and Kryo files to be deserialized. Also adds support for Geoshape serialization using GraphSON 2.0.

Fixes #374

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
@@ -16,11 +16,13 @@ package org.janusgraph.core.attribute;
 
 import com.google.common.base.Preconditions;
 import com.google.common.primitives.Doubles;
+import org.apache.tinkerpop.gremlin.structure.io.graphson.AbstractObjectDeserializer;
+import org.apache.tinkerpop.shaded.kryo.KryoException;
+import org.janusgraph.diskstorage.util.ReadArrayBuffer;
 import org.locationtech.spatial4j.context.SpatialContext;
 import org.locationtech.spatial4j.distance.DistanceUtils;
 import org.locationtech.spatial4j.shape.Circle;
 import org.locationtech.spatial4j.shape.Shape;
-import org.locationtech.spatial4j.shape.ShapeFactory;
 import org.locationtech.spatial4j.shape.SpatialRelation;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectReader;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectWriter;
@@ -159,7 +161,7 @@ public class Geoshape {
      * @return
      */
     public String toGeoJson() {
-        return GeoshapeGsonSerializer.toGeoJson(this);
+        return GeoshapeGsonSerializerV1d0.toGeoJson(this);
     }
 
     public Map<String,Object> toMap() throws IOException {
@@ -427,7 +429,6 @@ public class Geoshape {
 
         @Override
         public Geoshape convert(Object value) {
-
             if(value instanceof Map) {
                 return convertGeoJson(value);
             }
@@ -497,7 +498,7 @@ public class Geoshape {
 
         private Geoshape convertGeometry(Map<String, Object> geometry) throws IOException, ParseException {
             String type = (String) geometry.get("type");
-            List<Object> coordinates = (List) geometry.get("coordinates");
+            List<Object> coordinates = (List) geometry.get(FIELD_COORDINATES);
 
             if ("Point".equals(type)) {
                 double[] parsedCoordinates = convertCollection(coordinates);
@@ -547,7 +548,16 @@ public class Geoshape {
             try {
                 return GeoshapeBinarySerializer.read(inputStream);
             } catch (IOException e) {
-                throw new RuntimeException("I/O exception reading geoshape");
+                // retry using legacy point deserialization
+                try {
+                    ((ReadArrayBuffer) buffer).movePositionTo(0);
+                    VariableLong.readPositive(buffer);
+                    final float lat = buffer.getFloat();
+                    final float lon = buffer.getFloat();
+                    return point(lat, lon);
+                } catch (Exception e2) { }
+                // throw original exception
+                throw new RuntimeException("I/O exception reading geoshape", e);
             }
         }
 
@@ -560,7 +570,7 @@ public class Geoshape {
                 VariableLong.writePositive(buffer,bytes.length);
                 buffer.putBytes(bytes);
             } catch (IOException e) {
-                throw new RuntimeException("I/O exception writing geoshape");
+                throw new RuntimeException("I/O exception writing geoshape", e);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


davidclement90:
I thought about another serialization issues.

1. A gryo or graphson file write in 0.1.x, will not be able to be read by > 0.2
2. A client with JanusGraph 0.1.x will not be able to communicate with a gremlin server with JanusGraph > 0.2

I think 1 will be easy to handle (like you do for AttributeSerializer), but 2 is just impossible.

sjudeng:
@davidclement90 Yeah that's a good point. I think we can live with (2) as the workaround is just to update the client, right?

I'll add updates here to support legacy points in gryo/graphson input. Thanks for following up on this.

davidclement90:
Yes for (2), you just have to update the client.
I think we just should write this in the 0.2.0 release note.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -584,23 +595,32 @@ public void write(Kryo kryo, Output output, Geoshape geoshape) {
...
+            } catch (IOException | KryoException e) {
+                // retry using legacy point deserialization
+                try {
+                    input.setPosition(0);
+                    input.readLong();
+                    return point(input.readFloat(), input.readFloat());
+                } catch (KryoException e2) {
+                    // throw original exception
+                }
+                throw new RuntimeException("I/O exception reading geoshape");
```
amcp: should you wrap the initial exception e here? `throw new RuntimeException("I/O exception reading geoshape", e);`

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -584,23 +595,32 @@ public void write(Kryo kryo, Output output, Geoshape geoshape) {
...
-            } catch (IOException e) {
-                throw new RuntimeException("I/O exception reding geoshape");
+            } catch (IOException | KryoException e) {
+                // retry using legacy point deserialization
+                try {
+                    input.setPosition(0);
+                    input.readLong();
+                    return point(input.readFloat(), input.readFloat());
+                } catch (KryoException e2) {
+                    // throw original exception
```
amcp: is this a TODO or ??

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -547,6 +548,16 @@ public Geoshape read(ScanBuffer buffer) {
...
+                    // retry using legacy point deserialization
+                    ((ReadArrayBuffer) buffer).movePositionTo(0);
+                    VariableLong.readPositive(buffer);
+                    try {
+                        return point(buffer.getFloat(), buffer.getFloat());
+                    } catch (Exception e2) {
+                        // throw original exception
+                    }
+                }
                 throw new RuntimeException("I/O exception reading geoshape");
```
amcp: say if (!(buffer instances ReadArrayBuffer) || l != 1) throw new RuntimeException("I/O exception reading geoshape", e)
then when you catch (Exception e2), you can throw a more specific exception like "unable to create a point from buffer"

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -547,6 +548,16 @@ public Geoshape read(ScanBuffer buffer) {
             try {
                 return GeoshapeBinarySerializer.read(inputStream);
             } catch (IOException e) {
+                if (buffer instanceof ReadArrayBuffer && l == 1) {
+                    // retry using legacy point deserialization
+                    ((ReadArrayBuffer) buffer).movePositionTo(0);
+                    VariableLong.readPositive(buffer);
+                    try {
+                        return point(buffer.getFloat(), buffer.getFloat());
```
amcp: Can you add a new point() method that takes one argument instead of two? seems like you are calling the same method twice here.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -678,6 +698,63 @@ public Geoshape deserialize(JsonParser jsonParser, DeserializationContext deseri
...
+        public void serializeWithType(Geoshape geoshape, JsonGenerator jgen, SerializerProvider serializerProvider,
+                                      TypeSerializer typeSerializer) throws IOException {
+
+            jgen.writeStartObject();
+            if (typeSerializer != null) jgen.writeStringField(GraphSONTokens.VALUETYPE, "janusgraph:Geoshape");
+            jgen.writeFieldName(GraphSONTokens.VALUEPROP);
+            GraphSONUtil.writeStartObject(geoshape, jgen, typeSerializer);
+            final Map json = mapReader.readValue(toGeoJson(geoshape));
+            if (geoshape.getType() == Type.POINT) {
+                final double[] coords = ((List<Number>) json.get("coordinates")).stream().mapToDouble(i -> i.doubleValue()).toArray();
```
amcp: extract a string constant for coordinates and reuse throughout this file.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
sjudeng: @amcp It's reading from a buffer so it's actually getting a different float (e.g. lat/lon) on subsequent calls. But I updated to make this more explicit.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
sjudeng: I took out this if and put all statements in the try-catch. I thought it makes sense to throw the original exception since in the case of an error I don't think it's possible to be sure whether or not the unserializable geoshape is a legacy point or not.

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
sjudeng: I moved the comment

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
sjudeng: Yep thanks

janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
sjudeng: Done

sjudeng:
@amcp At your convenience can you confirm all your requests are in here? I'm waiting on submitting another PR until I can resolve conflicts once this is merged.

amcp:
@sjudeng I think this PR is a good approach. In the future, i would like to investigate if a more deterministic approach to determining whether an entity is a legacy type other than depending on a validation exception being thrown. The input could truly be invalid for both current and legacy, or what if the exception is thrown for some other reason. I wonder if it is possible to inspect the entity and definitively reach the conclusion that it is legacy or not? no need to address this here but please make an issue.

PR #377, opened 2017-06-30
Merge pull request #370 from pluradj/cssfix

pluradj:
Remove 3rd party stylesheet and font
(cherry picked from commit 0b77da81bf44c23dbfa6e20d04cdec6cf771a471)

Signed-off-by: Jason Plurad <pluradj@us.ibm.com>


pluradj:
@mbrukman I did the merge from the GitHub Web UI, and it ended up making a commit with an email different than the CLA, that's why the CLA checker is flagging this. Do I need to add that address to the CLA_SIGNERS.yaml or is there another way around this?

mbrukman:
@pluradj — yes, the only way is to add an alternate email address to the `CLA_SIGNERS.yaml` file; there isn't support for a single person to have multiple email addresses; please add another person record with your name / email / GitHub id to match this commit.

mbrukman:
@pluradj — thinking about this some more, perhaps we can make a special case for GitHub username `foobar` to also automatically allow `foobar@users.noreply.github.com` as an email address for patch submissions / merges; that will allow not having to add an alternate email address for everyone individually. I'll think about this some more; for now, adding another email address as I'm doing in https://github.com/JanusGraph/legal/pull/46 is the way to go; I'll do the same for you shortly.

PR #378, opened 2017-06-30
Clean up EntryMetaData enum and interned its state

amcp:
These are the cleanup tasks we discovered during the HBase TTL implementation.

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/EntryMetaData.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/EntryMetaData.java
@@ -18,59 +18,51 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import org.janusgraph.util.encoding.StringEncoding;
 
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


amcp:
@sjudeng I can rework this to not use lombok.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/EntryMetaData.java
```diff
@@ -18,59 +18,51 @@
...
-            default: throw new AssertionError("Unexpected meta data: " + this);
-        }
+        return identifying;
     }
 
-    public static final List<EntryMetaData> IDENTIFYING_METADATA = new ArrayList<EntryMetaData>(3) {{
-        for (EntryMetaData meta : values()) if (meta.isIdentifying()) add(meta);
-    }};
+    /**
+     * Validates a datum according to the metadata type.
```
sjudeng: What is a datum?

amcp: One datum two pieces of data.

Issue #372, opened 2017-06-27
Pull up privilegedMode setting to pipeline JSON file to make ES tests pass


amcp:
ES will not start if run as root. Allow pipeline build actions to choose whether or not to run as root.

PR #379, opened 2017-07-01
[skip ci] Set privileged mode on per build action basis.

amcp:

### DIFF
--- a/janusgraph-codepipelines-ci/README.md
+++ b/janusgraph-codepipelines-ci/README.md
@@ -21,8 +21,8 @@ and a default [output format](http://docs.aws.amazon.com/cli/latest/userguide/co
 of your choice (`json`, `text`, and `table` are available).
 4. Create an IAM policy for CodeBuild and associate it to a new service role as described
 [here](http://docs.aws.amazon.com/codebuild/latest/userguide/setting-up.html#setting-up-service-role). Select the
-__Amazon EC2__ role type as CodeBuild is not yet available on this page. Edit the new role's trust relationship
-and replace it with the text in step 16 of the CodeBuild user guide.
+__AWS CodeBuild__ role type. The documentation linked here says the CodeBuild service role type is not available, but
+it actually is.
 5. Create an IAM policy for CodePipelines and associate it to a new service role as described
 [here](http://docs.aws.amazon.com/codepipeline/latest/userguide/iam-identity-based-access-control.html#view-default-service-role-policy).
 Select the __Amazon EC2__ role type as CodePipeline is not yet available on this page. Edit the new role's trust
@@ -72,44 +72,54 @@ per region per account. If you need more, you can
 Here is an example of a file that defines two pipelines with parallel build actions.
 
 ```yaml
+defaultComputeImage: maven:3-jdk-8
+defaultComputeType: BUILD_GENERAL1_LARGE
+defaultPrivilegedMode: true
 pipelines:
 - name: j1
   parallelBuildActions:
   - name: bdb
+    timeout: 30
     env:
     - name: MODULE
       value: janusgraph-berkeleyje
   - name: h-l-s
+    timeout: 42
     env:
     - name: MODULE
       value: janusgraph-hadoop-parent/janusgraph-hadoop-2,janusgraph-lucene,janusgraph-solr
   - name: cassandra
+    timeout: 220
     env:
     - name: MODULE
       value: janusgraph-cassandra
   - name: test
+    timeout: 43
     env:
     - name: MODULE
       value: janusgraph-test
 - name: j2
   parallelBuildActions:
   - name: hbase098
+    timeout: 130
     env:
     - name: MODULE
       value: janusgraph-hbase-parent/janusgraph-hbase-098
   - name: hbase10
+    timeout: 79
     env:
     - name: MODULE
       value: janusgraph-hbase-parent/janusgraph-hbase-10
   - name: es
+    timeout: 100
     env:
     - name: MODULE
       value: janusgraph-es
   - name: cql
+    timeout: 134
     env:
     - name: MODULE
       value: janusgraph-cql
-
 ```
 The first pipeline is called `j1` and the second pipeline is called `j2`. Each of these pipelines have four
 parallel build actions defined. Each build action is keyed at the action name and includes the environment
--- a/janusgraph-codepipelines-ci/src/main/java/org/janusgraph/codepipelines/AwsCodePipelinesCi.java
+++ b/janusgraph-codepipelines-ci/src/main/java/org/janusgraph/codepipelines/AwsCodePipelinesCi.java
@@ -1,72 +1,55 @@
+// Copyright 2017 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 package org.janusgraph.codepipelines;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
-import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableList;
 
-import lombok.ToString;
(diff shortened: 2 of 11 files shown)


amcp:
@sjudeng ready for you to take a look.

janusgraph-codepipelines-ci/README.md
```diff
@@ -72,44 +72,68 @@ per region per account. If you need more, you can
 Here is an example of a file that defines two pipelines with parallel build actions.
 
 ```yaml
+defaultComputeImage: stephenreed/jenkins-java8-maven-git:latest
```
sjudeng: I was able to run tests using the `maven:3-jdk-8` image. Apparently git isn't required ... maybe code is checked out in pipeline external to compute image? I'd prefer this if possible because it's an official image and Apache licensed (I don't see license information on the `stephenreed/jenkins-java8-maven-git` image).

sjudeng:
Also while you're in here can you update step 4 in the README to indicate to select the AWS CodeBuild role type (which does exist)? Then you can also remove the last sentence in the item since the trust relationship will already be correct.

janusgraph-codepipelines-ci/src/main/java/org/janusgraph/codepipelines/AwsCodePipelinesCi.java
```diff
@@ -325,4 +136,23 @@ public static void main(String... args) {
...
+    private static Option createRequiredOneArgOption(final String longOpt, final String description) {
+        return createOneArgOption(longOpt, description, true);
+    }
+
+    private static Option createOptionalOneArgOption(final String longOpt, final String description) {
+        return createOneArgOption(longOpt, description, false);
+    }
+
+    private static Option createOneArgOption(final String longOpt, final String description, final boolean required) {
+        final Option o = new Option(null /*shortOpt*/, longOpt, true /*hasArg*/, description);
```
amcp: fixed

janusgraph-codepipelines-ci/src/main/java/org/janusgraph/codepipelines/AwsCodePipelinesLogic.java
```diff
@@ -0,0 +1,188 @@
...
+import software.amazon.awssdk.services.s3.model.CreateBucketConfiguration;
+import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
+import software.amazon.awssdk.services.s3.model.PutBucketTaggingRequest;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.model.Tag;
+import software.amazon.awssdk.services.s3.model.Tagging;
+
+/**
+ * Created by amcp on 2017/07/01.
+ */
```
amcp: fixed

janusgraph-codepipelines-ci/src/main/java/org/janusgraph/codepipelines/AwsCodePipelinesLogic.java
```diff
@@ -0,0 +1,188 @@
...
+                .privilegedMode(action.isPrivilegedMode())
+                .environmentVariables(action.getEnv().stream()
+                    .map(e -> EnvironmentVariable.builder().name(e.getName()).value(e.getValue()).build())
+                    .collect(Collectors.toList())).build())
+            .source(ProjectSource.builder().type(SourceType.CODEPIPELINE).build()).build();
+    }
+
+    /**
+     *
+     */
```
amcp: added

janusgraph-codepipelines-ci/README.md
amcp: True. git is avoided be cause the source step earlier in the pipeline does the pulling bit from git. good catch!

amcp:
@sjudeng Fixed the documentation issue in step 4 and I filed a ticket to get the AWS documentation fixed.

sjudeng:
Were you able to reproduce/resolve the issue with test errors ("connection refused") when running ES pipeline?

amcp:
@sjudeng I am doing a new test run now after all the changes we discussed to get fresh results. I feel like I will still get the ES "connection refused" errors. I have not had a chance to dig deep into those. 2AM here now so I'll take a look at the results again when I wake up tomorrow.

sjudeng:
It looks like it's still running as root even with privilegedMode=false.

```
 [2017-07-08T01:46:51,177][WARN ][o.e.b.ElasticsearchUncaughtExceptionHandler] [] uncaught exception in thread [main]
org.elasticsearch.bootstrap.StartupException: java.lang.RuntimeException: can not run elasticsearch as root
```

amcp:
@sjudeng I agree with your findings. Is there a way we can run ES in a docker container that is not started as a root user for the purpose of the build?

sjudeng:
@amcp You read my mind. I think so. I'll take a look. It'll be similar to what we do with ScyllaDB tests (e.g. run all tests under failsafe and start/stop container as part of pre/post-integration-test.

amcp:
Related answer on stack overflow: https://stackoverflow.com/questions/44458934/creating-a-codebuild-project-using-an-aws-curated-docker-image-with-privilegedmo

Seems like we may need to build a Dockerfile container off aws/codepipelines/docker and build it and run es inside it to have effective control over the execution environment.

amcp:
@sjudeng also what i do here for DynamoDB local: https://github.com/awslabs/dynamodb-janusgraph-storage-backend/blob/master/pom.xml#L286

amcp:
#403 for containerizing ES during tests

amcp:
@sjudeng Seems like this PR achieves its intended purpose, customizing compute type, image and privileged mode at the build action level. Maybe we can track fixing the ES build in a different issue, and move on with this PR?

sjudeng:
I thought the purpose of these updates is to get ES test pipeline to pass, to fix #372. If the ES test pipelines were passing I don't think we would have pursued these updates, correct?

I did add the update in #405 to support running tests against a Dockerized ES instance. Having said that I"m not sure this will fully solve the issue here.

sjudeng:
On second thought I didn't catch that you've also added customization of the compute type and image here. If you want to create another issue for that to tie this to then separating it seems okay to me.

sjudeng:
I was able to get the ES test pipeline to run successfully using the commit in #405 and the `docker:dind` compute image (see https://github.com/sjudeng/janusgraph/commit/542cd40f1361a40221b282bdb5ae8f5649697c91). If the buildspec/pipeline updates look okay you're welcome to just squash them into your commit here. Note this does start the Docker service unconditionally in the build (install phase) and so requires privilegedMode=true for all pipelines.

Also as you're probably already aware builds like ES can fail non-deterministically. I had a sporadic failure (testIndexReplay) that resolved on a subsequent run. Eventually it might be nice to add implementation equivalent to travis_retry to account for sporadic errors in the `mvn verify` call.

amcp:
@sjudeng Thank you for isolating ES. I can cherry pick your change into my PR after merging the failsafe change and rebasing.

janusgraph-codepipelines-ci/tp-pipe.yml
```diff
@@ -1,13 +1,35 @@
...
+# In general, set the timeouts to twice the usual execution time for the compute type.
+# 480 minutes is the maximum build time now.
+#
+defaultComputeImage: docker:dind
+defaultComputeType: BUILD_GENERAL1_LARGE
+defaultPrivilegedMode: true
 pipelines:
 - name: jtp1
   parallelBuildActions:
   - name: e-h-l-s-tp
```
sjudeng: You can remove this pipeline since none of these modules have TinkerPop tests and you're excluding the default tests.

amcp:
@sjudeng Builds including j2-es passed with these changes, with the following exceptions. j1-h-l-s and j2-hbase098 and j2-hbase10 are failing with these changes.

janusgraph-codepipelines-ci/tp-pipe.yml
amcp: removing

sjudeng:
Try setting `JAVA_HOME` as part of the [verify call](https://github.com/amcp/janusgraph/blob/81aa65404ccedaa3cb0098920158c87336181aa2/buildspec.yml#L29). I think embedded HBase (also started in hadoop tests) needs `JAVA_HOME` to be set and for some reason build commands don't seem to pick up top level `env` or even settings from previous build commands.

```
JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk mvn verify --projects $MODULE $ARGS
```

amcp:
@sjudeng Didn't work as expected so I just conditioned the install steps on janusgraph-es and set the default image back to maven:3-jdk-8

amcp:
@sjudeng 

All the following now pass: j1-bdb, j1-h-l-s, j1-test, j2-hbase098, j2-hbase10, j2-es, j2-cql. j1-cassandra is still in progress but I think it will pass based on past test runs.

amcp:
I squashed all the commits but kept your GitHub identifier in the commit message. Merging.

PR #384, opened 2017-07-03
Static cluster/session management in CQL backend

### LINKED ISSUE #383: Improve test performance in CQL backend
Currently CQL tests, especially the TinkerPop tests, take considerably longer than for other modules. Below are sample runtimes for default (`mvn verify`) and default+TP (`mvn verify -Dtest.skip.tp=false`).

| Module | Default (hours) | Default+TP (hours) |
| ---- | ---- | ---- |
| berkeleyje | 0.2 | 2.3 |
| cassandra | 1.9 | 4.3 |
| hbase-10 | 0.6 | 8.9 |
| cql | 2.5 | 15.1 |

The longer runtimes are causing problems for autmoated testing including [Travis](https://travis-ci.org/JanusGraph/janusgraph/jobs/249256585) and [AWS CodePipelines](https://github.com/JanusGraph/janusgraph/tree/master/janusgraph-codepipelines-ci).

sjudeng:
Makes the following updates to fix #383. With these updates the full (default+TinkerPop) test suite runtime for the CQL module is decreased from 15.1 hours to 3.3 hours.

* A single static cluster instance and a single static session per keyspace managed in CQLStoreManager.
* Truncate tables instead of drop keyspace in CQLStoreManager.clearStorage(). There was an issue with test failures when switching to truncation that was resolved by deleting the commitlog directory prior to starting embedded test instance in CassandraStorageSetup.
* Test parallelism was removed to resolve test failures associated with the above updates and for consistency with test executions in other modules.
* Minor updates to split Travis build matrix for CQL tests and reduce logging (also renamed logging appender names for consistency with other modules).

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
@@ -15,7 +15,7 @@
 package org.janusgraph.diskstorage.cql;
 
 import static com.datastax.driver.core.schemabuilder.SchemaBuilder.createKeyspace;
-import static com.datastax.driver.core.schemabuilder.SchemaBuilder.dropKeyspace;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.truncate;
 import static io.vavr.API.$;
 import static io.vavr.API.Case;
 import static io.vavr.API.Match;
@@ -115,7 +115,7 @@ public class CQLStoreManager extends DistributedStoreManager implements KeyColum
     private final int batchSize;
     private final boolean atomicBatch;
 
-    private final ExecutorService executorService;
+    final ExecutorService executorService;
 
     private final Cluster cluster;
     private final Session session;
@@ -189,7 +189,7 @@ public class CQLStoreManager extends DistributedStoreManager implements KeyColum
         this.openStores = new ConcurrentHashMap<>();
     }
 
-    private Cluster initializeCluster() throws PermanentBackendException {
+    Cluster initializeCluster() throws PermanentBackendException {
         final Configuration configuration = getStorageConfig();
 
         final List<InetSocketAddress> contactPoints;
@@ -249,7 +249,7 @@ public class CQLStoreManager extends DistributedStoreManager implements KeyColum
         return builder.build();
     }
 
-    private Session initializeSession(final String keyspaceName) {
+    Session initializeSession(final String keyspaceName) {
         final Configuration configuration = getStorageConfig();
         final Map<String, Object> replication = Match(configuration.get(REPLICATION_STRATEGY)).of(
                 Case($("SimpleStrategy"), strategy -> HashMap.<String, Object> of("class", strategy, "replication_factor", configuration.get(REPLICATION_FACTOR))),
@@ -328,7 +328,10 @@ public class CQLStoreManager extends DistributedStoreManager implements KeyColum
 
     @Override
     public void clearStorage() throws BackendException {
-        this.session.execute(dropKeyspace(this.keyspace));
+        final Future<Seq<ResultSet>> result = Future.sequence(
+            Iterator.ofAll(this.cluster.getMetadata().getKeyspace(this.keyspace).getTables())
+                .map(table -> Future.fromJavaFuture(this.session.executeAsync(truncate(this.keyspace, table.getName())))));
+        result.await();
     }
 
     @Override
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,10 @@ env:
     - MODULE='hbase-parent/janusgraph-hbase-10' ARGS='-Dtest=**/graphdb/hbase/*'
     - MODULE='hbase-parent/janusgraph-hbase-098' ARGS='-Dtest=**/diskstorage/hbase/*'
     - MODULE='hbase-parent/janusgraph-hbase-098' ARGS='-Dtest=**/graphdb/hbase/*'
-    - MODULE='cql'
+    - MODULE='cql' ARGS='-Dtest=**/diskstorage/cql/* -Dtest.skip.murmur=true'
+    - MODULE='cql' ARGS='-Dtest=**/diskstorage/cql/* -Dtest.skip.byteorderedpartitioner=true -Dtest.skip.murmur-serial=true -Dtest.skip.murmur-ssl=true'
+    - MODULE='cql' ARGS='-Dtest=**/graphdb/cql/* -Dtest.skip.murmur=true'
+    - MODULE='cql' ARGS='-Dtest=**/graphdb/cql/* -Dtest.skip.byteorderedpartitioner=true -Dtest.skip.murmur-serial=true -Dtest.skip.murmur-ssl=true'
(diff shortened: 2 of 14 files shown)


sjudeng:
@pkendall64 Would appreciate you taking a look at this to make sure the updates here seem reasonable. One of the main performance improvements came from switching back to table truncation (which I know you had originally) in clearStorage. As mentioned above the errors that occurred when making this switch were resolved by ensuring commitlog directory was deleted. Not sure why this is required for CQL and not Thrift.

amcp:
@sjudeng instead of one build target now there are four travis builds for cql. How long does each of the lines 43-46 take? let me see if i can find that in the build.

amcp:
15, 10, 16 and 14 minutes. OK, great. https://travis-ci.org/JanusGraph/janusgraph/builds/249466821?utm_source=github_status&utm_medium=notification

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -111,14 +111,14 @@
 
     private static final int DEFAULT_PORT = 9042;
 
+    private static Cluster cluster;
+    private static Map<String,Session> sessions = new ConcurrentHashMap<>();
```
amcp: We are not running tests in parallel anymore as per the xml file. why does this need to be concurrent? As on line 117 there is only one keyspace variable for this store manager so what else will go in the sessions map? I searched the original file and this file and I do not see any other uses of the sessions map. Finally, initializeSession is private and it seems to be the only method that mutates the session map.

sjudeng:
@amcp Yep. The Travis jobs for CQL (like Cassandra and HBase) are split by test execution and package. Test timing in Travis can vary considerably based on time of day and other external factors. Also by keeping individual Travis job times below 15 minutes they can get the full benefit of travis_retry, which will retry up to three times but only within the overall 50 minute job timeout.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -189,99 +189,101 @@ public CQLStoreManager(final Configuration configuration) throws BackendExceptio
...
-    private Session initializeSession(final String keyspaceName) {
-        final Configuration configuration = getStorageConfig();
-        final Map<String, Object> replication = Match(configuration.get(REPLICATION_STRATEGY)).of(
-                Case($("SimpleStrategy"), strategy -> HashMap.<String, Object> of("class", strategy, "replication_factor", configuration.get(REPLICATION_FACTOR))),
+    private synchronized void initializeSession(final String keyspaceName) {
+        if (!sessions.containsKey(keyspaceName)) {
+            sessions.put(keyspaceName, cluster.connect());
+            final Configuration configuration = getStorageConfig();
+            final Map<String, Object> replication = Match(configuration.get(REPLICATION_STRATEGY)).of(
+                Case($("SimpleStrategy"), strategy -> HashMap.<String, Object>of("class", strategy, "replication_factor", configuration.get(REPLICATION_FACTOR))),
```
amcp: What is the dollar sign for? Do we need to use it?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -328,7 +322,10 @@ public StoreTransaction beginTransaction(final BaseTransactionConfig config) thr
 
     @Override
     public void clearStorage() throws BackendException {
-        this.session.execute(dropKeyspace(this.keyspace));
+        final Future<Seq<ResultSet>> result = Future.sequence(
+            Iterator.ofAll(cluster.getMetadata().getKeyspace(this.keyspace).getTables())
+                                .map(table -> Future.fromJavaFuture(sessions.get(this.keyspace).executeAsync(truncate(this.keyspace, table.getName())))));
+        result.await();
```
amcp: no timeout for await?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
amcp: @sjudeng I get it, its static. You are considering the case of multiple cql store manager instances in the same JVM. When did this come up in testing?

sjudeng: @amcp No it might not be a problem in testing. Do you think it might cause issues to use concurrent data structure and/or synchronized modifier here?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
sjudeng: @amcp These are showing as changes but it's just whitespace (indented under the cluster/session==null check). This is original code from #166.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
sjudeng: Same comment as above

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
pkendall64: The $ turns a predicate into a pattern and most of the vavr APIs take patterns in the 0.9 version.

pkendall64:
Yeah, looks good to me. I have to wonder whether it would be better to use a wrapper to provide the caching in the test area rather than baking it into the main code directly though.

sjudeng:
@pkendall64 That's not a bad idea. Probably would be a little ugly but at least it would contain it to test code. I'll take a look.

The suggestion to use the singleton cluster/session comes from http://www.datastax.com/dev/blog/4-simple-rules-when-using-the-datastax-drivers-for-cassandra. But now that I think about it if there needs to be support for connecting to different Cassandra cluster's in the same JVM this would require more updates than provided here and would probably get pretty messy.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
amcp: no problem here.

amcp:
@sjudeng maybe for testing it is useful to connect to be able to connect to different Cassandra clusters, but I think we can just convert the same-jvm-multiple-cluster-cql support issue into an issue on GitHub and move on with these changes?

sjudeng:
@amcp I think it's moot now ... I just pushed updates that refactor to use a wrapper class during testing (thanks @pkendall64). As expected it's a little ugly (needed some reflection on StandardStoreManager) but almost all updates are in test code only. I think it makes a lot more sense this way as it allows clients to build in caching themselves if necessary (e.g. caching CQLStoreManager instances by physical cluster and/or keyspace).

PR #389, opened 2017-07-04
Fix IllegalAccessError for Iterators.emptyIterator() when using Guava >= 20.0

mikesager:
Since Guava 20.0, Iterators.emptyIterator() is no longer public, resulting in an IllegalAccessError when trying to use various traversal iteration steps which call it.  Per the deprecation notice in the documentation for Guava 19.0 (https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Iterators.html#emptyIterator()), this commit replaces all occurrences of Iterators.emptyIterator() with Collections.emptyIterator().

Signed-off-by: Mike Sager <mike@mikesager.name>


mikesager:
Also, I submitted my CLA a few days ago.  Haven't heard back yet, but wanted to get this submitted sooner than later.

PR #393, opened 2017-07-06
Merge 0.1 branch

sjudeng:
A number of commits in 0.1 are missing in master. @srosenthal Thanks for catching.

### DIFF
--- a/CHANGELOG.asc
+++ b/CHANGELOG.asc
@@ -2,6 +2,31 @@
 [appendix]
 == Release Notes
 
+=== Version 0.1.1 (Release Date: May 11, 2017) 
+
+[source, m2]
+<dependency>
+   <groupId>org.janusgraph</groupId>
+   <artifactId>janusgraph-core</artifactId>
+   <version>0.1.1</version>
+</dependency>
+
+*Tested Compatibility:*
+
+* Apache Cassandra 2.1.9
+* Apache HBase 0.98.8-hadoop2, 1.0.3, 1.1.8, 1.2.4
+* Google Bigtable 0.9.5.1
+* BerkeleyJE 7.3.7
+* Elasticsearch 1.5.1
+* Apache Lucene 4.10.4
+* Apache Solr 5.2.1
+* Apache TinkerPop 3.2.3
+* Java 1.8
+
+For more information on features and bug fixes in 0.1.1, see these GitHub milestone:
+
+* https://github.com/JanusGraph/janusgraph/milestone/3?closed=1
+
 === Version 0.1.0 (Release Date: April 11, 2017) 
 
 [source, m2]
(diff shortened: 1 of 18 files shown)


jerryjch:
Some of the commits are already in master.  e.g. 

> Add Spark/Scala version to version-compat list

> Get rid of the Titan-oriented introduction

Is it possible to do cherry-pick, not a whole-sale merge?
I hope we don't need to do this often in the future.

sjudeng:
What's happened is that the same commits went into 0.1 and master with different SHA-1 hashes. Git merges okay in this case but I agree it's not ideal. The [merge diff](https://github.com/JanusGraph/janusgraph/pull/393/files) looks clean to me but I have no objections if someone wants to cherry pick out commits. I do think we want all the (non-duplicate) commits, including the 0.1.1 release notes and compatibility matrix updates (master was incorrect on this one). 

Incidentally I think we're making a lot of (error prone) work for ourselves with the current process of merging into master and cherry-picking into release branches.

srosenthal:
I'm not a contributor to JanusGraph, just a user, but merging from a release branch to master is totally normal in my experience, and less error-prone than cherry-picking. I don't see any problem with a full merge.

mbrukman:
Looks like the `[cla: no]` label is there because some of the commits are actually merges, and the email used for a web-UI-based merge commit is the account's default email address, which is not always the same one as what's on the CLA. I'll add the relevant emails for @jerryjch and @amcp to clear this up.

amcp:
@sjudeng please rebase.

amcp:
Since we were not religions about cherry picking, is it OK to bite the bullet and squash and merge all these changes? I think it is OK to have a commit in master that refers to a commit range on a different branch without actually adding extra 20+ commits to master.

sjudeng:
I don't think we can merge the commits as we'd lose author attribution. If the merge is not desirable I think the missing commits should be identified/cherry-picked manually. Personally I prefer the approach of merging the release branch into master. It's a little messy this time because of the way we were doing things, but once it's done master will be up-to-date with the release branch and it'll be easier to keep it current going forward.

jerryjch:
By going thru the list, I think only a few are missing in master and are needed.
It will be rare cases in the future where we have to merge commits from branches to master in the future.
Generally, we can continue to discuss the approach we should adopt on the dev list again.

sjudeng:
@jerryjch Sounds good to me. I can leave this open to help evaluate the eventual cherry-pick-based merge request to make sure nothing gets missed.

amcp:
@sjudeng excited to get this in. Are we going to work on the cherry-pick-based merge request next?

sjudeng:
Yes I think @jerryjch is working on that as an alternative to the branch merge here.

jerryjch:
I will work on it.   Is your branch open to push, @sjudeng?  Or I should get another PR?

amcp:
@jerryjch to reduce clutter and to allow comparing commit lists, perhaps you should create a new PR.

amcp:
closing this as the cleaner PR #406 is now merged

Issue #398, opened 2017-07-07
CQL storage adapter doesn't work with default configs and bundled Cassandra


twilmes:
The bundled Cassandra 2.1.9 requires Cassandra binary protocol version 3 but the `CQLStoreManager` defaults to the newest protocol the bundled driver supports (v4) if no protocol version is specified.

PR #399, opened 2017-07-07
Issue 398 - CQL storage adapter doesn't work with default configs and bundled Cassandra

twilmes:
If no protocol is specified, the CQL driver no longer defaults to the highest driver
protocol version. Instead, the driver will negotiate with Cassandra and select the
highest mutually supported protocol version.


sjudeng:
Let me run through TinkerPop tests once before merging (unless you already did). Also @samant-orionhealth / @pkendall64 please let us know if you see any issue with this.

pkendall64:
Yes, that looks good. It's better that it negotiate than use the highest in the driver because that's newer than the cassandra that you can connect to.

sjudeng:
Thanks @pkendall64. All TinkerPop tests passed as well. I thought it was worth checking given the "if we let it negotiate then there are spurious errors" comment in the code.

robertdale:
Fixes issue #398 
(I know it's stated in the title but it's nice to have a link and link back to/from the issue)

twilmes:
Thanks for the TinkerPop and extra C* version testing @sjudeng

PR #400, opened 2017-07-07
Use non-JTS SpatialContextFactory in Solr config

### LINKED ISSUE #316: Make JTS jar dependency for solr index optional
Currently as described in 
http://docs.janusgraph.org/0.1.0/solr.html
24.2.3. JTS ClassNotFoundException with Geo Data

you always get errors when trying to use solr with schema.xml provided by Janus.

Problem is GEO Data is NOT a primary use case and it would be good NOT to require/enable it by default.

Some downsides include:
1) Manual patching of Solr classpath complicates automatic deployment scripts (e.g. using docker containers / docker compose) as you need to add own bash scripts to facilitate that

2) Adding 3rd party jars, which do not belong either to Janus or Solr distro may require additional legal approval / license scanning.

Either comment out certain schema.xml section or provide alternative solr core configs to copy.

sjudeng:
Fixes #316

### DIFF
--- a/janusgraph-dist/src/assembly/static/conf/solr/schema.xml
+++ b/janusgraph-dist/src/assembly/static/conf/solr/schema.xml
@@ -533,10 +533,10 @@
 
     <fieldType name="geo"
                class="solr.SpatialRecursivePrefixTreeFieldType"
-               spatialContextFactory="com.spatial4j.core.context.jts.JtsSpatialContextFactory"
+               spatialContextFactory="org.locationtech.spatial4j.context.SpatialContextFactory"
                distErrPct="0.000"
                maxDistErr="0.000009"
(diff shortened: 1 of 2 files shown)


amcp:
@sjudeng I think this has to do with a package name change. Which test is this configuration used in?

sjudeng:
@amcp The schema in janusgraph-solr/test is used in all solr tests. The legacy spatial4j package name just printed a warning but I figured I'd fix that while I was in here.

The update to the schema in janusgrpah-dist resolves errors when using the default solr release configs (e.g. janusgraph-cassandra-solr.propertes, etc.).

wojciechwojcik:
@sjudeng that great news, thanks :)

PR #401, opened 2017-07-08
docs: Add solr, remove data types [skip ci]

robertdale:
Syncing docs to match current state.

### DIFF
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -90,6 +90,7 @@ JanusGraph's modular architecture supports third-party adapters.
 ** <<bdb, BerkeleyDB>>
 * Indices, which speed up and enable more complex queries:
 ** <<elasticsearch, Elasticsearch>>
+** <<solr, Solr>>
 ** <<lucene, Lucene>>
 
 Broadly speaking, applications can interact with JanusGraph in two ways:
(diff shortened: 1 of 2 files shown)


robertdale:
@amcp code -  https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java#L130

robertdale:
@amcp not sure what you mean about generating the configuration reference.  The lists I'm referring to are static lists in the docs.  These data types were previously removed from the graph section in PR #170 but were not removed from the search predicate docs.

robertdale:
@amcp on generating the supported types list, I would say no to graph because there are many types for internal use beyond what is supported for user. [See here. ](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java#L76) 

For indexes, also no, because there's no succinct list, just if statements checking type.

robertdale:
@amcp T[he full list of types.](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/StandardSerializer.java)

amcp:
@robertdale Understood. Why is UUID not a a supported type for users, even though it is a primitive type? seem like most of the primitive types are in the documentation's list.

robertdale:
@amcp Good catch.  UUID is a supported property key type, but not listed in **search** data types. It appears that all search backends do support UUID.  The only caveat is that only `eq()` and `neq()` predicates are supported.   I've added it to the search docs.

Issue #411, opened 2017-07-12
Paginate column access in AstyanaxKeyValueColumnStore


dpitera:
JanusGraph has a tendency to make queries to Cassandra with _very_ large
limits. For example, any query without a `.limit()` clause will fetch
all matching columns from a given row. These requests are usually
range-bound (e.g., only requesting columns that are edges, based on the
serialization logic) but are not limited.

Fetching millions of columns in a single request is not very friendly to
Cassandra; in production we've observed memory allocation spikes in
excess of 10GB in a 30s period. On my local machine, calling

`graph.traversal().V(id).outE().count()`

prior to this commit will cause my Cassandra instance to OOM in
approximately 10 seconds.

jerryjch:
HBase has something similar with a build-in ColumnPaginationFilter.
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-hbase-parent/janusgraph-hbase-core/src/main/java/org/janusgraph/diskstorage/hbase/HBaseKeyColumnValueStore.java#L137

PR #412, opened 2017-07-12
Paginate column access in AstyanaxStore

dpitera:
#411 

JanusGraph has a tendency to make queries to Cassandra with _very_ large
limits. For example, any query without a `.limit()` clause will fetch
all matching columns from a given row. These requests are usually
range-bound (e.g., only requesting columns that are edges, based on the
serialization logic) but are not limited.

Fetching millions of columns in a single request is not very friendly to
Cassandra; in production we've observed memory allocation spikes in
excess of 10GB in a 30s period. On my local machine, calling

graph.traversal().V(id).outE().count()

prior to this commit will cause my Cassandra instance to OOM in
approximately 10 seconds.

This patch places a hard cap on the number of columns returned (per row)
via a single request to Cassandra. If, for a given request, a row has
more columns available in the requested range than the cap, the system
makes additional requests for that row until the requested number of
columns is returned or no more columns are available.

This code is somewhat more subtle that may be expected due to Titan's
use of the Astyanax API - it's not uncommon for Titan to make a request
for multiple keys concurrently, in which case the `limit` parameter
applies to each key - not the request as a whole. This has two
implications:

1.  We need to use a different API (getRow, rather than getKeySlice) to
fetch columns for rows that met the limit on the initial query, since
the starting column must be manipulated on a per-row basis

2. The page size needs to be relatively conservative, in case a single
request is made for hundreds of keys at once. This is relatively
unlikely, especially given the natural infrequency of supernodes,
but it's something that we must be aware of.

### DIFF
--- a/janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxKeyColumnValueStore.java
+++ b/janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxKeyColumnValueStore.java
@@ -15,22 +15,43 @@
 package org.janusgraph.diskstorage.cassandra.astyanax;
 
 import com.google.common.base.Predicate;
-import com.google.common.collect.*;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.netflix.astyanax.ExceptionCallback;
 import com.netflix.astyanax.Keyspace;
 import com.netflix.astyanax.connectionpool.OperationResult;
 import com.netflix.astyanax.connectionpool.exceptions.ConnectionException;
-import com.netflix.astyanax.model.*;
+import com.netflix.astyanax.model.Column;
+import com.netflix.astyanax.model.ColumnFamily;
+import com.netflix.astyanax.model.ColumnList;
+import com.netflix.astyanax.model.Rows;
+import com.netflix.astyanax.model.Row;
 import com.netflix.astyanax.query.AllRowsQuery;
 import com.netflix.astyanax.query.RowSliceQuery;
 import com.netflix.astyanax.retry.RetryPolicy;
 import com.netflix.astyanax.serializers.ByteBufferSerializer;
-import org.janusgraph.diskstorage.*;
+import com.netflix.astyanax.query.RowQuery;
+import org.janusgraph.diskstorage.BackendException;
+import org.janusgraph.diskstorage.PermanentBackendException;
+import org.janusgraph.diskstorage.TemporaryBackendException;
+import org.janusgraph.diskstorage.StaticBuffer;
+import org.janusgraph.diskstorage.Entry;
+import org.janusgraph.diskstorage.EntryList;
+import org.janusgraph.diskstorage.EntryMetaData;
 import org.janusgraph.diskstorage.cassandra.utils.CassandraHelper;
-import org.janusgraph.diskstorage.keycolumnvalue.*;
+import org.janusgraph.diskstorage.keycolumnvalue.KeySliceQuery;
+import org.janusgraph.diskstorage.keycolumnvalue.StoreTransaction;
+import org.janusgraph.diskstorage.keycolumnvalue.SliceQuery;
+import org.janusgraph.diskstorage.keycolumnvalue.KeyIterator;
+import org.janusgraph.diskstorage.keycolumnvalue.KeyRangeQuery;
+import org.janusgraph.diskstorage.keycolumnvalue.KeyColumnValueStore;
+import org.janusgraph.diskstorage.keycolumnvalue.KCVMutation;
 import org.janusgraph.diskstorage.util.RecordIterator;
 import org.janusgraph.diskstorage.util.StaticArrayBuffer;
 import org.janusgraph.diskstorage.util.StaticArrayEntry;
+import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalInterruptedException;
 
 import javax.annotation.Nullable;
 import java.nio.ByteBuffer;
@@ -46,6 +67,7 @@ public class AstyanaxKeyColumnValueStore implements KeyColumnValueStore {
     private final ColumnFamily<ByteBuffer, ByteBuffer> columnFamily;
     private final RetryPolicy retryPolicy;
     private final AstyanaxStoreManager storeManager;
+    private final int readPageSize;
     private final AstyanaxGetter entryGetter;
 
     AstyanaxKeyColumnValueStore(String columnFamilyName,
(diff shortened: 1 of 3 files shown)


dpitera:
Pardon my naiveté, but what sort of test are you looking for other than the passing of the cassandra astyanax test suite?

sjudeng:
I'd expect to see tests with smaller page size or larger test graphs (in terms of properties+edges since it's column pagination) to make sure the pagination implementation is being exercised. I'd also look at testing boundary conditions (e.g. number of columns equals page size).

dpitera:
Excellent suggestions, tests have been added!

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxKeyColumnValueStore.java
```diff
@@ -133,8 +146,69 @@ public EntryList getSlice(KeySliceQuery query, StoreTransaction txh) throws Back
 
         for (Row<ByteBuffer, ByteBuffer> row : rows) {
             assert !result.containsKey(row.getKey());
-            result.put(StaticArrayBuffer.of(row.getKey()),
-                  CassandraHelper.makeEntryList(row.getColumns(),entryGetter, query.getSliceEnd(), query.getLimit()));
+            ByteBuffer key = row.getKey();
+            ColumnList<ByteBuffer> pageColumns = row.getColumns();
+            ArrayList<Column<ByteBuffer>> queryColumns = new ArrayList();
```
sjudeng: `List` on LHS

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxKeyColumnValueStore.java
```diff
@@ -110,16 +113,26 @@ public EntryList getSlice(KeySliceQuery query, StoreTransaction txh) throws Back
...
-                .withRetryPolicy(retryPolicy.duplicate())
-                .getKeySlice(CassandraHelper.convert(keys));
 
-        // Thank you, Astyanax, for making builder pattern useful :(
-        rq.withColumnRange(query.getSliceStart().asByteBuffer(),
-                query.getSliceEnd().asByteBuffer(),
-                false,
-                query.getLimit() + (query.hasLimit()?1:0)); //Add one for potentially removed last column
+         // Add one for last column potentially removed in CassandraHelper.makeEntryList
+        int queryLimit = query.getLimit() + (query.hasLimit() ? 1 : 0);
```
sjudeng: `final` here and throughout wherever possible

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxKeyColumnValueStore.java
```diff
@@ -133,8 +146,69 @@ public EntryList getSlice(KeySliceQuery query, StoreTransaction txh) throws Back
...
+            Iterables.addAll(queryColumns, pageColumns);
+            while (pageColumns.size() == pageLimit && queryColumns.size() < queryLimit) {
+                Column<ByteBuffer> lastColumn = queryColumns.get(queryColumns.size() - 1);
+                sliceStart = lastColumn.getName();
+                // No possibility of two values at the same column name, so start the
+                // next slice one bit after the last column found by the previous query.
+                // byte[] is little-endian
+                Integer position = null;
+                for (int i = sliceStart.array().length - 1; i >= 0; i--) {
+                    if (sliceStart.array()[i] < 127) {
```
sjudeng: /127/Byte.MAX_VALUE/

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxKeyColumnValueStore.java
```diff
@@ -133,8 +146,69 @@ public EntryList getSlice(KeySliceQuery query, StoreTransaction txh) throws Back
...
+
+                OperationResult<ColumnList<ByteBuffer>> pageResult;
+                try {
+                    pageResult = (OperationResult<ColumnList<ByteBuffer>>) pageQuery.execute();
+                } catch (ConnectionException e) {
+                    throw new TemporaryBackendException(e);
+                }
+
+                if (Thread.interrupted()) {
+                    throw new RuntimeException(new InterruptedException());
```
sjudeng: Should this be a `TraversalInterruptedException`?

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxStoreManager.java
```diff
@@ -270,6 +279,7 @@
 
     private final RetryPolicy retryPolicy;
 
+    final int readPageSize;
```
amcp: private

janusgraph-cassandra/src/test/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxColumnPaginationTest.java
```diff
@@ -0,0 +1,153 @@
...
+    @Test
+    public void ensureReadPageSizePropertySetCorrectly() {
+        assertEquals(((AstyanaxStoreManager) manager).readPageSize, 4096);
+    }
+
+    @Test
+    public void retrieveLessThanBoundaryColumnPaginationProperties() {
+        Graph graph = JanusGraphFactory.open(getBaseStorageConfiguration());
+        Vertex v = graph.addVertex();
+        for(int i = 0; i < 4095; i++) {
```
amcp: refactor 4095 and reuse here and below

janusgraph-cassandra/src/test/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxColumnPaginationTest.java
```diff
@@ -0,0 +1,153 @@
...
+        graph.tx().commit();
+
+        assertEquals(4097, graph.traversal().V(v).valueMap().next().keySet().size());
+    }
+
+    @Test
+    public void retrieveWayBeyondBoundaryColumnPaginationProperties() {
+        Graph graph = JanusGraphFactory.open(getBaseStorageConfiguration());
+        Vertex v = graph.addVertex();
+        for (int i = 0; i < 20000; i++) {
```
amcp: refactor a constant for 20000

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxStoreManager.java
dpitera: this has no modifier on purpose to allow access for the `AstyanaxKeyColumnValueStore` in the same package.

dpitera:
@sjudeng @amcp Updated with your feedback. Thanks for your reviews

sjudeng:
I'm not seeing the updates. Did you commit/push?

dpitera:
Sorry I don't know what happened; all updates should be in place now.

Issue #111, opened 2017-02-13
Reduce amount of time to run tinkerpop tests from 40 hours to 2.5 hours


amcp:
Status of builds in janusgraph-codepipelines-ci/pipe.yml/j1:
- [x] bdb - Passed in 15 min 10 secs
- [x] hadoop-lucene-solr - Passed in 21 minutes 12 seconds 
- [x] cassandra - Passed in 110 minutes
- [x] test - Passed in 21 minutes 37 seconds

Status of builds in janusgraph-codepipelines-ci/pipe.yml/j2:
- [x] hbase098 - Passed in 65 minutes
- [x] hbase10 - Passed in 39 minutes 44 seconds
- [x] es - Passed in 18 minutes
- [x] cql - Passed in 67 minutes

Status of builds in janusgraph-codepipelines-ci/tp-pipe.yml/jtp1:
- [x] bdb - Passed in 118 minutes
- [x] test - Passed in 32 minutes 4 seconds
- [x] **cassandra - Failed in 131 minutes (known failures)**

Status of builds in janusgraph-codepipelines-ci/tp-pipe.yml/jtp2:
- [x] **hbase098 - Failed in 409 minutes (known failures)**
- [x] **hbase10 - Passed in 467 minutes (known failures)**
- [x] cql - Passed in 2 hours

amcp:
@sjudeng do you have any insight as to why these tests take so long? any clues on getting the time to complete faster?
@mbrukman our development and release velocity will continue to be adversely affected until we can get this time down.

sjudeng:
There are a large number of tests (see below). The total time is driven by the Cassandra and HBase modules. If you exclude those the total runtime is around 5 hours. 

| Module | Test Count | Time (sec) |
| ------ | ---------- | ---------- |
| JanusGraph-Test | 4496 | 2892 |
| JanusGraph-BerkeleyJE | 3191 | 12272 |
| JanusGraph-Cassandra | 3109 | 50577 |
| JanusGraph-ElasticSearch | 75 | 696 |
| JanusGraph-HBase | 2164 | 77801 |
| JanusGraph-Hadoop | 17 | 424 |
| JanusGraph-Lucene | 39 | 107 |
| JanusGraph-Solr | 68 | 455 |
| Total | 13159 | 145224 |

[...]

PR #413, opened 2017-07-16
Scan and delete rather than truncate in HBase 1.0 clearTable

sjudeng:
The existing `HBaseAdmin1_0.clearTable()` implementation disables, truncates and then re-enables the table, which performs poorly for integration tests. This update scans and deletes all rows instead of truncating, following the existing [`HBaseAdmin0_98.clearStorage()`](https://github.com/JanusGraph/janusgraph/blob/84ead7d907b62061e30c5d3210e5e8422fd4648c/janusgraph-hbase-parent/janusgraph-hbase-098/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin0_98.java#L65-L95) implementation.

With these updates the full (default+TinkerPop) test suite runtime for the janusgraph-hbase-10 module is decreased from 8.9 hours to 2.7 hours.

This supports #111.

### DIFF
--- a/janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
+++ b/janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
@@ -15,14 +15,22 @@
 package org.janusgraph.diskstorage.hbase;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.hadoop.hbase.HColumnDescriptor;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
-import org.apache.hadoop.hbase.TableNotDisabledException;
 import org.apache.hadoop.hbase.TableNotFoundException;
 import org.apache.hadoop.hbase.client.Admin;
 
@@ -37,6 +45,13 @@ public class HBaseAdmin1_0 implements AdminMask
     {
         this.adm = adm;
     }
+
+    /**
+     * Delete all rows from the given table. This method is intended only for development and testing use.
+     * @param tableString
+     * @param timestamp
+     * @throws IOException
+     */
     @Override
     public void clearTable(String tableString, long timestamp) throws IOException
     {
@@ -47,21 +62,26 @@ public class HBaseAdmin1_0 implements AdminMask
             return;
         }
 
-        if (!adm.isTableDisabled(tableName))
-            adm.disableTable(tableName);
+        // Unfortunately, linear scanning and deleting rows is faster in HBase when running integration tests than
+        // disabling and deleting/truncating tables.
+        final Scan scan = new Scan();
+        scan.setCacheBlocks(false);
+        scan.setCaching(2000);
+        scan.setTimeRange(0, Long.MAX_VALUE);
+        scan.setMaxVersions(1);
 
-        if (!adm.isTableDisabled(tableName))
-            throw new RuntimeException("Unable to disable table " + tableName);
-
-        // This API call appears to both truncate and reenable the table.
-        log.info("Truncating table {}", tableName);
-        adm.truncateTable(tableName, true /* preserve splits */);
-
-        try {
-            adm.enableTable(tableName);
-        } catch (TableNotDisabledException e) {
-            // This triggers seemingly every time in testing with 1.0.2.
-            log.debug("Table automatically reenabled by truncation: {}", tableName, e);
+        try (final Table table = adm.getConnection().getTable(tableName);
+             final ResultScanner scanner = table.getScanner(scan)) {
+            final Iterator<Result> iterator = scanner.iterator();
+            final int batchSize = 1000;
+            final List<Delete> deleteList = new ArrayList<>();
+            while (iterator.hasNext()) {
+                deleteList.add(new Delete(iterator.next().getRow(), timestamp));
+                if (!iterator.hasNext() || deleteList.size() == batchSize) {
+                    table.delete(deleteList);
+                    deleteList.clear();
+                }
+            }
         }
     }
 
--- a/janusgraph-hbase-parent/janusgraph-hbase-098/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin0_98.java
+++ b/janusgraph-hbase-parent/janusgraph-hbase-098/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin0_98.java
@@ -15,6 +15,9 @@
 package org.janusgraph.diskstorage.hbase;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -42,6 +45,12 @@ public class HBaseAdmin0_98 implements AdminMask
         this.adm = adm;
     }
 
+    /**
+     * Delete all rows from the given table. This method is intended only for development and testing use.
+     * @param tableName
+     * @param timestamp
+     * @throws IOException
+     */
     @Override
     public void clearTable(String tableName, long timestamp) throws IOException
     {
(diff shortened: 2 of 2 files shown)


janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
```diff
@@ -47,21 +53,29 @@ public void clearTable(String tableString, long timestamp) throws IOException
...
-
-        try {
-            adm.enableTable(tableName);
-        } catch (TableNotDisabledException e) {
-            // This triggers seemingly every time in testing with 1.0.2.
-            log.debug("Table automatically reenabled by truncation: {}", tableName, e);
+        // Unfortunately, linear scanning and deleting rows is faster in HBase when running integration tests than
+        // disabling and deleting/truncating tables.
+        final Scan scan = new Scan();
+        scan.setBatch(100);
```
amcp: is this batch size the same as the batch size on line 66?

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
```diff
@@ -47,21 +53,29 @@ public void clearTable(String tableString, long timestamp) throws IOException
...
+        scan.setCacheBlocks(false);
+        scan.setCaching(2000);
+        scan.setTimeRange(0, Long.MAX_VALUE);
+        scan.setMaxVersions(1);
+
+        final List<Delete> deleteList = new ArrayList<>();
+        final int batchSize = 100;
+        try (final Table table = adm.getConnection().getTable(tableName);
+             final ResultScanner scanner = table.getScanner(scan)) {
+            for (Result result = scanner.next(); result != null; ) {
```
amcp: if you don't use the post-loop statement, perhaps a while loop is better?

amcp: Also can you write this loop using a .stream() approach?

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
```diff
@@ -47,21 +53,29 @@ public void clearTable(String tableString, long timestamp) throws IOException
...
+                final Delete delete = new Delete(result.getRow());
+                delete.setTimestamp(timestamp);
```
amcp: is the timestamp necessary to delete?

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
```diff
@@ -47,21 +53,29 @@ public void clearTable(String tableString, long timestamp) throws IOException
...
-        // This API call appears to both truncate and reenable the table.
-        log.info("Truncating table {}", tableName);
-        adm.truncateTable(tableName, true /* preserve splits */);
-
-        try {
-            adm.enableTable(tableName);
-        } catch (TableNotDisabledException e) {
-            // This triggers seemingly every time in testing with 1.0.2.
-            log.debug("Table automatically reenabled by truncation: {}", tableName, e);
+        // Unfortunately, linear scanning and deleting rows is faster in HBase when running integration tests than
```
amcp: is this true in general or only for integration tests?

jerryjch:
Wow, it is a big reduction!  How big is the table data size?  The disable table and truncate table probably has some overhead, which makes the full scan faster when the table size is relatively small.

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
jerryjch: We don't need to set this batch size in scan unless it is hugely wide row.

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
```diff
@@ -47,21 +53,29 @@ public void clearTable(String tableString, long timestamp) throws IOException
...
+            for (Result result = scanner.next(); result != null; ) {
+                final Delete delete = new Delete(result.getRow());
+                delete.setTimestamp(timestamp);
+                deleteList.add(delete);
+                result = scanner.next();
+                if (result == null || deleteList.size() == batchSize) {
+                    table.delete(deleteList);
+                    deleteList.clear();
+                }
+            }
```
jerryjch: We need an extra table.delete(deleteList) to send the last batch which can be less than the batchSize?

jerryjch: Could you also update the HBaseAdmin0_98 version so that they are consistent?

mbrukman:
I appreciate the need to improve testing performance, but this will affect actual production performance on HBase and Bigtable quite a bit — it may be a good trade-off for tests where the number of rows is small (and hence, disable/enable dominates the test time), but in production use cases, scans+deletes over millions of rows are much slower than truncation; I believe this to be true for both HBase and Bigtable. (FWIW, Bigtable does not have enable/disable table, so those operations are no-ops at all times, so it wouldn't affect Bigtable anyway.)

Would it be possible to run HBase tests on an implementation of HBase where the enable/disable operations are mocked out or (ahem) disabled, as they seem to be the culprits here, but don't add any functionality that's being tested?

sjudeng:
There's a highly relevant discussion going on in https://groups.google.com/forum/#!topic/janusgraph-dev/hYFY-_nshyg. In summary the existing backend clearStorage implementations are designed specifically for the testing rather than production use case. This is based on the fact that for all storage backends the operation is to truncate rather than drop the keyspace/table, which I think is the expected operation in production (do you agree?). I'm thinking that's why the documentation on the existing utility method that exposes the backend clear functionality (see [`JanusGraph.clear(Graph)`](https://github.com/JanusGraph/janusgraph/blob/bfa3c970555932d4983155168322f72297aaee85/janusgraph-core/src/main/java/org/janusgraph/core/util/JanusGraphCleanup.java#L35-L36)) states "*This method is intended only for development and testing use.*" Additionally it's worth nothing that the scan-and-delete implementation is actually already present in HBase098, with a comment that reinforces that the testing use case is what's implemented (see [HBaseAdmin0_98.clearTable()](https://github.com/JanusGraph/janusgraph/blob/bfa3c970555932d4983155168322f72297aaee85/janusgraph-hbase-parent/janusgraph-hbase-098/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin0_98.java#L65-L66))

I think we should continue to support the testing use case in the clearStorage implementations but we can look at adding true production delete support as a new feature. The other option would be to update the existing clearStorage implementations to support the production instead of test use case (e.g. drop keyspaces/tables) but at the expense of test performance.

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
sjudeng: @amcp I think for integration tests since the table sizes are small.

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
sjudeng: I updated to use a while loop. I was interested in a stream approach too but I needed support for windowing the stream to support batching and it wasn't obvious how to (cleanly) do this.

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
sjudeng: I don't think it's necessary but it might be more correct. The timestamp is a method argument and it is set on the delete in the [HBase098 implementation](https://github.com/JanusGraph/janusgraph/blob/84ead7d907b62061e30c5d3210e5e8422fd4648c/janusgraph-hbase-parent/janusgraph-hbase-098/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin0_98.java#L87).

janusgraph-hbase-parent/janusgraph-hbase-10/src/main/java/org/janusgraph/diskstorage/hbase/HBaseAdmin1_0.java
sjudeng: @jerryjch HBaseAdmin0_98 actually already does scan-and-delete but I updated the implementation to be consistent with this.

sjudeng:
@jerryjch (and @mbrukman) You're definitely right ... the table sizes in tests (at least based on the couple dozen I checked) were no more than about 50 or so rows, which is why scan-and-delete is better. It is strange how much less efficient truncation is (at least for small tables like these) in HBase versus Cassandra.

jerryjch:
Ok.  
I did notice it is only used by tests.
Please add comment to the low level methods as well that the states "This method is intended only for development and testing use." so it is more clear.
Also, you can increase the batch delete size to a larger number, say 1000 or 2000

sjudeng:
@jerryjch Done

PR #414, opened 2017-07-16
Add link to API documentation [skip ci]

### LINKED ISSUE #213: Documentation is missing link to API docs
The [javadoc page](http://docs.janusgraph.org/0.1.0-SNAPSHOT/javadoc.html) should include a link to the API documentation (see corresponding [Titan page](http://s3.thinkaurelius.com/docs/titan/1.0.0/javadoc.html)).

sjudeng:
Fixes #213. Thanks @mbrukman for providing the correct link.

### DIFF
--- a/docs/appendices.txt
+++ b/docs/appendices.txt
@@ -5,7 +5,7 @@
 [appendix]
 == API Documentation (JavaDoc)
 
-Refer to the JanusGraph API documentation for a complete documentation of all core APIs exposed by JanusGraph.
+Refer to the JanusGraph https://javadoc.io/doc/org.janusgraph/janusgraph-core[API documentation] for a complete documentation of all core APIs exposed by JanusGraph.
 We *strongly* encourage all users of JanusGraph to use the Gremlin query language for any queries executed on JanusGraph and to not use JanusGraph's APIs outside of the management system.
 
 


docs/appendices.txt
```diff
@@ -5,7 +5,7 @@
 [appendix]
 == API Documentation (JavaDoc)
 
-Refer to the JanusGraph API documentation for a complete documentation of all core APIs exposed by JanusGraph.
+Refer to the JanusGraph https://javadoc.io/doc/org.janusgraph/janusgraph-core[API documentation] for a complete documentation of all core APIs exposed by JanusGraph.
```
mbrukman: Since this will be compiled with the rest of the docs, what about changing the link here to:

```
https://javadoc.io/doc/org.janusgraph/janusgraph-core/$MAVEN{{project.version}}
```

instead, as that will automatically point to the same version of the API docs as the docs themselves relate to? All other versions are still discoverable from that same page, but the default also matters.

sjudeng: @mbrukman It definitely would be better to link to the correct version. In the case where it's a snapshot version the page will show a "version does not exist" message, though the user could still then change versions on the same page. Is this okay?

https://javadoc.io/doc/org.janusgraph/janusgraph-core/0.2.0-SNAPSHOT

sjudeng: For the record I think it would be better to link to the correct version as you suggest even if there's the error message for the snapshot case.

mbrukman: I think the "does not exist" message is fine, because in that case, javadoc.io provides a superb error message, pointing the user to the latest available version.

The other thing is that any user looking at a previously-released version gets API docs relevant to just that version, rather than the latest, so they're not confused about missing or changed features.

PR #421, opened 2017-07-20
Add IBM to list of JanusGraph users

dpitera:


dpitera:
@mbrukman LMK what else we need to do; thanks!

Issue #338, opened 2017-06-13
Expose native index client for direct queries


wojciechwojcik:
See #318. Still there are more use cases that require direct read-access to index. 

E.g. with Solr if you want to use different parser than standard lucene query parser e.g. dismax or edismax, there are many more fields that can be set that just the query field:
https://cwiki.apache.org/confluence/display/solr/The+DisMax+Query+Parser
https://cwiki.apache.org/confluence/display/solr/The+Extended+DisMax+Query+Parser

With current ``janusGraph.indexQuery(String queryText)`` approach utilizing such features is impossible.

wojciechwojcik:
Some propositions:

A) get native client directly for read-only operations
```
interface Trasaction {
   /** existing interface: just indexName and queryString, all results are always fetched */
    public JanusGraphIndexQuery indexQuery(String indexName, String query);

   /** proposed interface: */
   public JanusGraphNativeIndexQueryHelper indexNativeQuery(String indexBackendName)
}

class JanusGraphNativeIndexQueryHelper {

    /**
    @returns depending on backend type an instance of:
      org.apache.solr.client.solrj.SolrClient
      org.elasticsearch.client.Client

      The client should only be used for read-only query operations.
      This could be used to facilitate count only queries, or setting extra request fields other than just the query string.
    @throws UnsupportedOperationException for embeded lucene backend
    @throws ClassCastException in case of type mismatch
    */
    public <T> T getNativeIndexClient(String indexBackendName)  
  
   /** translate janus management index name to index store name */
   public String getNativeIndexName(String indexName)

   /** substitute v.x, p.x, e.x occurrences in the query, same way org.janusgraph.graphdb.database.IndexSerializer::execute() does */
   public String transtaleQuery(String query)
}
```

B) fix existing extra parameter passing framework, which currently does nothing + allow count only queries

```
org.janusgraph.graphdb.query.graph.IndexQueryBuilder::addParameter()
org.janusgraph.graphdb.query.graph.IndexQueryBuilder::getParameters()
org.janusgraph.diskstorage.indexing.RawQuery::getParameters()
```

This should be used e.g in
``org.janusgraph.diskstorage.solr.SolrIndex::query(RawQuery query)``
to pass further down to: ``SolrQuery::setParam()``

Example parameters (single value, may be comma separated):
````
"defType": "edismax",
 "fl": "*,score"
"qf": "main, extra"
````

This should also support parameters with distinct multiple values e.g.:
````
      "defType": [
[...]

Issue #329, opened 2017-06-09
Make mixed index updates behave in transactional way


wojciechwojcik:
Configuration:
```
# storage backend
gremlin.graph=org.janusgraph.core.JanusGraphFactory
storage.backend=cassandrathrift
storage.hostname=9.128.105.70
schema.default=none

# index backend
index.search.backend=solr
index.search.solr.mode=cloud
index.search.solr.zookeeper-url=9.128.105.70:2182
```

Mixed Index creation:
```
JanusGraphManagement mgmt = = graph.openManagement();
mgmt.buildIndex("indexByName", Vertex.class).addKey("name").buildMixedIndex("search");
```

When you try to do:

```
// add new data
Vertex v1 = graph.addVertex();
v.property("name", "abc")
...
graph.tx().commit();

// since the transaction has been committed, you expect new data to be readable from index
Iterable<Result<JanusGraphVertex>> it = graph.indexQuery("indexByName', "v.name=abc").vertices();
it.next()  // false!
```

In reality when solr server is over the network you will get empty results. 
To workaround this issue you need to add arbitrary sleep time to wait for the data to be propagated to the index.

```
// workaround
LOG.info("wait for full text index refresh");
TimeUnit.SECONDS.sleep(5);

// now if you set your sleep right, you will see the new data
Iterable<Result<JanusGraphVertex>> it = graph.indexQuery("indexByName', "v.name=abc").vertices();
it.next() // true
```

It seems that despite the fact:
1) Janus IS transactional by default and requires you to manually commit() the operations
2) Solr IS transactional by default and requires you to manually commit() the operations

the index updates happen in non-transactional asynchronous fashion.

Now if you look at:
https://github.com/JanusGraph/janusgraph/blob/ec441ac6d87f3359c9740075786ec6de3651745e/janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java#L78

The javadoc states:
_Note, that using indexQuery will may ignore modifications in the current transaction._

That is fine, but still if I have already committed the transaction I expect my changes to be visible as soon as the commit() call returns.

This either has to be changed or at least a new property should be introduced which would allow to set the desired behavior e.g. 
`index.XXX.transactionalUpdates=true/false`
to switch between sync/asnych modes.

[...]

PR #423, opened 2017-07-22
Actually process raw query parameters.  Issue 338

holleyism:
Signed-off-by: Adam Holley <holleya@us.ibm.com>

### DIFF
--- a/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
+++ b/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
@@ -14,36 +14,42 @@
 
 package org.janusgraph.diskstorage.solr;
 
-import com.google.common.base.Joiner;
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Sets;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INDEX_MAX_RESULT_SET_SIZE;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INDEX_NS;
 
-import org.janusgraph.core.Cardinality;
-import org.janusgraph.graphdb.internal.Order;
-import org.janusgraph.core.JanusGraphElement;
-import org.janusgraph.core.attribute.*;
-import org.janusgraph.core.schema.Mapping;
-import org.janusgraph.diskstorage.*;
-import org.janusgraph.diskstorage.configuration.ConfigNamespace;
-import org.janusgraph.diskstorage.configuration.ConfigOption;
-import org.janusgraph.diskstorage.configuration.Configuration;
-import org.janusgraph.diskstorage.indexing.*;
-import org.janusgraph.diskstorage.solr.transform.GeoToWktConverter;
-import org.janusgraph.diskstorage.util.DefaultTransaction;
-import org.janusgraph.graphdb.configuration.PreInitializeConfigOptions;
-import org.janusgraph.graphdb.database.serialize.AttributeUtil;
-import org.janusgraph.graphdb.query.JanusGraphPredicate;
-import org.janusgraph.graphdb.query.condition.*;
+import java.io.IOException;
+import java.io.StringReader;
+import java.lang.reflect.Constructor;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
-import org.janusgraph.graphdb.types.ParameterType;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.http.client.HttpClient;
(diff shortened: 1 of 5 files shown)


jerryjch:
Hi, @holleyism 
Could you open an Issue with description of the problem you want to solve or the improvement you want to make, and reasons?  Then reference this PR to the Issue.
Thanks.

holleyism:
Fixes issue #338

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -988,6 +992,57 @@ public void testRawQueries() {
...
+     * Tests query parameters with raw indexQuery
+     */
+    @Test
+    public void testRawQueriesWithParameters() throws Exception {
+        if (!supportsLuceneStyleQueries()) return;
+        Parameter asc_sort_p = null;
+        Parameter desc_sort_p = null;
+
+        // ElasticSearch and Solr have different formats for sort parameters
+        if (this.getClass().getSimpleName().toLowerCase().contains("elasticsearch")) {
```
jerryjch: Is it possible to use Configuration to get the index backend and do things accordingly?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -988,6 +992,57 @@ public void testRawQueries() {
...
+        JanusGraphVertex v2 = tx.addVertex();
+        JanusGraphVertex v3 = tx.addVertex();
+
+        v1.property("field1", "Hello Hello Hello Hello Hello Hello Hello Hello");
+        v2.property("field1", "Hello blue and yellow meet green");
+        v3.property("field1", "Hello");
+
+        tx.commit();
+
+        Thread.sleep(5000);
```
jerryjch: What is this sleep for?

holleyism: The sleep is to give the index and graph time to commit since the commit() is not guaranteed to be complete upon return.  This pattern was copied from "testIndexQueryWithScore()" and used other places in the unit test.  I've opened a separate issue #429 to replace the use of Thread.sleep with tx().addTransactionListener() as documented in my response on issue #329.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -988,6 +993,60 @@ public void testRawQueries() {
...
+     * Tests query parameters with raw indexQuery
+     */
+    @Test
+    public void testRawQueriesWithParameters() throws Exception {
+        if (!supportsLuceneStyleQueries()) return;
+        Parameter asc_sort_p = null;
+        Parameter desc_sort_p = null;
+
+        // ElasticSearch and Solr have different formats for sort parameters
+        String backend = config.get(ConfigElement.getPath(INDEX_BACKEND, INDEX), String.class);
```
jerryjch: Try this to see if it can make it more explicit and clear. It would retrieve the default if necessary, I hope.
String backend = readConfig.get(INDEX_BACKEND, INDEX);

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -988,6 +993,60 @@ public void testRawQueries() {
...
+
+        tx.commit();
+
+        Thread.sleep(5000);
+
+        List<JanusGraphVertex> vertices = new ArrayList<JanusGraphVertex>();
+        for (JanusGraphIndexQuery.Result<JanusGraphVertex> r : graph.indexQuery("store1", "v.field1:(Hello)").addParameter(asc_sort_p).vertices()) {
+            vertices.add(r.getElement());
+        }
+        int idx=vertices.size()-1;
```
jerryjch: Can we assert the size of the result set so that we are sure we got things back?

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -23,6 +23,7 @@
 import org.janusgraph.core.JanusGraphElement;
 import org.janusgraph.core.attribute.*;
```
amcp: please replace the wildcards with explicit imports

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -515,6 +516,13 @@ private QueryResponse runCommonQuery(RawQuery query, KeyInformation.IndexRetriev
                                 .setIncludeScore(true)
                                 .setStart(query.getOffset())
                                 .setRows(query.hasLimit() ? query.getLimit() : maxResults);
+        for(final Parameter parameter: query.getParameters()) {
+            if (parameter.value() instanceof String[]) {
+                solrQuery = solrQuery.setParam(parameter.key(), (String[]) parameter.value());
```
amcp: does setParam return this? if so, do you even need to re-assign solrQuery?

holleyism: I guess I was doing this out of safety since while unlikely, if setParam changes its implementation and returns a new object then it would be broken.  I'll update it.

jerryjch:
The latest PR has addressed all the comments.  @amcp  If there is no objection or comment, I will merge it shortly. Thanks.

PR #436, opened 2017-07-30
Resolve read-only transaction errors in TinkerPop GraphComputer tests

### LINKED ISSUE #427: JanusGraph read only transaction does not throw any exception on write
[JanusTransactionTest.zip](https://github.com/JanusGraph/janusgraph/files/1173837/JanusTransactionTest.zip)

The following code (see attachment)
````
            System.out.println("Base transaction");
            
            TransactionBuilder builder = graph.buildTransaction();
            JanusGraphTransaction transaction = builder.readOnly().start();
            //JanusGraphTransaction transaction = builder.start();
            GraphTraversalSource g = transaction.traversal();
            
            System.out.println(g.V().count().next());
            g.V().drop().tryNext();
            System.out.println(g.V().count().next());
            transaction.commit();

            System.out.println("Other transaction");
            
            transaction = graph.newTransaction();
            System.out.println(transaction.traversal().V().count().next());
            transaction.tx().close();

            System.out.println("Done");
````

simply prints:
````
Base transaction
[main] WARN org.janusgraph.graphdb.transaction.StandardJanusGraphTx - Query requires iterating over all vertices [()]. For better performance, use indexes
3
[main] WARN org.janusgraph.graphdb.transaction.StandardJanusGraphTx - Query requires iterating over all vertices [()]. For better performance, use indexes
[main] WARN org.janusgraph.graphdb.transaction.StandardJanusGraphTx - Query requires iterating over all vertices [()]. For better performance, use indexes
0
Other transaction
[main] WARN org.janusgraph.graphdb.transaction.StandardJanusGraphTx - Query requires iterating over all vertices [()]. For better performance, use indexes
3
Done
[main] INFO org.janusgraph.diskstorage.cassandra.thrift.CassandraThriftStoreManager - Closed Thrift connection pooler.
````

which correctly prevents the ``drop()`` statement from modifications, still NO exception, log or any indication of error is presented to the user.

If you look at the javadoc of ``TransactionBuilder::readOnly()`` it clearly states:
_Makes the transaction read only. **Any writes will cause an exception**.._

This is further misleading as ``drop()`` effects are still immediately visible in its own transaction context, which means that e.g. trying to access a vertex property you will get such error:
````
java.lang.IllegalStateException: Vertex with id 4224 was removed.
	at org.apache.tinkerpop.gremlin.structure.Element$Exceptions.elementAlreadyRemoved(Element.java:154) ~[gremlin-core-3.2.5.jar:3.2.5]
	at org.janusgraph.core.InvalidElementException.removedException(InvalidElementException.java:58) ~[janusgraph-core-0.2.0-SNAPSHOT.jar:?]
	at org.janusgraph.graphdb.vertices.AbstractVertex.verifyAccess(AbstractVertex.java:86) ~[janusgraph-core-0.2.0-SNAPSHOT.jar:?]
	at org.janusgraph.graphdb.vertices.AbstractVertex.query(AbstractVertex.java:133) ~[janusgraph-core-0.2.0-SNAPSHOT.jar:?]
	at org.janusgraph.graphdb.vertices.AbstractVertex.properties(AbstractVertex.java:173) ~[janusgraph-core-0.2.0-SNAPSHOT.jar:?]
	at org.apache.tinkerpop.gremlin.structure.Vertex.property(Vertex.java:70) ~[gremlin-core-3.2.5.jar:3.2.5]
````
stating that a vertex was already removed inside a read only transaction (!), which is NOT true.

sjudeng:
The original (pre-TinkerPop-3.2.4) implementation didn't check whether transaction was read-only and (as expected) threw UnsupportedOperationExceptions when calling CacheVertexProperty.remove() during OLAP traversals. This behavior causes numerous TinkerPop test failures starting in v3.2.4,  where TraversalVertexProgram.HALTED_TRAVERSALS are cleaned up in [TraversalVertexProgram](https://github.com/apache/tinkerpop/blob/3.2.4/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/TraversalVertexProgram.java#L294-L296). Interestingly the logic is in 3.2.3 as well so I'm not sure why tests only started failing on this in 3.2.4.

Initial updates to always silently ignore the read-only transaction errors resulted in issues as identified in #427. The update here fixes #427 by limiting updates to ignore read-only transaction errors to just FulgoraGraphComputer, which itself is used/supported only for testing. All default/TinkerPop tests are passing (no additional test opt-outs were required).

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/ReadOnlyTransactionException.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/ReadOnlyTransactionException.java
@@ -1,0 +1,27 @@
+// Copyright 2017 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.core;
+
+public class ReadOnlyTransactionException extends UnsupportedOperationException {
+
+    public ReadOnlyTransactionException(String msg) {
+        super(msg);
+    }
+
+    public ReadOnlyTransactionException(String msg, Throwable cause) {
+        super(msg, cause);
+    }
+
+}
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/VertexProgramScanJob.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/VertexProgramScanJob.java
@@ -16,6 +16,7 @@ package org.janusgraph.graphdb.olap.computer;
 
 import org.janusgraph.core.JanusGraph;
 import org.janusgraph.core.JanusGraphVertex;
+import org.janusgraph.core.ReadOnlyTransactionException;
 import org.janusgraph.diskstorage.EntryList;
 import org.janusgraph.diskstorage.configuration.Configuration;
 import org.janusgraph.diskstorage.keycolumnvalue.SliceQuery;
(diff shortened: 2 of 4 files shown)


pluradj:
Do you prefer these updates rather than opting out of the test cases? My gut reaction is that it seems unusual to have a reference to a specific `VertexProgram` in `CacheVertex`.

sjudeng:
@pluradj Yeah definitely same here. Unfortunately I wasn't able to isolate a change I could make somewhere more appropriate, say in JanusGraph olap classes, to resolve the failures. Are we okay opting-out of the tests?

sjudeng:
On second thought it looks like opting-out might not work as I'm seeing many more test failures if I restore the original CacheVertexProperty implementation (`if (!tx().isRemovedRelation(super.longId())) {...}`). I'll try to find another option here.

pluradj:
@sjudeng bummer, thanks for the update

sjudeng:
@pluradj I just an update to put a try-catch in [VertexProgramScanJob](https://github.com/JanusGraph/janusgraph/blob/01fc5b66725443a67e549be6a05c54e71eec655a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/VertexProgramScanJob.java#L106) to silently ignore the error under FulgoraGraphComputer. Still ugly but it does limit the impact to FulgoraGraphComputer (e.g. testing).

Issue #442, opened 2017-08-01
Cannot create vertex-centric-index on edgeLabel with SIMPLE multiplicity


dpitera:
This issue existed in Titan: https://github.com/thinkaurelius/titan/issues/1275 and still exists here.

In the issue listed above, someone alluded to a possible solution being to update the `isContrained` function [here](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/core/Multiplicity.java#L67-L72) however I an edgeLabel is technically `constrained` if it is `SIMPLE`-- the constraint being that it can only have one edge between any two vertices. 

Rather than updating that code, I think it makes more sense and will upkeep system integrity by instead changing the caller to open up vertex-centric-index creation to work with SIMPLE by checking [the isUnique function](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/core/Multiplicity.java#L81-L91) instead-- which should correlate to the proper constraints for a vertex-centric-index in relation to an edgeLabel.

pluradj:
#443 merged

PR #443, opened 2017-08-01
Fix vertex-centric-index multiplicity constraints

dpitera:
Issues: https://github.com/JanusGraph/janusgraph/issues/442

### DIFF
--- a/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -1360,6 +1360,51 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
     }
 
     @Test
+    public void testVertexCentricEdgeIndexOnSimpleMultiplicityShouldWork() {
+        clopen(option(LOG_SEND_DELAY, MANAGEMENT_LOG), Duration.ofMillis(0),
+                option(KCVSLog.LOG_READ_LAG_TIME, MANAGEMENT_LOG), Duration.ofMillis(50),
+                option(LOG_READ_INTERVAL, MANAGEMENT_LOG), Duration.ofMillis(250)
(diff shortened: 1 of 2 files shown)


janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -1360,6 +1360,29 @@ public void testGotGIndexRemoval() throws InterruptedException, ExecutionExcepti
...
 
     @Test
+    public void testVertexCentricIndexOnSimpleMultiplicityShouldWork() {
+        clopen(option(LOG_SEND_DELAY, MANAGEMENT_LOG), Duration.ofMillis(0),
+                option(KCVSLog.LOG_READ_LAG_TIME, MANAGEMENT_LOG), Duration.ofMillis(50),
+                option(LOG_READ_INTERVAL, MANAGEMENT_LOG), Duration.ofMillis(250)
+        );
+        PropertyKey time = mgmt.makePropertyKey("time").dataType(String.class).cardinality(Cardinality.SINGLE).make();
+        EdgeLabel friend = mgmt.makeEdgeLabel("friend").multiplicity(Multiplicity.SIMPLE).make();
+        mgmt.buildEdgeIndex(friend, "byTime", Direction.OUT, decr, time);
```
pluradj: I think we'd need a test for `buildPropertyIndex()` too

dpitera: these indexes don't make usage of `edgeLabels` which deal with `MULTIPLICITY` concerns

dpitera:
Thanks @pluradj. Updated code.

krlohnes:
+1 (non-binding)

The difference between unique and constrained is a little subtle, but this is definitely correct. Looks to me like this was just a `method-name-looks-right` type of bug.

Issue #418, opened 2017-07-18
Failure to correctly set ids.store-name=titan_ids can corrupt a Titan Cassandra


smcquillan:
I've been running into issues while connecting JanusGraph 0.1.2-SNAPSHOT to instances of Cassandra which have graphs previously populated by Titan 1.0.0. Initially things looked to be working fine as JanusGraph could read the data but things began to go wrong when I created a new property with something like the following:

```
JanusGraphManagement mgmt = g.openManagement();
mgmt.makePropertyKey("newProperty").dataType(type).make();
```

The `newProperty` which had a data type of `String` got created fine. However, I later discovered that one of my existing properties which had previously been of type `Long` had now changed to type `String` causing many of our application tests to fail and effectively **the graph within Cassandra was now corrupted**.

After digging around a bit found issue #228 about ID_STORE_NAME table renamed from Titan and thought that must be where I was going wrong. Looking in Cassandra I saw that in addition to the existing `titan_ids` a new `janusgraph_ids` had been created and that confirmed it. But that wasn't the whole story because just adding the following to the configuration when connecting:

```
graph.titan-version=1.0.0
ids.store-name=titan_ids
```

Works fine when connecting to an empty Cassandra, but if you connect to one which had previously been populated by Titan you'll get the following:

```
WARN [2017-07-18 12:03:43,880] [main] o.j.g.c.GraphDatabaseConfiguration -  Local setting ids.store-name=titan_ids (Type: GLOBAL_OFFLINE) is overridden by globally managed value (janusgraph_ids).  Use the ManagementSystem interface instead of the local configuration to control this setting. 

```

The message is questionable because the implicit value of `ids.store-name` in the Cassandra being connected to is actually `titan_ids` so we're not overridding anything. And worse still if you don't notice the message and then go on to change the schema you'll risk corrupting the graph because the changes will to into `janusgraph_ids` instead. I worked around the issue by using the  ManagementSystem interface as the warning suggested:

```
mgmt = g.openManagement();
mgmt.set(ConfigElement.getPath(GraphDatabaseConfiguration.IDS_STORE_NAME), "titan_ids");
mgmt.commit();
```

Even this had its complications as you need to bring down the cluster and connect with just a single JanusGraph / Titan as any other connected instances will prevent the change from being applied.

[...]

PR #445, opened 2017-08-03
Failure to correctly set ids.store-name=titan_ids can corrupt a Titan Cassandra Issue #418

smcquillan:
To fix Issue #418 added a check for the following:

1. `graph.titan-version` is set and is '1.0.0'
2. `ids.store-name` is set in LOCAL and equal to 'titan_ids'
3. `ids.store-name` is not set in GLOBAL ( as read from KCVSConfiguration )

If all the above are true then set `ids.store-name` in the overwrite modifiable configuration to be `titan_ids`. This then prevents the global configuration inheriting the default value of janusgraph_ids but only in the case where ids.store-name had not already been set.

It will now throw an exception and abort if `graph.titan-version` has been set and `ids.store-name` was not also set to the value `titan_ids` which should help preserve data integrity.

### DIFF
--- a/docs/migrating.txt
+++ b/docs/migrating.txt
@@ -1,0 +1,31 @@
+[[migrating-titan]]
+== Migrating from Titan
+
+This page describes some of the Configuration options that JanusGraph provides to allow migration of data from a data store which had previously been created by Titan.
+
+=== Configuration
+
+When connecting to an existing Titan data store the `graph.titan-version` property should be set to the Titan version used to create that store.
+
+[source, properties]
+----
+graph.titan-version=1.0.0
+----
+
+The ID store name in JanusGraph is configurable via the `ids.store-name` property whereas in Titan it was a constant. If the `graph.titan-version` has been set, as in the example above, then you do **not** need to explicitly set this value as it will default to `titan_ids`.
+
+=== Cassandra
+
+The default keyspace used by Titan was `titan` and in order to reuse that existing keyspace the `storage.cassandra.keyspace` property needs to be set accordingly.
+
+[source, properties]
+----
+graph.titan-version=1.0.0
+storage.cassandra.keyspace=titan
+----
+
+These configuration options allow JanusGraph to read data from a Cassandra database which had previously been created by Titan.  However, once JanusGraph writes back to that database it will register additional serializers which mean that it will no longer be compatible with Titan. Users are therefore encouraged to backup the data in Casssandra before attempting to use it with the JanusGraph release. 
+
+=== BerkeleyDB
+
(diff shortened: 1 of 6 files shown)


pluradj:
Thanks @smcquillan . I think this looks good.

It would be worthwhile to start a documentation item under Section V: Advanced Topics on migrating an existing Titan graph to JanusGraph. In the new doc, you can spell out the requirements for these 2 configuration parameters (which would be easily lost in Chapter 12. Configuration Reference).

pluradj:
Another thought: is there any particular reason to require `graph.titan-version` and `ids.store-name`? The expected `ids.store-name` value for a Titan migration is a constant, and with your code fixes, it can be set on the initial startup of JanusGraph.

smcquillan:
@pluradj Apologies for the delayed response, have been on vacation last week. There's no particular reason to require both `graph.titan-version` and `ids.store-name` be set because as far as I can see the value of `ids.store-name` would always be `titan_ids` if Titan had created the graph.

If as suggested in your earlier comment we add a new advanced topic on migrating an existing Titan graph to JanusGraph then we could drop the need to explicitly set `ids.store-name` and explain that if you did choose to explicitly set its value then it should be to `titan_ids`. Unless there are any objections I'll look into making these changes this week and push them up into this pull request.

smcquillan:
@pluradj I've merged up a changes to remove the restriction that `ids.store-name` must be explicitly
set. So now if `graph.titan-version` **is** set and `ids.store-name` is not already in the global config then the default will become `titan_ids`.

Also added a new section on migrating data from Titan under Section V: Advanced Topics and updated the appendix on upgrading to reference it.

docs/migrating.txt
```diff
@@ -0,0 +1,31 @@
...
+----
+graph.titan-version=1.0.0
+storage.cassandra.keyspace=titan
+----
+
+These configuration options allow JanusGraph to read data from a Cassandra database which had previously been created by Titan.  However, once JanusGraph writes back to that database it will register additional serializers which mean that it will no longer be compatible with Titan. Users are therefore encouraged to backup the data in Casssandra before attempting to use it with the JanusGraph release. 
+
+=== BerkeleyDB
+
+The BerkeleyDB version has been updated, and it contains changes to the file format stored on disk. This file format change is forward compatible with previous versions of BerkeleyDB, so existing graph data stored with Titan can be read in. However, once the data has been read in with the newer version of BerkeleyDB, those files can no longer be read by the older version. Users are encouraged to backup the BerkeleyDB storage directory before attempting to use it with the JanusGraph release.
```
pluradj: @jerryjch are there any specific HBase migration concerns from Titan that should be documented here?

pluradj: You can mention the `storage.hbase.table=titan` as the previous default value, similar to what is in the Cassandra section.

smcquillan: @pluradj I've added a section for HBase which basically just covers the `storage.hbase.table=titan` property and I've included a mention about serialisers which I assume will behave the same with HBase as the did with Cassandra. However, I'm not in a position to test with HBase and so will await feedback from @jerryjch on whether this new section is sufficient.

Issue #408, opened 2017-07-11
Error while commiting index mutations for removing properties that are indexed


scottrpatterson:
I am running es embedded in the same jvm as JanusGraph 0.1.1. 
```
index.fts.backend=elasticsearch
index.fts.directory= /test/es
index.fts.elasticsearch.client-only=false
index.fts.elasticsearch.local-mode= true
index.fts.elasticsearch.interface=NODE
```

While performing a commit for a bunch of mutations, one of my changes to remove a property throws a exception complaining about a variable name in es. 

> Script1.groovy: 1: The current scope already contains a variable of the name index

Since the script that is executed in elasticsearch is generated by janusgraph, where is the suspect variable name "index" coming from? If there are multiple properties().drop() in the same commit, will that variable "index" be reused in the same es script?

The traversal is:
```
g.traversal().V().has("propA", "valueA").hasLabel("art")
                        .has("propB", "valueB"
                        .properties("propB").drop().iterate()
```

The backend exception:
```
[0]: index [janusgraph], type [FTSccde89523844234e61420a730ba96894], id [2wejoi8], message [ElasticsearchIllegalArgumentException[failed to execute script]; nested: GroovyScriptCompilationException[MultipleCompilationErrorsException[startup failed:
Script1.groovy: 1: The current scope already contains a variable of the name index
 @ line 1, column 156.
   propB"].remove(index);def index = ct
                                 ^

1 error
]]; ]
    at org.janusgraph.diskstorage.es.ElasticSearchIndex.mutate(ElasticSearchIndex.java:628)
    at org.janusgraph.diskstorage.indexing.IndexTransaction$1.call(IndexTransaction.java:137)
    at org.janusgraph.diskstorage.indexing.IndexTransaction$1.call(IndexTransaction.java:134)
    at org.janusgraph.diskstorage.util.BackendOperation.executeDirect(BackendOperation.java:69)
    at org.janusgraph.diskstorage.util.BackendOperation.execute(BackendOperation.java:55)
    at org.janusgraph.diskstorage.indexing.IndexTransaction.flushInternal(IndexTransaction.java:134)
    at org.janusgraph.diskstorage.indexing.IndexTransaction.commit(IndexTransaction.java:115)
    at org.janusgraph.diskstorage.BackendTransaction.commitIndexes(BackendTransaction.java:140)
    at org.janusgraph.graphdb.database.StandardJanusGraph.commit(StandardJanusGraph.java:743)
[...]

PR #449, opened 2017-08-07
ES: Ensure unique script var names

robertdale:
Fixes Issue #408

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -577,7 +577,9 @@ public class ElasticSearchIndex implements IndexProvider {
     }
 
     private String getDeletionScript(KeyInformation.IndexRetriever informations, String storename, IndexMutation mutation) throws PermanentBackendException {
-        StringBuilder script = new StringBuilder();
+        final StringBuilder script = new StringBuilder();
+        final String INDEX_NAME = "index";
+        int i = 0;
         for (IndexEntry deletion : mutation.getDeletions()) {
             KeyInformation keyInformation = informations.get(storename).get(deletion.field);
 
@@ -591,12 +593,13 @@ public class ElasticSearchIndex implements IndexProvider {
                 case SET:
                 case LIST:
                     String jsValue = convertToJsType(deletion.value, compat.scriptLang(), Mapping.getMapping(keyInformation));
-                    script.append("def index = ctx._source[\"").append(deletion.field).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(deletion.field).append("\"].remove(index);");
+                    String index = INDEX_NAME + i++;
+                    script.append("def ").append(index).append(" = ctx._source[\"").append(deletion.field).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(deletion.field).append("\"].remove(").append(index).append(");");
                     if (hasDualStringMapping(informations.get(storename, deletion.field))) {
-                        script.append("def index = ctx._source[\"").append(getDualMappingName(deletion.field)).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(getDualMappingName(deletion.field)).append("\"].remove(index);");
+                        index = INDEX_NAME + i++;
+                        script.append("def ").append(index).append(" = ctx._source[\"").append(getDualMappingName(deletion.field)).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(getDualMappingName(deletion.field)).append("\"].remove(").append(index).append(");");
                     }
                     break;
-
             }
         }
         return script.toString();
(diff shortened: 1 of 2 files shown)


janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1810,7 +1811,22 @@ private void testCollection(Cardinality cardinality, String property, Object val
...
+        g = graph.traversal();
+        v1 = g.addV().property(property, value1).property(property, value2).next();
+        Vertex v2 = g.addV().property(property, value1).property(property, value2).next();
+        clopen();//Flush the index
+        g = graph.traversal();
+        assertEquals(2, g.V().has(property, value1).toList().size());
+        g.V().properties().drop().iterate();
+        clopen();//Flush the index
+        g = graph.traversal();
+        assertEquals(false, g.V().has(property, value1).hasNext());
```
mbrukman: `assertEquals(false, ...)` == `assertFalse(...)` but the latter is shorter and easier to read.

Similarly below.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1810,7 +1811,22 @@ private void testCollection(Cardinality cardinality, String property, Object val
             assertEquals(v1, getOnlyElement(graph.query().has(property, Geo.WITHIN, Geoshape.circle(2.0, 2.0, 0.1)).vertices()));
         }
 
-
+        
+        // Test traversal property drop Issue #408
+        GraphTraversalSource g = graph.traversal();
+        g.V().drop().iterate();
+        clopen();//Flush the index
```
mbrukman: Add space between code and comment for readability? Similarly below.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -591,13 +592,16 @@ private String getDeletionScript(KeyInformation.IndexRetriever informations, Str
                 case SET:
                 case LIST:
                     String jsValue = convertToJsType(deletion.value, compat.scriptLang(), Mapping.getMapping(keyInformation));
-                    script.append("def index = ctx._source[\"").append(deletion.field).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(deletion.field).append("\"].remove(index);");
+                    String index = "index" + i;
+                    script.append("def ").append(index).append(" = ctx._source[\"").append(deletion.field).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(deletion.field).append("\"].remove(").append(index).append(");");
                     if (hasDualStringMapping(informations.get(storename, deletion.field))) {
-                        script.append("def index = ctx._source[\"").append(getDualMappingName(deletion.field)).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(getDualMappingName(deletion.field)).append("\"].remove(index);");
+                    	i++;
```
mbrukman: Fix indent, here and on the next line.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1810,7 +1811,22 @@ private void testCollection(Cardinality cardinality, String property, Object val
             assertEquals(v1, getOnlyElement(graph.query().has(property, Geo.WITHIN, Geoshape.circle(2.0, 2.0, 0.1)).vertices()));
         }
 
-
+        
+        // Test traversal property drop Issue #408
```
mbrukman: Should this be a separate test method to isolate it from other tests? Seems like it's independent from what's above.

robertdale: `testCollection()` is a common method for any (list,set) collection tests. It is called by `testIndexing()` which is called by `@Test testListIndexing()` and `@Test testSetIndexing()`. So I think of `testCollection()` as the gamut of all indexing tests on a collection.  That said, it could be pulled out into its own method `testMultiplePropertyDrop()` and add that to `testIndexing()` after the `testCollection()` calls.  But then do we call that for each property type or just one?  What's your preference?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -591,13 +592,16 @@ private String getDeletionScript(KeyInformation.IndexRetriever informations, Str
                 case SET:
                 case LIST:
                     String jsValue = convertToJsType(deletion.value, compat.scriptLang(), Mapping.getMapping(keyInformation));
-                    script.append("def index = ctx._source[\"").append(deletion.field).append("\"].indexOf(").append(jsValue).append("); ctx._source[\"").append(deletion.field).append("\"].remove(index);");
+                    String index = "index" + i;
```
sjudeng: Might be a little cleaner to replace with `"index" + i++` here and below and remove the standalone `i++` lines.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
robertdale: @mbrukman Do you have any additional comments or are we good to merge?

mbrukman: I don't have a strong opinion on this, just wanted to call out as a general anti-pattern from what I've seen elsewhere to cram too many test cases into a single test method, so wanted to suggest that we consider it here. If it does not make sense or would complicate the test cases in a different way, let's keep it as is.

Issue #456, opened 2017-08-10
Remove VertexCentricQueryBuilder orderBy cardinality constraint


dpitera:
Remove orderBy cardinality check

    While it is true that JanusGraph's VertexCentricQuery does not know how
    to orderBy multplicity > SINGLE, this check is actually acting a red
    herring.

    VertexCentricIndexes are meant to either return a list of
    vertexProperties or edges-- where we order and sort these returns by
    either a vertex property's properties (metaproperties) or vertex's
    adjacent edge's properties.

    Metaproperties and edge properties are-- by construction-- inherently
    single valued, i.e. we cannot store multiple values on a singular edge
    property or metaproperty key.

    However, you can create metaproperties and edge properties adhering to a
    given JanusGraph propertyKey whose cardinality > SINGLE, i.e. equal to
    LIST or SET. The current cardinality check is disallowing users to use
    vertex centric indexes on these existing edges and metaproperties
    without changing their schemae, which is an ~impossible effort on a large
    graph.

    Therefore, the check here-- to ensure that the VertexCentricQuery
    sorts by a propertyKey with a cardinality equivalent to SINGLE is
    superfluous and arguably incorrect; by design, these queries sort
    metaproperties and edge properties, which cannot store multiple values
    for a given key.

    Furthermore, any VertexCentricQuery constructed from a TinkerPop gremlin
    traversal will never add the orderBy step to a vertexProperty (the only
    type of property that can actually store multiple values):
    https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java#L181

    This just leaves us in a situation where you may argue that a user
    previously using the queryBuilder API directly to try to orderBy vertexProperties with
    cardinality > SINGLE will no longer receive an error. However, the
    alternative was _not_ true: a user trying to orderBy
    vertexProperties with cardinality == SINGLE did not receive an error,
    even though such a request on an ordering makes no sense. Similarly,
    ordering by a single propertyKey stored with multiple values makes
    sense, as we are only actually returning a singular propertyKey.

    In this scenario now, instead of getting an error, we will receive
    results with an undefined ordering, which actually matches the behavior
    as the behavior currently exists:

```
    gremlin> def v = graph.addVertex(); for (int i = 0; i < 3; i++) {
    v.property("sensor", i, "time", i); }; graph.tx().commit();
    ==>null
    gremlin> graph.vertices()
[...]

PR #457, opened 2017-08-10
Remove VertexCentricQueryBuilder orderBy cardinality check

dpitera:
While it is true that JanusGraph's VertexCentricQuery does not know how
to orderBy multplicity > SINGLE, this check is actually acting a red
herring.

VertexCentricIndexes are meant to either return a list of
vertexProperties or edges-- where we order and sort these returns by
either a vertex property's properties (metaproperties) or vertex's
adjacent edge's properties.

Metaproperties and edge properties are-- by construction-- inherently
single valued, i.e. we cannot store multiple values on a singular edge
property or metaproperty key.

However, you can create metaproperties and edge properties adhering to a
given JanusGraph propertyKey whose cardinality > SINGLE, i.e. equal to
LIST or SET. The current cardinality check is disallowing users to use
vertex centric indexes on these existing edges and metaproperties
without changing their schemae, which is an ~impossible effort on a large
graph.

Therefore, the check here-- to ensure that the VertexCentricQuery
sorts by a propertyKey with a cardinality equivalent to SINGLE is
superfluous and arguably incorrect; by design, these queries sort
metaproperties and edge properties, which cannot store multiple values
for a given key.

Furthermore, any VertexCentricQuery constructed from a TinkerPop gremlin
traversal will never add the orderBy step to a vertexProperty (the only
type of property that can actually store multiple values):
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java#L181

This just leaves us in a situation where you may argue that a user
previously using the queryBuilder API directly to try to orderBy vertexProperties with
cardinality > SINGLE will no longer receive an error. However, the
alternative was _not_ true: a user trying to orderBy
vertexProperties with cardinality == SINGLE did not receive an error,
even though such a request on an ordering makes no sense. Similarly,
ordering by a single propertyKey stored with multiple values makes
no sense, as we are only actually returning a singular propertyKey.

In this scenario now, instead of getting an error, we will receive
results with an undefined ordering, which actually matches the behavior
as the behavior currently exists:

gremlin> def v = graph.addVertex(); for (int i = 0; i < 3; i++) {
v.property("sensor", i, "time", i); }; graph.tx().commit();
==>null
gremlin> graph.vertices()
==>v[4184]
gremlin> graph.traversal().V(4184).properties() // we have 1 property with LIST value on our vertex
==>vp[sensor->0.0]
==>vp[sensor->1.0]
==>vp[sensor->2.0]
gremlin> graph.traversal().V(4184).properties().next().properties() // each value in list has a metaproperty of time equivalent to its own value
==>p[time->0]
gremlin> graph.traversal().V(4184).next().query().orderBy("time",
decr).properties()
==>vp[sensor->2.0]
==>vp[sensor->1.0]
==>vp[sensor->0.0]
gremlin> graph.traversal().V(4184).next().query().orderBy("sensor",
decr).properties() // we can sort vertexProperties by their metaProperties
==>vp[sensor->0.0]
==>vp[sensor->1.0]
==>vp[sensor->2.0]
gremlin> graph.traversal().V(4184).next().query().orderBy("time", decr).properties() // Notice, there is no defined ordering for properties without metapropertie
==>vp[sensor->10.0]
==>vp[sensor->2.0]
==>vp[sensor->1.0]
==>vp[sensor->0.0]
gremlin> graph.traversal().V(4184).next().query().orderBy("sensor", decr).properties() // Similarly there is no defined ordering if we sort by vertex properties themselves
==>vp[sensor->0.0]
==>vp[sensor->1.0]
==>vp[sensor->2.0]
gremlin> graph.traversal().V(4184).next().property("sensor", 10);
==>vp[sensor->10.0]
gremlin> graph.traversal().V(4184).next().property("sensor", -10, "time", -10);
==>vp[sensor->-10.0]
gremlin> graph.traversal().V(4184).next().query().orderBy("time", decr).properties() // however ordering definitely works when applicable, i.e. a vertex property has metaproperties in question
==>vp[sensor->10.0]
==>vp[sensor->2.0]
==>vp[sensor->1.0]
==>vp[sensor->0.0]
==>vp[sensor->-10.0]

Therefore, this undefined ordering is what we expect in this scenario
where we use the API incorrectly, i.e. trying to orderBy a
vertexProperty when the query is meant to orderBy a metaproperty or edge
property, which is guaranteed to only store a singular value and
therefore the check is not necessary.

Issues: #456

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BaseVertexCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BaseVertexCentricQueryBuilder.java
@@ -205,7 +205,6 @@ public abstract class BaseVertexCentricQueryBuilder<Q extends BaseVertexQuery<Q>
         Preconditions.checkArgument(key != null && order != null, "Need to specify and key and an order");
         Preconditions.checkArgument(Comparable.class.isAssignableFrom(key.dataType()),
                 "Can only order on keys with comparable data type. [%s] has datatype [%s]", key.name(), key.dataType());
-        Preconditions.checkArgument(key.cardinality() == Cardinality.SINGLE, "Ordering is undefined on multi-valued key [%s]", key.name());
         Preconditions.checkArgument(!(key instanceof SystemRelationType), "Cannot use system types in ordering: %s", key);
         Preconditions.checkArgument(!orders.containsKey(key));
         Preconditions.checkArgument(orders.isEmpty(), "Only a single sort order is supported on vertex queries");
--- a/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -1405,6 +1405,49 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
     }
 
     @Test
+    public void testVertexCentricIndexOrderingOnEdgePropertyWithCardinalityList() throws InterruptedException, ExecutionException {
+        clopen(option(LOG_SEND_DELAY, MANAGEMENT_LOG), Duration.ofMillis(0),
+                option(KCVSLog.LOG_READ_LAG_TIME, MANAGEMENT_LOG), Duration.ofMillis(50),
+                option(LOG_READ_INTERVAL, MANAGEMENT_LOG), Duration.ofMillis(250)
+        );
+        PropertyKey time = mgmt.makePropertyKey("time").dataType(Integer.class).cardinality(Cardinality.LIST).make();
+        EdgeLabel friend = mgmt.makeEdgeLabel("friend").multiplicity(Multiplicity.MULTI).make();
+        mgmt.buildEdgeIndex(friend, "byTime", Direction.OUT, decr, time);
+        finishSchema();
+        JanusGraphVertex v = tx.addVertex();
+        for (int i = 200; i < 210; i++) {
+            JanusGraphVertex o = tx.addVertex();
+            v.addEdge("friend", o, "time", i);
+        }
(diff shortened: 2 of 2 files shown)


dpitera:
> It seems that this boils down to a schema misconfiguration? Perhaps it should log a warning.

Due to the facts I've listed above, I think the check doesn't even belong here. If we wanted to check something, we would check something along the lines of [this](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java#L181), but we don't have this type of information in the internal query builder. And again, all internal queries constructed through the users' usage of the TinkerPop APIs will behave correctly.

pluradj:
OK, fair enough. We should be encouraging developers to use plain Gremlin (rather than the `query()` method), and the [docs](http://docs.janusgraph.org/latest/indexes.html#vertex-indexes) already show this. I agree -- not interested in changing the property key definitions to be specific for edges or meta-properties.

Looks good to me. +1

krlohnes:
+1 (non-binding)

I think this is good from a usability standpoint as well, where a user can't shoot themselves in the foot by using a multi-cardinality property on an edge and not being able to create a vertex centric index. Considering that schema properties are immutable, this is a particularly important consideration if one already has a large amount of data with a given schema already loaded before discovering this.

pluradj:
@sjudeng Could you double check this please? GitHub said there was a merge conflict on this one, and I used the GitHub UI to resolve the conflict. It looks like the 2 file updates are now good to merge, but the resulting [merge commit](https://github.com/JanusGraph/janusgraph/pull/457/commits/5d912d30bdb0bf85843a163b6d124f605e5c1f7f) seems to have a lot of extra gorp in it.

Issue #420, opened 2017-07-19
JanusGraphIoRegistry support of multiple serializer versions


robertdale:
Currently, there is only one JanusGraphIoRegistry.  There are older v1 of serializers. Then there are newer v2 of serializers.  For instance, there are GraphSON 1.0 and GraphSON 2.0.  Unfortunately, something like Geoshape is hardcoded in JanusGraphSONModule (to which JanusGraphIoRegistry delegates) to use GeoshapeGsonSerializerV1 .  I don't know the story behind Geoshape support for GraphSON 1.0. I could not get a remote client to work with a gremlin server with what's there.

However, if JanusGraphSONModule registers Geoshape with GeoshapeGsonSerializerV2d0, it all works wonderfully.

I propose that JanusGraphIoRegistry is split into v1 and v2 support much like TinkerIoRegistry is split into V1d0 and V2d0.

See also 
- https://groups.google.com/forum/#!topic/janusgraph-users/RswTmVraChg
- http://tinkerpop.apache.org/docs/3.2.5/dev/io/#graphson-2d0

sjudeng:
Not sure if it's relevant but recently-merged #376 (which added the support for GraphSON 2.0) included [some tests](https://github.com/JanusGraph/janusgraph/blob/78bbe5aa65dc2f31e13a004fa21a213de2741cc4/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java#L77-L82) for GraphSON 1.0/2.0 IO.

robertdale:
@sjudeng It is relevant. Note that only the [v1 serializers are registered with JanusGraphSONModule](https://github.com/JanusGraph/janusgraph/pull/376/files#diff-4a920cb0f4947a2678963f6822821a9fR53) (hence JanusGraphIoRegistry).  When using the GraphSONMessageSerializerGremlinV2d0, Geoshape is still locally serialized as GraphSON 1.0 which uses the `@class` key to determine type.  When v2 deserializes, the Geoshape is missing the v2 `@type` key to determine type, thus is deserialized as a Map.

sjudeng:
Got it. Your approach of separate JanusGraphIoRegistry sounds good to me. We'll have to add a V3d0 when we update to TinkerPop 3.3.

Separately it sounded like you weren't able to get 1.0 serialization to work properly through gremlin server? Is this a separate (bug) issue?

smcquillan:
[...]

PR #463, opened 2017-08-16
Errors deserialising Geoshape objects created by Titan

smcquillan:
This pull request is a continuation of a discussion @sjudeng from issue #420.

While testing I noticed that the latitude and longitude values being deserialised by JanusGraph didn't match those originally written by Titan. The problem looks like there's a bug in changes made in #376 when the repositioning of the buffer. The problem appears to be in the assumption that the buffer position always starts at zero. Adding a second geo point to the buffer used in the unit test [SerializerTest#testLegacyPointSerialization](https://github.com/JanusGraph/janusgraph/blob/78bbe5aa65dc2f31e13a004fa21a213de2741cc4/janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerTest.java#L380) will cause the original implementation to fail when attempting to deserialise the second geo point.

Fortunately there's a getPosition() method which can be used to store the position and changing the code to use this instead of rewinding back to zero resolves the issue.

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
@@ -554,14 +554,14 @@ public class Geoshape {
             long l = VariableLong.readPositive(buffer);
             assert l>0 && l<Integer.MAX_VALUE;
             int length = (int)l;
+            int position = ((ReadArrayBuffer) buffer).getPosition();
             InputStream inputStream = new ByteArrayInputStream(buffer.getBytes(length));
             try {
                 return GeoshapeBinarySerializer.read(inputStream);
             } catch (IOException e) {
                 // retry using legacy point deserialization
                 try {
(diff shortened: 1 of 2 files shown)


janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java
```diff
@@ -554,14 +554,14 @@ public Geoshape read(ScanBuffer buffer) {
             long l = VariableLong.readPositive(buffer);
             assert l>0 && l<Integer.MAX_VALUE;
             int length = (int)l;
+            int position = ((ReadArrayBuffer) buffer).getPosition();
```
dpitera: AFAICT this variable isn't used anywhere other than on L#564; should it be moved into the inner scope block?

sjudeng: We need the position before it's advanced in the buffer operation on the next line.

PR #464, opened 2017-08-17
Generate JaCoCo test coverage reports and upload to Codecov from Travis CI

### LINKED ISSUE #21: Add code coverage
We should use a service such as [Coveralls](https://coveralls.io/), [Codecov](https://codecov.io/), or [Circle CI](https://circleci.com/docs/code-coverage/)'s built-in support to track code coverage by tests.

Code coverage is an imperfect metric of test quality (necessary, but not sufficient), but it's an additional useful signal to keep in mind when assessing the quality of the overall test suite.

sjudeng:
Fixes #21. See https://codecov.io/gh/janusgraph/janusgraph/branch/master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,7 +99,15 @@ script:
       echo "Skipping script build step in Coverity branch/job";
     else
       travis_retry travis_wait 50 \
-          mvn clean verify --projects janusgraph-${MODULE} ${ARGS};
+          mvn clean verify --projects janusgraph-${MODULE} -Pcoverage ${ARGS};
+    fi
+
+after_success:
+  # Upload test coverage reports to Codecov
+  - if [ "${COVERITY_SCAN_BRANCH}" == 1 ] || [ -v COVERITY_ONLY ]; then
+      echo "Skipping test coverage report upload in Coverity branch/job";
+    else
+      bash <(curl -s https://codecov.io/bash);
     fi
 
 # Syntax and more info: https://docs.travis-ci.com/user/notifications
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,15 @@
+# Don't post a comment on pull requests.
+comment: off
+
+# Disable commit statuses
+coverage:
+  status:
+    project: no
+    patch: no
+    changes: no
+
+# Ignore test files
+ignore:
+  - "**/test/.*"
+  - "janusgraph-test/**"
+
--- a/janusgraph-cql/pom.xml
+++ b/janusgraph-cql/pom.xml
@@ -16,8 +16,11 @@
         <test.byteorderedpartitioner>byteorderedpartitioner</test.byteorderedpartitioner>
         <test.murmur>murmur</test.murmur>
         <test.murmur-ssl>murmur-ssl</test.murmur-ssl>
-        <test.jvm.opts>-Xms256m -Xmx1280m -ea -XX:+HeapDumpOnOutOfMemoryError -javaagent:${com.github.jbellis:jamm:jar}</test.jvm.opts>
+        <test.javaagent.opts>-javaagent:${com.github.jbellis:jamm:jar}</test.javaagent.opts>
+        <test.extra.jvm.opts>${test.javaagent.opts}</test.extra.jvm.opts>
+        <test.jvm.opts>-Xms256m -Xmx1280m -ea -XX:+HeapDumpOnOutOfMemoryError ${test.extra.jvm.opts}</test.jvm.opts>
         <test.excluded.groups>org.janusgraph.testcategory.MemoryTests,org.janusgraph.testcategory.PerformanceTests,org.janusgraph.testcategory.BrittleTests</test.excluded.groups>
+        <top.level.basedir>${basedir}/..</top.level.basedir>
     </properties>
 
     <dependencies>
@@ -316,6 +319,16 @@
         </plugins>
     </build>
     <profiles>
+        <profile>
+	          <id>coverage</id>
+	          <activation>
+	              <activeByDefault>false</activeByDefault>
+	          </activation>
+	          <properties>
+	              <test.extra.jvm.opts>${jacoco.opts} ${test.javaagent.opts}</test.extra.jvm.opts>
+	          </properties>
+        </profile>
+
         <profile>
             <!-- Run Thrift tests with ScyllaDB (requires Docker) -->
             <id>scylladb-test</id>
--- a/janusgraph-hbase-parent/janusgraph-hbase-098/pom.xml
+++ b/janusgraph-hbase-parent/janusgraph-hbase-098/pom.xml
@@ -13,7 +13,7 @@
     <properties>
         <jruby.version>1.6.8</jruby.version>
         <zookeeper.version>3.4.6</zookeeper.version>
-        <top.level.basedir>${project.parent.parent.basedir}</top.level.basedir>
+        <top.level.basedir>${basedir}/../..</top.level.basedir>
     </properties>
     
     <dependencies>
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/JanusGraphIoRegistryTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/JanusGraphIoRegistryTest.java
@@ -60,7 +60,7 @@ public class JanusGraphIoRegistryTest {
         assertEquals(15,
                 Stream.of(P.class.getDeclaredMethods()).filter(m -> Modifier.isStatic(m.getModifiers())).filter(p -> {
                     log.debug("Predicate: {}", p);
-                    return true;
+                    return !p.isSynthetic();
                 }).count());
 
         Graph graph = EmptyGraph.instance();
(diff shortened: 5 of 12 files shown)


sjudeng:
There was an additional CI check for Codecov which failed because there was no report on master to compare PR to. This is expected. I'll look at disabling the additional check and PR comment for now as we're probably not ready to start enforcing anything on test coverage just yet. I'll keep the above report comment for background.

codecov.yml
```diff
@@ -0,0 +1,15 @@
+# Don't post a comment on pull requests.
+comment: off
+
+# Disable commit statuses
+coverage:
+    status:
```
mbrukman: 2-space indent in this file? Seems to be 2-space indent below.

janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/JanusGraphIoRegistryTest.java
```diff
@@ -60,7 +60,7 @@ public void testTinkerPopPredicatesAsGryo() throws SerializationException {
         assertEquals(15,
                 Stream.of(P.class.getDeclaredMethods()).filter(m -> Modifier.isStatic(m.getModifiers())).filter(p -> {
                     log.debug("Predicate: {}", p);
-                    return true;
+                    return !p.isSynthetic();
```
mbrukman: Intentional change? It's unclear how this follows from the rest of the PR, and the comment seems ominous. I'm not sure how to interpret this. I'll add some more folks to review this and `pom.xml` changes as well.

sjudeng: JaCoCo instrumentation adds a synthetic method, `$jacocoInit()`, that gets picked up under reflection. See [JaCoCo FAQ](http://www.jacoco.org/jacoco/trunk/doc/faq.html) ("My code uses reflection. Why does it fail when I execute it with JaCoCo?").

sjudeng: @robertdale Since you added this test (and comment) does the update here seem okay?

robertdale: @sjudeng It's good. Probably even better this way.

.travis.yml
```diff
@@ -99,7 +99,15 @@ script:
       echo "Skipping script build step in Coverity branch/job";
     else
       travis_retry travis_wait 50 \
-          mvn clean verify --projects janusgraph-${MODULE} ${ARGS};
+          mvn clean verify --projects janusgraph-${MODULE} -Pcoverage ${ARGS};
+    fi
+
+after_success:
```
mbrukman: Are we going to have the same issue here as with Coverity, in that it will upload to Codecov for each of the individual jobs, i.e., a dozen times per build? Is it worth running this manually once in a while the same way as we're doing with Coverity, or do we want it to run automatically to get the report for every PR (to get incremental change in code coverage), or at least have one less thing to think about?

Will Codecov start rate-limiting the submissions?

sjudeng: @mbrukman The code coverage analysis happens during testing (through JaCoCo instrumentation) and all that's uploaded is the final report, which will just be for the given module in the Travis job. Codecov handles the aggregation of the reports.

https://github.com/codecov/support/wiki/Merging-Reports

sjudeng:
@pluradj, @amcp Can one of you take a look at the pom changes here? Once this gets merged we'll have visibility into test code coverage for newly submitted PRs, which will be especially helpful when reviewing larger PRs.

.travis.yml
```diff
@@ -99,7 +99,15 @@ script:
...
+  # Upload test coverage reports to Codecov
+  - if [ "${COVERITY_SCAN_BRANCH}" == 1 ] || [ -v COVERITY_ONLY ]; then
+      echo "Skipping test coverage report upload in Coverity branch/job";
+    else
+      bash <(curl -s https://codecov.io/bash);
```
hsaputra: Never use codecov before but is this the intended way (i.e: use curl) to upload test coverage to Codecov ?

mbrukman: Yes, see https://docs.codecov.io/v4.3.6/docs/about-the-codecov-bash-uploader -- that command is literally on that web page.

hsaputra: Ah, interesting. Thx!

janusgraph-hbase-parent/janusgraph-hbase-098/pom.xml
```diff
@@ -13,7 +13,7 @@
     <properties>
         <jruby.version>1.6.8</jruby.version>
         <zookeeper.version>3.4.6</zookeeper.version>
-        <top.level.basedir>${project.parent.parent.basedir}</top.level.basedir>
+        <top.level.basedir>${basedir}/../..</top.level.basedir>
```
pluradj: Why is this change needed? This replacement pattern is possible in other `pom.xml` files that weren't updated with this PR. It seems like something that would be a consistent change across the repo: `janusgraph-hadoop-parent/pom.xml`, `janusgraph-hbase-parent/janusgraph-hbase/pom.xml`, etc.

janusgraph-cql/pom.xml
```diff
@@ -16,8 +16,10 @@
         <test.byteorderedpartitioner>byteorderedpartitioner</test.byteorderedpartitioner>
         <test.murmur>murmur</test.murmur>
         <test.murmur-ssl>murmur-ssl</test.murmur-ssl>
-        <test.jvm.opts>-Xms256m -Xmx1280m -ea -XX:+HeapDumpOnOutOfMemoryError -javaagent:${com.github.jbellis:jamm:jar}</test.jvm.opts>
+        <test.extra.jvm.opts>-javaagent:${com.github.jbellis:jamm:jar}</test.extra.jvm.opts>
```
pluradj: How about instead:
```
<test.javaagent.opts>-javaagent:${com.github.jbellis:jamm:jar}</test.javaagent.opts>
<test.extra.jvm.opts>${test.javaagent.opts}</test.extra.jvm.opts>
```
to avoid repeating later on when overriding:
```
<test.extra.jvm.opts>${jacoco.opts} ${test.javaagent.opts}</test.extra.jvm.opts>
```

janusgraph-hbase-parent/janusgraph-hbase-098/pom.xml
sjudeng: @pluradj In Travis `${project.parent.parent.basedir}` didn't resolve properly but `${basedir}/../../` did. Originally I only updated in those modules with unit tests since jacoco reports were writing to the wrong spot but I just pushed updates to to use `${basedir}/../..` everywhere consistently.

hsaputra: This should work

hsaputra: @pluradj Are you ok with this update/change?

hsaputra:
I am +1 for the changes. But since @pluradj has concern I will let him finalize the review b4 merging.

janusgraph-hbase-parent/janusgraph-hbase-098/pom.xml
pluradj: LGTM +1

sjudeng:
@hsaputra, @pluradj, @mbrukman - Thanks for your reviews.

.travis.yml
```diff
@@ -99,7 +99,15 @@ script:
       echo "Skipping script build step in Coverity branch/job";
     else
       travis_retry travis_wait 50 \
-          mvn clean verify --projects janusgraph-${MODULE} ${ARGS};
+          mvn clean verify --projects janusgraph-${MODULE} -Pcoverage ${ARGS};
+    fi
+
+after_success:
+  # Upload test coverage reports to Codecov
+  - if [ "${COVERITY_SCAN_BRANCH}" == 1 ] || [ -v COVERITY_ONLY ]; then
```
mbrukman: @sjudeng — I think I must have missed this line in the original review of this PR (and this file hasn't changed since then), but what's the `-v` test do? I can't find it in the [`bash(1)` man page](http://man7.org/linux/man-pages/man1/bash.1.html#CONDITIONAL_EXPRESSIONS).

Or did you perhaps mean to use `[ -z "${COVERITY_ONLY:-}" ]` as is done elsewhere in this file?

Of course, it's typically used in the negated form: `! [ -z "${COVERITY_ONLY:-}" ]`; the single-character equivalent to that is `[ -n "${COVERITY_ONLY:-}" ]` if that's what you're looking for.

sjudeng: @mbrukman I might have missed this when you provided original feedback. I agree, the check here should be something like the below. 

```
  - if [ "${TRAVIS_BRANCH}" == "${COVERITY_BRANCH_NAME}" ] || ! [ -z "${COVERITY_ONLY:-}" ]; then
      echo "Skipping test coverage report upload in Coverity branch/job";
    else
      bash <(curl -s https://codecov.io/bash);
    fi
```

I can get a fix in but it might be a bit or else feel free to beat me to it. Currently reports are uploading to codecov properly. It's just attempting (and failing) to upload non-existent reports on regular module jobs on the coverity_scan branch (see https://travis-ci.org/JanusGraph/janusgraph/jobs/279705005#L488).

mbrukman: Sounds good; PR #761 sent with the fix.

Issue #52, opened 2017-01-19
Fix npe vunerabilities


amcp:
Fortify - High - can crash the program
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java#L347
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/EdgeSerializer.java#L165
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/EdgeSerializer.java#L173
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/core/attribute/Geoshape.java#L444
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexHelper.java#L58
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/QueryUtil.java#L299
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java#L675

Issue #50, opened 2017-01-19
Fix log forging vunerabilities


amcp:
Fortify - validate input from main or dont use input from main in log statement

- [x] [ConfigurationLint.java](https://github.com/JanusGraph/janusgraph/blob/84ead7d/janusgraph-core/src/main/java/org/janusgraph/util/system/ConfigurationLint.java#L46) - Fixed in #469 (verify) 
- [x] [GraphDatabaseConfiguration.java](https://github.com/JanusGraph/janusgraph/blob/84ead7d/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java#L1374)
- [x] [GraphDatabaseConfiguration.java](https://github.com/JanusGraph/janusgraph/blob/84ead7d/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java#L1377)
- [x] [JanusGraphFactory](https://github.com/JanusGraph/janusgraph/blob/84ead7d/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphFactory.java#L298) - Fixed in #469 (verify) 
- [x] [JanusGraphFactory](https://github.com/JanusGraph/janusgraph/blob/84ead7d/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphFactory.java#L301) - Fixed in #469 (verify)

PR #469, opened 2017-08-22
Resolve numerous high/medium Coverity findings

sjudeng:
Also fixes some of the Fortify findings in #50 and #52 but not all the findings in those issues are fixed so the issues should remain open.

### DIFF
--- a/janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxStoreManager.java
+++ b/janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxStoreManager.java
@@ -437,6 +437,10 @@ public class AstyanaxStoreManager extends AbstractCassandraStoreManager {
             if (this.storageConfig.get(GraphDatabaseConfiguration.DROP_ON_CLEAR)) {
                 ks.dropKeyspace();
             } else {
+                final KeyspaceDefinition keyspaceDefinition = cluster.describeKeyspace(keySpaceName);
+                if (keyspaceDefinition == null) {
+                    return;
+                }
                 for (final ColumnFamilyDefinition cf : cluster.describeKeyspace(keySpaceName).getColumnFamilyList()) {
                     ks.truncateColumnFamily(new ColumnFamily<Object, Object>(cf.getName(), null, null));
                 }
--- a/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphFactory.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphFactory.java
@@ -28,6 +28,7 @@ import org.janusgraph.diskstorage.configuration.backend.CommonsConfiguration;
 import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
 
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.*;
+import static org.janusgraph.util.system.LoggerUtil.sanitizeAndLaunder;
 
 import org.janusgraph.graphdb.database.StandardJanusGraph;
 
@@ -313,10 +314,10 @@ public class JanusGraphFactory {
         File storedir = new File(file);
         if (!storedir.isAbsolute()) {
             String newFile = configParent.getAbsolutePath() + File.separator + file;
-            log.debug("Overwrote relative path: was {}, now {}", file, newFile);
+            log.debug("Overwrote relative path: was {}, now {}", sanitizeAndLaunder(file), sanitizeAndLaunder(newFile));
             return newFile;
         } else {
-            log.debug("Loaded absolute path for key: {}", file);
+            log.debug("Loaded absolute path for key: {}", sanitizeAndLaunder(file));
             return file;
         }
     }
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexFeatures.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexFeatures.java
@@ -43,7 +43,7 @@ public class IndexFeatures {
                          String wildcardField, ImmutableSet<Cardinality> supportedCardinaities, boolean supportsNanoseconds,
                          boolean supportCustomAnalyzer, boolean supportsGeoContains) {
 
-        Preconditions.checkArgument(defaultMap!=null || defaultMap!=Mapping.DEFAULT);
+        Preconditions.checkArgument(defaultMap!=null && defaultMap!=Mapping.DEFAULT);
         Preconditions.checkArgument(supportedMap!=null && !supportedMap.isEmpty()
                                     && supportedMap.contains(defaultMap));
         this.supportsDocumentTTL = supportsDocumentTTL;
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
@@ -669,6 +669,7 @@ public class StandardJanusGraph extends JanusGraphBlueprintsGraph {
             if (logTransaction) {
                 //[FAILURE] Inability to log transaction fails the transaction by escalation since it's likely due to unavailability of primary
                 //storage backend.
+                Preconditions.checkState(txLog != null, "Transaction log is null");
                 txLog.add(txLogHeader.serializeModifications(serializer, LogTxStatus.PRECOMMIT, tx, addedRelations, deletedRelations),txLogHeader.getLogKey());
             }
 
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/FulgoraGraphComputer.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/FulgoraGraphComputer.java
@@ -366,6 +366,9 @@ public class FulgoraGraphComputer implements JanusGraphComputer {
             try {
                 for (Map.Entry<Long, Map<String, Object>> vprop : properties) {
                     Vertex v = tx.getVertex(vprop.getKey());
+                    if (v == null) {
+                        continue;
+                    }
                     for (Map.Entry<String, Object> prop : vprop.getValue().entrySet()) {
                         v.property(VertexProperty.Cardinality.single, prop.getKey(), prop.getValue());
                     }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
@@ -15,6 +15,7 @@
 package org.janusgraph.graphdb.relations;
 
 import com.carrotsearch.hppc.cursors.LongObjectCursor;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 import org.janusgraph.core.PropertyKey;
@@ -46,7 +47,9 @@ public class CacheEdge extends AbstractEdge {
     }
 
     public Direction getVertexCentricDirection() {
-        return data.getCache().direction;
+        final RelationCache cache = data.getCache();
+        Preconditions.checkState(cache != null, "Cache is null");
+        return cache.direction;
     }
 
     //############## Similar code as CacheProperty but be careful when copying #############################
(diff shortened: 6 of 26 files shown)


janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxStoreManager.java
```diff
@@ -433,8 +433,11 @@ public void clearStorage() throws BackendException {
             if (ks == null)
                 return;
 
-            for (ColumnFamilyDefinition cf : cluster.describeKeyspace(keySpaceName).getColumnFamilyList()) {
-                ks.truncateColumnFamily(new ColumnFamily<Object, Object>(cf.getName(), null, null));
+            final KeyspaceDefinition keyspaceDefinition = cluster.describeKeyspace(keySpaceName);
+            if (keyspaceDefinition != null) {
```
amcp: if (keyspaceDefinition == null) {
    return;
}

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/astyanax/AstyanaxStoreManager.java
```diff
@@ -433,8 +433,11 @@ public void clearStorage() throws BackendException {
...
+                for (ColumnFamilyDefinition cf : keyspaceDefinition.getColumnFamilyList()) {
```
amcp: un-nest from if

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
```diff
@@ -669,6 +669,7 @@ public void commit(final Collection<InternalRelation> addedRelations,
             if (logTransaction) {
                 //[FAILURE] Inability to log transaction fails the transaction by escalation since it's likely due to unavailability of primary
                 //storage backend.
+                assert txLog != null;
```
amcp: why assert and not Preconditions.checkState? recall we/I was burned a long time ago regarding silent asserts in custom vertex id logic.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/FulgoraGraphComputer.java
```diff
@@ -366,8 +366,10 @@ public void run() {
             try {
                 for (Map.Entry<Long, Map<String, Object>> vprop : properties) {
                     Vertex v = tx.getVertex(vprop.getKey());
-                    for (Map.Entry<String, Object> prop : vprop.getValue().entrySet()) {
-                        v.property(VertexProperty.Cardinality.single, prop.getKey(), prop.getValue());
+                    if (v != null) {
```
amcp: if v == null continue
and then un-nest the below.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
```diff
@@ -46,7 +46,9 @@ public CacheEdge(long id, EdgeLabel label, InternalVertex start, InternalVertex
     }
 
     public Direction getVertexCentricDirection() {
-        return data.getCache().direction;
+        final RelationCache cache = data.getCache();
+        assert cache != null;
```
amcp: Preconditions.checkState(cache != null) ?

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/MapReduceIndexManagement.java
```diff
@@ -163,10 +164,9 @@ public MapReduceIndexManagement(JanusGraph g) {
 
         // The index name and relation type name (if the latter is applicable)
         final String indexName = index.name();
-        final String relationTypeName =
-                RelationTypeIndex.class.isAssignableFrom(index.getClass()) ?
-                ((RelationTypeIndex)index).getType().name() :
-                "";
+        final RelationType relationType =
+            RelationTypeIndex.class.isAssignableFrom(index.getClass()) ? ((RelationTypeIndex) index).getType() : null;
+        final String relationTypeName = relationType == null ? "" : relationType.name();
```
amcp: externalize the empty string?

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/config/job/AbstractDistCacheConfigurer.java
```diff
@@ -146,7 +146,7 @@ private static Conf configureByClasspath(String mapredJarFilename) {
         for (String cpentry : classpath.split(File.pathSeparator)) {
             if (cpentry.toLowerCase().endsWith(".jar") || cpentry.toLowerCase().endsWith(".properties")) {
                 paths.add(new Path(cpentry));
-                if (cpentry.toLowerCase().endsWith(mrj));
+                if (cpentry.toLowerCase().endsWith(mrj))
```
amcp: use braces

amcp:
@sjudeng also can you update the linked issues and convert the lists to checkboxes so we can track what is done and what is not done yet?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexFeatures.java
```diff
@@ -43,7 +43,7 @@ public IndexFeatures(boolean supportsDocumentTTL, Mapping defaultMap, ImmutableS
                          String wildcardField, ImmutableSet<Cardinality> supportedCardinaities, boolean supportsNanoseconds,
                          boolean supportCustomAnalyzer, boolean supportsGeoContains) {
 
-        Preconditions.checkArgument(defaultMap!=null || defaultMap!=Mapping.DEFAULT);
```
mbrukman: Why are we dropping the second check here? Was this meant to be `&&` instead of `||`, perhaps?

janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/config/job/AbstractDistCacheConfigurer.java
mbrukman: This is a behavior change — while it's most likely correct and the `;` was a typo, it will now do assignments conditionally, whereas before, it would be unconditional.

Does this have potential to break some MapReduce workflows as a result? Should it be documented in release notes or otherwise that this is, while a correct change, is a potential issue to watch out for when upgrading from Titan to JanusGraph or from previous JanusGraph release to later ones?

sjudeng: @mbrukman Good point. But we might be okay. It definitely looks like a bug but more importantly it doesn't look like the actual JAR selected by this routine is relevant. The only usage originates from [Hadoop2Compat](https://github.com/JanusGraph/janusgraph/blob/bfa3c970555932d4983155168322f72297aaee85/janusgraph-hadoop-parent/janusgraph-hadoop-2/src/main/java/org/janusgraph/hadoop/compat/h2/Hadoop2Compat.java#L78-L84) where there's a comment that the actual JAR name (which corresponds to `mapredJarFilename` or `mrj` here) doesn't matter. I think if it did we would be seeing issues in Hadoop jobs since the incorrect job JAR is being used.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexFeatures.java
sjudeng: Thanks. I guess `defaultMap==Mapping.DEFAULT` wouldn't make sense. Incidentally I don't see any usages of the defaultMap and (accessor is not covered according to [coverity](https://codecov.io/gh/JanusGraph/janusgraph/src/7608d3325fe5d531647c6c9677fca54701ecdb93/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexFeatures.java#L63...65)).

sjudeng:
@amcp, @mbrukman - Updates are pushed. Thanks for the feedback.

@amcp I also updated the issues with current status. If you get a chance after this is merged it would be great if you could re-scan. Some of these were not indicated under Coverity.

sjudeng:
@amcp, @mbrukman Anything else on this? Unless there are objections I'll plan to merge this in 48 hours.

janusgraph-core/src/main/java/org/janusgraph/util/system/ConfigurationLint.java
```diff
@@ -40,7 +41,7 @@ public static void main(String args[]) throws IOException {
             System.exit(1);
         }
 
-        log.info("Checking " + args[0]);
+        log.info("Checking configuration file");
```
mbrukman: Would it make sense to still print out the value of `argv[0]` here so that it's clear from the logs which file path is being validated? Otherwise, this log line might not be useful in isolation, when the user doesn't know the full path to the file that's being checked.

sjudeng: @mbrukman I updated to instead sanitize the input (and make a deep copy of the characters to clear the Fortify finding).

janusgraph-test/src/test/java/org/janusgraph/util/StringEncodingTest.java
```diff
@@ -45,10 +46,15 @@ public void testAsciiStringEncoding() {
             try {
                 StringEncoding.getAsciiByteLength(s);
                 fail();
-            } catch (IllegalArgumentException e) {
-
-            } catch (NullPointerException e) {
-
+            } catch (IllegalArgumentException | NullPointerException e) { }
+
+            if (s != null) {
```
mbrukman: I understand that you're cleaning up Coverity findings here, but there's a cleaner way of writing tests that expect to fail, with a particular exception, e.g.,

* https://stackoverflow.com/a/156528
* https://stackoverflow.com/a/41031738
* more details on both: https://blog.goyello.com/2015/10/01/different-ways-of-testing-exceptions-in-java-and-junit/

My recommendation here would be to factor out the tests that are expected to fail into separate test methods, one for each invalid input (since each one fails in a different way). Since this is technically a Coverity cleanup PR, and not a test cleanup PR, it's up to you whether you want to fix it now, or add a TODO comment / file an issue to clean it up as a follow-up change.

sjudeng: @mbrukman Pushed fix to clean this up

janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphFactory.java
```diff
@@ -313,10 +314,10 @@ private static final String getAbsolutePath(final File configParent, String file
         File storedir = new File(file);
         if (!storedir.isAbsolute()) {
             String newFile = configParent.getAbsolutePath() + File.separator + file;
-            log.debug("Overwrote relative path: was {}, now {}", file, newFile);
+            log.debug("Overwrote relative path: was {}, now {}", sanitizeAndLaunder(file), sanitizeAndLaunder(newFile));
```
vin0010: Hi @sjudeng , 
I am trying to solve same problem in my project.
As fortify says "An attacker could take advantage of this behavior to forge log entries or inject malicious content into the log." and by looking at you code how com parsing each char of your string and build the string is going to help us prevent injecting? 
Or is this method is just to suppress fortify scan?

sjudeng: Hi @vin0010. As you note there are two issues, one is mitigating the risk of log forging and the other is resolving the Fortify finding. Currently the sanitize implementation is minimal and just [replaces newline characters](https://github.com/JanusGraph/janusgraph/blob/d12adfbf083f575fa48860daa37bfbd0e6095369/janusgraph-core/src/main/java/org/janusgraph/util/system/LoggerUtil.java#L27), though this could definitely be extended to be more robust. The second part of the implementation [recreates the string](https://github.com/JanusGraph/janusgraph/blob/d12adfbf083f575fa48860daa37bfbd0e6095369/janusgraph-core/src/main/java/org/janusgraph/util/encoding/StringEncoding.java#L73-L78), which was still needed to clear the Fortify finding in this case even after sanitizing.

Issue #429, opened 2017-07-28
Fix JanusGraphIndexTest to use transactionlistener instead of Thread.sleep


holleyism:
The use of Thread.sleep() in the unit tests is to give the index and store time to commit the changes since tx().commit() return does not mean the commit has completed.  As documented in response to #329, where applicable, the use of Thread.sleep() in the unit tests should be replaced with tx().addTransactionListener().  Also switch other tests as appropriate.

mbrukman:
Closing issue per [this comment](https://github.com/JanusGraph/janusgraph/pull/423#issuecomment-317487496) which says the PR fixes this issue.

sjudeng:
@mbrukman I think this issue was created to address feedback during review of #423 (see [this comment](https://github.com/JanusGraph/janusgraph/pull/423/files#r130005253)) and isn't yet resolved.

holleyism:
@mbrukman Can you please reopen this as it is not fixed.  Thanks.

mbrukman:
Apologies for the misunderstanding.

PR #473, opened 2017-08-25
Issue 329 - Added commits for Solr

holleyism:
Signed-off-by: Adam Holley <holleya@us.ibm.com>

### DIFF
--- a/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -1035,8 +1035,6 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
 
         tx.commit();
 
-        Thread.sleep(5000);
-
         List<JanusGraphVertex> vertices = new ArrayList<JanusGraphVertex>();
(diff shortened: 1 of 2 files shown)


holleyism:
Isn't this part of the commit contract?  Even in production I would think I should be able to know that a vertex has been committed and is available for query. If not then we should probably document that commit when using Solr is not deterministic.

sjudeng:
For performance some indexing/storage backends only guarantee eventual consistency as part of their distributed storage model. Do you know if waits are included in other indexing/storage backends?

robertdale:
There is already an option to wait for the index. See `wait-searcher` -   http://docs.janusgraph.org/latest/config-ref.html#_index_x_solr  

I know that elasticsearch also does not immediately commit or make visible.  It has an index refresh interval.

holleyism:
Ok. I'll remove this and see if I can update the unit test config to still avoid using sleep.

sjudeng:
Does it work to set `wait-searcher` in the test configs, or is the option not working? If not then it seems like your updates are the correct behavior if the option is set (good call @robertdale).

holleyism:
Going to look into this more as it appears the unit tests are already setting wait-searcher to true, and SolrIndex sets the commit to wait if true.

holleyism:
This PR should also close #429 as it's not necessary.

holleyism:
@sjudeng Can you please review again?  Removed all changes except those for unit-tests.  Thanks.

sjudeng:
I'm going to rerun the tests in Travis a few times. I want to make sure there's not a sporadic timing issue.

PR #467, opened 2017-08-21
Added a script for running a Coverity scan.

mbrukman:
This script will help address issue https://github.com/JanusGraph/janusgraph/issues/466 by running Coverity scans directly via our own script rather than the Travis CI plugin.

PR #479, opened 2017-08-27
Use coverity-scan script in Travis

### LINKED ISSUE #466: Use Coverity runner script instead of plugin in Travis CI
Travis configuration (see [.travis.yml](https://github.com/JanusGraph/janusgraph/blob/bab3937f601ec58636c8bde05b3e560283c577fd/.travis.yml)) attempts to run Coverity scans as part of the relevant job on the coverity_scan branch. But the current approach does not work.

* The Coverity plugin runs (even if the scan/upload doesn't) on all builds, not just the `COVERITY_ONLY` build (see [jobs/762.3](https://travis-ci.org/JanusGraph/janusgraph/jobs/266341893)). I don't think it counts against the project Coverity quota if scan/upload doesn't occur, but still. 
* The scan fails with a "syntax error near unexpected token 'then'" error (see [jobs/762.16](https://travis-ci.org/JanusGraph/janusgraph/jobs/266341906)) suggesting the approach of conditioning the scan using an if-then in `build_command_prepend` (see #300) doesn't work
* Even when temporarily forcing a scan (see https://github.com/JanusGraph/janusgraph/commit/026ab1f38a1f6059757f1e65bfbf4ff475ac9945) it fails during janusgraph-core compilation because of "no output". See [jobs/763](https://travis-ci.org/JanusGraph/janusgraph/builds/266349546). Normally `travis_wait` can be used but this may not be possible with the coverity_scan plugin (see @mbrukman's comment [here](https://github.com/JanusGraph/janusgraph/pull/300#issuecomment-306079362)).

@mbrukman As discussed in #300 it looks like your runner script is the only hope left on this. Can you add it into the project and get it running under the "if coverity branch and job" conditional in the .travis.yml script definition? You should then be able to wrap the call with `travis_wait` though of course we're out of luck if the scan takes longer than 50 minutes. Alternatively if you want to add the script I can look at getting it wired into .travis.yml.

sjudeng:
Fixes #466 using the the script added by @mbrukman in #467 instead of the plugin to run Coverity scans in Travis. Successfully tested by updating `COVERITY_BRANCH_NAME` in .travis.yml to point to a test branch and then pushing to same (see [build #805](https://travis-ci.org/JanusGraph/janusgraph/builds/268934336)). Travis jobs for all individual modules were skipped and the `COVERITY_ONLY` job ran the scan and uploaded results to Coverity (see [job #805.16](https://travis-ci.org/JanusGraph/janusgraph/jobs/268934352)). Some notes are below.

* As noted in #467, Coverity scan performance appears to be significantly better (e.g. 95% versus 68%) when run under Docker (specifically `centos:7`). As a result these updates include enabling Docker (`sudo: true`, `servies: docker`) and running the coverity-scan script within a Docker container.
* Coverity requires an email to download scan tool and upload results. @mbrukman can you add janusgraph-ci@googlegroups.com as a Coverity user or do you have other ideas how to handle this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 dist: trusty
 language: java
-sudo: false
-
+sudo: true
+services:
+  - docker
 jdk:
   - openjdk8
 
@@ -14,6 +15,8 @@ env:
     # This is the encrypted COVERITY_SCAN_TOKEN, created via the
     # `travis encrypt` command using the project repo's public key.
     - secure: "v5ixqTeb74y0vRuPcDbe3C28GDDYvqyEXA2dt+9UVU6GG7WpnmpkBf05gI1dIhp51lBhwx9WSlFBtzho+KdCBmNY/CzBRhVHe/lCQYK9Hb6uGPvuwBvC0WjJgJXsVrLFjppeRhcf+OAweVQ3uw2RPMDRvKIVMUcO1BTFjjJl6REJXNUdzGS57MtH2mmRyOEz250EwgqUELZvcOytG7fNrjMJKVK2nSsoxi0BqZIpItTWPWWeQ1wi1FplJ18A2qtD+MPfAGNSB+/a+r0Av+VCT2eGl06ZyZAzP3q/vG5IYjQ3AJsSPqcZUt4ms+2us1+kwuzXIILjzZmcfImu29+y/thndU5E5b2v+nZ4H69CUCc5OmKW2RwozLNmBIUhO0n+35va/J7FiPIqm3pwxCz5vWA3YTHDADxnIYe7+9uY/+dOK/AvP5fyu7u07vuF3liKNBdrX7ylP3kYc7FXGmYl8wCZv31iy1yTtndQ9qKef7bo8lM9Cdh39KyowrygH+Um7pr9gqf2S9jn99nQ3bib32fBWgBkLpJRwhZYHPUupZjZfgu/9woby0DuriuHZKMqZd7QUawYz6wXGlhzu78x5Tohlj1pGBwHYdcJ/Tm3PiEpyH4aYQLffkjGHJAcCW5tO8QbB0qrLYWC8xVMWuFz1TpSBRXOqVYdBfIa2UZDtOU="
+    - COVERITY_BRANCH_NAME="coverity_scan"
+    - COVERITY_EMAIL="sjudeng@gmail.com"
     # Default Elasticsearch heap size can be too large for Travis
     - ES_JAVA_OPTS="-Xms256m -Xmx512m"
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,6 +15,8 @@ env:
     # This is the encrypted COVERITY_SCAN_TOKEN, created via the
     # `travis encrypt` command using the project repo's public key.
     - secure: "v5ixqTeb74y0vRuPcDbe3C28GDDYvqyEXA2dt+9UVU6GG7WpnmpkBf05gI1dIhp51lBhwx9WSlFBtzho+KdCBmNY/CzBRhVHe/lCQYK9Hb6uGPvuwBvC0WjJgJXsVrLFjppeRhcf+OAweVQ3uw2RPMDRvKIVMUcO1BTFjjJl6REJXNUdzGS57MtH2mmRyOEz250EwgqUELZvcOytG7fNrjMJKVK2nSsoxi0BqZIpItTWPWWeQ1wi1FplJ18A2qtD+MPfAGNSB+/a+r0Av+VCT2eGl06ZyZAzP3q/vG5IYjQ3AJsSPqcZUt4ms+2us1+kwuzXIILjzZmcfImu29+y/thndU5E5b2v+nZ4H69CUCc5OmKW2RwozLNmBIUhO0n+35va/J7FiPIqm3pwxCz5vWA3YTHDADxnIYe7+9uY/+dOK/AvP5fyu7u07vuF3liKNBdrX7ylP3kYc7FXGmYl8wCZv31iy1yTtndQ9qKef7bo8lM9Cdh39KyowrygH+Um7pr9gqf2S9jn99nQ3bib32fBWgBkLpJRwhZYHPUupZjZfgu/9woby0DuriuHZKMqZd7QUawYz6wXGlhzu78x5Tohlj1pGBwHYdcJ/Tm3PiEpyH4aYQLffkjGHJAcCW5tO8QbB0qrLYWC8xVMWuFz1TpSBRXOqVYdBfIa2UZDtOU="
+    - COVERITY_BRANCH_NAME="coverity_scan"
+    - COVERITY_EMAIL="sjudeng@gmail.com"
```
mbrukman: How about using janusgraph-ci@googlegroups.com for this instead?

.travis.yml
```diff
@@ -66,38 +69,35 @@ matrix:
...
-          exit 0;
-        fi
-    build_command: mvn clean package -DskipTests=true --batch-mode
-    branch_pattern: coverity_scan
-
 install:
-  # Build and install current module and dependencies. Handle output timeouts and retry on error.
-  - if [ "${COVERITY_SCAN_BRANCH}" == 1 ] || [ -v COVERITY_ONLY ]; then
-      echo "Skipping install build step in Coverity branch/job";
+  - if [ "${TRAVIS_BRANCH}" == "${COVERITY_BRANCH_NAME}" ] && [ -v COVERITY_ONLY ]; then
```
mbrukman: Will this run twice, once on each Linux and macOS? Looks like `-v` is not supported in the `bash` I have on macOS:

```bash
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)
Copyright (C) 2007 Free Software Foundation, Inc.
```

The `-v` flag is not listed in the man page, and using it yields the error:

```
[: -v: unary operator expected
```

In any case, we don't need this to run on both platforms for Coverity; would just running it on Linux suffice?

.travis.yml
```diff
@@ -66,38 +69,35 @@ matrix:
...
+      echo "Building all modules for Coverity analysis";
+      travis_retry travis_wait \
+          mvn install -DskipTests=true -Dmaven.javadoc.skip=true --batch-mode --show-version;
```
mbrukman: If the modules are pre-built before Coverity even runs, will it be able to analyze the results? I ask because I've always built via the Coverity tool, not via Maven directly, and when I already had an existing build, Coverity was having issues building/analyzing from what I saw in the output.

analysis/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM centos:7
```
mbrukman: Please add standard copyright header here. You can use a tool I wrote for this: https://github.com/mbrukman/autogen :-)

.travis.yml
mbrukman: Sorry, missed the note about this in the PR description; let me see what we can do here.

.travis.yml
sjudeng: I'll switch to `-z ${COVERITY_ONLY:-}` instead. Build matrix doesn't vary OS (e.g. only dist Ubuntu Trusty is used).

.travis.yml
sjudeng: Since you added the `mvn clean` to the script it's okay. This is done upfront here to take care of dependency downloads so that's not included (not sure if it would matter) in the build through `cov-build`.

analysis/Dockerfile
pluradj: Isn't this already using the standard copyright header?

sjudeng: @pluradj The copyright header wasn't there originally when the comment was added.

pluradj: ah ok. I would have expected the comment to be collapsed with `Show outdated`. thanks.

.travis.yml
mbrukman: Given that janusgraph-ci@ is a public list (i.e., anyone can receive email there), I am wary of signing it up to be a member or admin/owner of the Coverity project, because I'm concerned that having email receive access would allow someone to modify the project.

If Coverity isn't able to send output to an email address that's different from the user's account email (which seems very strange), we can only do one of the following:

* one of the individuals on the project is receiving email
* we create a synthetic email address that's only used for automatically forwarding email to a mailing list but that seems overkill

Let's keep it as-is for now.

.travis.yml
```diff
@@ -66,38 +69,35 @@ matrix:
...
-          exit 0;
-        fi
-    build_command: mvn clean package -DskipTests=true --batch-mode
-    branch_pattern: coverity_scan
-
 install:
-  # Build and install current module and dependencies. Handle output timeouts and retry on error.
-  - if [ "${COVERITY_SCAN_BRANCH}" == 1 ] || [ -v COVERITY_ONLY ]; then
-      echo "Skipping install build step in Coverity branch/job";
+  - if [ "${TRAVIS_BRANCH}" == "${COVERITY_BRANCH_NAME}" ] && ! [ -z ${COVERITY_ONLY:-} ]; then
```
mbrukman: I may be paranoid, but I typically encode variables in double-quotes, i.e., ` [ -z "${COVERITY_ONLY:-}" ]`, here and below.

.travis.yml
```diff
@@ -66,38 +69,35 @@ matrix:
...
+      echo "Building all modules for Coverity analysis";
+      travis_retry travis_wait \
+          mvn install -DskipTests=true -Dmaven.javadoc.skip=true --batch-mode --show-version;
+    elif [ "${TRAVIS_BRANCH}" == "${COVERITY_BRANCH_NAME}" ] || ! [ -z ${COVERITY_ONLY:-} ]; then
```
mbrukman: Same comment here about double-quotes surrounding variable expansion/default value.

.travis.yml
```diff
@@ -66,38 +69,35 @@ matrix:
...
       travis_retry travis_wait \
           mvn install --projects janusgraph-${MODULE} --also-make \
           -DskipTests=true -Dmaven.javadoc.skip=true --batch-mode --show-version;
     fi
 
 script:
-  # Build and test module. Handle output timeouts and retry with a clean build on error.
-  - if [ "${COVERITY_SCAN_BRANCH}" == 1 ] || [ -v COVERITY_ONLY ]; then
-      echo "Skipping script build step in Coverity branch/job";
+  - if [ "${TRAVIS_BRANCH}" == "${COVERITY_BRANCH_NAME}" ] && ! [ -z ${COVERITY_ONLY:-} ]; then
```
mbrukman: And here too.

.travis.yml
```diff
@@ -66,38 +69,35 @@ matrix:
...
+      echo "Building Docker image for Coverity analysis";
+      docker build -t janusgraph/analysis analysis;
+      echo "Running Coverity scan";
+      travis_wait 50 \
+        docker run --rm \
+            -v ${HOME}/.m2:/root/.m2 -v ${PWD}:/opt/janusgraph \
+            -e COVERITY_SCAN_TOKEN="${COVERITY_SCAN_TOKEN}" \
+            -e COVERITY_EMAIL="${COVERITY_EMAIL}" \
+            -i janusgraph/analysis;
+    elif [ "${TRAVIS_BRANCH}" == "${COVERITY_BRANCH_NAME}" ] || ! [ -z ${COVERITY_ONLY:-} ]; then
```
mbrukman: And here.

sjudeng:
@mbrukman Thanks for providing the script for this. Glad it worked in Travis (I think we would have been out of options otherwise).

mbrukman:
@pluradj -- any other comments on this PR or can we merge it?

Issue #480, opened 2017-08-28
Stream mixed index result instead of one big list of results.


davidclement90:
Actually, JanusGraph do not use streaming capacity of mixed index backend and there is a limit number of results that JanusGraph can get from ES and Solr.

Issue #242, opened 2017-04-28
Add eclipse / IDEA formatters to repo and apply them.


amcp:
I think it is a good idea to derive/craft IDEA / Eclipse formatters and add them to source control here so everyone can use them.

PR #481, opened 2017-08-28
Stream the result from Search Backend instead of use List.

davidclement90:
Issue :  https://github.com/JanusGraph/janusgraph/issues/480

I can not use cursors for Solr because sort clause must include the unique key field
(https://cwiki.apache.org/confluence/display/solr/Pagination+of+Results)

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/profile/QueryProfiler.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/profile/QueryProfiler.java
@@ -15,6 +15,7 @@
 package org.janusgraph.graphdb.query.profile;
 
 import org.janusgraph.graphdb.query.Query;
+import org.janusgraph.graphdb.query.graph.JointIndexQuery.Subquery;
 
 import java.util.Collection;
 import java.util.function.Function;
@@ -90,16 +91,16 @@ public interface QueryProfiler {
     }
 
     public static<Q extends Query,R extends Collection> R profile(String groupName, QueryProfiler profiler, Q query, boolean multiQuery, Function<Q,R> queryExecutor) {
-        QueryProfiler sub = profiler.addNested(groupName);
+        final QueryProfiler sub = profiler.addNested(groupName);
         sub.setAnnotation(QUERY_ANNOTATION, query);
         if (query.hasLimit()) sub.setAnnotation(LIMIT_ANNOTATION,query.getLimit());
         sub.startTimer();
-        R result = queryExecutor.apply(query);
+        final R result = queryExecutor.apply(query);
         sub.stopTimer();
         long resultSize = 0;
         if (multiQuery && profiler!=QueryProfiler.NO_OP) {
             //The result set is a collection of collections, but don't do this computation if profiling is disabled
-            for (Object r : result) {
+            for (final Object r : result) {
                 if (r instanceof Collection) resultSize+=((Collection)r).size();
                 else resultSize++;
             }
@@ -110,4 +111,11 @@ public interface QueryProfiler {
         return result;
     }
 
+    public static QueryProfiler startProfile(QueryProfiler profiler, Subquery query) {
+        final QueryProfiler sub = profiler.addNested("backend-query");
+        sub.setAnnotation(QUERY_ANNOTATION, query);
+        if (query.hasLimit()) sub.setAnnotation(LIMIT_ANNOTATION,query.getLimit());
+        sub.startTimer();
+        return sub;
+    }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -15,7 +15,6 @@
 package org.janusgraph.graphdb.transaction;
 
 import com.carrotsearch.hppc.LongArrayList;
-import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.cache.Cache;
@@ -69,6 +68,7 @@ import org.janusgraph.graphdb.types.vertices.EdgeLabelVertex;
 import org.janusgraph.graphdb.types.vertices.PropertyKeyVertex;
 import org.janusgraph.graphdb.types.vertices.JanusGraphSchemaVertex;
 import org.janusgraph.graphdb.util.IndexHelper;
+import org.janusgraph.graphdb.util.SubqueryIterator;
 import org.janusgraph.graphdb.util.VertexCentricEdgeIterable;
 import org.janusgraph.graphdb.vertices.CacheVertex;
 import org.janusgraph.graphdb.vertices.PreloadedVertex;
@@ -89,6 +89,8 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
@@ -1201,7 +1203,7 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
                     }
                     vertices = vertexSet.iterator();
                 } else {
-                    vertices = com.google.common.collect.Iterators.transform(newVertexIndexEntries.get(standardIndexKey.getValue(), standardIndexKey.getKey()).iterator(), new Function<JanusGraphVertexProperty, JanusGraphVertex>() {
+                    vertices = com.google.common.collect.Iterators.transform(newVertexIndexEntries.get(standardIndexKey.getValue(), standardIndexKey.getKey()).iterator(), new com.google.common.base.Function<JanusGraphVertexProperty, JanusGraphVertex>() {
                         @Nullable
                         @Override
                         public JanusGraphVertex apply(@Nullable JanusGraphVertexProperty o) {
@@ -1256,7 +1258,7 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
             Iterator<JanusGraphElement> iter;
             if (!indexQuery.isEmpty()) {
                 List<QueryUtil.IndexCall<Object>> retrievals = new ArrayList<QueryUtil.IndexCall<Object>>();
-                for (int i = 0; i < indexQuery.size(); i++) {
+                for (int i = 1; i < indexQuery.size(); i++) {
                     final JointIndexQuery.Subquery subquery = indexQuery.getQuery(i);
                     retrievals.add(new QueryUtil.IndexCall<Object>() {
                         @Override
@@ -1266,7 +1268,7 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
                                 return indexCache.get(adjustedQuery, new Callable<List<Object>>() {
                                     @Override
                                     public List<Object> call() throws Exception {
-                                        return QueryProfiler.profile(subquery.getProfiler(), adjustedQuery, q -> indexSerializer.query(q, txHandle));
+                                        return QueryProfiler.profile(subquery.getProfiler(), adjustedQuery, q -> indexSerializer.query(q, txHandle).collect(Collectors.toList()));
                                     }
                                 });
                             } catch (Exception e) {
@@ -1275,10 +1277,8 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
                         }
                     });
                 }
-
-
-                List<Object> resultSet = QueryUtil.processIntersectingRetrievals(retrievals, indexQuery.getLimit());
-                iter = com.google.common.collect.Iterators.transform(resultSet.iterator(), getConversionFunction(query.getResultType()));
+                iter = new SubqueryIterator(indexQuery.getQuery(0), indexSerializer, txHandle, indexCache, indexQuery.getLimit(), getConversionFunction(query.getResultType()),
+                        retrievals.isEmpty() ? null: QueryUtil.processIntersectingRetrievals(retrievals, indexQuery.getLimit()));
             } else {
                 if (config.hasForceIndexUsage()) throw new JanusGraphException("Could not find a suitable index to answer graph query and graph scans are disabled: " + query);
                 log.warn("Query requires iterating over all vertices [{}]. For better performance, use indexes", query.getCondition());
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/StreamIterable.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/StreamIterable.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.janusgraph.graphdb.util;
+
+import java.util.Iterator;
+import java.util.stream.Stream;
+
+/**
+ * @author davidclement90@laposte.net
+ */
+public class StreamIterable<E> implements Iterable<E>{
+
+    private final Stream<E> stream;
+
+    public StreamIterable(Stream<E> stream) {
+        this.stream = stream;
+    }
+
+    @Override
+    public Iterator<E> iterator() {
+        return stream.iterator();
+    }
+}
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.janusgraph.graphdb.util;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import org.janusgraph.core.JanusGraphElement;
+import org.janusgraph.core.JanusGraphException;
+import org.janusgraph.diskstorage.BackendTransaction;
+import org.janusgraph.graphdb.database.IndexSerializer;
+import org.janusgraph.graphdb.query.graph.JointIndexQuery;
+import org.janusgraph.graphdb.query.profile.QueryProfiler;
+
+import com.google.common.cache.Cache;
+
+/**
+ * @author davidclement90@laposte.net
+ */
+public class SubqueryIterator implements Iterator<JanusGraphElement>, AutoCloseable {
+
+    private final JointIndexQuery.Subquery subQuery;
+
+    private final Cache<JointIndexQuery.Subquery, List<Object>> indexCache;
+
+    private Iterator<? extends JanusGraphElement> elementIterator;
+
+    private List<Object> currentIds;
+
+    private QueryProfiler profiler;
+
+    private boolean isTimerRunning;
+
+    public SubqueryIterator(JointIndexQuery.Subquery subQuery, IndexSerializer indexSerializer, BackendTransaction tx,
+            Cache<JointIndexQuery.Subquery, List<Object>> indexCache, int limit,
+            Function<Object, ? extends JanusGraphElement> function, List<Object> otherResults) {
+        this.subQuery = subQuery;
+        this.indexCache = indexCache;
+        final List<Object> cacheResponse = indexCache.getIfPresent(subQuery);
+        final Stream<?> stream;
+        if (cacheResponse != null) {
+            stream = cacheResponse.stream();
+        } else {
+            try {
+                currentIds = new ArrayList<>();
+                profiler = QueryProfiler.startProfile(subQuery.getProfiler(), subQuery);
+                isTimerRunning = true;
+                stream = indexSerializer.query(subQuery, tx).map(r -> {
+                    currentIds.add(r);
+                    return r;
+                });
+            } catch (final Exception e) {
+                throw new JanusGraphException("Could not call index", e.getCause());
+            }
+        }
+        elementIterator = stream.limit(limit).filter(e -> otherResults == null || otherResults.contains(e)).map(function).map(r -> (JanusGraphElement) r).iterator();
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (!elementIterator.hasNext() && currentIds != null) {
+            indexCache.put(subQuery, currentIds);
+            profiler.stopTimer();
+            isTimerRunning = false;
+            profiler.setResultSize(currentIds.size());
+        }
+        return elementIterator.hasNext();
+    }
+
+    @Override
+    public JanusGraphElement next() {
+        return this.elementIterator.next();
+    }
+
+    @Override
+    public void close() throws Exception {
+        if (isTimerRunning) {
+            profiler.stopTimer();
+        }
+    }
+
+}
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -72,6 +72,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -79,9 +80,12 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INDEX_MAX_RESULT_SET_SIZE;
@@ -163,6 +167,10 @@ public class ElasticSearchIndex implements IndexProvider {
             new ConfigOption<>(ELASTICSEARCH_NS, "use-deprecated-multitype-index",
             "Whether JanusGraph should group these indices into a single Elasticsearch index.", ConfigOption.Type.GLOBAL_OFFLINE, false);
 
+    public static final ConfigOption<Integer> ES_SCROLL_KEEP_ALIVE =
+            new ConfigOption<>(ELASTICSEARCH_NS, "scroll-keep-alive",
+            "How long (in secondes) elasticsearch should keep alive the scroll context.", ConfigOption.Type.GLOBAL_OFFLINE, 60);
+
     public static final int HOST_PORT_DEFAULT = 9200;
 
     /**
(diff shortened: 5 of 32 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1256,7 +1258,7 @@ else if (query.getResultType() == ElementCategory.VERTEX) {
             Iterator<JanusGraphElement> iter;
             if (!indexQuery.isEmpty()) {
                 List<QueryUtil.IndexCall<Object>> retrievals = new ArrayList<QueryUtil.IndexCall<Object>>();
-                for (int i = 0; i < indexQuery.size(); i++) {
+                for (int i = 1; i < indexQuery.size(); i++) {
```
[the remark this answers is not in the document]
davidclement90: Because I use the first query to initialize SubqueryIterator on line 1280

```
 iter = new SubqueryIterator(indexQuery.getQuery(0), indexSerializer, txHandle, indexCache, indexQuery.getLimit(), getConversionFunction(query.getResultType()),
                      retrievals.isEmpty() ? null: QueryUtil.processIntersectingRetrievals(retrievals, indexQuery.getLimit()));
```

davidclement90:
_It looks like the scroll api is always used now? Or can standard search api still be used with configuration update or based on situation (e.g. result count)? It seems that there are use cases to prefer search over scroll api. In particular Elasticsearch scroll documentation says "scrolling is not intended for real time user requests". Also would users of some raw index query features (see #423) require the search instead of scroll api?_
Yes but from my experience, they just the that because keep scrool context for nothing take memory and user do not delete scroll context.
But I just change the code to not use scroll during _totals_ method and if the limit is under nbResultByQuery.

_Depending on the answer to the above, are there tests that hit both scroll/non-scroll branches?_
There are already tests that cover both cases

_Why was max-result-size configuration item removed? I think having an overall max-result-size configuration is still valuable, regardless of whether results are being paginated or scrolled, or is this enforced somewhere else now?_
Why is still valuable ? Because for Elasticsearch with scroll it's useless. I already scroll over billions of results. For me _max-result-size_ is use because you can not ask in one query a lot of results.  
I did the PR for two goals : 
1. Speed-up the first result
2. Deep paging 

For Solr, I'm not sure because I can not use cursor. I may need to restore max-result-size for Solr ?

sjudeng:
Thanks for your responses, they all make good sense. I like the change to use the search api if result is less than `nbResultByQuery`. Rather than adding the new parameter what about keeping `max-result-size` and just updating documentation to indicate that for index backends that support scrolling it will represent the number of results in each batch?

Regarding deleting the scroll context, it looks like you're just clearing after competing iteration over the given batch? Do you handle cases where iteration does not complete (e.g. client code stops iterating early, exceptions occur during processing, etc.), in terms of clearing the last scroll? If not I'm wondering about implications for the case where the same graph stays open during the (potentially long) lifetime of an application, where these uncleared scrolls might stack up.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -432,26 +432,28 @@ private static Sort getSortOrder(IndexQuery query) {
...
-            if (searcher == null) return ImmutableList.of(); //Index does not yet exist
+            if (searcher == null) return Collections.unmodifiableList(new ArrayList<String>()).stream(); //Index does not yet exist
 
             Query q = searchParams.getQuery();
             if (null == q)
                 q = new MatchAllDocsQuery();
 
             long time = System.currentTimeMillis();
-            final TopDocs docs = searcher.search(q, query.hasLimit() ? query.getLimit() : Integer.MAX_VALUE - 1, getSortOrder(query));
+            TopDocs docs = searcher.search(q, query.hasLimit() ? query.getLimit() : Integer.MAX_VALUE - 1, getSortOrder(query));
```
sjudeng: Restore `final`

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -432,26 +432,28 @@ private static Sort getSortOrder(IndexQuery query) {
...
-            List<String> result = new ArrayList<String>(docs.scoreDocs.length);
-            for (int i = 0; i < docs.scoreDocs.length; i++) {
-                result.add(searcher.doc(docs.scoreDocs[i].doc).getField(DOCID).stringValue());
-            }
-            return result;
+            return Arrays.stream(docs.scoreDocs).map(match -> {
+                try {
+                    return searcher.doc(match.doc).getField(DOCID).stringValue();
+                } catch (IOException e) {
+                    throw new RuntimeException("Could not execute Lucene query", e);
```
sjudeng: In this case it might be worth sticking with a non-stream implementation so the IOException can be thrown and caught below for consistent exception handling (e.g. TemporaryBackendException) in the method.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -636,11 +638,13 @@ private final SearchParams convertQuery(Condition<?> condition, KeyInformation.S
...
-            List<RawQuery.Result<String>> result = new ArrayList<RawQuery.Result<String>>(docs.scoreDocs.length);
-            for (int i = offset; i < docs.scoreDocs.length; i++) {
-                result.add(new RawQuery.Result<String>(searcher.doc(docs.scoreDocs[i].doc).getField(DOCID).stringValue(),docs.scoreDocs[i].score));
-            }
-            return result;
+            return Arrays.stream(Arrays.copyOfRange(docs.scoreDocs, offset, docs.scoreDocs.length-offset)).map(match -> {
+                try {
+                    return new RawQuery.Result<>(searcher.doc(match.doc).getField(DOCID).stringValue(),match.score);
+                } catch (IOException e) {
+                    throw new RuntimeException("Could not execute Lucene query", e);
```
sjudeng: Ditto consistent exception handling

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -523,11 +528,11 @@ private void commitDeletes(String collectionName, List<String> deleteIds) throws
     }
 
     @Override
-    public List<String> query(IndexQuery query, KeyInformation.IndexRetriever informations, BaseTransaction tx) throws BackendException {
-        List<String> result;
+    public Stream<String> query(IndexQuery query, KeyInformation.IndexRetriever informations, BaseTransaction tx) throws BackendException {
         String collection = query.getStore();
         String keyIdField = getKeyFieldId(collection);
         SolrQuery solrQuery = new SolrQuery("*:*");
+        solrQuery.set(CommonParams.FL, keyIdField);
```
sjudeng: What's this update for?

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -539,84 +544,68 @@ private void commitDeletes(String collectionName, List<String> deleteIds) throws
...
-
-            if (logger.isDebugEnabled())
-                logger.debug("Executed query [{}] in {} ms", query.getCondition(), response.getElapsedTime());
-
-            int totalHits = response.getResults().size();
-
-            if (!query.hasLimit() && totalHits >= maxResults)
-                logger.warn("Query result set truncated to first [{}] elements for query: {}", maxResults, query);
+        if (query.hasLimit()) solrQuery.setRows(Math.min(query.getLimit(), nbResultByQuery));
+        else solrQuery.setRows(nbResultByQuery);
```
sjudeng: Add newlines and curly brackets

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -539,84 +544,68 @@ private void commitDeletes(String collectionName, List<String> deleteIds) throws
...
+
+    private SolrQuery runCommonQuery(RawQuery query, KeyInformation.IndexRetriever informations, BaseTransaction tx, String collection, String keyIdField) throws BackendException {
         SolrQuery solrQuery = new SolrQuery(query.getQuery())
                                 .addField(keyIdField)
                                 .setIncludeScore(true)
-                                .setStart(query.getOffset())
-                                .setRows(query.hasLimit() ? query.getLimit() : maxResults);
+                                .setStart(query.getOffset());
+        if (query.hasLimit()) solrQuery.setRows(Math.min(query.getLimit(), nbResultByQuery));
+        else solrQuery.setRows(nbResultByQuery);
```
sjudeng: Newlines and curly brackets

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -878,20 +885,20 @@ public void restore(Map<String,Map<String, List<IndexEntry>>> documents, KeyInfo
             }
         }
         sr.setFrom(0);
-        if (query.hasLimit()) sr.setSize(query.getLimit());
-        else sr.setSize(maxResultsSize);
+        if (query.hasLimit()) sr.setSize(Math.min(query.getLimit(), nbResultByQuery));
+        else sr.setSize(nbResultByQuery);
```
sjudeng: Newlines and curly braces

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchScroll.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+
+import org.janusgraph.diskstorage.indexing.RawQuery;
+import org.janusgraph.diskstorage.indexing.RawQuery.Result;
+
+/**
+ * @author David Clement (david.clement90@laposte.net)
+ */
+public class ElasticSearchScroll implements Iterator<RawQuery.Result<String>>{
```
sjudeng: White space before curly braces. Please fix here and throughout PR.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchScroll.java
```diff
@@ -0,0 +1,72 @@
...
+            if (isFinished) {
+                return false;
+            }
+            ElasticSearchResponse res = client.search(scrollId);
+            res.getResults().forEach(queue::add);
+            isFinished = res.nbResults() < nbDocByQuery;
+            if (isFinished) client.deleteScroll(scrollId);
+            return res.nbResults() > 0;
+        } catch (IOException e) {
+             throw new RuntimeException(e);
```
sjudeng: Consistent exception handling (see comments in SolrResultIterator)

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchScroll.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    @Override
+    public Result<String> next() {
+         try {
+             return queue.take();
+         } catch (InterruptedException e) {
+              throw new RuntimeException(e);
```
sjudeng: Consistent exception handling (see comments in SolrResultIterator)

davidclement90:
_Thanks for your responses, they all make good sense. I like the change to use the search api if result is less than nbResultByQuery. Rather than adding the new parameter what about keeping max-result-size and just updating documentation to indicate that for index backends that support scrolling it will represent the number of results in each batch?_
Ok

_Regarding deleting the scroll context, it looks like you're just clearing after competing iteration over the given batch? Do you handle cases where iteration does not complete (e.g. client code stops iterating early, exceptions occur during processing, etc.), in terms of clearing the last scroll? If not I'm wondering about implications for the case where the same graph stays open during the (potentially long) lifetime of an application, where these uncleared scrolls might stack up._
Elasticsearch do not keep the scroll context for life (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html#scroll-search-context), you can choose the duration with scroll-keep-alive by default it's 60 seconds.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrResultIterator.java
```diff
@@ -0,0 +1,105 @@
...
+    public SolrResultIterator(SolrClient solrClient, Integer limit, int offset, int nbDocByQuery, String collection, SolrQuery solrQuery, Function<SolrDocument, E> function) throws SolrServerException, IOException {
+        this.solrClient = solrClient;
+        cpt = 0;
+        this.offset = offset;
+        this.nbDocByQuery = nbDocByQuery;
+        queue = new LinkedBlockingQueue<>();
+        this.collection = collection;
+        this.solrQuery = solrQuery;
+        this.function = function;
+        long nbFound = solrClient.queryAndStreamResponse(collection, solrQuery, new SolrCallbackHandler(this, function)).getResults().getNumFound() - offset;
```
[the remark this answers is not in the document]
davidclement90: I'm in a stream inside a map or an iterator.
So I have to throw RuntimeException and It's have to be catch when the stream is collected ?
So I just create a specific RuntimeException

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrResultIterator.java
```diff
@@ -0,0 +1,105 @@
...
+        nbRequest = 1;
+    }
+
+    public BlockingQueue<E> getQueue(){
+        return queue;
+    }
+
+    @Override
+    public boolean hasNext() {
+        if(cpt!=0 && cpt % nbDocByQuery == nbDocByQuery/2 && nbDocByQuery * nbRequest < limit){
```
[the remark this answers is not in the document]
davidclement90: 1) It's not already tested ? I can add tests.

2) At the begin of the PR, i would like to delegate the search to a stream (so I asked to the stream when i consumed half of the previous bulk), but I have some issue so I split the PR in two.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
davidclement90: To not ask more information to Solr. With this line, Solr will just return the document id and not all the document. I did the same for ES, I add `"fields" : []` to each request.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchScroll.java
```diff
@@ -0,0 +1,73 @@
...
+            if (isFinished) {
+                return false;
+            }
+            ElasticSearchResponse res = client.search(scrollId);
+            res.getResults().forEach(queue::add);
+            isFinished = res.nbResults() < nbDocByQuery;
+            if (isFinished) client.deleteScroll(scrollId);
+            return res.nbResults() > 0;
+        } catch (IOException e) {
+             throw new PermanentBackendRuntimeException(e);
```
sjudeng: `throw new UncheckedIOException(e)` instead

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchScroll.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    @Override
+    public Result<String> next() {
+         try {
+             return queue.take();
+         } catch (InterruptedException e) {
+              throw new PermanentBackendRuntimeException(e);
```
sjudeng: `throw new UncheckedIOException(new IOException("Interrupted waiting on queue", e));` instead

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -540,83 +545,73 @@ private void commitDeletes(String collectionName, List<String> deleteIds) throws
...
 
-            result = new ArrayList<String>(totalHits);
-            for (SolrDocument hit : response.getResults()) {
-                result.add(hit.getFieldValue(keyIdField).toString());
-            }
+    private <E> Stream<E> executeQuery(Integer limit, int offset, String collection, SolrQuery solrQuery, Function<SolrDocument, E> function) throws PermanentBackendException {
+        try {
+            Stream<E> toReturn = StreamSupport.stream(Spliterators.spliteratorUnknownSize(new SolrResultIterator<>(solrClient,limit, offset, solrQuery.getRows(), collection, solrQuery, function), Spliterator.ORDERED), false);
+            return limit != null ? toReturn.limit(limit) : toReturn;
         } catch (IOException e) {
```
sjudeng: Add UncheckedIOException (e.g. `catch (IOException | UncheckedIOException e)`)

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrResultIterator.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public boolean hasNext() {
+        if(cpt!=0 && cpt % nbDocByQuery == 0 && nbDocByQuery * nbRequest < limit){
+            try {
+                solrQuery.setStart(nbRequest * nbDocByQuery + offset);
+                solrClient.queryAndStreamResponse(collection, solrQuery, new SolrCallbackHandler(this, function));
+                nbRequest++;
+            } catch (SolrServerException | IOException e) {
+                throw new PermanentBackendRuntimeException(e);
```
sjudeng: Add a new `UncheckedSolrException` class for wrapping the `SolrServerException`, then break this into two catch blocks to throw UncheckedSolrException or UncheckedIOException, respectively.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrResultIterator.java
```diff
@@ -0,0 +1,106 @@
...
+        return cpt < limit;
+    }
+
+    @Override
+    public E next() {
+        try {
+            cpt++;
+            return queue.take();
+        } catch (InterruptedException e) {
+             throw new PermanentBackendRuntimeException(e);
```
sjudeng: `throw new UncheckedIOException(new IOException("Interrupted waiting on queue", e));` instead

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -540,83 +545,73 @@ private void commitDeletes(String collectionName, List<String> deleteIds) throws
...
             logger.error("Query did not complete : ", e);
             throw new PermanentBackendException(e);
         } catch (SolrServerException e) {
```
sjudeng: Add `UncheckedSolrException` (e.g. `catch (SolrServerException | UncheckedSolrException e)`)

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -878,20 +885,23 @@ public void restore(Map<String,Map<String, List<IndexEntry>>> documents, KeyInfo
...
+            sr.setSize(Math.min(query.getLimit(), nbResultByQuery));
+        } else {
+            sr.setSize(nbResultByQuery);
+        }
 
         ElasticSearchResponse response;
         try {
-            response = client.search(getIndexStoreName(query.getStore()), useMultitypeIndex ? query.getStore() : null, compat.createRequestBody(sr, NULL_PARAMETERS));
+            response = client.search(getIndexStoreName(query.getStore()), useMultitypeIndex ? query.getStore() : null, compat.createRequestBody(sr, NULL_PARAMETERS), sr.getSize() >= nbResultByQuery);
         } catch (IOException e) {
```
sjudeng: Include the subsequent scrolls in the try-catch and add `UncheckedIOException` to the catch. Also can you split the long lines for clarity (maybe define indexName, typeName and ElasticSearchScroll iterator on separate lines)?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchResponse.java
```diff
@@ -42,11 +45,23 @@ public void setTotal(long total) {
...
-        return results;
+    public Stream<RawQuery.Result<String>> getResults() {
+        return results.stream();
     }
 
     public void setResults(List<RawQuery.Result<String>> results) {
         this.results = results;
     }
+
+    public int nbResults() {
```
sjudeng: How about `numResults` instead of `nbResults`?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchScroll.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @author David Clement (david.clement90@laposte.net)
+ */
+public class ElasticSearchScroll implements Iterator<RawQuery.Result<String>> {
+
+    private final BlockingQueue<RawQuery.Result<String>> queue;
+    private boolean isFinished;
+    private final ElasticSearchClient client;
+    private final String scrollId;
+    private final int nbDocByQuery;
```
sjudeng: /nbDocByQuery/batchSize/

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/profile/QueryProfiler.java
```diff
@@ -110,4 +111,11 @@ public void setResultSize(long size) {
         return result;
     }
 
+    public static QueryProfiler startProfile(QueryProfiler profiler, Subquery query) {
+        QueryProfiler sub = profiler.addNested("backend-query");
```
sjudeng: `final`

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/StreamIterable.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Iterator;
+import java.util.stream.Stream;
+
+/**
+ * @author davidclement90@laposte.net
+ */
+public class StreamIterable<E> implements Iterable<E>{
+
+    private Stream<E> stream;
```
sjudeng: `final`

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @author davidclement90@laposte.net
+ */
+public class SubqueryIterator implements Iterator<JanusGraphElement>, AutoCloseable {
+
+    private final JointIndexQuery.Subquery subQuery;
+
+    private final Cache<JointIndexQuery.Subquery, List<Object>> indexCache;
+
+    private Iterator<? extends JanusGraphElement> itElement;
```
sjudeng: /itElement/elementIterator/

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -0,0 +1,100 @@
...
+    private QueryProfiler profiler;
+
+    private boolean isTimerRunning;
+
+    public SubqueryIterator(JointIndexQuery.Subquery subQuery, IndexSerializer indexSerializer, BackendTransaction tx,
+            Cache<JointIndexQuery.Subquery, List<Object>> indexCache, int limit,
+            Function<Object, ? extends JanusGraphElement> function, List<Object> otherResults) {
+        this.subQuery = subQuery;
+        this.indexCache = indexCache;
+        List<Object> cacheResponse;
```
sjudeng: `final` here and throughout PR

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -187,7 +194,7 @@
     private final AbstractESCompat compat;
     private final ElasticSearchClient client;
     private final String indexName;
-    private final int maxResultsSize;
+    private final int nbResultByQuery;
```
sjudeng: /nbResultsByQuery/batchSize/

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
davidclement90: This one is not final because _cacheResponse_ is set on line 59 but for all the others and even more I update my eclipse save actions.
We need to work on https://github.com/JanusGraph/janusgraph/issues/242...

davidclement90:
I only add test for Solr implementation of Scroll, because ES one, it's base on Elasticsearch mechanism (respond empty when the scroll is over) and Lucene implementation does not change anymore.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -862,36 +871,38 @@ public void restore(Map<String,Map<String, List<IndexEntry>>> documents, KeyInfo
...
+            sr.setSize(Math.min(query.getLimit(), batchSize));
+        } else {
+            sr.setSize(batchSize);
+        }
 
         ElasticSearchResponse response;
         try {
-            response = client.search(getIndexStoreName(query.getStore()), useMultitypeIndex ? query.getStore() : null, compat.createRequestBody(sr, NULL_PARAMETERS));
-        } catch (IOException e) {
+            response = client.search(getIndexStoreName(query.getStore()), useMultitypeIndex ? query.getStore() : null, compat.createRequestBody(sr, NULL_PARAMETERS), sr.getSize() >= batchSize);
```
sjudeng: Define index and type on separate line to get this line length a little more manageable.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/UncheckedSolrException.java
```diff
@@ -0,0 +1,47 @@
...
+ * @author davidclement@laposte.net
+ */
+public class UncheckedSolrException extends RuntimeException {
+
+    private static final long serialVersionUID = -8688401420333013730L;
+
+    /**
+     * @param msg Exception message
+     */
+    public UncheckedSolrException(String msg) {
```
sjudeng: Change to `public UncheckedSolrException(SolrServerException cause)`.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/UncheckedSolrException.java
```diff
@@ -0,0 +1,47 @@
...
+        super(msg);
+    }
+
+    /**
+     * @param msg   Exception message
+     * @param cause Cause of the exception
+     */
+    public UncheckedSolrException(String msg, Throwable cause) {
```
sjudeng: Change to `public UncheckedSolrException(String message, SolrServerException cause)` (all constructors should take `SolrServerException`).

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/UncheckedSolrException.java
davidclement90: I keep this one but I change _public UncheckedSolrException(Throwable cause)_ to _public UncheckedSolrException(SolrServerException cause)_

sjudeng: I'm looking at [java.io.UncheckedIOException ](https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html) as a model here. Requiring SolrServerException in all constructors protects against the misuse of the class for something other than its intended purpose (e.g. to wrap a SolrServerException as a RuntimeException). It also provides an API guarantee for future clients that the cause of an UncheckedSolrException is always a SolrServerException.

davidclement90: Ok, I just make the change

davidclement90:
I find an easy fix. There have been breaking change in scroll API between ES 1 and ES 2.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java
```diff
@@ -91,7 +91,9 @@ public Connection connect(Configuration config) throws IOException {
             }
             RestClient rc = RestClient.builder(hosts.toArray(new HttpHost[hosts.size()])).build();
 
-            RestElasticSearchClient client = new RestElasticSearchClient(rc);
+            final int scrollKeepAlive = config.get(ElasticSearchIndex.ES_SCROLL_KEEP_ALIVE);
+            Preconditions.checkArgument(scrollKeepAlive >= 1, "Scroll Keep alive should be greater or equals than 1");
+            RestElasticSearchClient client = new RestElasticSearchClient(rc, scrollKeepAlive);
```
sjudeng: final

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -274,6 +279,38 @@ public RestSearchResponse search(String indexName, String type, Map<String,Objec
...
+    @Override
+    public RestSearchResponse search(String scrollId) throws IOException {
+        final String path;
+        final byte[] requestData;
+        if (ElasticMajorVersion.ONE == majorVersion) {
+             path = new StringBuilder(REQUEST_SEPARATOR).append("/_search/scroll?scroll=").append(scrollKeepAlive).toString();
+             requestData = scrollId.getBytes();
+        } else {
+            path = "_search/scroll";
+            Map<String, Object> request = new HashMap<>();
```
sjudeng: final

sjudeng:
Unless anyone wants to add their review or request more time to review I'll merge this in 48 hours.

PR #491, opened 2017-09-01
Bump jline to use same as TinkerPop version

robertdale:


robertdale:
@pluradj Agreed this is a better approach. Updated and verified.

PR #490, opened 2017-09-01
Enclosing transaction is closed error fix

xdev-developer:
Fixed enclosing transaction is closed error when multiple threads interact with graph
Fixes #422

PR #499, opened 2017-09-07
Support full drop/delete in storage and indexing backends

### LINKED ISSUE #462: Support graph/index deletion in core factory API
This issue is based on discussions in #392 and in https://groups.google.com/forum/#!topic/janusgraph-dev/hYFY-_nshyg (thanks @dpitera). To summarize the core graph factory API should support graph/index deletion with the expected result that keyspaces/tables/indices/etc. are dropped/deleted.

Currently storage and indexing backends implement deletion through [Backend#clearStorage()](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreManager.java#L48) and [IndexProvider#clearStorage()](https://github.com/JanusGraph/janusgraph/blob/7f924fade7cc36912a9ad9f2883011c7c511d72c/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexProvider.java#L131), respectively, which is exposed at a higher level through [JanusGraphCleanup#clear()](https://github.com/JanusGraph/janusgraph/blob/bfa3c970555932d4983155168322f72297aaee85/janusgraph-core/src/main/java/org/janusgraph/core/util/JanusGraphCleanup.java#L42). But existing delete implementations are not generally consistent with the expected behavior as it appears they were written to support the testing rather than production use case (`JanusGraphCleanup#clear()` documentation indicates that the method "is intended only for development and testing use"). For example the Cassandra implementations (CQL, Thrift, etc.) truncate instead of drop keyspaces, the HBase implementation deletes rows using an implementation that does not scale, and the Solr implementation does not appear to delete collections as would be expected. Also tests are missing across all storage/indexing backends to validate expected delete behavior.

The goal in addressing this issue should be to support dropping keyspaces/etc. while still maintaining the existing clearStorage implementations for testing. The latter is critical because it's been found that dropping keyspaces/tables during testing has a significant impact on performance (see #384 which showed truncation, among other updates, decreased runtime for the TinkerPop test suite in the CQL module from 15 hours to 3 hours).

sjudeng:
Fixes #462

### DIFF
--- a/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CachingCQLStoreManager.java
+++ b/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CachingCQLStoreManager.java
@@ -9,9 +9,12 @@ import org.janusgraph.diskstorage.configuration.Configuration;
 import java.util.HashMap;
 import java.util.Map;
 
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.DROP_ON_CLEAR;
+
 public class CachingCQLStoreManager extends CQLStoreManager {
 
     private static Cluster cluster;
+
     private static Map<String,Session> sessions = new HashMap<>();
 
     public CachingCQLStoreManager(final Configuration configuration) throws BackendException {
@@ -36,6 +39,10 @@ public class CachingCQLStoreManager extends CQLStoreManager {
 
     @Override
     public void close() {
+        if (this.storageConfig.get(DROP_ON_CLEAR)) {
+            sessions.values().forEach(Session::close);
+            sessions.clear();
+        }
         this.executorService.shutdownNow();
     }
 
--- a/janusgraph-berkeleyje/src/main/java/org/janusgraph/diskstorage/berkeleyje/BerkeleyJEStoreManager.java
+++ b/janusgraph-berkeleyje/src/main/java/org/janusgraph/diskstorage/berkeleyje/BerkeleyJEStoreManager.java
@@ -267,6 +267,11 @@ public class BerkeleyJEStoreManager extends LocalStoreManager implements Ordered
     }
 
     @Override
+    public boolean exists() throws BackendException {
+        return !environment.getDatabaseNames().isEmpty();
+    }
+
+    @Override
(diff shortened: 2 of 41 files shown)


sjudeng:
The build failure is due to Codecov. I thought the status message/checks were disabled in [codecov.yml](https://github.com/JanusGraph/janusgraph/blob/master/codecov.yml). Will look into that separately from this PR.

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CachingCQLStoreManager.java
```diff
@@ -36,6 +40,13 @@ Session initializeSession(final String keyspaceName) {
 
     @Override
     public void close() {
+        if (this.storageConfig.get(DROP_ON_CLEAR)) {
+            final Iterator<Map.Entry<String, Session>> iterator = sessions.entrySet().iterator();
+            while (iterator.hasNext()) {
```
davidclement90: You can use stream to make it more compact.

sjudeng:
@davidclement90 Did you receive (and approve) the invitation mentioned in https://github.com/JanusGraph/janusgraph/pull/490#issuecomment-327361360? @mbrukman Anything else needed to get David write access?

mbrukman:
@sjudeng — it appears that @davidclement90 hasn't yet accepted his invitation; once that's done, he automatically gets write access and his checkmarks get a shiny coat of green paint.

sjudeng:
@davidclement90 Please do approve that invitation since otherwise unfortunately your review does not count for the GitHub merge check.

mbrukman:
@sjudeng – per @davidclement90's [comment on another PR](https://github.com/JanusGraph/janusgraph/pull/490#issuecomment-329805917), the GitHub committer invitation ended up in his spam folder, so he did not notice it earlier. Now that he's accepted it, his approval checkmarks are green everywhere, and he should have write / merge access to this repo.

davidclement90:
The mail was in my spam...
I just accept it.

PR #501, opened 2017-09-07
Revisit disabling Codecov comments and build status updates

### LINKED ISSUE #502: Codecov is adding comments and affecting PR checks
Test coverage reports can be uploaded to Codecov without adding comments and checks to the PR. The Codecov PR feedback should be disabled for now until test coverage stabilizes and we assess whether/how to include Codecov in the PR review process.

sjudeng:
Fixes #502

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,12 +1,15 @@
 # Don't post a comment on pull requests.
-comment: off
+comment: false
 
 # Disable commit statuses
 coverage:
   status:
-    project: no
-    patch: no
-    changes: no
+    project:
+      default: false
+    patch:
+      default: false
+    changes:
+      default: false
 
 # Ignore test files
 ignore:


codecov.yml
```diff
@@ -1,12 +1,18 @@
 # Don't post a comment on pull requests.
-comment: off
+comment: false
```
mbrukman: It's ironic that YAML is written to be easy-to-write and according to http://yaml.org/type/bool.html, any possible negative string like "no", "off", and "false" are equivalent to `false`, but then some parsers are specific—that said, if that's what they need, that's fine. The config examples appear to use `false` so this change makes sense.

codecov.yml
```diff
@@ -1,12 +1,18 @@
...
 
 # Disable commit statuses
 coverage:
   status:
-    project: no
-    patch: no
-    changes: no
+    project:
```
mbrukman: The [docs](https://docs.codecov.io/v4.3.6/docs/commit-status) provide the following example:

```yaml
coverage:
  status:
    project:
      default: false  # disable the default status that measures entire project
```

I can't find any docs that use the format `enabled: no`, can you point me to where you found an example that this config is following?

sjudeng: @mbrukman I'll try `default: false`, that's still different than what was there and the docs look more current. I was looking at https://github.com/codecov/support/wiki/Codecov-Yaml#full-yaml.

codecov.yml
sjudeng: They have a validator too (`curl --data-binary @codecov.yml https://codecov.io/validate`) but it might only be checking for general YAML syntax because all variants I tried validated fine.

sjudeng:
This PR is unnecessary. Codecov comments/checks are disabled with the current codecov.yml. See comment in #502.

PR #511, opened 2017-09-12
Add support for Elasticsearch 6.0.0-beta2 and cleanup

### LINKED ISSUE #510: Update Elasticsearch client and add support for 6.0.0-beta2
Update to latest stable Elasticsearch client (currently 5.5.2), add support for latest Elasticsearch 6.0.0 pre-release (currently 6.0.0-beta2) and restore automated tests for Elasticsearch 1.x (requires Docker). Also cleanup tests as described below.

* Use `indes.search.hostname` argument instead of `is.external.es` for testing against an externally-manged server instance
* Remove lingering transport client references from ElasticSearchSetup
* Consolidate Elasticsearch test configuration to remove redundancy
* Cleanup ElasticSearchConfigTest
* Fix typo in ElasticSearchMultiTypeIndexTest name

sjudeng:
Fixes #510


davidclement90:
For ES6, we may use the new Java RestClient API https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-compatibility.html ?

This api was released in 5.6.0 but "The 5.6 client can communicate with any 5.6.x Elasticsearch node. Previous 5.x minor versions like 5.5.x, 5.4.x etc. are not (fully) supported." So I think we should stay with the actual code for version < 6 but we should use this API for 6

sjudeng:
@davidclement90 Once they release a stable 6.0.0 we can see if the provided low level client still allows communication with previous versions (ideally at least through 2.x but at least 5.x). If so we could update to it and then create another ElasticSearchClient implementation for use when version is 6.x, with the current RestElasticSearchClient being used otherwise. 

Can we create a separate issue for this? The bulk of the updates here are configuration cleanup unrelated to ES 6.x support. The actual updates to add support for ES 6.x to the low level client are pretty minor and allow full support for ES 1.x-5.x to be maintained while using the latest stable (5.5.2) client release.

davidclement90:
Ok I create the issue https://github.com/JanusGraph/janusgraph/issues/520

janusgraph-es/pom.xml
```diff
@@ -11,12 +11,13 @@
     <url>http://janusgraph.org</url>
     <properties>
         <top.level.basedir>${basedir}/..</top.level.basedir>
-        <elasticsearch.test.version>${elasticsearch.version}</elasticsearch.test.version>
-        <!-- Groovy dynamic scripting must be enabled when testing on ES 1.x/2.x servers.
-             From ES 5.0 Painless inline scripts are used which are enabled by default. -->
-        <elasticsearch.test.groovy.inline>false</elasticsearch.test.groovy.inline>
+        <elasticsearch1.test.version>1.7.6</elasticsearch1.test.version>
+        <elasticsearch2.test.version>2.4.6</elasticsearch2.test.version>
+        <elasticsearch5.test.version>5.5.2</elasticsearch5.test.version>
```
davidclement90: Do you test with 5.6.0 ?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -408,6 +404,19 @@ private void pushMapping(String store, String key, KeyInformation information) t
             properties.put(key, compat.createKeywordMapping());
         }
 
+        if (useAllField && client.getMajorVersion().getValue() >= 6) {
+            // add custom all field mapping if it doesn't exist
```
davidclement90: Copy_to has been added in ES 2, so your code should work in ES 2, 5 and 6

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -199,10 +190,12 @@
     private final boolean useExternalMappings;
     private final Map<String,Object> indexSetting;
     private final long createSleep;
+    private final boolean useAllField;
     private final boolean useMultitypeIndex;
 
     public ElasticSearchIndex(Configuration config) throws BackendException {
         indexName = config.get(INDEX_NAME);
+        useAllField = config.get(USE_ALL_FIELD);
         useExternalMappings = config.get(USE_EXTERNAL_MAPPINGS);
```
davidclement90: As you add an _all_ field, I think for _query_string_, you should set the _default_field_ in the query_string to _all_ for ES >=2 because copy_to was add in ES 2

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -163,9 +147,16 @@
             new ConfigOption<>(ES_CREATE_NS, "use-external-mappings",
             "Whether JanusGraph should make use of an external mapping when registering an index.", ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<Boolean> USE_ALL_FIELD =
+        new ConfigOption<>(ELASTICSEARCH_NS, "use-all-field",
+            "Whether JanusGraph should add an \"all\" field mapping. When enabled field mappings will " +
+            "include a \"copy_to\" parameter referencing the \"all\" field. This is supported since Elasticsearch 6.x " +
+            " and is required when using wildcard fields starting in Elasticsearch 6.x.", ConfigOption.Type.GLOBAL_OFFLINE, true);
+
```
davidclement90: Copy_to has been added in ES 2, so your code should work in ES 2, 5 and 6 ?
And where do you want to use wildcard fields ? In query_string ?

sjudeng: @davidclement90 It comes up in [JanusGraphIndexTest#testWidcardQuery](https://github.com/JanusGraph/janusgraph/blob/1563460babc6ebd658ac3dd4c687d7abdebb82b2/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java#L1667-L1689).

sjudeng:
@davidclement90 I was thinking it's not necessary to add the custom "all" field in ES 2.x/5.x since there's already the default `_all` field in those versions. What do you think? I wasn't sure if there would be a performance impact of an additional "all" field. It's also why I added the `USE_ALL_FIELD` configuration item, so even in ES 6.x users could opt-out of the custom "all" field if there was a performance issue.

If we do want to support the custom "all" field in ES 2.x/5.x the easist way would be to remove the `USE_ALL_FIELD` configuration item and then always use the custom field in ES 2.x-6.x. It would get more complicated to implement optionally via `USE_ALL_FIELD`, as we'd need to pass the configuration value in to the ESCompat instance for use in setting the `wildcardField` in the IndexFeatures.

janusgraph-es/pom.xml
sjudeng: Thanks, I just updated the client and default 5.x test version to 5.6.0. I hadn't seen this version was released.

davidclement90:
Ok, I was thinking that it's was a new feature and not a fix for ES 6.
So the original code it's ok for me

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -163,9 +147,16 @@
             new ConfigOption<>(ES_CREATE_NS, "use-external-mappings",
             "Whether JanusGraph should make use of an external mapping when registering an index.", ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<Boolean> USE_ALL_FIELD =
+        new ConfigOption<>(ELASTICSEARCH_NS, "use-all-field",
+            "Whether JanusGraph should add an \"all\" field mapping. When enabled field mappings will " +
+            "include a \"copy_to\" parameter referencing the \"all\" field. This is supported since Elasticsearch 6.x " +
+            " and is required when using wildcard fields starting in Elasticsearch 6.x.", ConfigOption.Type.GLOBAL_OFFLINE, true);
```
sjudeng: @davidclement90 Now I think this should be `ConfigOption.type.FIXED` (see [here](http://docs.janusgraph.org/latest/config-ref.html#cfg-mutability)), since I don't think it should be changed once the index is created. Do you agree? Should your `USE_DEPRECATED_MULTITYPE_INDEX` be "FIXED" as well?

davidclement90: No I think is GLOBAL_OFFLINE because you can stop all JanusGraph instances. Then you can upgrade ES or change directly on ES the mapping. But you may have to drop indexes for that. 
Then you change JanusGraph properties and you use reindex job if you had to drop indexes.

So I think is not FIXED because you do not change the primary storage so you can always reindex. And for the reason I think that no backend search property has to be FIXED

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java
```diff
@@ -101,96 +78,6 @@ public Connection connect(Configuration config) throws IOException {
...
-        settings.put("client.transport.ignore_cluster_name", true);
-
-        settings.put("path.home", System.getProperty("java.io.tmpdir"));
-
-        // Apply overrides from ES conf file
-        applySettingsFromFile(settings, config, INDEX_CONF_FILE);
-
-        // Apply ext.* overrides from JanusGraph conf file
-        applySettingsFromJanusGraphConf(settings, config, ElasticSearchIndex.ES_EXTRAS_NS);
-
```
davidclement90: @sjudeng since this PR, ES_EXTRA_NS is still in ElasticSearchIndex class and in doc. But JanusGraph never use it.
I miss something, or it's had to be removed ?

sjudeng: @davidclement90 This is in a private method block (`Map<String,Object> settingsBuilder(Configuration config)`) that had no usages, which is why it was removed. But the [ElasticSearchSetup#applySettingsFromJanusGraphConf](https://github.com/JanusGraph/janusgraph/blob/b9e83bd56dc90d57d4e84c0c96fe649bf6787488/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java#L81-L83) method itself is still there and is used in [ElasticSearchIndex](https://github.com/JanusGraph/janusgraph/blob/b9e83bd56dc90d57d4e84c0c96fe649bf6787488/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java#L235). But is something breaking for you since the update?

davidclement90: For me in ElasticSearchIndex, when we call applySettingsFromJanusGraphConf, we use ES_CREATE_EXTRAS_NS and not ES_EXTRAS_NS. 
ES_EXTRAS_NS is no longer used because Elasticsearch can only be query with Rest so we can not overrides ES yaml settings.
So for me, ES_EXTRAS_NS has to be removed and the doc has to be updated. 
What do you think ?

sjudeng: @davidclement90 So you're recommending we confirm ES_EXTRA_NS is no longer supported and then remove it and update docs, right? It makes sense to me.

davidclement90: Yes, I grep over all the code and I just find a match for the definition in ElasticSearchIndex and in one comment.
`grep -ri ES_EXTRAS_NS .
./janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchSetup.java: * Any key-value pairs under the {@link org.janusgraph.diskstorage.es.ElasticSearchIndex#ES_EXTRAS_NS} namespace
./janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java:    public static final ConfigNamespace ES_EXTRAS_NS =`

I just open an issue : https://github.com/JanusGraph/janusgraph/issues/545

Issue #535, opened 2017-09-21
Mitigate recent Coverity findings


sjudeng:
Latest Coverity scan report (see [here](https://scan.coverity.com/projects/janusgraph-janusgraph)) showed a number of new findings. Issues should be addressed or dismissed as false positives as appropriate.

chupman:
@sjudeng were there any outstanding issues to resolve from this scan or can this issue be closed out?

sjudeng:
@chupman It can be closed. Looks like the coverity service might be offline or experiencing issues currently but in any case we can open future new issues for specific findings if needed.

PR #536, opened 2017-09-21
Resolve numerous Coverity findings.

sjudeng:
Relates to #535

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/internal/AbstractElement.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/internal/AbstractElement.java
@@ -67,11 +67,11 @@ public abstract class AbstractElement implements InternalElement, Comparable<Jan
             return getCompareId()==((AbstractElement)other).getCompareId();
         } else if (other instanceof JanusGraphElement) {
             return ((JanusGraphElement) other).hasId() && getCompareId()==((JanusGraphElement)other).longId();
-        } else if (other instanceof Element) {
+        } else {
             Object otherId = ((Element)other).id();
             if (otherId instanceof RelationIdentifier) return ((RelationIdentifier) otherId).getRelationId()==getCompareId();
             else return otherId.equals(getCompareId());
-        } else return false;
+        }
     }
 
 
--- a/janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
+++ b/janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
@@ -45,7 +45,6 @@ public class JanusGraphVertexDeserializer implements AutoCloseable {
     private final TypeInspector typeManager;
     private final SystemTypeInspector systemTypes;
     private final IDManager idManager;
-    private final boolean verifyVertexExistence = false;
 
     private static final Logger log =
             LoggerFactory.getLogger(JanusGraphVertexDeserializer.class);
@@ -90,8 +89,6 @@ public class JanusGraphVertexDeserializer implements AutoCloseable {
         // Create TinkerVertex
         TinkerGraph tg = TinkerGraph.open();
 
-        boolean foundVertexState = !verifyVertexExistence;
-
         TinkerVertex tv = null;
 
         // Iterate over edgestore columns to find the vertex's label relation
@@ -103,14 +100,12 @@ public class JanusGraphVertexDeserializer implements AutoCloseable {
                 long vertexLabelId = relation.getOtherVertexId();
                 VertexLabel vl = typeManager.getExistingVertexLabel(vertexLabelId);
                 // Create TinkerVertex with this label
-                //tv = (TinkerVertex)tg.addVertex(T.label, vl.label(), T.id, vertexId);
                 tv = getOrCreateVertex(vertexId, vl.name(), tg);
             }
         }
 
         // Added this following testing
         if (null == tv) {
-            //tv = (TinkerVertex)tg.addVertex(T.id, vertexId);
             tv = getOrCreateVertex(vertexId, null, tg);
         }
 
@@ -121,9 +116,6 @@ public class JanusGraphVertexDeserializer implements AutoCloseable {
             try {
                 RelationReader relationReader = setup.getRelationReader(vertexId);
                 final RelationCache relation = relationReader.parseRelation(data, false, typeManager);
-                if (systemTypes.isVertexExistsSystemType(relation.typeId)) {
-                    foundVertexState = true;
-                }
 
                 if (systemTypes.isSystemType(relation.typeId)) continue; //Ignore system types
                 final RelationType type = typeManager.getExistingRelationType(relation.typeId);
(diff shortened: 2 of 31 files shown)


janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
```diff
@@ -219,10 +219,7 @@ public TinkerVertex readHadoopVertex(final StaticBuffer key, Iterable<Entry> ent
 
         /*Since we are filtering out system relation types, we might end up with vertices that have no incident relations.
          This is especially true for schema vertices. Those are filtered out.     */
-        if (!foundVertexState) {
-            log.trace("Vertex {} has unknown lifecycle state", vertexId);
-            return null;
-        } else if (!tv.edges(Direction.BOTH).hasNext() && !tv.properties().hasNext()) {
+        if (!tv.edges(Direction.BOTH).hasNext() && !tv.properties().hasNext()) {
```
davidclement90: Why did you remove the log ?
And can you delete the dead code just above ?

sjudeng: @davidclement90 This section is unreachable because it's conditioned on `!foundVertexState` and `foundVertexState` is set to `!verifyVertexExistence` ([L93](https://github.com/JanusGraph/janusgraph/blob/84ead7d907b62061e30c5d3210e5e8422fd4648c/janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java#L93)), which is always true since `verifyVertexExistence` is `false` ([L48](https://github.com/JanusGraph/janusgraph/blob/84ead7d907b62061e30c5d3210e5e8422fd4648c/janusgraph-hadoop-parent/janusgraph-hadoop-core/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java#L48)). I never would have found this if it wasn't for Coverity (and it took me awhile to see it even after it was flagged).

I just pushed updates to further remove the useless `foundVertexState`, which hopefully makes this clearer. Also removed the commented out code.

janusgraph-core/src/main/java/org/janusgraph/graphdb/internal/AbstractElement.java
```diff
@@ -67,7 +67,7 @@ public boolean equals(Object other) {
             return getCompareId()==((AbstractElement)other).getCompareId();
         } else if (other instanceof JanusGraphElement) {
             return ((JanusGraphElement) other).hasId() && getCompareId()==((JanusGraphElement)other).longId();
-        } else if (other instanceof Element) {
+        } else if (other != null) {
```
mbrukman: Why this change from `if (other instanceof Element)` to `if (other != null)`?

The new check is always true, because the first line checks for `other == null`; however, the next line casts `other` to `Element`, so we really should have checked whether `other instanceof Element` is true.

Should this line be reverted to the original?

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -121,8 +121,7 @@
     public LuceneIndex(Configuration config) {
         String dir = config.get(GraphDatabaseConfiguration.INDEX_DIRECTORY);
         File directory = new File(dir);
-        if (!directory.exists()) directory.mkdirs();
-        if (!directory.exists() || !directory.isDirectory() || !directory.canWrite())
+        if ((!directory.exists() && !directory.mkdirs()) || !directory.isDirectory() || !directory.canWrite())
```
mbrukman: Since you're here, would you mind enclosing the `if` statement's single-line body in `{ ... }` so that it's clear that it refers to a single statement, please? We've had other situations in this code base where indentation was counter to the actual syntax; would be nice to avoid that going forward.

Incidentally, I like how Go enforces braces for `if` statements, even for a single line, to avoid this ambiguity.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -133,8 +132,7 @@ private Directory getStoreDirectory(String store) throws BackendException {
         String dir = basePath + File.separator + store;
         try {
             File path = new File(dir);
-            if (!path.exists()) path.mkdirs();
-            if (!path.exists() || !path.isDirectory() || !path.canWrite())
+            if ((!path.exists() && !path.mkdirs()) || !path.isDirectory() || !path.canWrite())
```
mbrukman: Please add `{...}` around `if`'s statement body.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrResultIterator.java
```diff
@@ -63,7 +63,7 @@ public SolrResultIterator(SolrClient solrClient, Integer limit, int offset, int
 
     @Override
     public boolean hasNext() {
-        if (count != 0 && count % batchSize == 0 && batchSize * numBatches < limit) {
+        if (count != 0 && count % batchSize == 0 && count < limit) {
```
mbrukman: This change _seems_ reasonable, but was this a Coverity-specified error or a separate change? I had to convince myself that this is correct, but this is a behavior change, so I'm curious how none of the unit tests failed before with this, and none required to be changed given this change (i.e., it should be possible to reproduce this error in a unittest such that it fails before, and succeeds after this change; or, that it succeeds before with an erroneous assertion, and this change requires changing said assertion to be correct).

Should this be a separate change?

janusgraph-core/src/main/java/org/janusgraph/graphdb/internal/AbstractElement.java
sjudeng: @mbrukman At this point other is an instance of Element since otherwise it wouldn't pass the check [here](https://github.com/JanusGraph/janusgraph/blob/bfa3c970555932d4983155168322f72297aaee85/janusgraph-core/src/main/java/org/janusgraph/graphdb/internal/AbstractElement.java#L61-L63). But I did just push an update to remove the null check since you're right null was checked above.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrResultIterator.java
sjudeng: @mbrukman The Coverity finding was that `batchSize*numBatches` is performed at int32 but limit is int64. Upon review though I realized when `count % batchSize == 0` (e.g. at the start of a new batch), `count` is equal to `batchSize * numBatches`. I was familiar with this line from reviewing when it was added (see https://github.com/JanusGraph/janusgraph/pull/481#discussion_r136208445). @davidclement90 Does this seem okay?

mbrukman: Agreed, LGTM.

davidclement90: Yes it's ok LGTM

PR #573, opened 2017-09-29
Update Elasticsearch client to 5.6.2 and ES 6.x test version to 6.0.0-rc1

### LINKED ISSUE #572: Update Elasticsearch client to 5.6.2 and ES 6.x test version to 6.0.0-rc1

sjudeng:
Fixes #572

### DIFF
--- a/janusgraph-es/pom.xml
+++ b/janusgraph-es/pom.xml
@@ -13,8 +13,8 @@
         <top.level.basedir>${basedir}/..</top.level.basedir>
         <elasticsearch1.test.version>1.7.6</elasticsearch1.test.version>
         <elasticsearch2.test.version>2.4.6</elasticsearch2.test.version>
-        <elasticsearch5.test.version>5.6.0</elasticsearch5.test.version>
-        <elasticsearch6.test.version>6.0.0-beta2</elasticsearch6.test.version>
+        <elasticsearch5.test.version>5.6.2</elasticsearch5.test.version>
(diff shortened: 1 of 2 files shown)


janusgraph-es/pom.xml
```diff
@@ -13,8 +13,8 @@
         <top.level.basedir>${basedir}/..</top.level.basedir>
         <elasticsearch1.test.version>1.7.6</elasticsearch1.test.version>
         <elasticsearch2.test.version>2.4.6</elasticsearch2.test.version>
-        <elasticsearch5.test.version>5.6.0</elasticsearch5.test.version>
-        <elasticsearch6.test.version>6.0.0-beta2</elasticsearch6.test.version>
+        <elasticsearch5.test.version>5.6.2</elasticsearch5.test.version>
```
pluradj: Wow, 3 ES 5.6.x releases in September. They certainly are cranking them out. @sjudeng do you know if they are planning to keep pushing releases this fast? ES 5.5.x releases were about once a month.

sjudeng: @pluradj I'm not sure but I don't see them slowing down anytime soon. One thing is that as far as JanusGraph dependencies we just use the [rest client](https://github.com/elastic/elasticsearch/tree/v5.6.2/client/rest) (org.elasticsearch.client:elasticsearch-rest-client), which doesn't look like it changes within a given minor version (e.g. 5.6.x). So going forward we don't need to update the client version except for minor version bumps (e.g. 5.6->5.7). I think it's still a good idea to keep the test versions up-to-date.

PR #646, opened 2017-10-17
Fix syntax highlighting for XML. [skip ci]

mbrukman:
This is the only file that marks XML code snippets with `m2` rather than `xml`
and is not properly highlighted in the HTML output as a result.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -4,7 +4,7 @@
 
 === Version 0.2.0 (Release Date: October 11, 2017)
 
-[source, m2]
+[source, xml]
 <dependency>
    <groupId>org.janusgraph</groupId>
    <artifactId>janusgraph-core</artifactId>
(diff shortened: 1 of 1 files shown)


sjudeng:
@mbrukman Unrelated question, I thought if the PR author is a committer than only one other approval is required to merge since it's assumed the author approves their own PR. I know I read that when I started committing but now I'm not sure where. In practice if the PR is larger I too will look for at least two other approvals but otherwise I'll usually only look for one approval if author is a committer. Is this wrong?

mbrukman:
@sjudeng — this isn't exactly clarified, the [development process](https://github.com/JanusGraph/janusgraph/blob/master/docs/development.txt) suggests 2 reviews and I wasn't counting the status of the PR author as committer as counting towards the +2 of a PR, but that's an interesting way to look at it. (I did merge a few of my trivial PRs with only a +1, but only because it's silly to wait for two people to review a simple text change; in this case, I just haven't gotten around to merging it, not that I am not comfortable merging it with only a +1 on the PR).

I see your point and I like where you're going with this. Would you like to propose this as a clarification on the dev list and see what the general opinion is on this? I would be +1 on such a PR / proposal, which I think is sufficient to merge as it's a docs change, right? :-)

Issue #612, opened 2017-10-09
JanusGraph with ElasticSearch authentication/SSL


techpanga:
Hi 

Our ElasticSearch has Shield Protected. So Janusgraph configuration to be supplied with ElasticSearch userid and password. 

what are the configuration keys to be supplied.

Please advise.

sjudeng:
It's not currently supported but it shouldn't be too hard to add this support in the future. Would also need to add support for HTTPS connections to Elasticsearch to ensure credentials are encrypted.

davidclement90:
I think is not that easy.
We will have to support shield syntax for ES 1 and 2 and X-pack syntax for ES 5 and 6.

For the test, shield plugin can be found in maven central. X-pack is not published in maven central. But X-pack is in official elasticsearch docker image. So we have to find a properly way to get X-pack for testing.

And the initialisation of shield and X-pack are completely different.

sjudeng:
@davidclement90 It's a good point testing will be the hardest part. Maybe we can live with Docker for the tests. For the implementation I was thinking basic auth and SSL support could be added at the HttpClient level when we're building the RestClient and we wouldn't have to get into Elasticsearch specifics (yet another benefit of dropping the Transport client).

davidclement90:
Ok the implementation is not that hard with HttpClient.
If docker is ok and works on Travis I do not know any alternative.

SatyaNarayan1:
@ngrigoriev thanks for working on it. Auth was working for us earlier ( 0.1.1) with transport client and shield with below setting

# Management System to read or modify this value after bootstrapping.
index.search.elasticsearch.client-only=true
index.search.elasticsearch.interface=TRANSPORT_CLIENT
index.search.elasticsearch.ext.shield.user=elastic:changeme

ngrigoriev:
@SatyaNarayan1 

Most recent version of JG requires the migration to REST_CLIENT, e.g. it does not support TRANSPORT_CLIENT anymore.

Ref: https://groups.google.com/forum/#!topic/janusgraph-dev/F-zfIsuiIoQ

https://www.elastic.co/blog/state-of-the-official-elasticsearch-java-clients

SatyaNarayan1:
@ngrigoriev correct and due to that we are currently not able to upgrade JG. 
Thanks for this , it will help us to migrate.

sjudeng:
Fixed in #671

PR #671, opened 2017-10-29
HTTPS, HTTP Basic auth and custom authentication support (#612, #645)

### LINKED ISSUE #645: ES REST client does not support HTTPS
In ElasticSearchSetup class I see this:

```
for (String host : config.get(INDEX_HOSTS)) {
                String[] hostparts = host.split(":");
                String hostname = hostparts[0];
                int hostport = defaultPort;
                if (hostparts.length == 2) hostport = Integer.parseInt(hostparts[1]);
                log.debug("Configured remote host: {} : {}", hostname, hostport);
                hosts.add(new HttpHost(hostname, hostport, "http"));
            }
```

I think we should have an option for HTTPS as well.

ngrigoriev:
Fixes #645 and #612: multiple changes for ES REST transport settings

- refactor client set-up code to make it more testable
- extract REST client set-up logic to a separate class
- add a set of unit tests for client configuration
- add new options for enabling SSL (e.g. HTTPS) for the client (off by
  default) and optionally configuring SSL trust store and its password
- implement HTTP basic authentication
- implement support for custom HTTP authenticator callback - to be used
  externally for other authentication types such as AWS SigV4

NOTE: This PR is primarily for preview purposes, to get the sense if the approach selected is not out of the line. If it is acceptable in general, I will update the documentation and resubmit.

### DIFF
--- a/docs/elasticsearch.txt
+++ b/docs/elasticsearch.txt
@@ -15,7 +15,7 @@ JanusGraph supports https://www.elastic.co/[Elasticsearch] as an index backend. 
 * *Temporal*: Nanosecond granularity temporal indexing.
 * *Custom Analyzer*: Choose to use a custom analyzer
 
-Please see <<version-compat>> for details on what versions of ES will work with JanusGraph.
+Please see <<version-compat>> for details on what versions of Elasticsearch will work with JanusGraph.
 
 [IMPORTANT]
 ===============================
@@ -38,7 +38,7 @@ For security reasons Elasticsearch must be run under a non-root account
 
 === Elasticsearch Configuration Overview
 
-JanusGraph supports HTTP client connections to a running Elasticsearch cluster. Please see <<version-compat>> for details on what versions of ES will work with the different client types in JanusGraph.
+JanusGraph supports HTTP(S) client connections to a running Elasticsearch cluster. Please see <<version-compat>> for details on what versions of Elasticsearch will work with the different client types in JanusGraph.
 
 [NOTE]
 JanusGraph's index options start with the string "`index.[X].`" where "`[X]`" is a user-defined name for the backend. This user-defined name must be passed to JanusGraph's ManagementSystem interface when building a mixed index, as described in <<index-mixed>>, so that JanusGraph knows which of potentially multiple configured index backends to use. Configuration snippets in this chapter use the name `search`, whereas prose discussion of options typically write `[X]` in the same position. The exact index name is not significant as long as it is used consistently in JanusGraph's configuration and when administering indices.
@@ -84,6 +84,135 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
 
 The REST client accepts the `index.[X].bulk-refresh` option. This option controls when changes are made visible to search. See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html[?refresh documentation] for more information.
 
+==== REST Client HTTPS Configuration
+
+SSL support for HTTP can be enabled by setting the `index.[X].elasticsearch.ssl.enabled` configuration option to `true`. Note that depending on your configuration you may need to change the value of `index.[X].port` if your HTTPS port number is different from the default one for the REST API (9200).
+
+When SSL is enabled you may also configure the location and password of the truststore. This can be done as follows:
+
+[source, properties]
+----
+index.search.elasticsearch.ssl.truststore.location=/path/to/your/truststore.jks
+index.search.elasticsearch.ssl.truststore.password=truststorepwd
+----
+
+Note that these settings apply only to Elasticsearch REST client and do not affect any other SSL connections in JanusGraph.
+
+Configuration of the client keystore is also supported:
+
+[source, properties]
+----
+index.search.elasticsearch.ssl.keystore.location=/path/to/your/keystore.jks
+index.search.elasticsearch.ssl.keystore.storepassword=keystorepwd
+index.search.elasticsearch.ssl.keystore.keypassword=keypwd
+----
+
+Any of the passwords can be empty.
+
+If needed, the SSL hostname verification can be disabled by setting the `index.[X].elasticsearch.ssl.disable-hostname-verification` property value to `true` and the support for self-signed SSL certificates can be enabled by setting `index.[X].elasticsearch.ssl.allow-self-signed-certificates` property value to `true`.
+
+[TIP]
+It is not recommended to rely on the self-signed SSL certificates or to disable the hostname verification for a production system as it significantly limits the client's ability to provide the secure communication channel with the Elasticsearch server(s). This may result in leaking the confidential data which may be a part of your JanusGraph index.
+
+==== REST Client HTTP Authentication
+
+REST client supports the following authentication options: Basic HTTP Authentication (username/password) and custom authentication based on the user-provided implementation.
+
+These authentication methods are independent from SSL client authentication described above.
+
+===== REST Client Basic HTTP Authentication
+
+Basic HTTP Authentication is available regardless of the state of SSL support.  Optionally, an authentication realm can be specified via `index.[X].elasticsearch.http.auth.basic.realm` property.
+
+
+[source, properties]
+----
+index.search.elasticsearch.http.auth.type=basic
+index.search.elasticsearch.http.auth.basic.username=httpuser
+index.search.elasticsearch.http.auth.basic.password=httppassword
+----
+
+[TIP]
+It is highly recommended to use SSL (e.g. setting `index.[X].elasticsearch.ssl.enabled` to `true`) when using this option as the credentials can be intercepted when sent over an unencrypted connection!
+
+===== REST Client Custom HTTP Authentication
+
+Additional authentication methods can be implemented by providing your own implementation. The custom authenticator is configured as follows:
+
+[source, properties]
+----
+index.search.elasticsearch.http.auth.custom.authenticator-class=fully.qualified.class.Name
+index.search.elasticsearch.elasticsearch.http.auth.custom.authenticator-args=arg1,arg2,...
+----
+
+Argument list is optional and can be empty.
+
+The class specified there has to implement the `org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticator` interface or extend `org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticatorBase` convenience class. The implementation gets access to HTTP client configuration and can customize the client as needed. Refer to <<javadoc>> for more information.
+
+For example, the following code snippet implements an authenticator allowing the
+Elasticsearch REST client to authenticate and get authorized against AWS IAM:
+
+[source,java]
+----
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
+import org.apache.http.HttpRequestInterceptor;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticatorBase;
+
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
+import com.amazonaws.regions.DefaultAwsRegionProviderChain;
+import com.google.common.base.Supplier;
+
+import vc.inreach.aws.request.AWSSigner;
+import vc.inreach.aws.request.AWSSigningRequestInterceptor;
+
+/**
+ * <p>
+ * Elasticsearch REST HTTP(S) client callback implementing AWS request signing.
+ * </p>
+ * <p>
+ * The signer is based on AWS SDK default provider chain, allowing multiple options for providing
+ * the caller credentials. See {@link DefaultAWSCredentialsProviderChain} documentation for the details.
+ * </p>
+ */
+public class AWSV4AuthHttpClientConfigCallback extends RestClientAuthenticatorBase {
+
+    private static final String AWS_SERVICE_NAME = "es";
+    private HttpRequestInterceptor awsSigningInterceptor;
+
+    public AWSV4AuthHttpClientConfigCallback(final String[] args) {
+        // does not require any configuration
+    }
+
+    @Override
+    public void init() throws IOException {
+        DefaultAWSCredentialsProviderChain awsCredentialsProvider = new DefaultAWSCredentialsProviderChain();
+        final Supplier<LocalDateTime> clock = () -> LocalDateTime.now(ZoneOffset.UTC);
+
+        // using default region provider chain
+        // (http://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-region-selection.html)
+        DefaultAwsRegionProviderChain regionProviderChain = new DefaultAwsRegionProviderChain();
+        final String awsRegion = regionProviderChain.getRegion();
+
+        final AWSSigner awsSigner = new AWSSigner(awsCredentialsProvider, awsRegion, AWS_SERVICE_NAME, clock);
+        this.awsSigningInterceptor = new AWSSigningRequestInterceptor(awsSigner);
+    }
+
+    @Override
+    public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
+        return httpClientBuilder.addInterceptorLast(awsSigningInterceptor);/
+    }
+}
+
+----
+
+This custom authenticator does not use any constructor arguments.
+
+
+
 ==== Ingest Pipelines
 If using Elasticsearch 5.0 or higher, a different ingest pipelines can be set for each mixed index.
 Ingest pipeline can be use to pre-process documents before indexing. A pipeline is composed by a series of processors. Each processor transforms the document in some way.
@@ -95,14 +224,16 @@ See https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html[
 
 === Secure Elasticsearch
 
-Elasticsearch does not perform authentication or authorization. A client that can connect to ES is trusted by ES. When Elasticsearch runs on an unsecured or public network, particularly the Internet, it should be deployed with some type of external security. This is generally done with a combination of firewalling and tunneling of Elasticsearch's ports. Elasticsearch has two client-facing ports to consider:
+Elasticsearch does not perform authentication or authorization. A client that can connect to Elasticsearch is trusted by Elasticsearch. When Elasticsearch runs on an unsecured or public network, particularly the Internet, it should be deployed with some type of external security. This is generally done with a combination of firewalling, tunneling of Elasticsearch's ports or by using Elasticsearch extensions such as https://www.elastic.co/guide/en/x-pack/current/index.html[X-Pack]. Elasticsearch has two client-facing ports to consider:
 
 * The HTTP REST API, usually on port 9200
 * The native "transport" protocol, usually on port 9300
 
 A client uses either one protocol/port or the other, but not both simultaneously. Securing the HTTP protocol port is generally done with a combination of firewalling and a reverse proxy with SSL encryption and HTTP authentication. There are a couple of ways to approach security on the native "transport" protocol port:
 
-Tunnel ES's native "transport" protocol:: This approach can be implemented with SSL/TLS tunneling (for instance via https://www.stunnel.org/index.html[stunnel]), a VPN, or SSH port forwarding. SSL/TLS tunnels require non-trivial setup and monitoring: one or both ends of the tunnel need a certificate, and the stunnel processes need to be configured and running continuously. The setup for most secure VPNs is likewise non-trivial. Some Elasticsearch service providers handle server-side tunnel management and provide a custom Elasticsearch `transport.type` to simplify the client setup.
+In addition to that, some hosted Elasticsearch services offer other methods of authentication and authorization. For example, AWS Elasticsearch Service requires the use of HTTPS and offers an option for using IAM-based access control. For that the requests sent to this service must be signed. This can be achieved by using a custom authenticator (see above).
+
+Tunnel Elasticsearch's native "transport" protocol:: This approach can be implemented with SSL/TLS tunneling (for instance via https://www.stunnel.org/index.html[stunnel]), a VPN, or SSH port forwarding. SSL/TLS tunnels require non-trivial setup and monitoring: one or both ends of the tunnel need a certificate, and the stunnel processes need to be configured and running continuously. The setup for most secure VPNs is likewise non-trivial. Some Elasticsearch service providers handle server-side tunnel management and provide a custom Elasticsearch `transport.type` to simplify the client setup.
 Add a firewall rule that allows only trusted clients to connect on Elasticsearch's native protocol port:: This is typically done at the host firewall level. Easy to configure, but very weak security by itself.
 
 [[es-cfg-index-create]]
@@ -116,9 +247,9 @@ JanusGraph supports customization of the index settings it uses when creating it
 
 Settings customized through this mechanism are only applied when JanusGraph attempts to create its index in Elasticsearch. If JanusGraph finds that its index already exists, then it does not attempt to recreate it, and these settings have no effect.
 
-==== Embedding ES index creation settings with `create.ext`
+==== Embedding Elasticsearch index creation settings with `create.ext`
 
-JanusGraph iterates over all properties prefixed with `index.[X].elasticsearch.create.ext.`, where `[X]` is an index name such as `search`. It strips the prefix from each property key. The remainder of the stripped key will be interpreted as an Elasticsearch index creation setting. The value associated with the key is not modified. The stripped key and unmodified value are passed as part of the `settings` object in the Elasticsearch create index request that JanusGraph issues when bootstrapping on ES. This allows embedding arbitrary index creation settings settings in JanusGraph's properties. Here's an example configuration fragment that customizes three Elasticsearch index settings using the `create.ext` config mechanism:
+JanusGraph iterates over all properties prefixed with `index.[X].elasticsearch.create.ext.`, where `[X]` is an index name such as `search`. It strips the prefix from each property key. The remainder of the stripped key will be interpreted as an Elasticsearch index creation setting. The value associated with the key is not modified. The stripped key and unmodified value are passed as part of the `settings` object in the Elasticsearch create index request that JanusGraph issues when bootstrapping on Elasticsearch. This allows embedding arbitrary index creation settings settings in JanusGraph's properties. Here's an example configuration fragment that customizes three Elasticsearch index settings using the `create.ext` config mechanism:
 
 [source, properties]
 ----
--- a/janusgraph-es/pom.xml
+++ b/janusgraph-es/pom.xml
@@ -21,6 +21,7 @@
         <skip.es.test>${skipTests}</skip.es.test>
         <es.docker.image>docker.elastic.co/elasticsearch/elasticsearch</es.docker.image>
         <elasticsearch.docker.test.version>${elasticsearch.test.version}</elasticsearch.docker.test.version>
+        <powermock.version>1.6.1</powermock.version>
     </properties>
     <dependencies>
         <dependency>
@@ -101,6 +102,31 @@
             <artifactId>logback-classic</artifactId>
         </dependency>
         <!-- End logging backends. -->
+
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-core</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-reflect</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-module-junit4</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-api-mockito</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
     </dependencies>
     <build>
         <resources>
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -21,6 +21,7 @@ import com.google.common.collect.Iterators;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
 import org.janusgraph.diskstorage.es.compat.ES6Compat;
+import org.janusgraph.diskstorage.es.rest.util.HttpAuthTypes;
 import org.locationtech.spatial4j.shape.Rectangle;
 import org.apache.commons.lang.StringUtils;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
@@ -166,6 +167,86 @@ public class ElasticSearchIndex implements IndexProvider {
 
     public static final ConfigNamespace ES_INGEST_PIPELINES =
             new ConfigNamespace(ELASTICSEARCH_NS, "ingest-pipeline", "Ingest pipeline applicable to a store of an index.");
+
+    public static final ConfigNamespace SSL_NS =
+            new ConfigNamespace(ELASTICSEARCH_NS, "ssl", "Elasticsearch SSL configuration");
+
+    public static final ConfigOption<Boolean> SSL_ENABLED =
+            new ConfigOption<>(SSL_NS, "enabled",
+            "Controls use of the SSL connection to Elasticsearch.", ConfigOption.Type.LOCAL, false);
+
+    public static final ConfigOption<Boolean> SSL_DISABLE_HOSTNAME_VERIFICATION =
+            new ConfigOption<>(SSL_NS, "disable-hostname-verification",
+            "Disables the SSL hostname verification if set to true. Hostname verification is enabled by default.",
+            ConfigOption.Type.LOCAL, false);
+
+    public static final ConfigOption<Boolean> SSL_ALLOW_SELF_SIGNED_CERTIFICATES =
+            new ConfigOption<>(SSL_NS, "allow-self-signed-certificates",
+            "Controls the accepting of the self-signed SSL certificates.",
+            ConfigOption.Type.LOCAL, false);
+
+    public static final ConfigNamespace SSL_TRUSTSTORE_NS =
+            new ConfigNamespace(SSL_NS, "truststore", "Configuration options for SSL Truststore.");
+
+    public static final ConfigOption<String> SSL_TRUSTSTORE_LOCATION =
+            new ConfigOption<>(SSL_TRUSTSTORE_NS, "location",
+            "Marks the location of the SSL Truststore.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> SSL_TRUSTSTORE_PASSWORD =
+            new ConfigOption<>(SSL_TRUSTSTORE_NS, "password",
+            "The password to access SSL Truststore.", ConfigOption.Type.LOCAL, "", val -> val != null);
+
+    public static final ConfigNamespace SSL_KEYSTORE_NS =
+            new ConfigNamespace(SSL_NS, "keystore", "Configuration options for SSL Keystore.");
+
+    public static final ConfigOption<String> SSL_KEYSTORE_LOCATION =
+            new ConfigOption<>(SSL_KEYSTORE_NS, "location",
+            "Marks the location of the SSL Keystore.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> SSL_KEYSTORE_PASSWORD =
+            new ConfigOption<>(SSL_KEYSTORE_NS, "storepassword",
+            "The password to access SSL Keystore.", ConfigOption.Type.LOCAL, "", val -> val != null);
+
+    public static final ConfigOption<String> SSL_KEY_PASSWORD =
+            new ConfigOption<>(SSL_KEYSTORE_NS, "keypassword",
+            "The password to access the key in the SSL Keystore. If the option is not present, the value of \"storepassword\" is used.",
+            ConfigOption.Type.LOCAL, "", val -> val != null);
+
+    public static final ConfigNamespace ES_HTTP_NS =
+            new ConfigNamespace(ELASTICSEARCH_NS, "http", "Configuration options for HTTP(S) transport.");
+
+    public static final ConfigNamespace ES_HTTP_AUTH_NS =
+            new ConfigNamespace(ES_HTTP_NS, "auth", "Configuration options for HTTP(S) authentication.");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_TYPE =
+            new ConfigOption<>(ES_HTTP_AUTH_NS, "type",
+            "Authentication type to be used for HTTP(S) access.", ConfigOption.Type.LOCAL, HttpAuthTypes.NONE.toString());
+
+    public static final ConfigNamespace ES_HTTP_AUTH_BASIC_NS =
+            new ConfigNamespace(ES_HTTP_AUTH_NS, "basic", "Configuration options for HTTP(S) Basic authentication.");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_USERNAME =
+            new ConfigOption<>(ES_HTTP_AUTH_BASIC_NS, "username",
+            "Username for HTTP(S) authentication.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_PASSWORD =
+            new ConfigOption<>(ES_HTTP_AUTH_BASIC_NS, "password",
+            "Password for HTTP(S) authentication.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_REALM = new ConfigOption<>(ES_HTTP_AUTH_BASIC_NS,
+            "realm", "Realm value for HTTP(S) authentication. If empty, any realm is accepted.",
+            ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigNamespace ES_HTTP_AUTH_CUSTOM_NS =
+            new ConfigNamespace(ES_HTTP_AUTH_NS, "custom", "Configuration options for custom HTTP(S) authenticator.");
+
+    public static final ConfigOption<String> ES_HTTP_AUTHENTICATOR_CLASS = new ConfigOption<>(
+            ES_HTTP_AUTH_CUSTOM_NS, "authenticator-class", "Authenticator fully qualified class name.",
+            ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String[]> ES_HTTP_AUTHENTICATOR_ARGS = new ConfigOption<>(
+            ES_HTTP_AUTH_CUSTOM_NS, "authenticator-args", "Comma-separated custom authenticator constructor arguments.",
+            ConfigOption.Type.LOCAL, new String[0]);
 
     public static final int HOST_PORT_DEFAULT = 9200;
 
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
@@ -1,0 +1,256 @@
+// Copyright 2017 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.es.rest;
+
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INDEX_HOSTS;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INDEX_PORT;
+
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.http.HttpHost;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.elasticsearch.client.RestClient;
+import org.elasticsearch.client.RestClientBuilder;
+import org.elasticsearch.client.RestClientBuilder.HttpClientConfigCallback;
+import org.elasticsearch.client.RestClientBuilder.RequestConfigCallback;
+import org.janusgraph.diskstorage.configuration.ConfigOption;
+import org.janusgraph.diskstorage.configuration.Configuration;
+import org.janusgraph.diskstorage.es.ElasticSearchClient;
+import org.janusgraph.diskstorage.es.ElasticSearchIndex;
+import org.janusgraph.diskstorage.es.rest.util.BasicAuthHttpClientConfigCallback;
+import org.janusgraph.diskstorage.es.rest.util.HttpAuthTypes;
+import org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticator;
+import org.janusgraph.diskstorage.es.rest.util.SSLConfigurationCallback;
+import org.janusgraph.diskstorage.es.rest.util.SSLConfigurationCallback.Builder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Create an instance of Elasticsearch REST {@link org.elasticsearch.client.RestClient} from a JanusGraph
+ * {@link org.janusgraph.diskstorage.configuration.Configuration}.
+ */
+public class RestClientSetup {
+
+    private static final Logger log = LoggerFactory.getLogger(RestClientSetup.class);
+
+    public ElasticSearchClient connect(Configuration config) throws IOException {
+        log.debug("Configuring RestClient");
+
+        final List<HttpHost> hosts = new ArrayList<>();
+        final int defaultPort = config.has(INDEX_PORT) ? config.get(INDEX_PORT) : ElasticSearchIndex.HOST_PORT_DEFAULT;
+        final String httpScheme = config.get(ElasticSearchIndex.SSL_ENABLED).booleanValue() ? "https" : "http";
+        for (String host : config.get(INDEX_HOSTS)) {
+            String[] hostparts = host.split(":");
+            String hostname = hostparts[0];
+            int hostport = defaultPort;
+            if (hostparts.length == 2) hostport = Integer.parseInt(hostparts[1]);
+            log.debug("Configured remote host: {} : {}", hostname, hostport);
+            hosts.add(new HttpHost(hostname, hostport, httpScheme));
+        }
+        final RestClient rc = getRestClient(hosts.toArray(new HttpHost[hosts.size()]), config);
+
+        final int scrollKeepAlive = config.get(ElasticSearchIndex.ES_SCROLL_KEEP_ALIVE);
+        Preconditions.checkArgument(scrollKeepAlive >= 1, "Scroll keep-alive should be greater than or equal to 1");
+        final RestElasticSearchClient client = getElasticSearchClient(rc, scrollKeepAlive);
+        if (config.has(ElasticSearchIndex.BULK_REFRESH)) {
+            client.setBulkRefresh(config.get(ElasticSearchIndex.BULK_REFRESH));
+        }
+
+        return client;
+    }
+
+    protected RestClient getRestClient(HttpHost[] hosts, Configuration config) {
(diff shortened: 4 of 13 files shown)


ngrigoriev:
We will have CCLA signed early next week. Sorry, takes some time, our legal specialist was not available last week. This PR is primarily a preview.

The configuration now may look like this (example):

```
index.search.backend=elasticsearch
index.search.elasticsearch.interface=REST_CLIENT
index.search.hostname=localhost
index.search.port=443
index.search.elasticsearch.ssl.enabled=true
index.search.elasticsearch.ssl.truststore.location=/Users/nik/Documents/test-CA/MyOwnCaCATrustStore.jks
index.search.elasticsearch.ssl.truststore.password=TrustStorePwd
index.search.elasticsearch.http.auth.type=basic
index.search.elasticsearch.http.auth.basic.username=estest
index.search.elasticsearch.http.auth.basic.password=espwd
```

I have done integration testing with this configuration, could connect successfully to ES 5.6 with X-Pack over HTTPS with basic authentication.

sjudeng:
Thanks for your work on this! I'll post some comments/questions once the CLA is in place.

mbrukman:
FYI, CCLA is being added in https://github.com/JanusGraph/legal/pull/65

mbrukman:
@ngrigoriev – you're all set with the CLA for this PR! Please consider squashing your commits, such as the merge commit. Also, in the future, consider rebasing your PR on `master` rather than merging from `master` so that the PR stays cleaner.

@sjudeng – go ahead and review the PR when you get a chance, please!

ngrigoriev:
- I will add more tests now. The inner logic of BasicAuthClientConfigCallback is indirectly tested in the main test class (RestClientSetupTest) but I will make it similar: mocking the actual callback and testing it separately. This will be cleaner for sure.
- I was thinking about the system properties, but then I found that Cassandra storage backend uses the same pair of settings. Plus, I am not sure if I would dare to force the people to use the same pair of values for any instance of HttpClient used inside of JG. At the end, I am not sure if, for example, DynamoDB driver is using HttpClient. If for whatever reason people need to use non-default trust store - they are most likely doing something unusual and may need more flexibility. So, this is why I went with the separate properties for ES REST HTTPS/SSL.
- Of course. I was waiting for the initial feedback to see if I should proceed with modifying the docs or I will need to work more on this code.

sjudeng:
Thanks for the explanation on the SSL system properties, I think keeping support for explicit truststore parameters makes sense. What about adding another property (e.g. `elasticsearch.ssl.useSystemProperties` or something like that) for the system properties case? Otherwise could you add support for additional client keystore parameters? It would be nice to include support for two-way client authentication as part of this update.

I think documentation will be helpful as part of the review to see how this comes together from a user perspective (especially for the AWS auth case). I might have some minor feedback on the code but on first pass it looks really solid to me. Great work and thanks again for contributing this back to the project!

ngrigoriev:
@sjudeng For client-side options, are you referring to supporting https://www.elastic.co/guide/en/x-pack/current/pki-realm.html#_pki_realm_configuration ? Because it seems to be the only use case I am aware of.

sjudeng:
In general it's just the case where the server requires the client to present a valid certificate (e.g. `shield.http.client.auth: required`). The server may or may not authenticate users based on the content of that certificate (as described in the above link) but either way it doesn't add any additional requirements on the client side beyond presenting a valid certificate. This can be provided through `javax.net.ssl.keyStore` and `javax.net.ssl.keyStorePassword` system properties or else using [SSLContextBuilder.html#loadKeyMaterial](https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/conn/ssl/SSLContextBuilder.html#loadKeyMaterial(java.security.KeyStore,%20char[])) if you add support for the keystore parameters like is done for truststore.

ngrigoriev:
Yes, I was thinking about something like this :) Please see my updated PR. I have also added the documentation to cover all this new stuff.

ngrigoriev:
Squashed the commits into one, updated the email address.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,28 @@
...
+ * constructor. The constructor must not perform any blocking operations.
+ * </p>
+ */
+public interface RestClientAuthenticator extends HttpClientConfigCallback, RequestConfigCallback {
+
+    /**
+     * Do all your initialization (reading configuration files etc) here.
+     * @throws IOException
+     */
+    public void init() throws IOException;
```
sjudeng: It might be clearer to implement the `String[]` args in init instead of the constructor.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallback.java
```diff
@@ -0,0 +1,143 @@
...
+        this.trustStorePassword = trustStorePassword;
+
+        this.keyStoreFile = keyStoreFile;
+        this.keyStorePassword = keyStorePassword;
+        this.keyPassword = keyPassword;
+    }
+
+    @Override
+    public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
+        SSLContext sslcontext;
```
sjudeng: final

docs/appendices.txt
```diff
@@ -8,6 +8,69 @@
 Refer to the JanusGraph https://javadoc.io/doc/org.janusgraph/janusgraph-core/$MAVEN{project.version}[API documentation] for a complete documentation of all core APIs exposed by JanusGraph.
 We *strongly* encourage all users of JanusGraph to use the Gremlin query language for any queries executed on JanusGraph and to not use JanusGraph's APIs outside of the management system.
 
+[[escustomauth]]
+[appendix]
+
+== Elasticsearch Custom Authenticator Example
+
+The following code snippet implements an authenticator allowing
```
sjudeng: "allowing **the** Elasticsearch REST client ..."

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
 
 The REST client accepts the `index.[X].bulk-refresh` option. This option controls when changes are made visible to search. See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html[?refresh documentation] for more information.
 
+===== REST Client HTTPS support
+
+SSL support for HTTP can be enabled by setting the `index.[X].elasticsearch.ssl.enabled` configuration option to `true`. Note that depending on your configuration you may need to change the value of `index.[X].port` if your HTTPS port number is different from the default one for the REST API (9200).
+
+In addition to enabling SSL support, you may use non-default location for SSL Trust Store - for example if you run your own CA etc. This can be done as follows:
```
sjudeng: "When SSL is enabled you may also configure the location and password of the truststore. This can be done as follows:"

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+
+[source, properties]
+----
+index.search.elasticsearch.ssl.truststore.location=/path/to/your/truststore.jks
+index.search.elasticsearch.ssl.truststore.password=truststorepwd
+----
+
+Note that these settings apply only to Elasticsearch REST client and do not affect any other SSL connections in JanusGraph.
+
+===== SSL client authentication for HTTPS
```
sjudeng: I don't think this needs it's own section. If it were to be its own section I'd think it would be under the HTTPS section but the nesting is already pretty deep here.

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+
+Regardless of HTTP authentication settings, the client may be configured to use SSL client authentication.
```
sjudeng: "Configuration of the client keystore is also supported:"

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+
+[source, properties]
+----
+index.search.elasticsearch.http.auth.custom.authenticator-class=fully.qualified.class.Name
+index.search.elasticsearch.elasticsearch.http.auth.custom.authenticator-args=arg1,arg2,...
+----
+
+Argument list is optional and can be empty.
+
+The class specified there has to implement `org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticator` interface or extend `org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticatorBase` convenience class. The implementation gets access to HTTP client configuration and can customize the client as needed. Refer to <<javadoc>> for more information.
```
sjudeng: "implement **the** `org....`"

docs/elasticsearch.txt
```diff
@@ -95,13 +162,15 @@ See https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html[
...
 
 * The HTTP REST API, usually on port 9200
 * The native "transport" protocol, usually on port 9300
 
 A client uses either one protocol/port or the other, but not both simultaneously. Securing the HTTP protocol port is generally done with a combination of firewalling and a reverse proxy with SSL encryption and HTTP authentication. There are a couple of ways to approach security on the native "transport" protocol port:
 
+In addition to that, some hosted Elasticsearch services offer other methods of authentication and authorization. For example, AWS ES Service requires to use HTTPS and offers an option for using IAM-based access control. For that the requests sent to this service must be signed. This can be achieved by using a custom authenticator (see above).
```
sjudeng: "requires the use of HTTPS..."

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -0,0 +1,228 @@
...
+    protected RestClientBuilder getRestClientBuilder(HttpHost[] hosts) {
+        return RestClient.builder(hosts);
+    }
+
+    protected RestElasticSearchClient getElasticSearchClient(RestClient rc, int scrollKeepAlive) {
+        return new RestElasticSearchClient(rc, scrollKeepAlive);
+    }
+
+    /**
+     * Returns the callback used the default {@link RequestConfig} being set to
```
sjudeng: `"Returns the {@link RequestConfig} callback used to customize the {@link CloseableHttpClient}..."`?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.io.IOException;
+
+import org.elasticsearch.client.RestClientBuilder.HttpClientConfigCallback;
+import org.elasticsearch.client.RestClientBuilder.RequestConfigCallback;
+
+/**
+ * <p>
+ * A class instance implementing this interface may customize some aspects of
+ * JanusGraph ES REST client to implement a custom authentication method.
```
sjudeng: "JanusGraph Elasticsearch REST client..."

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,28 @@
...
+ * </p>
+ * <p>
+ * A class implementing instance must provide a constructor that accepts a String[]. This
```
sjudeng: "Implementations must provide..."

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,28 @@
...
+ * array will contain the values provided as constructor parameters in JanusGraph graph index
+ * backend configuration. If the list of parameters is empty, a zero-size array is passed to the
```
sjudeng: Replace "JanusGraph graph index backend configuration" with `"{@link ElasticSearchIndex#ES_HTTP_AUTHENTICATOR_ARGS}"`? Would need to import ElasticSearchIndex if you do that though.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,28 @@
...
+ * constructor. The constructor must not perform any blocking operations.
+ * </p>
+ */
+public interface RestClientAuthenticator extends HttpClientConfigCallback, RequestConfigCallback {
+
+    /**
+     * Do all your initialization (reading configuration files etc) here.
```
sjudeng: "Initialize the authenticator."

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticatorBase.java
```diff
@@ -0,0 +1,42 @@
...
+package org.janusgraph.diskstorage.es.rest.util;
+
+import java.io.IOException;
+
+import org.apache.http.client.config.RequestConfig.Builder;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+
+/**
+ * Convenience class for using as parent when implementing custom REST client authenticator. By default
+ * this class does not do anything.
```
sjudeng: "Base class for custom REST client authenticators."

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
 
 The REST client accepts the `index.[X].bulk-refresh` option. This option controls when changes are made visible to search. See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html[?refresh documentation] for more information.
 
+===== REST Client HTTPS support
```
sjudeng: "REST Client HTTPS Configuration". Also this is getting pretty nested ... I'd consider moving up a level (e.g. "=== REST...").

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+----
+index.search.elasticsearch.ssl.keystore.location=/path/to/your/keystore.jks
+index.search.elasticsearch.ssl.keystore.storepassword=keystorepwd
+index.search.elasticsearch.ssl.keystore.keypassword=keypwd
+----
+
+Any of the passwords can be empty.
+
+
+===== REST Client HTTP Authentication
```
sjudeng: Consider moving up a level (e.g. "=== REST..."). Likewise for the below sub-sections.

docs/elasticsearch.txt
```diff
@@ -84,6 +84,73 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+----
+index.search.elasticsearch.http.auth.custom.authenticator-class=fully.qualified.class.Name
+index.search.elasticsearch.elasticsearch.http.auth.custom.authenticator-args=arg1,arg2,...
+----
+
+Argument list is optional and can be empty.
+
+The class specified there has to implement `org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticator` interface or extend `org.janusgraph.diskstorage.es.rest.util.RestClientAuthenticatorBase` convenience class. The implementation gets access to HTTP client configuration and can customize the client as needed. Refer to <<javadoc>> for more information.
+
+You can see an example of a custom authenticator for Elasticsearch REST client in <<escustomauth>>.
```
sjudeng: I think you could put the AWS example here instead of in the appendix.

docs/appendices.txt
```diff
@@ -8,6 +8,69 @@
 Refer to the JanusGraph https://javadoc.io/doc/org.janusgraph/janusgraph-core/$MAVEN{project.version}[API documentation] for a complete documentation of all core APIs exposed by JanusGraph.
 We *strongly* encourage all users of JanusGraph to use the Gremlin query language for any queries executed on JanusGraph and to not use JanusGraph's APIs outside of the management system.
 
+[[escustomauth]]
+[appendix]
+
+== Elasticsearch Custom Authenticator Example
```
sjudeng: I think you could put this in the Elasticsearch section instead of the appendix.

ngrigoriev: I have no preferences - I thought that this sort of stuff looks better in an appendix. Moving.

docs/appendices.txt
ngrigoriev: See above

docs/elasticsearch.txt
ngrigoriev: moving

docs/elasticsearch.txt
ngrigoriev: amended

docs/elasticsearch.txt
ngrigoriev: removed

docs/elasticsearch.txt
ngrigoriev: amended

docs/elasticsearch.txt
ngrigoriev: moved

docs/elasticsearch.txt
ngrigoriev: done

docs/elasticsearch.txt
ngrigoriev: Amended. Grammarly failed me :)

janusgraph-es/pom.xml
```diff
@@ -101,6 +102,34 @@
...
         </dependency>
         <!-- End logging backends. -->
+
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-core</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
+
```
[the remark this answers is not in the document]
ngrigoriev: removed

janusgraph-es/pom.xml
```diff
@@ -101,6 +102,34 @@
...
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-reflect</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
+
```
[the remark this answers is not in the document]
ngrigoriev: removed

janusgraph-es/pom.xml
```diff
@@ -101,6 +102,34 @@
...
+		<dependency>
+			<groupId>org.powermock</groupId>
+			<artifactId>powermock-module-junit4</artifactId>
+			<version>${powermock.version}</version>
+			<scope>test</scope>
+		</dependency>
+
```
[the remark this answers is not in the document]
ngrigoriev: removed

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -107,6 +108,9 @@
     public static final ConfigNamespace ELASTICSEARCH_NS =
             new ConfigNamespace(INDEX_NS, "elasticsearch", "Elasticsearch index configuration");
 
+    public static final ConfigNamespace SSL_NS =
+            new ConfigNamespace(ELASTICSEARCH_NS, "ssl", "Elasticsearch SSL configuration");
+
```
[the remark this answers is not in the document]
ngrigoriev: done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,72 @@
...
+    public static final ConfigOption<Boolean> SSL_ENABLED =
+            new ConfigOption<Boolean>(SSL_NS, "enabled",
+            "Controls use of the SSL connection to Elasticsearch.", ConfigOption.Type.LOCAL, false);
+
+    public static final ConfigNamespace SSL_TRUSTSTORE_NS =
+            new ConfigNamespace(SSL_NS, "truststore", "Configuration options for SSL Trust Store.");
+
+    public static final ConfigOption<String> SSL_TRUSTSTORE_LOCATION =
+            new ConfigOption<String>(SSL_TRUSTSTORE_NS, "location",
+            "Marks the location of the SSL Trust Store.", ConfigOption.Type.LOCAL, "");
```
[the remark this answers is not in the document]
ngrigoriev: I have found all the different ways to refer to these...https://docs.oracle.com/cd/E19509-01/820-3503/6nf1il6er/index.html uses "TrustStore" and "Key Store", for example...I have decided to with "Truststore" to be consistent with the corresponding term in Cassandra driver (existing code).

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,72 @@
...
+
+    public static final ConfigOption<String> SSL_TRUSTSTORE_PASSWORD =
+            new ConfigOption<String>(SSL_TRUSTSTORE_NS, "password",
+            "The password to access SSL Trust Store.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigNamespace SSL_KEYSTORE_NS =
+            new ConfigNamespace(SSL_NS, "keystore", "Configuration options for SSL Key Store.");
```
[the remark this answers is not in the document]
ngrigoriev: see above

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
ngrigoriev: rephrased both get...Callback descriptions.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -0,0 +1,228 @@
...
+            boolean configureSSL = false;
+            if (config.has(ElasticSearchIndex.SSL_TRUSTSTORE_LOCATION)) {
+                sslConfCBBuilder.withTrustStore(config.get(ElasticSearchIndex.SSL_TRUSTSTORE_LOCATION),
+                        config.get(ElasticSearchIndex.SSL_TRUSTSTORE_PASSWORD));
+                configureSSL = true;
+            }
+            if (config.has(ElasticSearchIndex.SSL_KEYSTORE_LOCATION)) {
+                sslConfCBBuilder.withKeyStore(config.get(ElasticSearchIndex.SSL_KEYSTORE_LOCATION),
+                        config.get(ElasticSearchIndex.SSL_KEYSTORE_PASSWORD),
+                        config.get(ElasticSearchIndex.SSL_KEY_PASSWORD));
```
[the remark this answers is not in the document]
ngrigoriev: OK. I was looking for a recommendation or common practices for this and found this: "When you are prompted for the key password, you are given the option of letting it be the same as the keystore password." (https://docs.oracle.com/javase/tutorial/security/toolsign/step3.html). Most of the APIs, however, require a separate key password, so I am placing the logic of defaulting to the keystore password right here.
Meanwhile, I have also discovered that my code did not support the empty passwords. Fixed and added the tests.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
ngrigoriev: fixed

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
ngrigoriev: not with the fully qualified name - only the javadoc will need the class I refer to :)

ngrigoriev: updated

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
ngrigoriev: updated

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
ngrigoriev: I would prefer to keep it this way unless there are strong arguments against it. This allows the constructor to fail early if some of the values are invalid or inconsistent. And then init() may fail if the second step of the initialization fails.

In fact, I have found that my AWS authenticator may violate my own rule because some of these SDK methods might do I/O under the hood. So I have updated the sample code as well :)

sjudeng: Understood. Keeping it in the constructor makes sense.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticatorBase.java
ngrigoriev: updated

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallback.java
ngrigoriev: thanks!

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,28 @@
+package org.janusgraph.diskstorage.es.rest.util;
```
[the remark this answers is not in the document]
ngrigoriev: added

docs/elasticsearch.txt
```diff
@@ -84,6 +84,126 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+
+Basic HTTP Authentication is available regardless of the state of SSL support.  Authentication realm value is optional. If it is not provided, the credentials will match any authentication realm value.
+
+
+[source, properties]
+----
+index.search.elasticsearch.http.auth.type=basic
+index.search.elasticsearch.http.auth.basic.username=httpuser
+index.search.elasticsearch.http.auth.basic.password=httppassword
+index.search.elasticsearch.http.auth.basic.realm=myesrealm
```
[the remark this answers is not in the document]
ngrigoriev: Thanks a lot for additional testing! I was thinking about the hostname verification too. I am not a big fan of this (kind of defeats the purpose of SSL) and I prefer to test with my own CA and the certificates issued by this CA - since I have the configurable trust store :) But it is a very simple change, adding it now.

docs/elasticsearch.txt
```diff
@@ -84,6 +84,126 @@ JanusGraph only uses default values for `index-name` and `health-request-timeout
...
+----
+index.search.elasticsearch.ssl.keystore.location=/path/to/your/keystore.jks
+index.search.elasticsearch.ssl.keystore.storepassword=keystorepwd
+index.search.elasticsearch.ssl.keystore.keypassword=keypwd
+----
+
+Any of the passwords can be empty.
+
+
+==== REST Client Basic HTTP Authentication
```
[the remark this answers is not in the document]
ngrigoriev: They were nested before but I thought you wanted to avoid a.b.c.d sections so I have removed the umbrella section :) Restoring.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallback.java
```diff
@@ -0,0 +1,159 @@
...
+        this.disableHostNameVerification = disableHostNameVerification;
+    }
+
+    @Override
+    public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
+        final SSLContext sslcontext;
+        try {
+            if (StringUtils.isNotEmpty(trustStoreFile)) {
+                sslContextBuilder.loadTrustMaterial(new File(trustStoreFile), trustStorePassword.toCharArray());
+            }
```
sjudeng: I still get verification errors with `disableHostNameVerifcation=true`. To resolve I also needed a no-op trust material in this case.

```java
if (StringUtils.isNotEmpty(trustStoreFile) && !disableHostNameVerification) {
    sslContextBuilder.loadTrustMaterial(new File(trustStoreFile), trustStorePassword.toCharArray());
} else if (disableHostNameVerification) {
    sslContextBuilder.loadTrustMaterial(null, (chain,authType) -> true);
}
```

ngrigoriev: Well, then I think we are talking about different features really. Is it possible that you are talking about self-signed certificate support for testing? Because they usually require more than just disabling the host verification. I implemented only the host verification control, which is one of the steps of the SSL strategy.

I have just performed a simple test. I have ES running on a remote host. I use SSH tunnel to connect my JG to it. I have my own custom CA certificate and that server has a certificate for "localhost" signed with this CA - because I intended to connect to it via my localhost (redirected). So, for fun, I have attempted to use an alias to for one of my interfaces instead:

```
Caused by: org.janusgraph.diskstorage.PermanentBackendException: Host name 'mbp-en7' does not match the certificate subject provided by the peer (EMAILADDRESS=ngrigoriev@newforma.com, CN=localhost, OU=Engineering, O=Newforma, L=Montreal, ST=QC, C=CA)
        at org.janusgraph.diskstorage.es.ElasticSearchIndex.<init>(ElasticSearchIndex.java:307)
        ... 31 more
Caused by: javax.net.ssl.SSLPeerUnverifiedException: Host name 'mbp-en7' does not match the certificate subject provided by the peer (EMAILADDRESS=ngrigoriev@newforma.com, CN=localhost, OU=Engineering, O=Newforma, L=Montreal, ST=QC, C=CA)
        at org.apache.http.nio.conn.ssl.SSLIOSessionStrategy.verifySession(SSLIOSessionStrategy.java:208)
        at org.apache.http.nio.conn.ssl.SSLIOSessionStrategy$1.verify(SSLIOSessionStrategy.java:188)
        at org.apache.http.nio.reactor.ssl.SSLIOSession.doHandshake(SSLIOSession.java:368)
        ...
        at java.lang.Thread.run(Thread.java:745)
```

As expected - the CN mismatch. Now, if I enable my newly introduced option, it works perfectly fine without any other changes, it ignores the fact that the certificate presented for "localhost" does not match the host "mbp-en7" I am trying to connect to.

ngrigoriev: Re: PowerMock and coverage - I could not find an easy way to address it. Unfortunately, I had no choice but to use PowerMock for my tests. Some of the classes I needed to deal with, particularly HttpAsyncClientBuilder and RestClientBuilder, are not well designed for testing :( I find it problematic that they do not implement any interfaces, which would allow easy mocking. 

It seems that later versions of PowerMock offer more options - like a  java agent that allows it to coexist with JaCoCo, the current version of PowerMock that does not use Javassist or, worst case, offline instrumentation. But all this also requires newer Mockito. Probably we should create an issue to bump up the versions of the test tools for JG in general.

ngrigoriev: To avoid the arguments, I am implementing another option to enable self-signed certificates.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +163,86 @@
...
+
+    public static final ConfigOption<Boolean> SSL_DISABLE_HOSTNAME_VERIFICATION =
+            new ConfigOption<Boolean>(SSL_NS, "disable-hostname-verification",
+            "Disables the SSL hostname verification if set to true. Hostname verification is enabled by default.",
+            ConfigOption.Type.LOCAL, false);
+
+    public static final ConfigOption<Boolean> SSL_ALLOW_SELF_SIGNED_CERTIFICATES =
+            new ConfigOption<Boolean>(SSL_NS, "allow-self-signed-certificates",
+            "Controls the accepting of the self-signed SSL certificates.",
+            ConfigOption.Type.LOCAL, false);
```
sjudeng: Can we combine these last two properties into one? Maybe `rejectUnauthorized` like in the [Javascript client](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/auth-reference.html#_https_ssl).

ngrigoriev: I can probably imagine the usage of one of these options in some environments - for initial discovery or for on-going load balancing. In case of TCP-based load balancer (e.g. the one that does not terminate SSL/HTTP) I can imagine that disabling just the hostname validation while keeping the rest of the validation logic active may make sense - in some scenarios, of course.  

As long as it is only two safety switches, not 10, I personally prefer keeping them separate. As long as they are not hidden Java properties like in some other products. Each of these properties disables one particular check. To make your connections completely unprotected you need to use both.

Honestly, these days, with the free providers like letsencrypt, sslforfree etc I do not even know why one would bother setting a server with a self-signed certificate, even internally :) Especially considering that for ES you actually need to do extra work to get SSL support. Not like a product that only supports SSL and requires you to provide a valid certificate without offering plain HTTP option. SSL is no longer a luxury like 10 years ago :)

sjudeng: Okay I was thinking the implementation would be the equivalent of the [TrustAllStrategy](https://issues.apache.org/jira/browse/HTTPCLIENT-1817) (added in 4.5.4) at which point the two properties would essentially represent the same thing and could be merged. But the slightly more restrictive TrustSelfSignedStrategy should still allow most if not all cases (it worked for my test case) so I'm okay with it.

As with most SSL client implementations supporting various even unrecommended configurations allows the final decisions to be made at the user level. There's documentation in the project (thanks to you) and elsewhere with recommendations and warnings on configuration best practices. By exposing the option (or options) we're giving users access to all the knobs in SSL configuration, including those we might recommend against using.

ngrigoriev:
Thanks for your support and comments.

Commits have been squashed.

janusgraph-es/pom.xml
```diff
@@ -21,6 +21,7 @@
         <skip.es.test>${skipTests}</skip.es.test>
         <es.docker.image>docker.elastic.co/elasticsearch/elasticsearch</es.docker.image>
         <elasticsearch.docker.test.version>${elasticsearch.test.version}</elasticsearch.docker.test.version>
+        <powermock.version>1.6.1</powermock.version>
```
davidclement90: Normally for 1.8.5 Mockito you should have powermock between 1.3.9 and 1.4.9.
Either way powermock 1.6.1 is not reference as stable.
(https://github.com/powermock/powermock/wiki/Mockito)

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/rest/RestClientSetupTest.java
```diff
@@ -0,0 +1,562 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.es.rest;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.*;
```
davidclement90: Can you put exact import instead of using * import ?

ngrigoriev: Not to reopen the eternal debate...but

```
~/git/janusgraph nik$ find . -name \*.java | xargs egrep ^import | grep -v generated-sources | grep '\*' | wc -l
     400
```

I think you should probably set the rules for the coding style somewhere. And set up checkstyle plugin, or a similar one. From my personal experience, I always preferred the exact imports for the code but specifically for the unit tests and specifically for Mockito and JUnit I was making the exception.  Tests frequently change and people need new assertXYZ() or new method from Mockito - and this messes the imports on every commit. Not a huge deal, of course, they are for the compiler, not for the humans :)

janusgraph-es/pom.xml
ngrigoriev: I know, I remember I was having problems at some point with my tests and Java8.  Turned out, it was a problem with Javassist and Java8. And to solve it I needed to bump Powermock version. And going to the latest 1.6.x was causing some other horrible issues. So I have kept 1.6.1 (worked) hoping that we can revisit it once Mockito version is updated. I am going to retest with 1.4.9 now to see if I am still hitting that bug.

ngrigoriev: Yes, I hit this problem: https://github.com/powermock/powermock/issues/551. It seems until there is a new release of JUnit, it can be worked around either by downgrading JUnit to 4.11 or by bumping up PowerMock to 1.6.1 I have tried the latter approach and it worked. Which one would you prefer?

amcp:
@ngrigoriev can you upgrade the version of powermock? I suspect the scope of changing powermock version is less than junit version.

ngrigoriev:
@amcp You mean up from 1.6.1?

ngrigoriev:
@amcp I think 1.6.1 is the sweetspot. Anything lower than this fails because of that JUnit change. 1.6.5 fails the compilation - the enforcer plug-in does not like the dependency conflict in mockito-core. Upgrading Mockito would most likely address this...

ngrigoriev:
Could you please review my responses? It does not seem I can move PowerMock up or down without affecting other dependencies and I'd like to confirm if you do want to remove the wildcard imports from the unit tests. Thanks.

amcp:
I’m ok with your recommendation for *mock and junit versions

sjudeng:
@ngrigoriev Can you rebase and resolve the conflict in ElasticSearchSetup?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/BasicAuthHttpClientConfigCallback.java
```diff
@@ -0,0 +1,53 @@
...
+            authScope = new AuthScope(AuthScope.ANY_HOST, AuthScope.ANY_PORT, realm, AuthScope.ANY_SCHEME);
+        } else {
+            authScope = AuthScope.ANY;
+        }
+        credentialsProvider.setCredentials(authScope, new UsernamePasswordCredentials(username, password));
+    }
+
+    @Override
+    public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
+        // TODO: clarify about disabling the cache and preemptive authentication
```
amcp: do we have a GitHub issue for this TODO?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
```diff
@@ -0,0 +1,42 @@
...
+ * constructor. The constructor must not perform any blocking operations.
+ * </p>
+ */
+public interface RestClientAuthenticator extends HttpClientConfigCallback, RequestConfigCallback {
+
+    /**
+     * Initializes the authenticator. This method may perform the blocking I/O operations if needed.
+     * @throws IOException
+     */
+    public void init() throws IOException;
```
amcp: interfaces do not need access modifiers on methods

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallback.java
```diff
@@ -0,0 +1,176 @@
...
+
+        this.disableHostNameVerification = disableHostNameVerification;
+        this.allowSelfSignedCertificates = allowSelfSignedCertificates;
+    }
+
+    @Override
+    public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
+        final SSLContext sslcontext;
+
+        final TrustStrategy trustStrategy = (allowSelfSignedCertificates ? new TrustSelfSignedStrategy() : null);
```
amcp: why the parentheses around the ternary operation?

docs/elasticsearch.txt
```diff
@@ -95,13 +224,15 @@ See https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html[
...
 
 * The HTTP REST API, usually on port 9200
 * The native "transport" protocol, usually on port 9300
 
 A client uses either one protocol/port or the other, but not both simultaneously. Securing the HTTP protocol port is generally done with a combination of firewalling and a reverse proxy with SSL encryption and HTTP authentication. There are a couple of ways to approach security on the native "transport" protocol port:
 
+In addition to that, some hosted Elasticsearch services offer other methods of authentication and authorization. For example, AWS ES Service requires the use of HTTPS and offers an option for using IAM-based access control. For that the requests sent to this service must be signed. This can be achieved by using a custom authenticator (see above).
```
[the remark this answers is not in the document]
ngrigoriev: OK, fixed that and 7 other occurrences of "ES" in the same document.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
 
+    public static final ConfigOption<Boolean> SSL_ENABLED =
+            new ConfigOption<Boolean>(SSL_NS, "enabled",
+            "Controls use of the SSL connection to Elasticsearch.", ConfigOption.Type.LOCAL, false);
+
+    public static final ConfigNamespace SSL_TRUSTSTORE_NS =
+            new ConfigNamespace(SSL_NS, "truststore", "Configuration options for SSL Truststore.");
+
+    public static final ConfigOption<String> SSL_TRUSTSTORE_LOCATION =
+            new ConfigOption<String>(SSL_TRUSTSTORE_NS, "location",
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            "Marks the location of the SSL Truststore.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> SSL_TRUSTSTORE_PASSWORD =
+            new ConfigOption<String>(SSL_TRUSTSTORE_NS, "password",
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            "The password to access SSL Truststore.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigNamespace ES_HTTP_NS =
+            new ConfigNamespace(ELASTICSEARCH_NS, "http", "Configuration options for HTTP(S) transport.");
+
+    public static final ConfigNamespace ES_HTTP_AUTH_NS =
+            new ConfigNamespace(ES_HTTP_NS, "auth", "Configuration options for HTTP(S) authentication.");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_TYPE =
+            new ConfigOption<String>(ES_HTTP_AUTH_NS, "type",
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            "Authentication type to be used for HTTP(S) access.", ConfigOption.Type.LOCAL, HttpAuthTypes.NONE.toString());
+
+    public static final ConfigNamespace ES_HTTP_AUTH_BASIC_NS =
+            new ConfigNamespace(ES_HTTP_AUTH_NS, "basic", "Configuration options for HTTP(S) Basic authentication.");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_USERNAME =
+            new ConfigOption<String>(ES_HTTP_AUTH_BASIC_NS, "username",
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            "Username for HTTP(S) authentication.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_PASSWORD =
+            new ConfigOption<String>(ES_HTTP_AUTH_BASIC_NS, "password",
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            "Password for HTTP(S) authentication.", ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String> ES_HTTP_AUTH_REALM = new ConfigOption<String>(ES_HTTP_AUTH_BASIC_NS,
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            "realm", "Realm value for HTTP(S) authentication. If empty, any realm is accepted.",
+            ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigNamespace ES_HTTP_AUTH_CUSTOM_NS =
+            new ConfigNamespace(ES_HTTP_AUTH_NS, "custom", "Configuration options for custom HTTP(S) authenticator.");
+
+    public static final ConfigOption<String> ES_HTTP_AUTHENTICATOR_CLASS = new ConfigOption<String>(
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -162,6 +166,57 @@
...
+            ES_HTTP_AUTH_CUSTOM_NS, "authenticator-class", "Authenticator fully qualified class name.",
+            ConfigOption.Type.LOCAL, "");
+
+    public static final ConfigOption<String[]> ES_HTTP_AUTHENTICATOR_ARGS = new ConfigOption<String[]>(
```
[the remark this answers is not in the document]
ngrigoriev: Updated. Copy-pasted from another initializer and then replicated the bad pattern.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -0,0 +1,215 @@
...
+            String hostname = hostparts[0];
+            int hostport = defaultPort;
+            if (hostparts.length == 2) hostport = Integer.parseInt(hostparts[1]);
+            log.debug("Configured remote host: {} : {}", hostname, hostport);
+            hosts.add(new HttpHost(hostname, hostport, httpScheme));
+        }
+        RestClient rc = getRestClient(hosts.toArray(new HttpHost[hosts.size()]), config);
+
+        final int scrollKeepAlive = config.get(ElasticSearchIndex.ES_SCROLL_KEEP_ALIVE);
+        Preconditions.checkArgument(scrollKeepAlive >= 1, "Scroll Keep alive should be greater or equals than 1");
```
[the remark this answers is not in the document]
ngrigoriev: Actually, this is not the part of my commit (moved the code from another class as-is). Fixing the message while I am there :)

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -0,0 +1,252 @@
...
+     * </p>
+     * <p>
+     * See {@link RestClientBuilder#setRequestConfigCallback(RequestConfigCallback) for more details.
+     * </p>
+     *
+     * @param config
+     *            ES index configuration
+     * @return callback or null if the request customization is not needed
+     */
+    protected RequestConfigCallback getRequestConfigCallback(Configuration config) {
```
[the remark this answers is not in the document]
ngrigoriev: This method is intended for internal use but made protected for testability - so the class can be mocked/spied on without the nasty things like PowerMock. Or probably I do not understand this comment...

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -0,0 +1,252 @@
...
+                sslConfCBBuilder.allowSelfSignedCertificates();
+                configureSSL = true;
+            }
+
+            if (configureSSL) {
+                callbackList.add(sslConfCBBuilder.build());
+            }
+        }
+
+        if (callbackList.size() > 0) {
```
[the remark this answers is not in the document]
ngrigoriev: Updated

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -0,0 +1,252 @@
...
+                    authClassName, authClassConstructurArgList, e);
+            throw new RuntimeException(e);
+        }
+
+        try {
+            authenticator.init();
+        } catch (IOException e) {
+            log.error("Unable to instantiate the custom authenticator {} with constructor arguments \"{}\"",
+                    authClassName, authClassConstructurArgList, e);
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
ngrigoriev: Updated with a simple message. In general, I do not know which pattern is better - I was searching through the existing JG code and found all 3 possible cases.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/BasicAuthHttpClientConfigCallback.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public BasicAuthHttpClientConfigCallback(final String realm, final String username, final String password) {
+        Preconditions.checkArgument(StringUtils.isNotEmpty(username), "HTTP Basic Authentication: username must be provided");
+        Preconditions.checkArgument(StringUtils.isNotEmpty(password), "HTTP Basic Authentication: password must be provided");
+
+        credentialsProvider = new BasicCredentialsProvider();
+
+        final AuthScope authScope;
+        if (StringUtils.isNotEmpty(realm)) {
+            authScope = new AuthScope(AuthScope.ANY_HOST, AuthScope.ANY_PORT, realm, AuthScope.ANY_SCHEME);
```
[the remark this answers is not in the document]
ngrigoriev: Because the cluster is discovered by connecting to one node. The client may (and should not) know about all the nodes in the cluster. As well as the port numbers, I believe (that I am not 100% sure about). So the client cannot have a list of hosts at this point. Even configurable realm here is probably an overkill but I thought it may be useful in some cases as the only common "filter" for the nodes in the cluster.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/RestClientAuthenticator.java
ngrigoriev: Fixed, thanks.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallback.java
ngrigoriev: Looks like without a good reason, thanks!

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/util/BasicAuthHttpClientConfigCallback.java
ngrigoriev: Thanks, I actually missed this one. Took a note for myself, I was planning to find out if it is worth implementing preemptive authentication or not.  Pros: it would be faster on the initial request, cons - potential leak of credentials if misconfigured. After reviewing HttpClient code I have concluded that RequestAuthCache (which is NOT disabled by default) will do just fine for the subsequent requests. So, removing this TODO.

ngrigoriev:
Addressed the review comments, will rebase, test and squash the commits again today.

ngrigoriev:
Squashed, rebased, ready to be merged once the CI build passes.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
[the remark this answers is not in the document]
amcp: @ngrigoriev you could use the `@VisibleForTesting` annotation

Issue #648, opened 2017-10-18
Excessive of array resizing in MultiKeySliceQuery.java


pencal:
Result place holder is set to size of 4 in MultiKeySliceQuery.java:56. For a query with large number of results being returned, i.e. tx.query().has("customId", Contain.IN, customIds10000), it will incur an excessive amount of array resizing. Proposing to fix it by switching it to a LinkedList or an ArrayList with initial size of Math.min(getLimit(), queries.size())

robertdale:
I think we should stick with ArrayList unless we're absolutely certain that no downstream usage requires positional index. e.g. range(x,y)

pencal:
submitted PR: https://github.com/JanusGraph/janusgraph/pull/668
trying to figure out why I get the cla:no at the moment.

robertdale:
@pencal Please stop creating new PRs. You can `git push -f` on an existing one.   Looks like you're not listed as a contributor - https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTORS.txt - or you are not using the email address that is listed.

PR #682, opened 2017-10-31
Issue 648 - avoid excessive array list resizing 

pencal:
Issue: #648 

Signed-off-by: Kim Pang Lei <ckplei@gmail.com>

Thank you for contributing to JanusGraph.

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/MultiKeySliceQuery.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/MultiKeySliceQuery.java
@@ -15,12 +15,12 @@
 package org.janusgraph.graphdb.query.graph;
 
 import com.google.common.base.Preconditions;
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.janusgraph.diskstorage.BackendTransaction;
 import org.janusgraph.diskstorage.EntryList;
 import org.janusgraph.diskstorage.keycolumnvalue.KeySliceQuery;
 import org.janusgraph.graphdb.query.BackendQuery;
 import org.janusgraph.graphdb.query.BaseQuery;
-import org.apache.commons.lang.builder.HashCodeBuilder;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -46,7 +46,7 @@ public class MultiKeySliceQuery extends BaseQuery implements BackendQuery<MultiK
 
     public List<EntryList> execute(final BackendTransaction tx) {
         int total = 0;
-        List<EntryList> result = new ArrayList<EntryList>(4);
+        final List<EntryList> result = new ArrayList<>(Math.min(getLimit(), queries.size()));
         for (KeySliceQuery ksq : queries) {
             EntryList next =tx.indexQuery(ksq.updateLimit(getLimit()-total));
             result.add(next);
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/query/QueryTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/query/QueryTest.java
@@ -14,7 +14,9 @@
 
 package org.janusgraph.graphdb.query;
 
+import com.google.common.collect.Iterators;
 import org.janusgraph.core.*;
+import org.janusgraph.core.attribute.Contain;
 import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
 import org.janusgraph.diskstorage.keycolumnvalue.inmemory.InMemoryStoreManager;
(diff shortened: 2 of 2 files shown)


pluradj:
@pencal To get the CLA checker to verify you correctly, you need to update your commit so that the `Author:` field matches your name and email as on the CLA. Currently it is:

```
Author: pencal <ckplei@gmail.com>
```

but it should be

```
Author: Kim Pang Lei <ckplei@gmail.com>
```

https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla

pencal:
@pluradj  thanks. I submitted the PR initially using my github user name instead of my name. Since then I have fixed it but I didn't commit with --reset-author. Let me give that a try.

pluradj:
latest commit looks correct to me, might need to pull in @mbrukman if the CLA checker doesn't clear up

pencal:
thanks @pluradj  I wasn't sure if the check will run for each push. But looks like it would according to the conversation I have in other threads. Or may be someone does have to clear manually first?

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/MultiKeySliceQuery.java
```diff
@@ -46,7 +46,7 @@ public MultiKeySliceQuery updateLimit(int newLimit) {
 
     public List<EntryList> execute(final BackendTransaction tx) {
         int total = 0;
-        List<EntryList> result = new ArrayList<EntryList>(4);
+        List<EntryList> result = new ArrayList(Math.min(getLimit(), queries.size()));
```
robertdale: Should be typed. e.g. `new ArrayList<>`

robertdale: And `final`.

mbrukman:
The issue is that there's a difference in GitHub in terms of who the "author" and "committer" are (as they could be different people); the CLA checker is very simple: it verifies that both of these are each on the CLA verbatim; in this case, they are:

```
author: Kim Pang Lei <ckplei@gmail.com>, GitHub: pencal
committer: pencal <ckplei@gmail.com>, GitHub: pencal
```

which means the first one is OK, and the second one is not. You can see this yourself for the most-recent commit in your git tree via:

```
$ git show --format=full
```

or for the full repo via:

```
$ git log --format=full
```

Note the `Author:` and `Commit:` lines.

In addition, your `Signed-off-by` line is non-standard, because it doesn't include `<` and `>` around the email address. This suggests you did not follow the commands verbatim as suggested in the documentation, because the `-s` flag to `git commit` adds the `<` and `>`.

You need to copy-paste and run these commands as-is, verbatim, as per [the docs](https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla), here I've adjusted them for you to make it easier:

```
$ git config user.name "My Name"
$ git config user.email "my-email@example.com"
$ git commit --amend -s --reset-author
$ git push -f
```

And possibly remove / adjust any config that could be overriding this. Note that the repo-local config should be overriding your global config in this case, so be sure to set both name and email in the repo to take precedence.

If you're still having an issue, we can look into it further, but we haven't had this happen and it was fixed after using these commands, so I'm hopeful this should solve the problem for you.

mbrukman:
@pencal – congrats! You now have a green CLA tag on this PR. Please keep doing what you did in this PR going forward, this is the correct course of action.

pencal:
@sjudeng  the added test is more of a regression. That said, I shall add a test case to test out the case where the limit is accounted for.

robertdale:
@pencal Can you rebase this on 0.2 and then merge to master on a new PR?

pencal:
@robertdale  why is this necessary? This PR is already merging to PR and my branch was based off Master right after 0.2 was cut.

robertdale:
@pencal We would like to target the lowest maintenance branch and then merge up. You should only need to change the base of this PR to 0.2 [1]. I can handle the merge.  Also, understand that JanusGraph 0.2 will be on TinkerPop 3.2.x while JanusGraph 0.3 will be on TinkerPop 3.3.x which are somewhat incompatible and could have different release timelines.  

1. https://help.github.com/articles/changing-the-base-branch-of-a-pull-request/

pencal:
I see. I will rebase, close this PR and open a new one as suggested. Thanks for giving me the context.

robertdale:
@pencal No need to close. Just change the base branch as shown in that link!

pencal:
@robertdale  done.

Issue #337, opened 2017-06-12
Publish Docker Images in Post Build Step


chrislbs:
Now that a docker image build exists, it would be useful to publish them to docker hub. 

This will require the creation of a docker hub account and then either encrypting those credentials in the travis.yml file or adding those environment variables through the travis project itself. If that information can be updated, I can add the necessary changes to the travis.yml file and pom to handle the publishing.

amcp:
I've squatted for janusgraph organization in my dockerhub account a while back. How to proceed

chrislbs:
@amcp We'll need whoever has access to the travis ci build to add some environment variables that specify the docker hub account credentials. I'm working on a more thorough docker image that supports multiple persistence configurations. I can likely update the travis yml to support publishing but it will still require someone with access to the travis build to do some updates.

FlorianHockmann:
Superseded by janusgraph/janusgraph-docker#9.

PR #700, opened 2017-11-06
Updating JanusGraph Docker Image Build For #337

chrislbs:
Updating JanusGraph Docker Image Build for #337 

- Utilizing spotify maven dockerfile plugin to replace deprecated plugin

- Adding environment variable configurations to enable multiple storage
and indexing backends

- Updating README with examples of running the container

Signed-off-by: Chris Pounds <chris.pounds@experoinc.com>


mbrukman:
@chrislbs – you probably want to add a reference to the issue # in your original message, so that it gets cross-linked. References to the issue in the subject do not get cross-linked.

janusgraph-dist/README.md
```diff
@@ -68,6 +68,178 @@ To build the .deb and .rpm packages:
...
+First, build the zip archive from the previous section. Next run the command
+`mvn dockerfile:build -Pjanusgraph-docker -pl janusgraph-dist` from the root directory of this 
+repository. This command will build a JanusGraph docker image tagged with the current project 
+version. If you want to specify a different tag, you can run the command as
+`mvn dockerfile:build -Ddocker.tag=latest -Pjanusgraph-docker -pl janusgraph-dist`.
+
+Running the Docker Image
+------------------------
+
+By default, the docker image is built using JanusGraph's [in memory](http://docs.janusgraph.org/latest/inmemorystorage.html)
```
sjudeng: I think the default should be Cassandra+Elasticsearch

FlorianHockmann:
Any updates on this? This would be really helpful.

janusgraph-dist/src/docker/gremlin-server.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - { 
+      className: org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0,
+      config: { ioRegistries: [org.janusgraph.graphdb.tinkerpop.JanusGraphIoRegistry] }
+    }
+  - { 
+      className: org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV2d0,
+      config: { ioRegistries: [org.janusgraph.graphdb.tinkerpop.JanusGraphIoRegistry] }
+    }
+processors:
+  - {
```
farodin91: I added to this config
````yaml
  - { 
      className: org.apache.tinkerpop.gremlin.server.op.traversal.TraversalOpProcessor, 
      config: { cacheExpirationTime: 600000, cacheMaxSize: 1000 }
    }
```
Before, I can't add Vertex to the graph.

amcp:
@chrislbs please rebase.

chrislbs:
@amcp I'll rebase and ensure it's working with current develompent. As well as address open issues.

janusgraph-dist/README.md
```diff
@@ -68,6 +68,178 @@ To build the .deb and .rpm packages:
...
+storage engine. You can run an instance of JanusGraph with the command
+`docker run -p8182:8182 experoinc/janusgraph:latest`. It also supports integration with other
```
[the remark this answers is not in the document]
lucamaraschi: I think distributing an official image it's crucial for the developer experience and the growth of JanusGraph. @sjudeng and @twilmes can you ping me offline please?

tamethecomplex:
Hi @chrislbs, I just pulled your branch and tried to follow the build instructions [here](https://github.com/chrislbs/janusgraph/blob/comprehensive-docker-image/janusgraph-dist/README.md), but ran into a roadblock.  Specifically:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project janusgraph-codepipelines-ci: Fatal error compiling: java.lang.ExceptionInInitializerError: com.sun.tools.javac.code.TypeTags

I'm new to janusgraph and the build toolchain in general....  I was hoping to do a quick pull from dockerhub and get a container running just to try out the system.  Question for you...  Is there a way you could publish a docker image to dockerhub for use by the community, maybe under your user ID, while this pull request is still in progress?

tamethecomplex:
Hi @chrislbs , please disregard the above.  I see (by looking at the build README in your branch) there's already a version pushed to https://hub.docker.com/r/experoinc/janusgraph/tags/, although a little older.  Also, please disregard my comment about the build error - this was a java config error.  I'm able to build janusgraph and the docker image fine now using the instructions in your README.

chrislbs:
If anyone comes across this branch I created a dedicated repository here https://github.com/experoinc/janusgraph-images to host images for different versions / platforms at some point in the future. I believe there is still some desire to have a docker build within the source repository as well for local testing purposes.

FlorianHockmann:
> I believe there is still some desire to have a docker build within the source repository as well for local testing purposes.

Definitely! I would suggest that we publish something like CI or nightly builds of JanusGraph to Docker Hub so that the new language drivers can use these images to always test against the latest version of JanusGraph. That way we could quickly notice when a change in JanusGraph breaks one of the drivers.

janusgraph-dist/README.md
chrislbs: @sjudeng I'm working on taking the implementation I moved outside of this repository to incorporate it in this PR. 

Here are my thoughts on why I had the initial implementation as in memory although I'm actually planning on changing it to be berkleydb + lucene.

Docker images are supposed to start a single process only per their best practices. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/. We could use cassandra embedded but we would still need an elastic search instance to be running.

When I consume other database implementations docker images, I'm normally using them as some sort of integration layer for testing my applications against them. In that context, I normally just desire that the interaction with the database behaves as expected rather than it being representative of a potential production deployment. I'm looking for speed rather than a true end to end representation. I think this also rings true for anyone who is just looking to "play" with JanusGraph.

I also might use those docker images to persist data during my day to day application development without having to have an actual implementation installed on my machine. For instance, I have one docker compose file that I use for my fast integration tests (in memory) and another that I use for my typical development where I want to interact with data I've accumulated during development that persists between restarts (berkley db + lucene). We can accomplish the persistence by mounting a directory as a data volume that will store the data between restarts of the containers themselves.

Example:

```yml
version: "3"

services:
  janusgraph:
    image: experoinc/janusgraph:0.2.0
    container_name: janusgraph-mount
    ports:
      - "8182:8182"
    volumes:
      # named volume mounts
      - "janusgraph-mount-data:/var/lib/janusgraph"
      # bind mounts for configs; use read only so not overridden by environment variables
      - "./example-config:/etc/opt/janusgraph/:ro"

# use a named volume to maintain database data between restarts
volumes:
  janusgraph-mount-data:
```

We can provide docker-compose files that represent a more complex deployment using Cassandra and Elasticsearch but I don't think that should be the initial implementation of the image.

https://github.com/experoinc/janusgraph-images/blob/master/docker-compose-cassandra-es.yml

Let me know what you think.

chrislbs:
@sjudeng Can you take a look at how I'm handling the configuration of the server.yml and graph.properties here?

https://github.com/experoinc/janusgraph-images/blob/master/0.2.0/docker-entrypoint.sh

Would providing a properties file and yml specifically for the docker image in ./janusgraph-dist/src/assembly be sufficient?

chrislbs:
I was unaware of the new repository https://github.com/JanusGraph/janusgraph-docker that @twilmes opened a PR for with the work I had done in the other repository. I think once the work there is complete, we can revisit this PR as to what makes it into the image that exists in the source repository.

janusgraph-dist/README.md
sjudeng: @chrislbs Thanks for the explanation. You're right the image should be functional without additional (Cassandra, Elasticsearch, etc.) containers. Only other thing would be supporting data persistence but I think you're on the right track with berkleydb+lucene.

farodin91:
@sjudeng Should we close this PR due to the offical docker repo?

sjudeng:
In case anyone is here looking for JanusGraph Docker images please see the separate repo: https://github.com/JanusGraph/janusgraph-docker.

Issue #716, opened 2017-11-15
Sync EvictionTrigger's openInstances on schedule


dpitera:
Any job to REGISTER an index requires
acknowledgement from every server on which the graph was open at the
time of index creation, however the index state change would never
fulfill should, for example, the DBA removed a JG node from the cluster after the index creation but  before the move to REGISTER completed.

The EvictionTrigger should not take into account newly
added JG nodes because these nodes would never receive the message that
the index was is being registered and thus could never send an
acknowledgement message.

We should also add better configuration options around managing graph servers.

PR #717, opened 2017-11-15
Sync EvictionTrigger's openInstances on schedule

dpitera:
Without this change, any job to REGISTER an index requires
acknowledgement from every server on which the graph was open at the
time of index creation, however the index state change would never
fulfill should, for example, the DBA removed a JG node from the cluster.

Note that we do not make the EvictionTrigger take into account newly
added JG nodes because these nodes would never receive the message that
the index was is being registered and thus could never send an
acknowledgement message.

Issues: #716

Create ConfigOption REPLACE_INSTANCE_IF_EXISTS

If a user supplies their instance name (or suffix) at graph open time,
as well as use this new configuration option with a value of true, then
a graph trying to open with the same instance-id as one previously
opened will still open.

Issues: #716

ConfigOption to set graph-instance to hostname

This is useful if you want to track JG nodes by their hostnames, with
the option of appending suffix values to the end of the hostname,
whereas previously the suffix option is appended to a hex encoded local address.

This is particularly useful if you are trying to set a graph's instance
name based on the server the graph is open on, while managing a cluster
through the ConfigurationManagementGraph APIs).

Issues: #716

Thank you for contributing to JanusGraph.

### For all changes:

### For code changes:

### Note:


janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -1580,27 +1590,49 @@ public boolean apply(@Nullable Map.Entry<ConfigElement.PathIdentifier, Object> e
...
-        } catch (UnknownHostException e) {
-            throw new JanusGraphConfigurationException("Cannot determine local host", e);
+    private static String getUid(Configuration config) {
+        final String uid;
+        if (config.has(GraphDatabaseConfiguration.UNIQUE_INSTANCE_ID_HOSTNAME)
+          && config.get(GraphDatabaseConfiguration.UNIQUE_INSTANCE_ID_HOSTNAME)) {
+            try {
+                uid = Inet4Address.getLocalHost().getHostName();
+            } catch (UnknownHostException e) {
+                throw new JanusGraphConfigurationException("Cannot determine local host", e);
```
amcp: reuse the exception message as a string constant

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
```diff
@@ -122,38 +124,70 @@ public void sendCacheEviction(Set<JanusGraphSchemaVertex> updatedTypes,
...
                 }
             }
+            log.info("Received all acknowledgements for eviction [{}]",evictionId);
+            evictionTriggerMap.remove(evictionId,this);
+        }
+
+        int removeDroppedInstances() {
+            final JanusGraphManagement mgmt = graph.openManagement();
+            final Set<String> updatedInstances = ((ManagementSystem) mgmt).getOpenInstancesInternal();
+            instancesToBeAcknowledged.forEach(it -> {
```
amcp: instancesToBeAcknowledged.stream().filter(it -> !updatedInstances.contains(it)).filter(instancesToBeAcknowledged::remove).forEach(it -> log.debug())

dpitera:
Thanks @amcp I updated according to ur feedback and pushed new code.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/log/kcvs/KCVSLog.java
```diff
@@ -645,6 +651,15 @@ public synchronized boolean unregisterReader(MessageReader reader) {
         return this.readers.remove(reader);
     }
 
+    private class MessageReaderStateUpdater implements Runnable {
+        @Override
+        public void run() {
+                for (MessageReader reader : readers) {
```
amcp: indentation is off here

dpitera: fixed and pushed, thanks

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
```diff
@@ -157,8 +160,12 @@ public StandardJanusGraph(GraphDatabaseConfiguration configuration) {
         //Register instance and ensure uniqueness
         String uniqueInstanceId = configuration.getUniqueGraphId();
         ModifiableConfiguration globalConfig = GraphDatabaseConfiguration.getGlobalSystemConfig(backend);
-        if (globalConfig.has(REGISTRATION_TIME, uniqueInstanceId)) {
+        boolean instanceExists = globalConfig.has(REGISTRATION_TIME, uniqueInstanceId);
+        boolean replaceExistingInstance = configuration.getConfiguration().get(REPLACE_INSTANCE_IF_EXISTS);
```
pluradj: add `final` on both booleans

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
```diff
@@ -157,8 +160,12 @@ public StandardJanusGraph(GraphDatabaseConfiguration configuration) {
...
+        if (instanceExists && !replaceExistingInstance) {
```
pluradj: based on the config description for `REPLACE_INSTANCE_IF_EXISTS` (_this should be only be a concern..._), you'd need to check that either `UNIQUE_INSTANCE_ID_SUFFIX` or `UNIQUE_INSTANCE_ID` is also set.

janusgraph-test/src/main/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -0,0 +1,80 @@
...
+        assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
+        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
+
+        thrown.expect(JanusGraphException.class);
+        final String err = "A JanusGraph graph with the same instance id [not-unique] is already open. Might required forced shutdown.";
+        thrown.expectMessage(equalTo(err));
+        final StandardJanusGraph graph2 = new StandardJanusGraph(new GraphDatabaseConfiguration(new CommonsConfiguration(config)));
+
+        graph1.close();
+    }
```
pluradj: need a test to cover the new config option `UNIQUE_INSTANCE_ID_HOSTNAME `

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -1764,6 +1764,67 @@ public void testIndexUpdateSyncWithMultipleInstances() throws InterruptedExcepti
 
     }
 
+    @Category({BrittleTests.class})
+    @Test
+    public void testIndexSHouldRegisterWhenWeRemoveAnInstance() throws InterruptedException {
```
pluradj: lowercase `h` in `SHould`

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -1764,6 +1764,67 @@ public void testIndexUpdateSyncWithMultipleInstances() throws InterruptedExcepti
...
+        newTx();
+        tx.addVertex("name", "v3");
+        tx.commit();
+
+        Thread.sleep(2000); //Wait for the index to register in graph2
+        finishSchema();
+        try {
+            mgmt.updateIndex(mgmt.getGraphIndex("theIndex"), SchemaAction.ENABLE_INDEX);
+            fail(); //Open tx2 should not make this possible
+        } catch (IllegalArgumentException e) {
```
pluradj: I'm trying to follow this... the attempt to move `theIndex` to ENABLED throws an exception because the transaction is still open. That makes sense, but is `theIndex` in REGISTERED status at this point, or is it INSTALLED status? The comment below on line 1807 (close second graph instance, so index can move to REGISTERED) is adding to my confusion since I would think that it is already REGISTERED based on the comment on line 1798 (Wait for the index to register in graph2). https://github.com/JanusGraph/janusgraph/wiki/Indexing

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
dpitera: I think I should remove that sentence from the ConfigurationOption description; I was merely making a testament to the fact that if you let JanusGraph set the id attribute, you will not run into overlapping ids, but if you start meddling with the graph instance id, then we may run into overlapping ids.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
dpitera: I think this test was failing for one of the backends; if I remember correctly, the backend returning the `graph.getOpenInstances()` wasn't returning "graph2" yet, and we needed to allow it time to catch up (otherwise the index will indeed move to REGISTERED and the index will not throw an exception when we try to move it to ENABLED). So this comment is _not_ a testament to the index status rather the state of the system's open instances. 

I am going to try removing the line for now and see what happens.

dpitera: removed the line and tests pass so all is good in the world

PR #806, opened 2017-12-07
Fix #777: Allow external indexing of properties with spaces

### LINKED ISSUE #777: Property name with spaces cannot be indexed in ElasticSearch
If have a property with a space in the name, I cannot use that property in an ElasticSearch-backed mixed index.

The issue is down to this code in ElasticSearchIndex.java:

```
    @Override
    public String mapKey2Field(String key, KeyInformation information) {
        Preconditions.checkArgument(!StringUtils.containsAny(key,new char[]{' '}),"Invalid key name provided: %s",key);
        return key;
    }
```

Instead of mapping the key to a legal field name, this method simply rejects any property key with a space. IMO this is completely unacceptable because it's easy to be stuck in my case, where you have a bunch of existing property keys with spaces and want to add a new ES index on them - and there's no indication anywhere else in the system that properties with spaces might be bad.

My suggested solution is to replace all spaces with an obscure unicode character that isn't ES-prohibited and is extremely unlikely to lead to collisions with other property keys (simply removing spaces would lead to collisions between 'My Key' and 'MyKey', but 'My•Key' (Unicode+2022 bullet) should be safe enough. I might be able to whip up a PR for this soon.

sharpau:
Fixes #777 

Replace spaces in property names with an obscure unicode character in mapKey2Field() so that external indexing backends (Lucene, ElasticSearch, Solr) that disallow spaces in field names can still index those properties.

Added unit tests for all 3 backends and tested with a production ES index.

Also added another commit to update AUTHORS.txt and CONTRIBUTORS.txt to reflect this PR and 260e6fc7bb54f818a10f4c34938a512ebfaf22cc.

I'm targeting the 0.2 branch because for our use of JanusGraph this is a blocking bug, and so it would be quite nice to have it in an 0.2.1 release so we don't have to keep using our own snapshot build.

Since this was my first attempt at contributing, I'll note a couple interesting things I discovered:

- The ES tests (at least) don't run correctly on Windows. I tried fixing this (detecting OS and using elasticsearch.bat instead, taskkill instead of kill) but wasn't able to get it working consistently so just used an Ubuntu VM.
- On that Ubuntu VM, the BerkeleyElasticsearchTest failed when doing 'mvn clean install in janusgraph-es project. Not sure why, but maybe there's some non-obvious dependency that maven didn't pull down?
- Importing the JanusGraph pomfile into IntelliJ worked flawlessly 👍 

PR Template checklist:

### For all changes:
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)? *As noted above, 0.2 rather than master.*
- [ ] Is your initial contribution a single, squashed commit? *No, but I think the commits make sense to be separated.*

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? *N/A*
- [x] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository? *N/A*
- [x] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository? *N/A*

### DIFF
--- a/docs/directindex.adoc
+++ b/docs/directindex.adoc
@@ -59,6 +59,10 @@ Names of property keys that contain non-alphanumeric characters must be placed i
 [source, java]
 graph.indexQuery("vertexByText", "v.\"first_name\":john").vertices()
 
+Some property key names may be transformed by the JanusGraph indexing backend implementation. For instance, an indexing backend that does not permit spaces in field names may transform "My Field Name" to "My•Field•Name", or an indexing backend like Solr may append type information to the name, transforming "myBooleanField" to "myBooleanField_b". These transformations happen in the index backend's implementation of IndexProvider, in the "mapKey2Field" method. Indexing backends may reserve special characters (such as '•') and prohibit indexing of fields that contain them. For this reason it is recommended to avoid spaces and special characters in property names.
+
+In general, making direct index queries depends on implementation details of JanusGraph indexing backends that are normally hidden from users, so it's best to verify a query empirically against the indexing backend in use.
+
 ==== Element Identifier Collision
 
 The strings "v.", "e.", and "p." are used to identify a vertex, edge or property element respectively in a query. If the field name or the query value contains the same sequence of characters, this can cause a collision in the query string and parsing errors as in the following example:
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexProvider.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexProvider.java
@@ -14,6 +14,7 @@
 
 package org.janusgraph.diskstorage.indexing;
 
+import org.apache.commons.lang.StringUtils;
 import org.janusgraph.diskstorage.BackendException;
 import org.janusgraph.diskstorage.BaseTransaction;
 import org.janusgraph.diskstorage.BaseTransactionConfig;
@@ -23,6 +24,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
 
+import com.google.common.base.Preconditions;
+
 /**
  * External index for querying.
  * An index can contain an arbitrary number of index stores which are updated and queried separately.
@@ -31,6 +34,16 @@ import java.util.stream.Stream;
  */
 
 public interface IndexProvider extends IndexInformation {
+    /*
+     * An obscure unicode character (•) provided as a convenience for implementations of {@link #mapKey2Field}, for
+     * instance to replace spaces in property keys. See #777.
+     */
+    char REPLACEMENT_CHAR = '\u2022';
+
+    static void checkKeyValidity(String key) {
+        Preconditions.checkArgument(!StringUtils.containsAny(key, new char[]{ IndexProvider.REPLACEMENT_CHAR }),
+            "Invalid key name containing reserved character %c provided: %s", IndexProvider.REPLACEMENT_CHAR, key);
+    }
 
     /**
      * This method registers a new key for the specified index store with the given data type. This allows the IndexProvider
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -22,7 +22,6 @@ import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
 import org.janusgraph.diskstorage.es.compat.ES6Compat;
 import org.locationtech.spatial4j.shape.Rectangle;
-import org.apache.commons.lang.StringUtils;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectWriter;
 import org.apache.tinkerpop.shaded.jackson.databind.SerializationFeature;
@@ -70,7 +69,6 @@ import org.janusgraph.graphdb.query.condition.Not;
 import org.janusgraph.graphdb.query.condition.Or;
 import org.janusgraph.graphdb.query.condition.PredicateCondition;
 import org.janusgraph.graphdb.types.ParameterType;
-import org.javatuples.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -1054,8 +1052,8 @@ public class ElasticSearchIndex implements IndexProvider {
 
     @Override
     public String mapKey2Field(String key, KeyInformation information) {
-        Preconditions.checkArgument(!StringUtils.containsAny(key,new char[]{' '}),"Invalid key name provided: %s",key);
-        return key;
+        IndexProvider.checkKeyValidity(key);
+        return key.replace(' ', IndexProvider.REPLACEMENT_CHAR);
     }
 
     @Override
--- a/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
+++ b/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
@@ -69,6 +69,8 @@ public class ElasticSearchIndexTest extends IndexProviderTest {
     static CloseableHttpClient httpClient;
     static ObjectMapper objectMapper;
 
+    private static char REPLACEMENT_CHAR = '\u2022';
+
     @BeforeClass
     public static void startElasticsearch() throws Exception {
         esr = new ElasticsearchRunner();
@@ -246,5 +248,16 @@ public class ElasticSearchIndexTest extends IndexProviderTest {
             assertEquals(1, tx.queryStream(new IndexQuery("ingestvertex", PredicateCondition.of(TEXT, Text.CONTAINS, "bob"))).count());
             assertEquals(1, tx.queryStream(new IndexQuery("ingestvertex", PredicateCondition.of(STRING, Cmp.EQUAL, "hello"))).count());
         }
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMapKey2Field_IllegalCharacter() {
+        index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
+    }
+
+    @Test
+    public void testMapKey2Field_MappingSpaces() {
+        String expected = "field" + REPLACEMENT_CHAR + "name" + REPLACEMENT_CHAR + "with" + REPLACEMENT_CHAR + "spaces";
+        assertEquals(expected, index.mapKey2Field("field name with spaces", null));
     }
 }
--- a/janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
+++ b/janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
@@ -29,6 +29,8 @@ import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
 import org.janusgraph.diskstorage.indexing.IndexProvider;
 import org.janusgraph.diskstorage.indexing.IndexProviderTest;
 import org.janusgraph.diskstorage.indexing.IndexQuery;
+import org.janusgraph.diskstorage.indexing.KeyInformation;
+import org.janusgraph.diskstorage.indexing.StandardKeyInformation;
 import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
 import org.janusgraph.graphdb.query.condition.PredicateCondition;
 import org.junit.AfterClass;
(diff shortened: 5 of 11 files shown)


robertdale:
What happened to the PR template?

sharpau:
@robertdale Oops, I kind of assumed that was an internal checklist for me - I'll bring it back to the PR description, sorry!

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
```diff
@@ -165,4 +167,19 @@ public void testSupport() {
...
+    public void testMapKey2Field_IllegalCharacter() {
+        KeyInformation keyInfo = new StandardKeyInformation(Boolean.class, Cardinality.SINGLE);
+
+        index.mapKey2Field("here is an illegal character: •", keyInfo);
+    }
+
+    @Test
+    public void testMapKey2Field_MappingSpaces() {
+        KeyInformation keyInfo = new StandardKeyInformation(Boolean.class, Cardinality.SINGLE);
+
```
mbrukman: Drop blank line?

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
```diff
@@ -165,4 +167,19 @@ public void testSupport() {
      */
     @Override @Test @Ignore
     public void clearStorageTest() throws Exception { }
+
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMapKey2Field_IllegalCharacter() {
+        KeyInformation keyInfo = new StandardKeyInformation(Boolean.class, Cardinality.SINGLE);
+
```
mbrukman: Drop blank line?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -181,6 +181,11 @@
      */
     public static final double DEFAULT_GEO_DIST_ERROR_PCT = 0.025;
 
+    /*
+     * A special character to use instead of whitespace, which is not allowed in field names. See #777.
+     */
+    private static char REPLACEMENT_CHAR = '\u2022';
```
mbrukman: This is replicated 3 ways across the 3 index backends; any way this can be kept in a single location?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -1054,8 +1059,9 @@ public boolean supports(KeyInformation information) {
 
     @Override
     public String mapKey2Field(String key, KeyInformation information) {
-        Preconditions.checkArgument(!StringUtils.containsAny(key,new char[]{' '}),"Invalid key name provided: %s",key);
-        return key;
+        Preconditions.checkArgument(!StringUtils.containsAny(key, new char[]{ REPLACEMENT_CHAR }),
+            "Invalid key name containing reserved character • provided: %s", key);
```
mbrukman: This string should use the constant `REPLACEMENT_CHAR`, no?

Same for the other 3 copies of this code.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sharpau: I'm happy to do so, but I wasn't sure where a good place was. The most relevant place seems like the IndexProvider interface alongside the mapKey2Field method, but I wasn't sure if others considered it clean to add a constant to an interface. Not to mention it's kind of an implementation detail of each index provider and it doesn't *have* to be the same...but if IndexProvider or another location works, I'll move it.

mbrukman: I don't have a strong opinion about this, and you make a good point.

Maybe @sjudeng and/or @davidclement90 have some thoughts on this?

davidclement90: Like others static constant, I may put it in org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration ?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
davidclement90: I'm agree

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sharpau: It could go there but that's mostly public constants and configuration, and this definitely isn't something anyone other than an implementer of IndexProvider should ever need to know about. I'd prefer IndexProvider over GraphDatabaseConfiguration.

sharpau: I decided to put it in IndexProvider, but am open to moving it elsewhere.

sharpau:
OK, I believe I've addressed all the requested changes so far.

amcp:
@sharpau For your first observation about tests not working in Windows, can you be more specific and raise a separate issue please? Also please confirm whether running in a docker container is enough to get the tests working even on a Windows host.

For your second observation about running the janusgraph-es tests in the janusgraph-es module, please check and see if running the module tests as a submodule of the core project fixes the issue?

sharpau:
@amcp Looks like running janusgraph-es tests from janusgraph directory means they all pass (on Ubuntu). I didn't try the Windows tests with Docker, so if I get the chance I'll characterize the problems I had there and file a separate issue.

@davidclement90 As far as I can tell, direct index queries totally bypass all the infrastructure in JanusGraph including the field name mapping. So someone writing a direct index query would have to manually apply mapKey2Field themselves. (This was already true but really only mattered for Solr before, as Lucene and ES backends didn't do any mapping.) That said, I don't think it's a huge issue because if you're writing a direct index query, you're probably testing it against a running ES instance directly first, and at that point the differences in field names should be obvious.

davidclement90:
Ok so can you add documentation here on direct index query section ?

sharpau:
@davidclement90 Ah yes, that's a good idea. I did so and pushed the change up, feel free to reword it if you like.

sharpau:
@davidclement90 Anything else you want to see before approving?

sharpau:
@amcp are we waiting for anything else?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexProvider.java
```diff
@@ -31,6 +31,11 @@
  */
 
 public interface IndexProvider extends IndexInformation {
+    /*
+     * An obscure unicode character (•) provided as a convenience for implementations of {@link #mapKey2Field}, for
+     * instance to replace spaces in property keys. See #777.
+     */
+    public static char REPLACEMENT_CHAR = '\u2022';
```
amcp: Interfaces do not need public keyword.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -1054,8 +1054,9 @@ public boolean supports(KeyInformation information) {
 
     @Override
     public String mapKey2Field(String key, KeyInformation information) {
-        Preconditions.checkArgument(!StringUtils.containsAny(key,new char[]{' '}),"Invalid key name provided: %s",key);
-        return key;
+        Preconditions.checkArgument(!StringUtils.containsAny(key, new char[]{ IndexProvider.REPLACEMENT_CHAR }),
```
amcp: Can you pull this check into a method and reuse it in all indexing backends?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexProvider.java
sharpau: Fixed, also removed 'static'

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
sharpau: Done. Though while doing this I realized that the check may not actually be necessary; I think we only need a one-way mapping from key to external indexing backend field, and having something that maps back the other way is not actually needed - in other words, we replace spaces with a special char, but we never have to replace the special char with spaces.

So I think these checks may be an abundance of caution, but it might be nice future-proofing. I'm slightly inclined to leave them as I doubt they'll inconvenience anyone, but I could go either way.

sharpau:
@amcp I chose the special character by looking at unicode characters and picking something that made a nice-looking separator (i.e. not too close to normal alphanumerics), wasn't unobservable (i.e. weird whitespace characters), and was not a standard accented character or similar that I could see someone using in a property name.

So it was pretty arbitrary and I'm open to changing it if we want (though I'd prefer not to, as we are running a patched JG that has this exact change already!)

sjudeng:
Is it possible and would it make sense to make the special character configurable as an index configuration option defaulting to the value you've selected?

sharpau:
@sjudeng It's probably possible (though I haven't tried). It does seem like something that should never change for a given JG instance though - is there a configuration level that is once-per-installation (or first-startup-only)?

sjudeng:
Definitely agree it shouldn't be changed after initialization. It looks like [ConfigOption.Type.FIXED](https://github.com/JanusGraph/janusgraph/blob/52975f940e67a8e1c303694b0d5f9604bbb9af22/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ConfigOption.java#L44-L48) is right type for this case.

docs/directindex.adoc
```diff
@@ -59,6 +59,10 @@ Names of property keys that contain non-alphanumeric characters must be placed i
 [source, java]
 graph.indexQuery("vertexByText", "v.\"first_name\":john").vertices()
 
+Some property key names may be transformed by the JanusGraph indexing backend implementation. For instance, an indexing backend that does not permit spaces in field names may transform "My Field Name" to "My•Field•Name", or an indexing backend like Solr may append type information to the name, transforming "myBooleanField" to "myBooleanField_b". These transformations happen in the index backend's implementation of IndexProvider, in the "mapKey2Field" method. Indexing backends may reserve special characters (such as '•') and prohibit indexing of fields that contain them.
+
+In general, making direct index queries depends on implementation details of JanusGraph indexing backends that are normally hidden from users, so it's best to verify a query empirically against the indexing backend in use.
```
pluradj: It's probably worthwhile to make a general recommendation in the docs: If possible, only use alphanumeric and the underscore for property names.

This might be more appropriate/visible in the [schema and data modeling](https://github.com/JanusGraph/janusgraph/blob/master/docs/basics.adoc#defining-property-keys) docs.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
```diff
@@ -247,4 +247,14 @@ public void testInjestPipeline() throws Exception {
             assertEquals(1, tx.queryStream(new IndexQuery("ingestvertex", PredicateCondition.of(STRING, Cmp.EQUAL, "hello"))).count());
         }
     }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMapKey2Field_IllegalCharacter() {
+        index.mapKey2Field("here is an illegal character: •", null);
```
pluradj: Others mentioned this previously, but replace the dot in all test usages with `REPLACEMENT_CHAR`

docs/directindex.adoc
amcp: @sharpau any updates on this point?

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
amcp: @sharpau I would probably copy the character to a constant in this class, so that we can detect when the replacement char is mistakenly changed in the interface.

pluradj:
@sharpau I don't think the CLA enforcer is happy because of the double signing on 14048d5ccdb15e064ef57029beb343caad70e197 so you'll need to get that cleaned up.

docs/directindex.adoc
sharpau: Added a recommendation here and in basics.adoc. I'm going with that in lieu of making it configurable, as it's now probably obvious that I'd just like to get this finished. 😄

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
sharpau: @amcp Done

pluradj:
We'll need this rolled up to `master` also.

sharpau:
@amcp Had a chance to take a look at things?

@pluradj What's the way of rolling up to master here? Merge 0.2 into master after this merges, or something else?

pluradj:
@sharpau right, merge into `0.2` first then roll up to `master`

pluradj:
@amcp waiting to clear your approval

pluradj:
@sharpau can you squash your commits please? thanks

sharpau:
I intentionally made the commits separate because one is just adding contributors and the other is all of the substantial changes. I think it makes sense the way it is...

pluradj:
ah ok, thanks @sharpau

PR #839, opened 2017-12-14
Check whether a polygon is closed or not

changzhihua:
To ensure a polygon is closed, its first point and last point should be equal.

### DIFF
--- a/janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
@@ -16,10 +16,19 @@ package org.janusgraph.core.attribute;
 
 import org.janusgraph.core.attribute.GeoshapeHelper;
 import org.janusgraph.core.attribute.JtsGeoshapeHelper;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.locationtech.spatial4j.context.jts.JtsSpatialContext;
 import org.locationtech.spatial4j.context.jts.JtsSpatialContextFactory;
 import org.locationtech.spatial4j.shape.ShapeFactory;
 import org.locationtech.spatial4j.shape.jts.JtsShapeFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.hamcrest.Matchers.equalTo;
 
 /**
  * Test of JtsGeoshapeHelper.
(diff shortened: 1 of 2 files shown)


changzhihua:
My ICLA has been submitted.

changzhihua:
Test cases are added. I hava submitted a CLA with the email chzh.nju@gmail.com, but it is still labelled as cla:no. Does my CLA have some problem?

robertdale:
@changzhihua I don't see your email added to CLA yet.  But your second commit isn't signed anyway.  It would be best if you could squash the commits to one and sign it exactly as you submitted in the CLA (name and email must match!). Then when the CLA is added, it should update to green.

mbrukman:
My apologies, I haven't gotten to handling the CLA submission yet. I will follow-up via email.

mbrukman:
The CLA is being taken care of in https://github.com/JanusGraph/legal/pull/82

mbrukman:
@changzhihua — the CLA has been added, but the reason that you still see the `[cla: no]` tag on this PR is because you did not associate the email address you're using in your commits with your GitHub profile.

Note that it does **NOT** have to be listed publicly on your profile page, it just has to be listed and confirmed as one of the valid emails associated with your account. Once you fix this, the CLA label will flip to `[cla: yes]`.

Please see these docs:
* https://help.github.com/articles/adding-an-email-address-to-your-github-account/

changzhihua:
@mbrukman  very thanks for your help. I have added my email in the commits to my profile and now the CLA label have changed to [cla: yes].

janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
```diff
@@ -42,4 +51,35 @@ public boolean supportJts() {
     public ShapeFactory getShapeFactory() {
         return new JtsShapeFactory((JtsSpatialContext) getHelper().context, (JtsSpatialContextFactory) getHelper().factory);
     }
+
+    @Test
+    public void testPolygon() {
```
mbrukman: Rename this to `testPolygonIsClosed()`?

mbrukman: Is my understanding correct that this would not have been valid in the original code (which checked for object equality, rather than array equality), and hence would have thrown an exception?

janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
```diff
@@ -16,11 +16,20 @@
 
 import org.janusgraph.core.attribute.GeoshapeHelper;
 import org.janusgraph.core.attribute.JtsGeoshapeHelper;
+import org.junit.Assert;
```
mbrukman: We seem to typically use `import static org.junit.Assert.assertEquals;` in this repo, so that we can write code such as:

```java
assertEquals(...);
```

instead of the longer

```java
Assert.assertEquals(...);
```

Can you make the change to use the shorter form? It doesn't matter much in this specific case, since you only use the assert once, but it will make future code easier to write if they follow the same pattern.

janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
changzhihua: Actually，the original code may be just a slip of pen. It does nothing at most of the time or may throw an unexpected exception just when the first and last points are the same object (in address). It can be seen that we want a closed polygon (whose first and last point must be equal). I think we should test equality instead of inequality. Second, the symbol == or object equality you mentioned actually compares the address of the objects instead of their values. So I think here array equality is more suitable.

janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
changzhihua: ok. Thanks for your review.

janusgraph-test/src/test/java/org/janusgraph/core/attribute/JtsGeoshapeHelperTest.java
changzhihua: ok. It has been renamed.

robertdale:
@mbrukman Please amend or approve your review. Thanks.

Issue #252, opened 2017-05-02
order().by(property) doesn't work with mixed index 


rkozmer:
Results from the mixed index are not sorted natively (in my case by Lucene). Instead the slow comparator is used.

Index looks like:
```
mgmt.buildIndex("typeName", Vertex.class).addKey(type).addKey(name).buildMixedIndex("typeName")
```
Query:
```
results = traversal().V().has("type", "person").order().by("name")
```

As already discussed with @twilmes on gitter, the LuceneIndex.getSortOrder(IndexQuery) is being called with this incomplete query, the order-part is missing:
```
[(type = person)](2000):typeName
```

twilmes:
Thanks for entering this @rkozmer. I tracked this down to JanusGraph checking to see if the incoming Gremlin order comparators are instances of `ElementValueComparator`. It appears that this changed at some point on the TinkerPop side so we need to instead expect an `ElementValueTraversal`. I'll finish change up and push for review.

severin83:
Is there any news on this issue? I'm experiencing the same problem with pre-release 0.2.0 (didn't try 0.1.1) and elasticsearch (bundled version).
I came across this discussion thread about a similar problem in TitanDB:
[https://groups.google.com/forum/#!topic/aureliusgraphs/JFNthKyiBRM](url)

I tried the test script that was posted by Daniel Kuppitz on 08.02.16, the script works in JanusGraph, I see the indices in elasticsearch being updated, and I get very bad timings (compare to the timings posted in above linked discussion thread):

== 10000 Vertices ==
[893.2463445299999, 1507908527096]

== 20000 Vertices ==
[2475.23476129, 1507908622394]

Thanks, br
Severin

pluradj:
Looks like this fell off the radar for @twilmes. Based on his previous comments, the [`HasStepFolder`](https://github.com/JanusGraph/janusgraph/blob/v0.1.1/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java#L69) class is where the fix needs to be made.

twilmes:
Yes, thanks for the bump @severin83 & @pluradj. I'll take a look at this again.

severin83:
I made a few changes, which fix the problem for me. I can't be sure if the code is totally correct, but it certainly works on some use cases I tried.
If you want to take over the changes, I attached the diff to this comment.
[...]

PR #842, opened 2017-12-14
Issue #252: order().by(property) doesn't work with mixed index

twilmes:
Issue #252 

With @severin83's help, the HasStepFolder has been updated to fold Janus compatible global ordering operations into JanusGraphSteps.

This re-enables the pushdown of ordering operations into indexing backends.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
@@ -236,7 +236,6 @@ public interface HasStepFolder<S, E> extends Step<S, E> {
 
             if (key != null ? !key.equals(that.key) : that.key != null) return false;
             return order == that.order;
-
         }
 
         @Override
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
@@ -15,6 +15,7 @@
 package org.janusgraph.graphdb.tinkerpop.optimize;
 
 import org.apache.tinkerpop.gremlin.process.traversal.P;
+import org.apache.tinkerpop.gremlin.process.traversal.lambda.ElementValueTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.GraphStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.NoOpBarrierStep;
 import org.apache.tinkerpop.gremlin.process.traversal.util.AndP;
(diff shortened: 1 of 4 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
```diff
@@ -227,6 +226,33 @@ public OrderEntry(String key, Order order) {
...
+
+            if (key != null ? !key.equals(that.key) : that.key != null) return false;
+            return order == that.order;
+
+        }
+
+        @Override
+        public int hashCode() {
+            int result = key != null ? key.hashCode() : 0;
+            result = 31 * result + (order != null ? order.hashCode() : 0);
```
mbrukman: Might be more readable as:

```java
int keyCode = (key != null) ? key.hashCode() : 0;
int orderCode = (order != null) ? order.hashCode() : 0;
return (31 * keyCode) + orderCode;
```

Because it crease parallel structure for the two elements.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
```diff
@@ -227,6 +226,33 @@ public OrderEntry(String key, Order order) {
...
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            OrderEntry that = (OrderEntry) o;
+
+            if (key != null ? !key.equals(that.key) : that.key != null) return false;
+            return order == that.order;
+
```
mbrukman: Extra blank line.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
```diff
@@ -227,6 +226,33 @@ public OrderEntry(String key, Order order) {
...
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            OrderEntry that = (OrderEntry) o;
+
+            if (key != null ? !key.equals(that.key) : that.key != null) return false;
```
mbrukman: This is very interesting; I hadn't seen a `?:` **inside** an `if` condition before. A bit complex to read, methinks.

I believe this is equivalent to the following but with fewer constructs per line/clause so should be easier to understand:

```java
if (key == null && that.key != null) return false;
if (!key.equals(that.key)) return false;
```

Thoughts?

twilmes: Ah yes, a ternary operator nested in an if! I use Intellij to generate `equals` and `hashCode` methods and I agree, it is a bit baroque. I'll update it per your suggestion for improved readability.

mbrukman: Hmm, thinking about it some more, that'll cause a NPE if `key == null` and `that.key == null` are both true. This needs some more thought; looks like the original one is well thought-through and correct but perhaps a bit complex.

May be best to leave it alone; if it's the IntelliJ auto-generated method, it's probably well-designed and well-written.

robertdale:
@twilmes does this need to be merged up to master?

twilmes:
@robertdale it does. I'll create a 0.2 -> master pr. Looking at it though, I think I screwed up last time and used a rebase when I went to master. Anyway, I'll submit a new PR and we can review.

PR #843, opened 2017-12-17
Fixed Titan compatibility if local configuration not set

### LINKED ISSUE #418: Failure to correctly set ids.store-name=titan_ids can corrupt a Titan Cassandra

rafatuita:
The fixes #418 and #445 might not have taken into account systems that do not have local titan versions. Added a try/catch on that specific check fixes the issue and won't prevent embedded applications to start.

Signed-off-by: Rafael Fernandes <luizrafael@gmail.com>

-----

### For all changes:

### For code changes:
- [x] Have you written and/or updated unit tests to verify your changes? [RF]: it already runs in many different unit tests with Restriction.LOCAL.

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
@@ -1385,7 +1385,12 @@ public class GraphDatabaseConfiguration {
 
         // When connecting to a store created by Titan the ID store name will not be in the
         // global configuration. To ensure compatibility override the default to titan_ids.
-        boolean localTitanConfigured = localBasicConfiguration.get(TITAN_COMPATIBLE_VERSIONS) != null;
+        boolean localTitanConfigured = false;
+        try {
+            localTitanConfigured = localBasicConfiguration.get(TITAN_COMPATIBLE_VERSIONS) != null;
+        } catch (IllegalStateException e) {
+            localTitanConfigured = false;
+        }
         boolean localIdStoreIsDefault = JanusGraphConstants.JANUSGRAPH_ID_STORE_NAME.equals(localBasicConfiguration.get(IDS_STORE_NAME));
 
         if (localTitanConfigured) {


smcquillan:
@rafatuita I was able to reproduce by first connecting with the following set in my config:
```
graph.titan-version=1.0.0
```
which was accessible in both the global and local config and was written to Cassandra. Then when I attempted to connect a second time but without ` graph.titan-version=1.0.0` in my config I got the same stack trace. 


However, when I put the code in from this pull request to catch the exception and re-ran it no longer ran following bit of the code
```
overwrite.set(IDS_STORE_NAME, JanusGraphConstants.TITAN_ID_STORE_NAME);
```

To override the ID store back to the Titan default of `titan_ids` and a new ID store created named `janusgraph_ids` was created. This will lead to the type of corruption originally raised in issue #418.

I’m wondering if its best just to remove the setting and checking of the `localTitanConfigured` Boolean entirely and always overwrite the ID store name, because it wasn’t something which you could configure in Titan.

smcquillan:
@pluradj @amcp @jerryjch @sjudeng I did some more testing on this today and I don't think we can fix it with the try /catch block currently being proposed as this will result in the [overwrite](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java#L1398) being missed and a second ID store named `janusgraph_ids` being created in addition to the existing `titan_ids` which is what #418 was intended to fix.

However, as @rafatuita points out we shouldn't force people to always include `titan-version` in their local configuration when its already there in the Global config. I think the fix is therefore just to remove the check for `titan-version` in the local config entirely as I've done on a branch in my fork  [here](https://github.com/JanusGraph/janusgraph/compare/master...smcquillan:titanCompatibleVersionsIsGlobal), I've left the white space to better illustrate what's changed.

This would mean that when `titan-version` is in the global config we'll **always** use `titan_ids` for the key store name. Which given that the key store name was not configurable in Titan I don't think would present a problem.

amcp:
@rafatuita seems like you should update the content of this PR with your [fork](https://github.com/JanusGraph/janusgraph/compare/master...smcquillan:titanCompatibleVersionsIsGlobal) or create a new PR.

pluradj:
Or @smcquillan can submit a PR with the fix proposed from his fork.

smcquillan:
@pluradj @rafatuita Bit tied up with other things this week but unless I hear different I'll look to open an new PR next week from my fork.

smcquillan:
@pluradj @rafatuita raised pull request #977 with the fix from my fork.

pluradj:
Closing in favor of #1005. @rafatuita please refer to Issue #1012 to make sure I captured your intent correctly. Thanks.

Issue #672, opened 2017-10-30
Custom analyzer for lucene


Lakedaemon:
Is there a way to use custom analyzer with a lucene index ?
I have properties that might be in 30 different languages (chinese, japanese, russian, french, urdu, lots of fancy languages that won't work best with the StandardAnalyzer shipped with lucene) and that need specific analyzers

I have been using lucene extensively for the last 5 years in some multi-lingual android app and I would like to know if I can use the full might of Lucene with JanusGraph or if I should start looking for a property graph replacement. 

If it is theoretically possible to make lucene work with custom analyzers and if It is in my capacity to implement it, I would also like to help (might take less time to implement the lucene features I need in JanusGraph than looking for better alternatives or getting burned reimplementing one)

Best regards,
Olivier

robertdale:
You could check or add support similar to ES, Solr - http://docs.janusgraph.org/latest/field-mapping.html#_custom_analyser

Lakedaemon:
Before posting, I  had a (fast) look at the code for Solr and gor the feeling it might be feasible (but I needed a confirmation before trying it)

I'll probably have a go at it (not immediately though). If I manage to make things work and to put it in a decent shape, I'll create a pr. Thanks

Lakedaemon:
It looks like I just managed to implement custom analyzer support for lucene and JanusGraph
I'll now take some time to test if it really works (it was relatively easy) and if it does I'll clean my stuff and make a pr

Lakedaemon:
It looks like my code produces an index with a custom analyzer (for french, chinese, english...)
I looked in the lucene index and all the terms are there
But when I'm doing a simple gremlin query on a field indexed with the custom analyzer, it warns that 

> Query requires iterating over all vertices [(fraG = livre)]. For better performance, use indexes

pondering now.... what am I doing wrong ?

robertdale:
@Lakedaemon  Where's your repo? I didn't see a fork of JanusGraph.

Lakedaemon:
I just forked janusgraph and sent a pr
How may I sign the CLA ?

robertdale:
@Lakedaemon  https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md

Lakedaemon:
thanks, I'm sending the email with the scans and the cla right now.

Lakedaemon:
figured out what was wrong with my test.
I was using operator has with a Mapping.Text field instead of Mapping.textContains
[...]

PR #854, opened 2017-12-20
issue #672 : Custom analyzer support for LuceneIndex

Lakedaemon:
Handling issue #672

There is (at least) one major issue with this pull request : 
you can use a custom analyzer with the lucene index backend through

        mgmt.buildIndex(someString, Vertex::class.java)
                .addKey(someProperty, Mapping.TEXT.asParameter(), Parameter.of("TEXT_ANALYZER", "org.apache.lucene.analysis.en.EnglishAnalyzer"))

It will use the right Analyzer instance (it I didn't mess the strings), 
it will tokenize the data sent by janusgraph and store them in the index (I checked on the lucene side)

BUT
When I'm querrying the graph, it looks like it doesn't find the field with the custom terms and 
warns that the index is looping through all vertices because it couldn't find an index.

So, it is not fully working yet. This pr is a call for help : 
I would like to get help to make the necessary changes to make it fully working. Please, someone knowledgeable with janusgraph, give me some pointers/advice !

Thanks, 
Olivier

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
+++ b/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
@@ -1,0 +1,119 @@
+// Copyright 2017 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.lucene;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.DelegatingAnalyzerWrapper;
+import org.janusgraph.core.schema.Mapping;
+import org.janusgraph.core.schema.Parameter;
+import org.janusgraph.diskstorage.indexing.KeyInformation;
+import org.janusgraph.graphdb.types.ParameterType;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Olivier binda (olivier.binda@wanadoo.fr)
+ */
+
+/**
+ * JanusGraph may open different index stores and needs an analyzer for each
+ * so an analyzer should be provided to build IndexWriters/IndexReaders that are used when an index is opened and documents are added
+ * BUT a janusGraph may add a new indexable field, with a custom tokenizing process through Parameters, at any moment, which means that it should be possible
+ * to map any field to the corresponding custom analyzer, though indexreaders are already in use.
+ * <p>
+ * A way to achieve that would be to use a delegating analyzer that delegates Tokenizing for a field to
+ * the corresponding Analyzer (whose class you get through a Mapping)
+ */
+public class CustomAnalyzer extends DelegatingAnalyzerWrapper {
+    private static final String KEYWORD_ANALYZER = org.apache.lucene.analysis.core.KeywordAnalyzer.class.getName();
+    private static final String STANDARD_ANALYZER = org.apache.lucene.analysis.standard.StandardAnalyzer.class.getName();
+
+    private final String store;
+    private final KeyInformation.IndexRetriever informations;
+
+    private final Map<String, Analyzer> analyzers = new HashMap<>();
+
+    public CustomAnalyzer(String store, KeyInformation.IndexRetriever informations, ReuseStrategy fallbackStrategy) {
+        super(fallbackStrategy);
+        this.store = store;
+        this.informations = informations;
+    }
+
+    @Override
+    final protected Analyzer getWrappedAnalyzer(String fieldName) {
+        Mapping mapping = null;
+        String stringAnalyzerString = null;
+        String textAnalyzerString = null;
+
+
+        Parameter[] parameters = null;
+        Class<?> clazz = null;
+
+
+        KeyInformation keyInformation = informations.get(store, fieldName);
+        if (keyInformation != null) {
+            parameters = keyInformation.getParameters();
+            clazz = keyInformation.getDataType();
+        }
+
+        if (parameters != null) {
+            // if mapping isn't present in parameters, we use Mapping.DEFAULT
+            mapping = ParameterType.MAPPING.findParameter(parameters, Mapping.DEFAULT);
+            stringAnalyzerString = ParameterType.STRING_ANALYZER.findParameter(parameters, null);
+            textAnalyzerString = ParameterType.TEXT_ANALYZER.findParameter(parameters, null);
+        }
+      
+        String analyzerString = null;
+
+        if (clazz != null && String.class.isAssignableFrom(clazz) && mapping != null) {
+            // at the moment, we only try to support custom analyzers for string data.
+            // everything else falls through a StandardAnalyzer as was the case before
+            switch (mapping) {
+                case TEXTSTRING:
+                    throw new RuntimeException("TextString is an unsupported mapping for string data & custom analyzers");
+                case PREFIX_TREE:
+                    throw new RuntimeException("Prefix-tree is an unsupported mapping for string data & custom analyzers");
+                case STRING:
+                    analyzerString = stringAnalyzerString != null ? stringAnalyzerString : KEYWORD_ANALYZER;
+                    break;
+                case TEXT:
+                case DEFAULT:// TEXT
+                    analyzerString = textAnalyzerString;
+                    break;
+                default:
+                    throw new RuntimeException("Not supported");
+            }
+        }
+
+        // the default (& fall through) Analyzer is a StandardAnalyzer
+        if ((clazz == null || !String.class.isAssignableFrom(clazz)) || mapping == null || analyzerString == null) {
+            analyzerString = STANDARD_ANALYZER;
+        }
+
+        final String analyzerStringFinal = analyzerString;
+
+        Analyzer analyzer = analyzers.get(analyzerString);
+        if (analyzer == null) {
+            try {
+                Class classDefinition = Class.forName(analyzerStringFinal);
+                analyzer = (Analyzer) classDefinition.newInstance();
+            } catch (Exception e) {
+                throw new RuntimeException("Analyzer cannot be instanciated for class " + analyzerStringFinal, e);
+            }
+            analyzers.put(analyzerString, analyzer);
+        }
+        return analyzer;
+    }
+}
--- a/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
+++ b/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
@@ -17,6 +17,9 @@ package org.janusgraph.diskstorage.lucene;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import org.apache.lucene.analysis.CachingTokenFilter;
+import org.apache.lucene.analysis.tokenattributes.TermToBytesRefAttribute;
+import org.janusgraph.core.schema.Parameter;
 import org.locationtech.spatial4j.context.SpatialContext;
 import org.locationtech.spatial4j.shape.Shape;
 
@@ -38,7 +41,6 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.lucene.analysis.Analyzer;
-import org.apache.lucene.analysis.standard.StandardAnalyzer;
 import org.apache.lucene.document.*;
 import org.apache.lucene.index.*;
 import org.apache.lucene.queryparser.classic.ParseException;
@@ -86,6 +88,7 @@ public class LuceneIndex implements IndexProvider {
     private static final IndexFeatures LUCENE_FEATURES = new IndexFeatures.Builder()
         .supportedStringMappings(Mapping.TEXT, Mapping.STRING)
         .supportsCardinality(Cardinality.SINGLE)
+        .supportsCustomAnalyzer()
         .supportsNanoseconds()
         .supportsGeoContains()
         .build();
@@ -101,8 +104,6 @@ public class LuceneIndex implements IndexProvider {
     public static final double DEFAULT_GEO_DIST_ERROR_PCT = 0.025;
 
     private static final Map<Geo, SpatialOperation> SPATIAL_PREDICATES = spatialPredicates();
-
-    private final Analyzer analyzer = new StandardAnalyzer();
 
     private final Map<String, IndexWriter> writers = new HashMap<>(4);
     private final ReentrantLock writerLock = new ReentrantLock();
@@ -137,10 +138,11 @@ public class LuceneIndex implements IndexProvider {
         }
     }
 
-    private IndexWriter getWriter(String store) throws BackendException {
+    private IndexWriter getWriter(String store, KeyInformation.IndexRetriever informations) throws BackendException {
         Preconditions.checkArgument(writerLock.isHeldByCurrentThread());
         IndexWriter writer = writers.get(store);
         if (writer == null) {
+            final Analyzer analyzer = new CustomAnalyzer(store, informations, Analyzer.PER_FIELD_REUSE_STRATEGY);
             IndexWriterConfig iwc = new IndexWriterConfig(analyzer);
             iwc.setOpenMode(IndexWriterConfig.OpenMode.CREATE_OR_APPEND);
             try {
@@ -157,9 +159,9 @@ public class LuceneIndex implements IndexProvider {
         SpatialStrategy strategy = spatial.get(key);
         Mapping mapping = Mapping.getMapping(ki);
         final int maxLevels = ParameterType.INDEX_GEO_MAX_LEVELS.findParameter(ki.getParameters(),
-                DEFAULT_GEO_MAX_LEVELS);
+            DEFAULT_GEO_MAX_LEVELS);
         final double distErrorPct = ParameterType.INDEX_GEO_DIST_ERROR_PCT.findParameter(ki.getParameters(),
-                DEFAULT_GEO_DIST_ERROR_PCT);
+            DEFAULT_GEO_DIST_ERROR_PCT);
         if (strategy == null) {
             synchronized (spatial) {
                 if (!spatial.containsKey(key)) {
@@ -181,11 +183,11 @@ public class LuceneIndex implements IndexProvider {
 
     private static Map<Geo, SpatialOperation> spatialPredicates() {
         return Collections.unmodifiableMap(Stream.of(
-                new SimpleEntry<>(Geo.WITHIN, SpatialOperation.IsWithin),
-                new SimpleEntry<>(Geo.CONTAINS, SpatialOperation.Contains),
-                new SimpleEntry<>(Geo.INTERSECT, SpatialOperation.Intersects),
-                new SimpleEntry<>(Geo.DISJOINT, SpatialOperation.IsDisjointTo))
-                .collect(Collectors.toMap(SimpleEntry::getKey, SimpleEntry::getValue)));
+            new SimpleEntry<>(Geo.WITHIN, SpatialOperation.IsWithin),
+            new SimpleEntry<>(Geo.CONTAINS, SpatialOperation.Contains),
+            new SimpleEntry<>(Geo.INTERSECT, SpatialOperation.Intersects),
+            new SimpleEntry<>(Geo.DISJOINT, SpatialOperation.IsDisjointTo))
+            .collect(Collectors.toMap(SimpleEntry::getKey, SimpleEntry::getValue)));
     }
 
     @Override
@@ -194,7 +196,7 @@ public class LuceneIndex implements IndexProvider {
         Mapping map = Mapping.getMapping(information);
         Preconditions.checkArgument(map == Mapping.DEFAULT || AttributeUtil.isString(dataType) ||
                 (map == Mapping.PREFIX_TREE && AttributeUtil.isGeo(dataType)),
-                "Specified illegal mapping [%s] for data type [%s]", map, dataType);
+            "Specified illegal mapping [%s] for data type [%s]", map, dataType);
     }
 
     @Override
@@ -217,7 +219,7 @@ public class LuceneIndex implements IndexProvider {
         IndexReader reader = null;
         try {
             String storeName = stores.getKey();
-            IndexWriter writer = getWriter(storeName);
+            IndexWriter writer = getWriter(storeName, information);
             reader = DirectoryReader.open(writer, true, true);
             IndexSearcher searcher = new IndexSearcher(reader);
             for (Map.Entry<String, IndexMutation> entry : stores.getValue().entrySet()) {
@@ -266,7 +268,7 @@ public class LuceneIndex implements IndexProvider {
         try {
             for (Map.Entry<String, Map<String, List<IndexEntry>>> stores : documents.entrySet()) {
                 String store = stores.getKey();
-                IndexWriter writer = getWriter(store);
+                IndexWriter writer = getWriter(store, information);
                 IndexReader reader = DirectoryReader.open(writer, true, true);
                 IndexSearcher searcher = new IndexSearcher(reader);
 
@@ -341,7 +343,7 @@ public class LuceneIndex implements IndexProvider {
                                KeyInformation.IndexRetriever information) {
         Preconditions.checkNotNull(doc);
         for (IndexEntry e : content) {
-            Preconditions.checkArgument(!e.hasMetaData(),"Lucene index does not support indexing meta data: %s",e);
+            Preconditions.checkArgument(!e.hasMetaData(), "Lucene index does not support indexing meta data: %s", e);
             if (log.isTraceEnabled())
                 log.trace("Adding field [{}] on document [{}]", e.field, docID);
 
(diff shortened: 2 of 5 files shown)


Lakedaemon:
this pr is related to issue #672
And how the hell may I sign the damn CLA ?

davidclement90:
@Lakedaemon — thank you for your contribution!

Please follow the instructions in [`CONTRIBUTING.md`](https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md) to sign the CLA.

Lakedaemon:
mmmh, lots of errors.
I forgot to register mappings probably...looking into that...

Lakedaemon:
Still got 10 errors.
First, let's try to fix the "Invalid mapping specified: null" errors
I get that when doing 
if (parameters != null) {
            analyzerString = ParameterType.TEXT_ANALYZER.findParameter(parameters, null);
        }
in the precondition there : 
if (p.key().equalsIgnoreCase(name)) {
                Object value = p.value();
                Preconditions.checkArgument(value!=null,"Invalid mapping specified: %s",value);
              
Which means that someone put a mapping for 
ParameterType.TEXT_ANALYZER
and it is null !!!!!!
Isn't it weird ? I'm lost

Maybee the test tells if the string should be tokenized with a Text analyzer instead of a string analyzer ?
And it interferes with the custom analyzer support that looks for an analyzer class name for the same key ?
But why, why, why should there be a null value if the key exists in the parameters ?  

-> I'll have a look at the test to see if I can get an insight there

Lakedaemon:
Ok, testCustomAnalyzers is launched and then, it builds a simple graph with "Tom and Jerry".

initialDoc.put(STRING, "Tom and Jerry");
        initialDoc.put(ANALYZED, "Tom and Jerry");
        if(indexFeatures.supportsStringMapping(Mapping.TEXTSTRING))
            initialDoc.put(FULL_TEXT, "Tom and Jerry");
        initialDoc.put(KEYWORD, "Tom and Jerry");
        add(store, "docId", initialDoc, true);
        clopen();
The error happens when closing, because I guess that the graph is mutated, the vertex is indexed...
But the configuration doesn't contain anything regarding custom Analyzer class names.

So, the way I have been trying to tell janusGraph to support custom Analyzer must be wrong (I tried to do it like elastic search or solr (don't remember which one) that allows adding indexing on the fly (and not putting it in stone in a schema first))

@robertdale can you (or someone else) give me a few hints how to do it right ?

Lakedaemon:
that's it : For some reason, Janusgraph or the test puts that in the mappings 
Caused by: java.lang.RuntimeException: [mapping->TEXT, text-analyzer->null]
and then, it makes findParameter crash because of a preconditions that the object of a parameter should not be null.

SERIOUSLY ?
(I hate it when not my code make my code crash. Especially when it looks like something really stupid I have no control on, and I have no clue why the other people write it like this in the first place).
It will make fixing this a grand pleasure..... :(

mbrukman:
The CLA issue is being addressed in https://github.com/JanusGraph/legal/pull/81.

mbrukman:
@Lakedaemon — the CLA PR has been submitted, but your git config must match the CLA agreement, as per these instructions:
https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla

If you look at https://github.com/JanusGraph/janusgraph/pull/854.patch (i.e., add the `.patch` suffix to the URL of this PR), it says:

> From: lakedaemon <olivier.binda@wanadoo.fr>

but it needs to look as follows:

> From: Olivier Binda <olivier.binda@wanadoo.fr>

for the automated CLA checker to verify this PR and all of its commits.

To fix it, please run the following commands, as documented in
https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla

```bash
git config user.name "Olivier Binda"
# You've already done `git config user.email` and it's correct
```

Note that since you have 8 commits on your branch, rather than just a single one, you need to squash your commits, and then update the authorship:

```bash
git rebase -i HEAD~8      # since you currently have 8 commits
```

Replace instances of `pick` with `squash`, and save the file. Edit the commit message and save.

Then, continue as documented in the `CONTRIBUTING.md` file:
https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla
 since you now have a single commit on this brach, namely:

```bash
git commit --amend --reset-author -s
```

Edit the message (remove the old Signed-off-by line if needed).

Then update the branch on this PR:

```bash
git push -f
```

Hope this helps; let me know if you have any questions.

davidclement90:
I think one of your problem with the custom analyzer test is that you do not update the name of English and Keyword analyzer in LuceneIndexTest (https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/LuceneIndexTest.java#L61)

Lakedaemon:
Many thanks for the advice. I'm looking into it and making the necessary changes. 
Thanks again. :)

Lakedaemon:
My latest commit fixed the NPE (by setting the keywoard and english analyzers for the test), but tests still fail. For example, I got : 
java.lang.AssertionError: [string = Tom Jerry]:vertex expected:<1> but was:<0>

For these tests : 
883  IndexQuery query = new IndexQuery(store, PredicateCondition.of(STRING, Cmp.EQUAL, "Tom and Jerry"));
884        assertEquals(query.toString(), 1, tx.queryStream(query).count());

// this is the test that fails
885        query = new IndexQuery(store, PredicateCondition.of(STRING, Cmp.EQUAL, "Tom Jerry"));
886        assertEquals(query.toString(), 1, tx.queryStream(query).count());

// I should understand how these tests are setup in order to understand why they fail.

On my machine, I'm still getting 
WARN org.janusgraph.graphdb.transaction.StandardJanusGraphTx - Query requires iterating over all vertices [(fraG = livre)]. For better performance, use indexes

So, it might be possible that janusGraph doesn't know that some fields are indexed (with a custom analyzer)

Also, I'll try to make some change to support Mapping .STRING and Mappinf.TEXT even when custom analyzers aren't set (with defaults to KeywordAnalyzer  for String and StandardAnalyzer for text)

Lakedaemon:
Wow, getting test fail in org.janusgraph.diskstorage.lucene.BerkeleyLuceneTest
in at org.janusgraph.graphdb.JanusGraphIndexTest.testSimpleUpdate(JanusGraphIndexTest.java:225)

This was completely unexpected and may not be related to this pr. 

In my opinion, the failed tests regarding custom analyzer support come from the fact that I never changed anything in the convertQuery routine that translates janus query stuff into lucene query stuff
Especially since the analyzers have changed since this was last written, so the TermQueries in there should completely wrong... the string stuff should be analyzed by the correct Analyzer before creating queries...

I'll look at it, in depth and investiguate if this is the reason my code change doesn't work as expected 
(basically the right terms are produced but searching fails epically)

Lakedaemon:
Also, now that I opened janusgraph project in Idea and that I am able to selectively launch tests (not relying on Travis).
I should be able to get things done correctly faster...

Lakedaemon:
mmh. First I had not fully understood the distinction between Mapping.STRING and Mapping.TEXT...
Got a lot of fixing to do....

Lakedaemon:
Coming from the Lucene world, I was expecting Mapping.STRING to be done by KeywordAnalizer which it is not at all... Strings are still tokenized around non alphanumeric characters and words less than 2 chars are removed... this is like a poor version of analyzing...and definately not like the full string is used as a token/keyword

Lakedaemon:
I have problems understanding why this test should pass : 
query = new IndexQuery(store, PredicateCondition.of(STRING, Cmp.EQUAL, "Tom Jerry"));
        assertEquals(query.toString(), 1, tx.queryStream(query).count());

Considering that the document mainly contains 1 vertex with a (STRING) mapped field : "Tom and Jerry"
and it says [here](http://docs.janusgraph.org/latest/index-parameters.html) that : 
eq: if the string is identical to the query string

Lakedaemon:
same thing about this one : 
        query = new IndexQuery(store, PredicateCondition.of(STRING, Cmp.EQUAL, "Tom or Jerry"));
        assertEquals(query.toString(), 1, tx.queryStream(query).count());

davidclement90:
Because STRING use English analyzer so 'and' and 'or' are stopwords 
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-test/src/main/java/org/janusgraph/diskstorage/indexing/IndexProviderTest.java#L107

Lakedaemon:
Thanks for the tip (again). Now, it makes more sense. 
(note to self, the english analyzer used for tests is expected to have "or" and "and" stopwords, which is the case with the classic lucene English Analyzer... Mapping.String can also use a custom analyzer BUT supports the same operators as keywords...eq, noteq...)

I'll think about it make find out what necessary changes I should make then.
Many thanks again. :)

Lakedaemon:
I have advanced some more and yet bumped into another weird difficulty.

To pass customAnalyzer tests, a graph with the "string" field (and the Mapping.STRING) is supposed to be used with an Analyzer that remove the stopwords "or" and "and".

YET, in the current LuceneIndex.class 
such fields are added to the lucene index in the addToDoc routine through
switch(mapping) {
                    case STRING:
                        field = new StringField(e.field, str, Field.Store.YES);
}

This is awkward, because, StringField are never analyzed in lucene : 
they are stored as is, as keywords.

Of course, I tried adding the value to the index with 
field = new TextField(e.field, str, Field.Store.YES);
and (with my code) it makes the problematic testCustomAnalyzer tests pass

YET it makes like 5 other tests in the LuceneIndexTest fail...

This is going to be painfull...

Lakedaemon:
Making the Mapping.STRING mapped fields fallback to KeywordAnalyzer and modifying the convertQuery routine relatively to my changes fixes some stuff.

At this point, I'm thinking that Mapping.STRING fields may use custom Analyzers that behave like a KeywordAnalyzer (or fall back to just indexing the full string but that is a performance optimization as it is usually best to normalize entries when indexing). They still need a TextField to work (in the custom analyzer case)
And  Mapping.Default or Text fields may use custom Analyzers... and fall back to the default tokenization process of Janusgraph when not expliciely set

Working further to make everything work (but 1 change has cascading effects and require more changes...It'll take me some time yet)

Lakedaemon:
Rebased my commits and awaiting advice on how to proceed next...

Lakedaemon:
Damned, forgot to remove some print statements... :/
I'll clean the code some more next, then

Lakedaemon:
Cleaned and rebase the code. Waiting for a review... and advice on how to proceed next

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+ * A way to achieve that would be to use a delegating analyzer that delegates Tokenizing for a field to
+ * the corresponding Analyzer (whose class you get through a Mapping)
+ */
+public class CustomAnalyzer extends DelegatingAnalyzerWrapper {
+    private String store;
+    private KeyInformation.IndexRetriever informations;
+
+    private Map<String, Analyzer> analyzers = new HashMap<>();
+
+    public CustomAnalyzer(String store, KeyInformation.IndexRetriever informations, ReuseStrategy fallbackStrategy) {
```
davidclement90: add final

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+ * so an analyzer should be provided to build IndexWriters/IndexReaders that are used when an index is opened and documents are added
+ * BUT a janusGraph may add a new indexable field, with a custom tokenizing process through Parameters, at any moment, which means that it should be possible
+ * to map any field to the corresponding custom analyzer, though indexreaders are already in use.
+ * <p>
+ * A way to achieve that would be to use a delegating analyzer that delegates Tokenizing for a field to
+ * the corresponding Analyzer (whose class you get through a Mapping)
+ */
+public class CustomAnalyzer extends DelegatingAnalyzerWrapper {
+    private String store;
+    private KeyInformation.IndexRetriever informations;
```
davidclement90: add final

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+    private Map<String, Analyzer> analyzers = new HashMap<>();
+
+    public CustomAnalyzer(String store, KeyInformation.IndexRetriever informations, ReuseStrategy fallbackStrategy) {
+        super(fallbackStrategy);
+        this.store = store;
+        this.informations = informations;
+    }
+
+    @Override
+    protected Analyzer getWrappedAnalyzer(String fieldName) {
```
davidclement90: add final

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+        if (clazz != null && String.class.isAssignableFrom(clazz) && mapping != null) {
+            // at the moment, we only try to support custom analyzers for string data.
+            // everything else falls through a StandardAnalyzer as was the case before
+            switch (mapping) {
+                case TEXTSTRING:
+                    throw new RuntimeException("TextString is an unsupported mapping for string data & custom analyzers");
+                case PREFIX_TREE:
+                    throw new RuntimeException("Prefix-tree is an unsupported mapping for string data & custom analyzers");
+                case STRING:
+                    if (stringAnalyzerString != null) analyzerString = stringAnalyzerString; else analyzerString = org.apache.lucene.analysis.core.KeywordAnalyzer.class.getName();
```
davidclement90: `private static String KEYWORD_ANALYZER = org.apache.lucene.analysis.core.KeywordAnalyzer.class.getName();
analyzerString = stringAnalyzerString != null ? stringAnalyzerString : KEYWORD_ANALYZER;`

 will be more readable

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+                    break;
+                case TEXT:
+                case DEFAULT:// TEXT
+                    analyzerString = textAnalyzerString;
+                    break;
+            }
```
davidclement90: Can you add default case, 'Not supported'

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        // the default (& fall through) Analyzer is a StandardAnalyzer
+        if ((clazz == null || !String.class.isAssignableFrom(clazz)) || mapping == null || analyzerString == null) {
+            analyzerString = org.apache.lucene.analysis.standard.StandardAnalyzer.class.getName();
```
davidclement90: add static attribut for StandardAnalyzer's name

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+      
+        String analyzerString = null;
+
+        if (clazz != null && String.class.isAssignableFrom(clazz) && mapping != null) {
+            // at the moment, we only try to support custom analyzers for string data.
+            // everything else falls through a StandardAnalyzer as was the case before
+            switch (mapping) {
+                case TEXTSTRING:
+                    throw new RuntimeException("TextString is an unsupported mapping for string data & custom analyzers");
```
davidclement90: Instead of Runtime, can you use PermanentBackendException, because it's a problem of configuration

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+        Analyzer analyzer = analyzers.computeIfAbsent(analyzerString, t -> {
+            try {
+                Class classDefinition = Class.forName(analyzerStringFinal);
+                return (Analyzer) classDefinition.newInstance();
+            } catch (Exception e) {
+                throw new RuntimeException("Analyzer cannot be instanciated for class " + analyzerStringFinal, e);
+            }
+        });
+        if (analyzer == null) {
+            throw new RuntimeException("Analyzer cannot be instanciated for class " + analyzerString);
```
davidclement90: Instead of Runtime, can you use PermanentBackendException, because it's a problem of configuration

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        // the default (& fall through) Analyzer is a StandardAnalyzer
+        if ((clazz == null || !String.class.isAssignableFrom(clazz)) || mapping == null || analyzerString == null) {
+            analyzerString = org.apache.lucene.analysis.standard.StandardAnalyzer.class.getName();
+        }
+
+        final String analyzerStringFinal = analyzerString;
+
+        Analyzer analyzer = analyzers.computeIfAbsent(analyzerString, t -> {
```
davidclement90: I'm not sure that lambda is the best here, because if you can not instantiate analyzer it's a problem of configuration. So it's a PermanentBackendException

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -73,6 +77,12 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.apache.lucene.analysis.Analyzer.PER_FIELD_REUSE_STRATEGY;
+import static org.janusgraph.core.attribute.Cmp.EQUAL;
+import static org.janusgraph.core.schema.Mapping.DEFAULT;
```
davidclement90: Can you remove static import

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -102,7 +113,7 @@
 
     private static final Map<Geo, SpatialOperation> SPATIAL_PREDICATES = spatialPredicates();
 
-    private final Analyzer analyzer = new StandardAnalyzer();
+    //private final Analyzer analyzer = new StandardAnalyzer();
```
davidclement90: dead code

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -137,10 +148,11 @@ private Directory getStoreDirectory(String store) throws BackendException {
         }
     }
 
-    private IndexWriter getWriter(String store) throws BackendException {
+    private IndexWriter getWriter(String store, KeyInformation.IndexRetriever informations) throws BackendException {
         Preconditions.checkArgument(writerLock.isHeldByCurrentThread());
         IndexWriter writer = writers.get(store);
         if (writer == null) {
+            Analyzer analyzer = new CustomAnalyzer(store, informations, PER_FIELD_REUSE_STRATEGY);
```
davidclement90: final

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -498,6 +515,116 @@ private static Query numericQuery(String key, Cmp relation, Number value) {
...
+        } else analyzerString = null;
+
+        List<String> terms;
+        if (analyzerString != null) {
+            try {
+                Class classDefinition = Class.forName(analyzerString);
+                Analyzer analyzer = (Analyzer) classDefinition.newInstance();
+                terms = customTokenize(analyzer, key, value);
+            } catch (Exception e) {
+                throw new RuntimeException("Analyzer cannot be instanciated for class " + analyzerString, e);
```
davidclement90: It's a permanent backend exception

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -498,6 +515,116 @@ private static Query numericQuery(String key, Cmp relation, Number value) {
...
+            return terms;
+        } catch (IOException e) {
+            throw new IllegalArgumentException(e.getMessage(), e);
+        } finally {
+            org.apache.commons.io.IOUtils.closeQuietly(stream);
+        }
+    }
+
+    private void tokenize(SearchParams params, KeyInformation.StoreRetriever information, String value, String key, JanusGraphPredicate janusgraphPredicate) {
+        Parameter[] parameters = information.get(key).getParameters();
```
davidclement90: add final

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -498,6 +515,116 @@ private static Query numericQuery(String key, Cmp relation, Number value) {
...
+                break;
+            default:
+                throw new RuntimeException("unsupported case " + mapping);
+        } else analyzerString = null;
+
+        List<String> terms;
+        if (analyzerString != null) {
+            try {
+                Class classDefinition = Class.forName(analyzerString);
+                Analyzer analyzer = (Analyzer) classDefinition.newInstance();
```
davidclement90: add final

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -498,6 +515,116 @@ private static Query numericQuery(String key, Cmp relation, Number value) {
...
+                q.add(new TermQuery(new Term(key, value)), BooleanClause.Occur.MUST_NOT);
+                params.addQuery(q.build(), BooleanClause.Occur.MUST);
+            } else if (janusgraphPredicate == Text.CONTAINS_PREFIX) {
+                BooleanQuery.Builder q = new BooleanQuery.Builder();
+                Term term;
+                if (mapping == STRING) term = new Term(key, terms.get(0)); else  term = new Term(key, terms.get(0).toLowerCase());
+                q.add(new TermQuery(term), BooleanClause.Occur.SHOULD);
+                q.add(new PrefixQuery(term), BooleanClause.Occur.SHOULD);
+                params.addQuery(q.build(), BooleanClause.Occur.MUST);
+
```
davidclement90: empty line

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -498,6 +515,116 @@ private static Query numericQuery(String key, Cmp relation, Number value) {
...
+                params.addQuery(new TermQuery(new Term(key, value)));
+            } else if (janusgraphPredicate == Cmp.NOT_EQUAL) {
+                BooleanQuery.Builder q = new BooleanQuery.Builder();
+                q.add(new MatchAllDocsQuery(), BooleanClause.Occur.MUST);
+                q.add(new TermQuery(new Term(key, value)), BooleanClause.Occur.MUST_NOT);
+                params.addQuery(q.build(), BooleanClause.Occur.MUST);
+            } else if (janusgraphPredicate == Text.CONTAINS_PREFIX) {
+                BooleanQuery.Builder q = new BooleanQuery.Builder();
+                Term term;
+                if (mapping == STRING) term = new Term(key, terms.get(0)); else  term = new Term(key, terms.get(0).toLowerCase());
```
davidclement90: It's not readable on github

Lakedaemon:
Thanks for the review. I'll make the necessary changes as soon as possible.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
Lakedaemon: If the lambda fails to instanciate the Analyzer, it's catch block will (from now on) throw a PermanentBackendException wrapping the original exception
If it succeeds, it will put the successfully built analyzer in the analyzers map cache

Lakedaemon: I can also remove it and use a few imperative commands... (it's less java 8-ish but only a few lines more)

Lakedaemon: Just done that, saving some object creations

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
Lakedaemon: I tried to but Idea is complaining because of the signature of getWrappedAnalyzer which is not supposed to throw checked exceptions and PermanentBackEndException extends Exception and not RuntimeException

I'm afraid the code won't compile if I do so (or is there another way ?)

Lakedaemon:
I made most of the changes (I also simplified the logic in two places in convertQuery, removing a BooleanQuery builder that wasn't really necessary) but could not use BackendPermanentException instead of RuntimeException because the class I use extends an abstract class whose methods cannot throw checked Exceptions.

Please advise on how to proceed next
Best regards,
Olivier

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -87,35 +83,37 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
...
-            throw new RuntimeException("Analyzer cannot be instanciated for class " + analyzerString);
+            analyzers.put(analyzerString, analyzer);
         }
         return analyzer;
     }
 
-}
-
+    private static String KEYWORD_ANALYZER = org.apache.lucene.analysis.core.KeywordAnalyzer.class.getName();
+    private static String STANDARD_ANALYZER = org.apache.lucene.analysis.standard.StandardAnalyzer.class.getName();
```
davidclement90: Can you put this with others class attribute on the top of the class ?

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
```diff
@@ -0,0 +1,119 @@
...
+ * JanusGraph may open different index stores and needs an analyzer for each
+ * so an analyzer should be provided to build IndexWriters/IndexReaders that are used when an index is opened and documents are added
+ * BUT a janusGraph may add a new indexable field, with a custom tokenizing process through Parameters, at any moment, which means that it should be possible
+ * to map any field to the corresponding custom analyzer, though indexreaders are already in use.
+ * <p>
+ * A way to achieve that would be to use a delegating analyzer that delegates Tokenizing for a field to
+ * the corresponding Analyzer (whose class you get through a Mapping)
+ */
+public class CustomAnalyzer extends DelegatingAnalyzerWrapper {
+    private static String KEYWORD_ANALYZER = org.apache.lucene.analysis.core.KeywordAnalyzer.class.getName();
```
davidclement90: final

Lakedaemon: Done. I'll also rebase these into a single commit too.

Many thanks for the kind review. 
Best regards,
Olivier

Lakedaemon:
Many thanks for the kind review.
Best regards,
Olivier

Lakedaemon:
Did I do something wrong ? A  merge aren't happening though David has approved the code. 
Should I get another approval for the code to be merged ?

Actually, I'm waiting for this to be merged to propose more pr (implementing Cardinality.Single & Cardinality.LIST for lucene, fixing the "lucene back end loses data  bug", improving test coverage, etc...)

Lakedaemon:
@sjudeng Many thanks for the feedback. Indeed, in my personal repo, my branch was called master.
Sorry about that. I'm not that good with git/github so, after I renamed the branch, I had to make another pr.
But luckily, github still recognizes the cla signature and david's approval.

Best regards, 
Olivier

sjudeng:
Closing this PR as the contribution is now being tracked under #909

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/CustomAnalyzer.java
Lakedaemon: @sjudeng the place where I was using computeIfAbsent

Issue #590, opened 2017-10-02
Evict graphs from JanusGraphManager cache on every node in cluster when applicable


dpitera:
Upon configuration updates and certain actions, graphs will need to be evicted from the JanusGraphManager cache to guarantee correctness when accessing graph references.

For example, if a user stores a configuration “Graph1” used to access “Graph1” in a 3 node JanusGraph cluster, and the user `drops` the graph using `ConfiguredGraphFactory.drop(“Graph1”)`, then the configuration for that graph will be deleted, and therefore the graph should be evicted from the cache on every node in the cluster. Otherwise, a user can use `ConfiguredGraphFactory.createConfiguration()` to create a new configuration for this graph, but then when they `ConfiguredGraphFactory.open(“Graph1”)`, the JanusGraphManager will already include the graph with the old configuration in the cache and the user will use that graph reference rather than the newly updated one. This gets really messy when the cache on each node in the cluster is not in the same state.

This can also happen when users 1. directly update a graph configuration or 2. update a template configuration (each graph configuration created using the template configuration must be updated). The graph should also probably be removed from all caches upon ConfiguredGraphFactory.close() method invocations as well, and perhaps as well for `JanusGraphFactory.close()`  and `.drop()` invocations (Although these calls have nothing to do with the ConfigurationManagementGraph updates or potential updates, so maybe cache eviction across the cluster is irrelevant for the JanusGraphFactory).

Currently, the documentation makes note of these scenarios and aids the user in ensuring they follow the right steps to guarantee correctness in the way they access graphs in this sticky areas, i.e. calling `JanusGraphFactory.close(graph);` or `ConfiguredGraphFactory.close("graphName");` on every node in the cluster. The reason JanusGraph did not automatically handle these situations themselves is because generally, JanusGraph nodes are not aware of each other and we did not want to make JanusGraph node-aware. And obviously to ensure consistent state across graph caches in the cluster, we need some form of communication between our JanusGraph  nodes. 

I propose that we make JanusGraph smart enough to handle these sticky situations. The method of communication can be the same that is used to move an index from the INSTALLED state to the REGISTERED state: the LogMessageQueue. In a nutshell, JanusGraph has the ability to push log messages onto a queue stored in the backend each graph is configured with, and every graph on every JanusGraph node is continuously polling the its backend’s message queue awaiting log messages and processing them when they arrive. We can use this mechanism to evict graphs from the JanusGraphManager cache.

PR #897, opened 2018-01-18
Evict graph from cache upon configuration updates

dpitera:
When a graph configuration is updated or removed from the
ConfigurationGraphManagement, the related graph is evicted from the
cache on all JanusGraph nodes in the cluster. This is done to prevent
users from opening graphs with old/incorrect configurations across the
cluster and finding themselves in an inconsistent state. Note that this
is only done for users updating the ConfigurationManagementGraph by
going through the ConfiguredGraphFactory APIs; this is done for two
reasons: 1. to allow more fine-tuned control for users who wish to use
the lower-levls CMG APIs and 2. to avoid a cyclical dependency between
the CMG and the JanusGraphManager.

Issues: #590

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/docs/configuredgraphfactory.adoc
+++ b/docs/configuredgraphfactory.adoc
@@ -85,7 +85,7 @@ for which you have instantiated _and_ the references are stored inside the `Janu
 
 IMPORTANT: This is an irreversible operation that will delete all graph and index data.
 
-IMPORTANT: To ensure all graph representations are consistent across all JanusGraph nodes in your cluster, remove the graph from the `JanusGraphManager` graph reference tracker on all nodes in your cluster: `ConfiguredGraphFactory.close("graphName");`.
+IMPORTANT: To ensure all graph representations are consistent across all JanusGraph nodes in your cluster, this removes the graph from the `JanusGraphManager` graph cache on every node in the cluster, assuming each node has been properly configured to use the `JanusGraphManager`.
 
 [[configuring-JanusGraph-server-for-configuredgraphfactory]]
 === Configuring JanusGraph Server for ConfiguredGraphFactory
@@ -210,16 +210,7 @@ the property `graph.graphname` go through the `JanusGraphManager` which
 keeps track of graph references created on the given JVM. Think of it as
 a graph cache. For this reason:
 
-IMPORTANT: Any updates to a configuration are not guaranteed to take effect until
-you remove the graph in question on every JanusGraph node in your
-cluster.
-
-You can do so by calling:
-
-[source, gremlin]
-----
-ConfiguredGraphFactory.close("graph2");
-----
+IMPORTANT: Any updates to a graph configuration results in the eviction of the relevant graph from the graph cache on every node in the JanusGraph cluster, assuming each node has been configured properly to use the `JanusGraphManager`.
 
 Since graphs created using the template configuration first create a
 configuration for that graph in question using a copy and create method,
@@ -232,8 +223,7 @@ configuration are not guaranteed to take effect on the specific graph
 until:
 
 1. The relevant configuration is removed: `ConfiguredGraphFactory.removeConfiguration("graph2");`
-2. The graph in question has been closed on every JanusGraph node: `ConfiguredGraphFactory.close("graph2");`
-3. The graph is recreated using the template configuration: `ConfiguredGraphFactory.create("graph2");`
+2. The graph is recreated using the template configuration: `ConfiguredGraphFactory.create("graph2");`
 ====
 
 [[update-examples]]
@@ -259,9 +249,6 @@ map.put("storage.hostname", "10.0.0.1");
 ConfiguredGraphFactory.updateConfiguration("graph1",
 map);
 
-// Close graph
-ConfiguredGraphFactory.close("graph1");
-
 // We are now guaranteed to use the updated configuration
 def g1 = ConfiguredGraphFactory.open("graph1");
 ----
@@ -286,9 +273,6 @@ map.put("index.search.hostname", "127.0.0.1");
 map.put("index.search.elasticsearch.transport-scheme", "http");
 ConfiguredGraphFactory.updateConfiguration("graph1",
 map);
-
-// Close graph
-ConfiguredGraphFactory.close("graph1");
 
 // We are now guaranteed to use the updated configuration
 def g1 = ConfiguredGraphFactory.open("graph1");
@@ -316,9 +300,6 @@ MapConfiguration(map));
 
 // Remove Configuration
 ConfiguredGraphFactory.removeConfiguration("graph1");
-
-// Close graph on all JanusGraph nodes
-ConfiguredGraphFactory.close("graph1");
 
 // Recreate
 ConfiguredGraphFactory.create("graph1");
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/GraphCacheEvictionAction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/GraphCacheEvictionAction.java
@@ -1,0 +1,22 @@
+// Copyright 2018 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.database.management;
+
+public enum GraphCacheEvictionAction {
+
+    EVICT,
+    DO_NOT_EVICT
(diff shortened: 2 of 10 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -744,6 +746,30 @@ public IndexJobFuture updateIndex(Index index, SchemaAction updateAction) {
         return future;
     }
 
+    /**
+     * Upon the open managementsystem's commit, this graph will be evicted from the cache on all JanusGraph nodes in your
+     * cluster, once there are no open transactions on this graph on each respective JanusGraph node.
```
dpitera: and assuming each node in the cluster is configured correctly to use the JGM

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
```diff
@@ -191,12 +208,24 @@ public void run() {
                         txStillOpen = true;
                     }
                 }
-                if (!txStillOpen) {
+                final JanusGraphManager jgm = JanusGraphManager.getInstance();
+                final boolean janusGraphManagerIsInBadState = null == jgm && action.equals(EVICT);
+                if (!txStillOpen && janusGraphManagerIsInBadState) {
+                    log.error("JanusGraphManager should be instantiated on this server, but it is not. " +
```
sjudeng: Can you add a test to cover this case?

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
```diff
@@ -191,12 +208,24 @@ public void run() {
                         txStillOpen = true;
                     }
                 }
-                if (!txStillOpen) {
+                final JanusGraphManager jgm = JanusGraphManager.getInstance();
+                final boolean janusGraphManagerIsInBadState = null == jgm && action.equals(EVICT);
```
sjudeng: If `jgm` is null couldn't that just mean that JanusGraph Server is not being used?

dpitera: The only way `action.equals(EVICT)` is if the user actively called `mgmt.evictGraphFromCache()` on their graph's mangementSystem, or if it was called by the ConfiguredGraphFactory [updateConfiguration](https://github.com/dpitera/janusgraph/blob/06d519e48dd64ae8df4dda5c1e94ca34d2eac250/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java#L203) or [removeConfiguration](https://github.com/dpitera/janusgraph/blob/06d519e48dd64ae8df4dda5c1e94ca34d2eac250/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java#L229). 

The issue in this scenario is indeed that a user is using the JanusGraph Server (more specifically configured to use the JanusGraphManager) for one of their JG nodes but not the other node (the one that _received_ this eviction request).

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
dpitera: Note that this would be hit asynchronously on a node other than the one the original action was made on, so this is a log action-- not the throwing of an error. I can add a test that ensures if the JGM is null that we don't remove it from the cache, but that is pretty obvious.

In terms of testing this functionality across multi-node deployments, I would like to avoid if we can as it would be ~non-trivial as there is no infra in place to do so, and there is also no infra in place to test the existing ManagementLogger.

sjudeng: ManagementLogger seems to be otherwise well covered in tests (see Codecov). If the test you describe hits this case I think it's worth including.

dpitera: That code-coverage analysis is pretty cool!

I can hit the case with the simpler test of opening a graph without configuring the JGM and trying to evict the graph from the cache.

dpitera:
@sjudeng Pushed new code with your suggested updates and I added 2 tests, both testing the graph management logger's graph cache eviction logic; one when the JGM is null (mostly an obvious test, but we will see the code get covered in the codecov analysis) and one when it is not (the test proving the functionality works, i.e. the log can successfully evict graphs from the cache).

dpitera:
> Also do you think we'll eventually want to move JanusGraphManager to the janusgraph-server module?

Sorry, I forgot to answer this question. Artificially, _maybe_ the JanusGraphManager should live in the `janusgraph-server` module. I think it's easy to sound like this is the right place to live, especially when one considers that the notion of a GraphManager is a TinkerPop ideation. Concretely, this gets complicated because things like the JanusGraphFactory, ConfiguredGraphFactory, things that live in the `janusgraph-core` module... rely on the JanusGraphManager; this would cause a cyclical dependency between the two modules. Even theoretically I am more of the notion that the `core` project is indeed the best place for the JGM to reside. What do you think? Might not be a bad thing over which to open a discussion. Either way, if the community deems it is best to move, then it would be best done in a separate PR with its own refactoring issue.

sjudeng:
Thanks for the explanation. It looks to me like ConfiguredGraphFactory could maybe be moved over as is. I don't see any integration to other classes in core. JGM would be more painful given integration in JanusGraphFactory and (with this PR) ManagementLogger. Can we create a separate issue to investigate further, maybe specifically for moving over ConfiguredGraphFactory if you agree that would be straight-forward, as well as for identifying/moving any other relevant classes if any?

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
```diff
@@ -191,12 +211,24 @@ public void run() {
...
+                if (!txStillOpen && janusGraphManagerIsInBadState) {
```
sjudeng: Looks like this branch still isn't getting hit (see [here](https://codecov.io/gh/JanusGraph/janusgraph/src/graph-cache-evictions/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java#L214...221)). It's not a big deal.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
```diff
@@ -30,12 +30,16 @@
...
+import org.janusgraph.graphdb.management.JanusGraphManager;
 import org.janusgraph.graphdb.types.vertices.JanusGraphSchemaVertex;
+import static org.janusgraph.graphdb.database.management.GraphCacheEvictionAction.EVICT;
+import static org.janusgraph.graphdb.database.management.GraphCacheEvictionAction.DO_NOT_EVICT;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
+import java.util.HashSet;
```
pluradj: this isn't used

docs/configuredgraphfactory.adoc
```diff
@@ -83,7 +83,7 @@ for which you have instantiated _and_ the references are stored inside the `Janu
 
 IMPORTANT: This is an irreversible operation that will delete all graph and index data.
 
-IMPORTANT: To ensure all graph representations are consistent across all JanusGraph nodes in your cluster, remove the graph from the `JanusGraphManager` graph reference tracker on all nodes in your cluster: `ConfiguredGraphFactory.close("graphName");`.
+IMPORTANT: To ensure all graph representations are consistent across all JanusGraph nodes in your cluster, this removes the graph from the `JanusGraphManager` graph cache on every node in the cluster, assuming each node has been properly configured to use the `JanusGraphManager`.
```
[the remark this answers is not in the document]
dpitera: I think this suggestion makes alot of sense; I prefer to leave the call out and link to the new section about using a multinode JG cluster. I also think the information in this PR [here](https://github.com/JanusGraph/janusgraph/pull/884#issuecomment-368149520) will be relevant, so I suggest creating a _separate_ PR with this new doc section that includes the information relevant to both PRs; this will avoid merge-conflicts between these two PRs.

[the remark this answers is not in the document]
pluradj: separate PR for docs update is fine, see https://github.com/JanusGraph/janusgraph/issues/938

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/GraphCacheEvictionAction.java
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.database.management;
+
+public enum GraphCacheEvictionAction {
+
+    EVICT,
```
amcp: consider documenting each of the enum values with a Javadoc.

Issue #948, opened 2018-03-13
add OrderPreservingSerializer support on UUIDSerializer


pluradj:
From @krlohnes on [janusgraph-dev](https://groups.google.com/d/msg/janusgraph-dev/BPyoK0KhvrI/RaVIIFr7AAAJ):

In the docs it states that

>The property keys used in a vertex-centric index must have an explicitly defined data type (i.e. not Object.class) which supports a native sort order.

I'm curious as to exactly what that means. As far as I can tell it just means "implements `Comparable`". If so, is there any reason to not have UUID implement `OrderPreservingSerializer`? 

If it's something that can/should be done, I'm happy to submit a PR for it.

PR #978, opened 2018-03-29
Add OrderPreservingSerializer on UUIDSerializer

krlohnes:
Issue #948

As far as I can tell, supporting "native sort order", as described in
the docs, simply means implementing `Comparable`, which UUID certainly
does. This adds support for using a UUID value in a Vertex Centric
Index.

Signed-off-by: Keith Lohnes <lohnesk@gmail.com>

-----

### For all changes:
- [ X ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ X ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [ X ] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [ X ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ X ] Have you written and/or updated unit tests to verify your changes?

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/attribute/UUIDSerializer.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/attribute/UUIDSerializer.java
@@ -14,17 +14,18 @@
 
 package org.janusgraph.graphdb.database.serialize.attribute;
 
-import com.google.common.base.Preconditions;
-import org.janusgraph.core.attribute.AttributeSerializer;
+import java.util.UUID;
+
 import org.janusgraph.diskstorage.ScanBuffer;
 import org.janusgraph.diskstorage.WriteBuffer;
+import org.janusgraph.graphdb.database.serialize.OrderPreservingSerializer;
 
-import java.util.UUID;
+import com.google.common.base.Preconditions;
 
 /**
  *  @author Bryn Cooke (bryn.cooke@datastax.com)
  */
-public class UUIDSerializer implements AttributeSerializer<UUID>  {
+public class UUIDSerializer implements OrderPreservingSerializer<UUID>  {
 
     @Override
     public UUID read(ScanBuffer buffer) {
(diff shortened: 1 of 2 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/attribute/UUIDSerializer.java
```diff
@@ -33,12 +34,6 @@ public UUID read(ScanBuffer buffer) {
         return new UUID(mostSignificantBits, leastSignificantBits);
     }
 
-    @Override
```
amcp: minor, it seems like the content of the write() method did not change. so, can you keep it in the same place to reduce the number of diffs in this PR?

dpitera:
LGTM; do we want to update docs so people know that they can now use UUIDs with VCIs?

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/serialize/attribute/UUIDSerializer.java
krlohnes: @acmp Sure thing. Fixed,

krlohnes:
@dpitera @pluradj The docs already have a "note"

> The property keys used in a vertex-centric index must have an explicitly defined data type (i.e. not Object.class) which supports a native sort order.

Do you want me to add a list to that section of types supported by VCIs? Or do you think it would be enough to change the above section to

>The property keys used in a vertex-centric index must have an explicitly defined data type (i.e. not Object.class) which supports a native sort order (implements `Comparable) and whose serializer implements OrderPreservingSerializer

I'm leaning towards just adding a list of officially supported datatypes for vertex centric indexes, but I'd like to sit down and write tests for that list if that's the way you want to go.

krlohnes:
The test failing right now is for Solr. It wasn't failing before. Can the just the solr tests be rerun?

sjudeng:
Restarted the Solr job

dpitera:
> Do you want me to add a list to that section of types supported by VCIs? Or do you think it would be enough to change the above section to

If you have the cycles to do the former suggestion in a timely fashion, then yeah that's great. The docs would be more user-friendly and those tests would be a nice-to-have anyway. But if not, I personally am cool with ur latter suggestion.

Issue #1103, opened 2018-06-08
Cassandra Privilege "Create on Table"


BeatSager:
The issue [Cassandra Privilege "Create on All Keyspaces"](https://github.com/JanusGraph/janusgraph/issues/933) has already been solved.
Unfortunately using cql the same issue is still preset for create table within a keyspace. 
```
CQLKeyColumnValueStore

    private static void initializeTable(final Session session, final String keyspaceName, final String tableName, final Configuration configuration) {
        session.execute(createTable(keyspaceName, tableName)
                .ifNotExists()
                .addPartitionKey(KEY_COLUMN_NAME, DataType.blob())
                .addClusteringColumn(COLUMN_COLUMN_NAME, DataType.blob())
                .addColumn(VALUE_COLUMN_NAME, DataType.blob())
                .withOptions()
                .compressionOptions(compressionOptions(configuration))
                .compactionOptions(compactionOptions(configuration))
                .compactStorage());
    }
```
We would like to run janusgraph without having to give the cassandra user this privilege.

Possible solution for discussion:
https://github.com/BeatSager/janusgraph/commit/d8e1a5008abab4e4fa22482d5295f73ff25a886e

pluradj:
@BeatSager Thank you for opening up the issue. It would be best if you submit your solution as a pull request. Please file an [individual or corporate CLA](https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md) first. Thanks.

BeatSager:
https://github.com/JanusGraph/janusgraph/pull/1110

PR #1110, opened 2018-06-13
Issue #1103 Cassandra Privilege "Create on Table"

BeatSager:
If security is enabled on Cassandra the user must not have the rights to create tables when they are created ahead of time by an administrator

-----

### For all changes:

### For code changes:

### Note:


docs/cassandra.adoc
```diff
@@ -19,7 +19,7 @@ If you are using Cassandra 2.2 or higher you need to explicitly enable thrift so
 Do so by running `./bin/nodetool enablethrift`.
 [NOTE]
 If security is enabled on Cassandra, the user must have `CREATE permission on <all keyspaces>`,
-otherwise the keyspace must be created ahead of time by an administrator.
+otherwise the keyspace must be created ahead of time by an administrator including the required tables or the user must have `CREATE permission on <the configured keyspace>`.
```
pluradj: We should create a cql file with the required tables in it to make it easier for the Cassandra administrator to create the tables on behalf of the JanusGraph user. Then add to these docs on where to find the cql file in the JanusGraph distribution and how to load it with cqlsh.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -143,16 +144,19 @@
      * @param tableName the name of the database table for storing the key/column/values
      * @param configuration data used in creating this store
      * @param closer callback used to clean up references to this store in the store manager
+     * @param shouldInitializeTable if true is provided the table gets initialized
      */
-    public CQLKeyColumnValueStore(final CQLStoreManager storeManager, final String tableName, final Configuration configuration, final Runnable closer) {
+    public CQLKeyColumnValueStore(final CQLStoreManager storeManager, final String tableName, final Configuration configuration, final Runnable closer, final Supplier<Boolean> shouldInitializeTable) {
```
pluradj: I'm curious why you chose to pass the `Supplier` from `CQLStoreManager` here. It seems like keeping the table existence checking logic contained in `CQLKeyColumnValueStore` would be consistent with how the keyspace is handled.

docs/cassandra.adoc
BeatSager: [janusgraphCassandraTables.txt](https://github.com/JanusGraph/janusgraph/files/2122228/janusgraphCassandraTables.txt)

I would say this is the current situation. The required tables are not defined in the cql module but in core (Backend.java). Placing the file in core seems to be wrong and placing it in cql module it might be hard to keep this document updated.

I started to use janusgraph not a long time ago. As a developer I set it up locally and janusgraph created the keyspace containing all the tables. This way I had the current schema. Bad would just be when janusgraph would create some tables later during runtime.

If a cassandra admin does not want janusgraph to do any schema modification, then I think the knowhow how to handle a cql file should already be there.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
BeatSager: I decided to do it that way for two reasons:
# The cluster containing the required metadata is private to the `CQLStoreManager` and can therefore not be accessed in `CQLKeyColumnValueStore`. Provide this information on the manager might be an option as well, but this would couple the manager an the store more together.
# Using `Supplier` instead of passing the information directly I chose because I did not want to  evaluate this for an open store. (`KeyColumnValueStore` already in `this.openStores`)

pluradj: Thanks for the explanation.

docs/cassandra.adoc
pluradj: I suggest placing the file under `janusgraph-dist/src/assembly/static/conf/cassandra/`, which would end up in the zip distribution under `conf/cassandra/`. I wouldn't expect the file to be updated frequently.

Would you be able to add the file and a note in the docs pointing to the file location? Thanks.

BeatSager: Thank you for your input. I added the file and updated the documentation.

pluradj:
@BeatSager could you please fetch the latest `0.2` branch and resolve the conflicts? thanks.

pluradj:
@BeatSager could you please go ahead and squash your commits and make sure the CI goes through? thanks.

BeatSager:
CI was only skipped for the second commit, since this one contains documentation only.

PR #1141, opened 2018-06-28
Delete only janusgraph-aliased indices in Elasticsearch clearStorage

### LINKED ISSUE #1139: drop graph, but drop all index. On es 6.0
【hbase + es 6.0】
when I execute JanusGraphFactory.drop(graph)
All indices are drop. 

screenshot from 2018-06-27 20-37-34

Is it bug? Why?

sjudeng:
Fixes #1139

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
@@ -217,20 +217,23 @@ public class RestElasticSearchClient implements ElasticSearchClient {
 
     @Override
     public void deleteIndex(String indexName) throws IOException {
-        if (majorVersion.getValue() < 6 && indexExists(indexName)) {
-            performRequest(REQUEST_TYPE_DELETE, REQUEST_SEPARATOR + indexName, null);
-        } else {
-            final Response response = performRequest(REQUEST_TYPE_GET, REQUEST_SEPARATOR + "_cat" + REQUEST_SEPARATOR + "aliases" + REQUEST_PARAM_BEGINNING + "format=json", null);
+        if (isAlias(indexName)) {
+            // aliased multi-index case
+            final String path = new StringBuilder(REQUEST_SEPARATOR)
+                .append("_alias").append(REQUEST_SEPARATOR).append(indexName).toString();
+            final Response response = performRequest(REQUEST_TYPE_GET, path, null);
             try (final InputStream inputStream = response.getEntity().getContent()) {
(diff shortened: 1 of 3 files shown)


janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -219,13 +219,14 @@ public IndexMapping getMapping(String indexName, String typeName) throws IOExcep
     public void deleteIndex(String indexName) throws IOException {
         if (majorVersion.getValue() < 6 && indexExists(indexName)) {
             performRequest(REQUEST_TYPE_DELETE, REQUEST_SEPARATOR + indexName, null);
-        } else {
-            final Response response = performRequest(REQUEST_TYPE_GET, REQUEST_SEPARATOR + "_cat" + REQUEST_SEPARATOR + "aliases" + REQUEST_PARAM_BEGINNING + "format=json", null);
+        } else if (isAlias(indexName)) {
```
jerryjch: We will always have alias for majorVersion >= 6?  If so, could you add a comment to explain the fact a little?

sjudeng: @jerryjch Thanks for the review. I updated the implementation and also added a note. Elasticsearch doesn't support multi-type indices since version 6.0, so the indices should always be aliased. But the JanusGraph [USE_DEPRECATED_MULTITYPE_INDEX](https://github.com/JanusGraph/janusgraph/blob/176f9db0d6f4e38a5f66214de59c28874e214316/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java#L165-L168) setting can be used to force the legacy non-aliased index, and technically this could be set even with ES 6.x as long as the user only has one type. So the case is valid. I added (updated) a test case for it.

jerryjch: What is the difference for the @Test @Override clearStorageTest() in ElasticSearchMultiTypeIndexTest?  The same in parent will be executed anyway?

sjudeng: @jerryjch ElasticSearchMultiTypeIndexTest overrides setUp and tearDown to add an explicit clear method to purge all janusgraph indices before and after all tests. But the clearStorageTest calls setUp and tearDown inside the test body and so it was passing in the multi-type index cases with ES6 not because clearStorage/deleteIndex was working (it wasn't) but because of that clear method in the test class. The difference in the overridden test is that  it calls the base (super) setUp and tearDown methods so the local clear call isn't executed and the test can properly test index.clearStorage.

Issue #1138, opened 2018-06-26
Add support for Kerberized SOLR


blindenvy:
Janus will fail to index data when using a kerberized SOLR as the indexing backend. For both modes: `CLOUD` and `HTTP` there is a need to set the appropriate configurations for kerberos in order for the SOLR client to successfully authenticate.

blindenvy:
This is similar but different then #1056

blindenvy:
The problematic code is in client creation starting around [line 242](https://github.com/JanusGraph/janusgraph/blob/6f7d300d8c739896eb4fb483d87535209353c9fc/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java#L242) in the 0.2.0 release branch.

hoda1369:
Any updates about this issue? I am trying to connect to kerberized Solr but it fails with:

Exception in thread "main" org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at http://10.0.130.40:8983/solr: Expected mime type application/octet-stream but got text/html. <html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 401 require authentication</title>
</head>
<body><h2>HTTP ERROR 401</h2>
<p>Problem accessing /solr/admin/collections. Reason:
<pre>    require authentication</pre></p>
</body>
</html>

	at org.apache.solr.client.solrj.impl.HttpSolrClient.executeMethod(HttpSolrClient.java:591)
	at org.apache.solr.client.solrj.impl.HttpSolrClient.request(HttpSolrClient.java:253)
	at org.apache.solr.client.solrj.impl.HttpSolrClient.request(HttpSolrClient.java:242)
	at org.apache.solr.client.solrj.impl.LBHttpSolrClient.doRequest(LBHttpSolrClient.java:483)
	at org.apache.solr.client.solrj.impl.LBHttpSolrClient.request(LBHttpSolrClient.java:413)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.sendRequest(CloudSolrClient.java:1121)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:862)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.request(CloudSolrClient.java:793)
	at org.apache.solr.client.solrj.SolrRequest.process(SolrRequest.java:178)
[...]

PR #1230, opened 2018-08-23
Add support for kerberized solr instance (#1138)

blindenvy:
Fixes issue #1138.

Signed-off-by: Christopher Jackson <chris.jackson@us.ibm.com>

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
+++ b/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
@@ -216,15 +216,15 @@ public class SolrIndex implements IndexProvider {
             "When mutating - wait for the index to reflect new mutations before returning. This can have a negative impact on performance.",
             ConfigOption.Type.LOCAL, false);
 
-    
+
     /** Security Configuration */
-    
+
     public static final ConfigOption<Boolean> KERBEROS_ENABLED = new ConfigOption<Boolean>(SOLR_NS,"kerberos-enabled",
             "Whether SOLR instance is Kerberized or not.",
-            ConfigOption.Type.GLOBAL_OFFLINE, false);
+            ConfigOption.Type.MASKABLE, false);
     public static final ConfigOption<String> KERBEROS_CONFIG = new ConfigOption<String>(SOLR_NS,"kerberos-config",
             "The absolute path to the JAAS configuration file for providing kerberos configuration to SOLR.",
-            ConfigOption.Type.GLOBAL_OFFLINE, String.class);
+            ConfigOption.Type.MASKABLE, String.class);
 
     private static final IndexFeatures SOLR_FEATURES = new IndexFeatures.Builder()
         .supportsDocumentTTL()
@@ -261,13 +261,13 @@ public class SolrIndex implements IndexProvider {
         batchSize = config.get(INDEX_MAX_RESULT_SET_SIZE);
         ttlField = config.get(TTL_FIELD);
         waitSearcher = config.get(WAIT_SEARCHER);
-        
+
         if (kerberosEnabled) {
             logger.debug("Kerberos is enabled. Configuring SOLR for Kerberos.");
             configureSolrClientsForKerberos();
         } else {
             logger.debug("Kerberos is NOT enabled.");
-            logger.debug("KERBEROS_ENABLED name is " + KERBEROS_ENABLED.getName() + " and it is" + (KERBEROS_ENABLED.isOption() ? " " : " not") + " an option."); 
+            logger.debug("KERBEROS_ENABLED name is " + KERBEROS_ENABLED.getName() + " and it is" + (KERBEROS_ENABLED.isOption() ? " " : " not") + " an option.");
             logger.debug("KERBEROS_ENABLED type is " + KERBEROS_ENABLED.getType().name());
         }
         final ModifiableSolrParams clientParams = new ModifiableSolrParams();
@@ -315,7 +315,7 @@ public class SolrIndex implements IndexProvider {
                 @Override
                 public void process(HttpRequest request, HttpContext context) throws HttpException, IOException {
                     if(request instanceof HttpEntityEnclosingRequest) {
-                        HttpEntityEnclosingRequest enclosingRequest = ((HttpEntityEnclosingRequest) request);  
+                        HttpEntityEnclosingRequest enclosingRequest = ((HttpEntityEnclosingRequest) request);
                         HttpEntity requestEntity = enclosingRequest.getEntity();
                         enclosingRequest.setEntity(new BufferedHttpEntity(requestEntity));
                     }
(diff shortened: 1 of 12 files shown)


blindenvy:
Background
===

HBase will use the typical hadoop configuration files to configure itself. For example the files founds in `/etc/hbase/conf` and `/etc/hadoop/conf`. If Hadoop Core Services (HDFS, YARN, etc.) and HBase are configured for Kerberos it will be reflected by the pertinent properties in these files. 

Both Zookeeper and SOLR gets its kerberos configuration from flags set when the corresponding service processes are started.

Each of these aforementioned services will have their own JAAS configuration files. In each of those files will be various JAAS configuration elements, some for the server processes of that service and potentially some for the client processes. Sometimes the client/server portions are split out into unique jass configuration files.

Keytabs aren't typically shared between the services. Each service will have a unique principal, sometimes multiple and the keytab is a stored credential(s) for a specific principal(s). 

Janus on Kerberized Hadoop
===
When working with Janus in a Hadoop environment, lets say using HBase as the storage backend and SOLR as the indexing back end there are two main things that need to happen for Janus to work properly:

1) The HBase and Hadoop configuration folders (ex. `/etc/hbase/conf` and `/etc/hadoop/conf`) must be on the classpath. This will ensure that when Janus attempts to communicate with HBase it knows that Kerberos is in play and does the necessary things automatically for authentication. 

2) Janus makes use of the SOLRJ library, and that library must also be made aware that Kerberos is in play. Hence this PR. The changes in this PR configure the solrj clients (both standalone and cloud) to make use of a JAAS configuration file to handle the necessary authentication.

    The JAAS configuration file needs one section called: `Client`. In Solr Cloud mode this configuration pulls double duty as it handles authentication for both internode requests and requests to ZooKeeper (Since Solr Cloud Mode uses Zookeeper).

Your Question
===
Even if SOLR and HBase share the same Zookeeper there should be no conflicts with using them together with Janus. As a user I can authenticate as a single principal by using a Keytab or TGT and then start using Janus which will in turn interact with HBase, Solr, Zookeeper, etc.

Sorry for the long winded answer but I thought the background and context would be helpful. Please let me know if you have any other questions or concerns.

jerryjch:
Thanks for the very good explanation!
Just to be clear.  Suppose Zookeeper server is a secure server.  HBase's Zookeeper client will share the the same JAAS configuration as the one introduced by this PR, and share the same 'client' section?

blindenvy:
I will answer your question in two parts.

1. The HBase daemons (Master, and RegionServer) will **NOT** share this configuration. They will be configured to use their own JAAS configuration files which will dictate authentication configuration with Zookeeper. 

2. It is not immediately clear to me if the HBase Client API (used by Janus Graph) will share the configuration. Once we call `System.setProperty("java.security.auth.login.config", kerberosConfig);` that property will persist for the lifecycle of that particular JVM or until that property is overwritten. However, I don't know that the HBase Client code uses that property to configure auth settings to interact with Zookeeper.

I took a look at the HBase documentation and from this [section](http://hbase.apache.org/book.html#_client_side_configuration_for_secure_operation) I took note of:

> Before 2.2.0 version, the client environment must be logged in to Kerberos from KDC or keytab via the kinit command before communication with the HBase cluster will be possible.

and this [section](http://hbase.apache.org/book.html#_securing_zookeeper_data) I took note of:

> In cases where ZooKeeper is used for service discovery or sharing state with the client, the znodes created by HBase will also allow anyone (regardless of authentication) to read these znodes (clusterId, master address, meta location, etc), but only the HBase user can modify them.

Based upon the information in the docs I think that the HBase Client code will also not share this configuration.

Further Notes:

The JAAS configuration file I have included in my PR is just for testing purposes and used for the unit tests and isn't shipped with the product. 

This PR expects the end user to add additional configuration in the `janusgraph.properties` file that specifies kerberos is being used and the location of the JAAS configuration file we want to use for authentication.

blindenvy:
Any more feedback @jerryjch?

jerryjch:
@blindenvy I will get back to it.  Sorry about the delay,

pluradj:
@blindenvy I think it would be worthwhile to add some pointers in the Solr docs to some of the Kerberos resources you mentioned in the discussion above. Although we don't need to completely duplicate those docs, making it easier for other JanusGraph users to find them would be helpful, especially with respect to the HBase, ZooKeeper, and Solr interplay.

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -205,6 +216,16 @@ public static Mode parse(String mode) {
             "When mutating - wait for the index to reflect new mutations before returning. This can have a negative impact on performance.",
             ConfigOption.Type.LOCAL, false);
 
+    
+    /** Security Configuration */
+    
+    public static final ConfigOption<Boolean> KERBEROS_ENABLED = new ConfigOption<Boolean>(SOLR_NS,"kerberos-enabled",
+            "Whether SOLR instance is Kerberized or not.",
+            ConfigOption.Type.GLOBAL_OFFLINE, false);
```
pluradj: I'm not sure how valuable this parameter is if it's configured as `GLOBAL_OFFLINE`. Do users typically toggle Kerberos on and off? If not, it seems like you could just use `KERBEROS_CONFIG` as the only config parameter for whether to use Kerberos or not.

blindenvy: > Do users typically toggle Kerberos on and off?
No I would say that is not a typical use case.

pluradj: If it makes sense to support changing the `KERBEROS_ENABLED` setting per connection (for example, if Solr or HBase support it also), then you should declare it as `ConfigOption.Type.MASKABLE` rather than `ConfigOption.Type.GLOBAL_OFFLINE`.

blindenvy: After discussion I will make the change to `MASKABLE`.

jerryjch:
@blindenvy I am still a little worried about setting the java.security.auth.login.config property for the whole JVM programmably for Solr only that may overwrite setting that may come from other places (e.g in the java command line from gremlin server) and/or for other clients.
Can we do a check in the new code to see if the system property java.security.auth.login.config is set or not, and if it is, give a warning but don't overwrite it?

jerryjch:
HBase secure config with Zookeeper is [here](https://hbase.apache.org/book.html#zk.sasl.auth).

blindenvy:
@jerryjch @pluradj I agree with both your assessments. There is some concern with that overriding a prior set configuration also its typical for jaas config to be set via command line through jvm options instead of programatically. 

Thus, I have removed the jaas config option from janusgraph properties. Instead `System.getProperty("java.security.auth.login.config")` will be used to pick up the appropriate jaas config. 

This will require that end users add this JVM option when invoking janus graph. 

For instance to run gremlin.sh you would need to do:
```
export JAVA_OPTIONS="-Djava.security.auth.login.config=/absolute/path/jaas.conf"
$JANUSGRAPH_HOME/bin/gremlin.sh
```

PR #1238, opened 2018-08-30
Simple-fix grammar in javadoc of indexQuery method

tonit:
I am not a native speaker, but i think my change is more correct..

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java
@@ -68,7 +68,7 @@ public interface Transaction extends Graph, SchemaManager {
      * Returns a {@link org.janusgraph.core.JanusGraphIndexQuery} to query for vertices or edges against the specified indexing backend using
      * the given query string. The query string is analyzed and answered by the underlying storage backend.
      * <p>
-     * Note, that using indexQuery will may ignore modifications in the current transaction.
+     * Note, that using indexQuery may ignore modifications in the current transaction.
      *
      * @param indexName Name of the index to query as configured
      * @param query Query string


pluradj:
@tonit I agree with your wording update. JanusGraph currently requires a CLA for contributions. Is that something you're interested in filing?

janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java
```diff
@@ -68,7 +68,7 @@
      * Returns a {@link org.janusgraph.core.JanusGraphIndexQuery} to query for vertices or edges against the specified indexing backend using
      * the given query string. The query string is analyzed and answered by the underlying storage backend.
      * <p>
-     * Note, that using indexQuery will may ignore modifications in the current transaction.
```
chupman: While it is an improvement grammatically I actually disagree with @pluradj on this one. I believe it should be that using indexQuery **will** ignore modifications in the current transaction. 

If you're doing a direct query against the index any changes made in the current open transaction will not be written yet, and therefore not available. I will yield and approve if Jason still thinks **may** is more appropriate.

tonit:
Will file the CLA since we are adopting JanusGraph in a project and its likely there is more contact ;)

tonit:
Update, CLA filed on Monday morning, but does not appear on https://github.com/JanusGraph/legal/blob/master/CLA_SIGNERS.yaml yet.

janusgraph-core/src/main/java/org/janusgraph/core/Transaction.java
pluradj: This comment has been in place going back to Titan. I haven't dug into the code to verify whether "will" or "may" is correct. My impression was that there could be some doubt, perhaps given different characteristics of the chosen storage backend, so that's why "may" might make more sense.

Issue #205, opened 2017-04-08
Add pointers for accessing JanusGraph via other libraries


mbrukman:
We've received a number of questions for how to access JanusGraph from different programming languages, or native libraries with bindings to access JanusGraph other than Java.

We should either collect a set of well-tested libraries that work with JanusGraph either via TinkerPop or Gremlin, or link to existing lists such as:

* https://github.com/mohataher/awesome-tinkerpop#wrappers
* http://tinkerpop.apache.org/

that are maintained elsewhere.

FlorianHockmann:
I agree that it would be a good idea to include this information in the documentation. It's a very important decision point for most people that evaluate different databases how good the support in their programming language is. Unfortunately, some people will probably not directly grasp completely how JanusGraph and TinkerPop are related and that they can check the TinkerPop website for such information.
But JanusGraph has a real advantage in this regard thanks to the Gremlin language variants (GLVs) which make it feel very natural to work with Gremlin compared to other databases where you usually have to construct some query string and send that manually to the database.
So the documentation could explain that JanusGraph can be queried with Gremlin and that therefore all Gremlin drivers can be used with JanusGraph. Then a short explanation of Gremlin language variants with a link to the respective section in the TinkerPop documentation would show how easy it is to work with Gremlin in one of the languages with a GLV (currently only Java, Groovy, and Python with JavaScript and C# under way).

babarinde:
+1

FlorianHockmann:
I wrote some documentation on how to connect to JanusGraph Server in general and from .NET in particular for the JanusGraph.Net library. Previews for these docs are available via GitHub:

* [Connecting to JanusGraph ](https://github.com/GDATASoftwareAG/janusgraph/blob/gremlin-dotnet-extensions/docs/connecting.adoc) is the main part of the added chapter.
* [Connecting from .NET](https://github.com/GDATASoftwareAG/janusgraph/blob/gremlin-dotnet-extensions/docs/connecting-via-dotnet.adoc) first explains how to connect to JanusGraph Server with just Gremlin.Net and the second part is for the JanusGraph.Net library.

If you think that this is going in the right direction, then I could add similar sections for Java and maybe Python and then create a PR for this ticket.

PR #1282, opened 2018-10-04
Support list for Zookeeper URL for Solr cloud and index job #1165

### LINKED ISSUE #1165: Support list for Zookeeper URL for Solr cloud and index job
Reported on [janusgraph-users](https://groups.google.com/d/msg/janusgraph-users/x5CnYVpr1ls/BNfwNQ9dCAAJ)

```
15:44:17 WARN  org.apache.hadoop.hbase.zookeeper.RecoverableZooKeeper  - Unable to create ZooKeeper Connection
java.net.UnknownHostException: [app01: invalid IPv6 address
```

Where the graph configuration properties

```
index.search.solr.zookeeper-url=app01:2181,app02:2181,app03:2181
```

The [`ZOOKEEPER_URL`](https://github.com/JanusGraph/janusgraph/blob/0.2/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java#L161-L163) is defined as a `String` property rather than a `String[]` (for example, `storage.hostname` and `index.[X].solr.http-urls` use `String[]`). If the property allows a list, it would allow for a quorum of Zookeeper nodes to be specified.

jerryjch:
Signed-off-by: Jerry He <jerryjch@apache.org>

Fixes #1165 

Fixed two things:
1.  The solr.zookeeper-url option should be String[] instead of String. See #1165 for error details.
2.  Fixed MapReduceIndex job's use and conversion of the String[] properties.
-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
+++ b/janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
@@ -26,6 +26,7 @@ import java.text.SimpleDateFormat;
 import java.time.Instant;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -157,10 +158,13 @@ public class SolrIndex implements IndexProvider {
             ConfigOption.Type.GLOBAL_OFFLINE, "ttl");
 
     /** SolrCloud Configuration */
-
-    public static final ConfigOption<String> ZOOKEEPER_URL = new ConfigOption<String>(SOLR_NS,"zookeeper-url",
+    /*
+     * TODO Rename ZOOKEEPER_URL and "zookeeper-url" to ZOOKEEPER_URLS and
+     * "zookeeper-urls" in future major releases.
+     */
+    public static final ConfigOption<String[]> ZOOKEEPER_URL = new ConfigOption<String[]>(SOLR_NS,"zookeeper-url",
             "URL of the Zookeeper instance coordinating the SolrCloud cluster",
-            ConfigOption.Type.MASKABLE, "localhost:2181");
+            ConfigOption.Type.MASKABLE, new String[] { "localhost:2181" });
 
     public static final ConfigOption<Integer> NUM_SHARDS = new ConfigOption<Integer>(SOLR_NS,"num-shards",
             "Number of shards for a collection. This applies when creating a new collection which is only supported under the SolrCloud operation mode.",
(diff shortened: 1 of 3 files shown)


janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -239,11 +240,24 @@ public SolrIndex(final Configuration config) throws BackendException {
...
-                .sendUpdatesOnlyToShardLeaders()
-                .build();
+            final String[] zookeeperUrl = config.get(SolrIndex.ZOOKEEPER_URL);
+            // Process possible zookeeper chroot. e.g. localhost:2181/solr
+            String chroot = null;
+            for (int i = 0; i < zookeeperUrl.length; i++) {
+                if (zookeeperUrl[i].contains("/")) {
+                    int chrootIndex = zookeeperUrl[i].indexOf("/");
+                    String hostAndPort = zookeeperUrl[i].substring(0, chrootIndex);
+                    chroot = zookeeperUrl[i].substring(chrootIndex);
```
porunov: Is different `chroot` possible? I.e. something like this?:
```
localhost:2181/solr
localhost:2181/mysolr
localhost:2181/mysolr2
```
If yes, the you rewrite your `chroot` here each time. If you need only the last `chroot` you can do a reverse loop (i.e. `for (int i = zookeeperUrl.length-1; i >= 0 ; i--)`) and then check `if(chroot==null){chroot = zookeeperUrl[i].substring(chrootIndex)}`.

porunov: Also, it's possible to do only one search for "/".
```
int chrootIndex = zookeeperUrl[i].indexOf("/");
if(chrootIndex!=-1){
...
}
```
But both 2 optimizations (reverse loop and one search) are not critical at all. I'm more interested if different `chroot` values possible or they all the same?

jerryjch: Different chroot is not supported.
I was trying to allow the chroot to be specified in any of the strings. Bu I like your suggestion.  Will update it.

jerryjch:
I have tested the PR  on my cluster.  I specified three zookeeper server strings, with the first string having wrong server name. It succeeded in connecting to the next and correct server string. 
Also tested a MR re-indexing job with three zookeeper hosts for hbase and solr.

jerryjch:
@amcp Thanks for the review. I will update the code.

jerryjch:
@amcp Do you have additional comments?  Thanks.

FlorianHockmann:
@jerryjch Could you please merge the changes from this PR also into `master`? I just wanted to merge #205 into `0.2` and afterwards `0.2` into `master`, but that led to a bunch of non-trivial merge conflicts mainly from Solr files which seem to be caused by this PR and PR #1230.
Otherwise I can also create a PR for the merge from `0.2` into `master`, but I would really prefer if we could merge each PR individually completely upstream to avoid these merge conflicts.

jerryjch:
Hi, @FlorianHockmann  Sure.  I will do that.

Issue #1276, opened 2018-09-29
Create docs for JanusGraph GLV clients


debasishdebs:
JanusGraph GLV clients have been in development since some time. Currently those drivers are available for Python and DotNet.

We will need to update the official JanusGraph docs to add a section for the same. 

A new Topic needs to be created after JanusGraph server, namely **JanusGraph Clients**. The topic will contain 2 section currently, one each for Python and DotNet. 

The docs will refer to Official JanusGraph DotNet repo for DotNet, and Python repo for Python. While the respective libs are hosted on their official channels (eg PyPi for Python) they will keep pointing to official repo. Once hosted, the docs needs to be updated.

Also, since Python GLV docs are available currently under readthedocs, they will need to add a section pointing to same.

FlorianHockmann:
I guess we need to decide first whether we just want to include the docs in the main JanusGraph docs or whether they should be independent. The first approach would of course be easier to implement but we could only release new client docs when a new version of JanusGraph is released which is currently only every few months.
We [already discussed on the mailing list](https://groups.google.com/forum/#!topic/janusgraph-dev/8npH66hMFU0) about creating dedicated sections for the different languages like `docs.janusgraph.org/dotnet/x.y.z/index.html`.

WolfgangFahl:
I never got janusgraph python glv  to work and the bug report #1826 is closed.  So what is the official way to make it work?

WolfgangFahl:
I  don't know how this is solve now and why this ticket is closed? I never retried again given the situation.

FlorianHockmann:
@WolfgangFahl You're talking about Python, but we still simply don't have a Python library for JanusGraph. So, there is also nothing we could document except for the official gremlin-python library and that is already documented here: https://docs.janusgraph.org/interactions/connecting/python/

We currently only have a library for .NET which is now documented: https://docs.janusgraph.org/master/interactions/connecting/dotnet/#janusgraphnet-for-janusgraph-specific-types-and-predicates

By the way, if you want to help improving the situation for Python, then contributions are of course welcome :-)

PR #1286, opened 2018-10-06
Add docs on how to connect to JanusGraph Server #205 [skip ci]

### LINKED ISSUE #205: Add pointers for accessing JanusGraph via other libraries

FlorianHockmann:
The new part in the docs simply shows examples for the most frequently used languages Java, Python and .NET/C# on how to connect to JanusGraph Server and points to the lists of available TinkerPop drivers and languages. So, users can also just pick one from there in case they are using a language that is not covered by these docs.

These changes intentionally don't mention the new libraries that we currently work on. I propose that we first just add a part to the docs that explains how to connect to JanusGraph from different languages by using the already available TinkerPop GLVs (this PR). We can then extend this part later when we have first official releases for these libraries with #1276.

A preview of the docs is available [here](https://florianhockmann.github.io/docs.janusgraph.org/0.2.2/) with its main new part [Connecting to JanusGraph](https://florianhockmann.github.io/docs.janusgraph.org/0.2.2/connecting.html).

I left out JavaScript simply because I don't have a JS environment set up and because I never used Gremlin-JavaScript. So, it would be good if someone else could add similar docs for JavaScript, but I think that we should do that with a separate PR later on.

Fixes #205
-----

### For all changes:

### For code changes:
- [] Have you written and/or updated unit tests to verify your changes?
- [] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### Note:

### DIFF
--- a/docs/basics.adoc
+++ b/docs/basics.adoc
@@ -345,9 +345,9 @@ This section is a brief overview of the Gremlin query language. For more informa
 * http://tinkerpop.apache.org/docs/{tinkerpop_version}/tutorials/the-gremlin-console/[Gremlin Console Tutorial]: Learn how to use the Gremlin Console effectively to traverse and analyze a graph interactively.
 * http://tinkerpop.apache.org/docs/{tinkerpop_version}/recipes/[Gremlin Recipes]: A collection of best practices and common traversal patterns for Gremlin.
 * http://tinkerpop.apache.org/index.html#language-drivers[Gremlin Language Drivers]: Connect to a Gremlin Server with different programming languages, including Go, JavaScript, .NET/C#, PHP, Python, Ruby, Scala, and TypeScript.
-* http://tinkerpop.apache.org/docs/{tinkerpop_version}/tutorials/gremlin-language-variants/[Gremlin Language Variants]: Learn how to embed Gremlin in a host programming language.
 * http://sql2gremlin.com[Gremlin for SQL developers]: Learn Gremlin using typical patterns found when querying data with SQL.
 
+In addition to these resources, <<connecting>> explains how Gremlin can be used in different programming languages to query a JanusGraph Server.
 
 === Introductory Traversals
 
@@ -446,7 +446,7 @@ The example above has 4 chained functions: `out`, `in`, `except`, and `values` (
 When chaining together functions, the incoming type must match the outgoing type, where `U` matches anything. Thus, the "co-battled/ally" traversal above is correct.
 
 [NOTE]
-The Gremlin overview presented in this section focused on the Gremlin-Groovy language implementation used in the Gremlin Console. Additional http://tinkerpop.apache.org/index.html#language-drivers[language drivers] and http://tinkerpop.apache.org/docs/{tinkerpop_version}/tutorials/gremlin-language-variants/[language variants] of Gremlin are available.
+The Gremlin overview presented in this section focused on the Gremlin-Groovy language implementation used in the Gremlin Console. Refer to <<connecting>> for information about connecting to JanusGraph with other languages than Groovy and independent of the Gremlin Console.
 
 === Iterating the Traversal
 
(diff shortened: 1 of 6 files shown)


FlorianHockmann:
I know I could merge this now via lazy consensus, but I think it would be better if someone would review this as it adds a completely new part to the docs. Again:
> A preview of the docs is available [here](https://florianhockmann.github.io/docs.janusgraph.org/0.2.2/) with its main new part [Connecting to JanusGraph](https://florianhockmann.github.io/docs.janusgraph.org/0.2.2/connecting.html).

docs/basics.adoc
```diff
@@ -344,10 +344,9 @@ This section is a brief overview of the Gremlin query language. For more informa
 * http://tinkerpop.apache.org/docs/{tinkerpop_version}/reference[Complete Gremlin Manual]: Reference manual for all of the Gremlin steps.
 * http://tinkerpop.apache.org/docs/{tinkerpop_version}/tutorials/the-gremlin-console/[Gremlin Console Tutorial]: Learn how to use the Gremlin Console effectively to traverse and analyze a graph interactively.
 * http://tinkerpop.apache.org/docs/{tinkerpop_version}/recipes/[Gremlin Recipes]: A collection of best practices and common traversal patterns for Gremlin.
-* http://tinkerpop.apache.org/index.html#language-drivers[Gremlin Language Drivers]: Connect to a Gremlin Server with different programming languages, including Go, JavaScript, .NET/C#, PHP, Python, Ruby, Scala, and TypeScript.
```
hsaputra: I think we could just leave the link to TinkerPop references for supported drivers.

The additional documentation you added in this PR on how specifically Gremlin can be used to query JanusGraph should help funnel the information from generic TInkerPop to JanusGraph.

docs/connecting-via-python.adoc
```diff
@@ -0,0 +1,49 @@
+[[connecting-via-python]]
+== Connecting from Python
+
+JanusGraph can be queried from a Python application with Apache TinkerPop's http://tinkerpop.apache.org/docs/{tinkerpop_version}/reference/#gremlin-python[Gremlin-Python].
+
+Gremlin traversals can be constructed with Gremlin-Python just like in Gremlin-Java or Gremiln-Groovy. Refer to <<gremlin>> for an introduction to Gremlin and pointers to further resources.
+
+IMPORTANT: Some Gremlin step and predicate names are reserved words in Python. Those are simply postfixed with `_` in Gremlin-Python, e.g., `in()` becomes `in_()`, `not()` becomes `not_()`, and so on.
```
chupman: I was trying to find a proper link to the reserved words, but [this](https://github.com/apache/tinkerpop/blob/3.3.4/gremlin-python/glv/generate.groovy#L29) is the closest I could find and what I used as a reference when testing the janusgraph-python driver PR. Not sure if this list is exhaustive, but if it is it would be great to include it since linking to source code is less than ideal.

docs/connecting-via-dotnet.adoc
```diff
@@ -0,0 +1,46 @@
...
+[source, bash]
+dotnet add package Gremlin.Net -v $MAVEN{tinkerpop.version}
+
+. Create a `GraphTraversalSource` which is the basis for all Gremlin traversals:
++
+[source, csharp]
+----
+var graph = new Graph();
+var client = new GremlinClient(new GremlinServer("localhost", 8182));
+// The client needs to be disposed on shut down to release resources and close open connections
```
chupman: In the python example you actually have a comment saying to use connection.close(). I think it would be nice to have a similar comments for .NET and Java. 

Also you used `close` instead of `dispose` in the other two examples. Not sure if dispose is more semantically correct for .NET, but there wasn't much to comment on so I figured I'd include it.

FlorianHockmann: I'll add explicit references to the methods that need to be called for Java and .NET.

Regarding `Dispose()` in .NET: This method is part of the `IDisposable` interface. It is basically always used in cases where resources are released on shut down (like open connections, file handles, ...). This allows to use `using` in C# which looks like this:
```c#
using (var client = new GremlinClient(new GremlinServer("localhost", 8182)))
{
    // do something with the client
}  // <-- `client.Dispose()` is called automatically here even if an exception was thrown
```
`Close()` would be more in line with Gremlin-Java, but it wouldn't feel natural to .NET developers and it couldn't be used with `using`.

docs/connecting-via-python.adoc
FlorianHockmann: The one you linked should be the best resource as that's what is used to generate Gremlin-Python. I added the affected names to the note.

FlorianHockmann:
Thanks for your reviews @hsaputra and @chupman. I think that I addressed your comments and while working again on this, I also noticed that the config for Java used Gryo v3 which isn't available here as this PR now targets 0.2. So, I changed that to Gryo v1. I'll change the version back to v3 for master during the merge.

hsaputra:
LGTM. Sheep it!

Thanks for the great work, Florian.

chupman:
I agree. Sheep it!

Issue #1261, opened 2018-09-18
Run Docker the docker image as non root user


graphdb2:
Hello,

is there a way with the current docker file to run janusgraph as a non root user ?

Best Regards

tonit:
#1272

tonit:
Actually, to explain, i think it is a good practice to run services as non-root user by default.

graphdb2:
Thanks for the sample

PR #1289, opened 2018-10-08
#1261 Makes Janusgraph docker run as a non-privileged user

tonit:
In this commit i also incorporated a small fix to the overall size by removing the zip file after extraction.

Retargeted modifictation for janusgraph 0.2 branch.

### DIFF
--- a/janusgraph-dist/janusgraph-dist-hadoop-2/Dockerfile
+++ b/janusgraph-dist/janusgraph-dist-hadoop-2/Dockerfile
@@ -6,6 +6,15 @@ ADD ${server_zip} /var
 RUN apt-get update -y && apt-get install -y zip && \
     server_base=`basename ${server_zip} .zip` && \
     unzip -q /var/${server_base}.zip -d /var && \
-    ln -s /var/${server_base} /var/janusgraph
+    rm /var/${server_base}.zip && \
+    ln -s /var/${server_base} /var/janusgraph && \
+    groupadd -g 999 janusgraph && \
+    useradd -d /home/janusgraph -m -r -u 999 -g janusgraph janusgraph && \
+    chown -R janusgraph:janusgraph /var/${server_base} && \
+    chmod 755 /var/${server_base} && \
+    chown -R janusgraph:janusgraph /var/janusgraph && \
+    chmod 755 /var/janusgraph
+
+USER janusgraph
 
 WORKDIR /var/janusgraph


janusgraph-dist/janusgraph-dist-hadoop-2/Dockerfile
```diff
@@ -6,6 +6,15 @@ ADD ${server_zip} /var
...
+    useradd -r -u 999 -g janusgraph janusgraph && \
+    chown -R janusgraph:janusgraph /var/${server_base} && \
+    chmod 755 /var/${server_base} && \
+    chown -R janusgraph:janusgraph /var/janusgraph && \
+    chmod 755 /var/janusgraph
 
-WORKDIR /var/janusgraph
+USER janusgraph
+
+WORKDIR /var/janusgraph
```
amcp: add a newline at the end of the file

porunov:
@tonit Could you squash the commits please?

janusgraph-dist/janusgraph-dist-hadoop-2/Dockerfile
```diff
@@ -6,6 +6,15 @@ ADD ${server_zip} /var
 RUN apt-get update -y && apt-get install -y zip && \
     server_base=`basename ${server_zip} .zip` && \
     unzip -q /var/${server_base}.zip -d /var && \
-    ln -s /var/${server_base} /var/janusgraph
+    rm /var/${server_base}.zip && \
+    ln -s /var/${server_base} /var/janusgraph && \
+    groupadd -g 999 janusgraph && \
```
FlorianHockmann: Not that important, but is there a reason why you specified the group and user ids explicitly?

janusgraph-dist/janusgraph-dist-hadoop-2/Dockerfile
```diff
@@ -6,6 +6,15 @@ ADD ${server_zip} /var
...
+    useradd -r -u 999 -g janusgraph janusgraph && \
```
chupman: add a `-d` flag here so that the home directory gets created.

chupman:
@tonit do you have any interest in finishing this PR out? It looks like it just needs to have a `-d` flag added to the user add command and have the commit signed.

tonit:
Sorry for not being on top of this. Somehow this has slipped my eyes. Let me finish this up..;)

tonit:
@chupman what should be the correct home directory in your oppinion? /home/janusgraph?

chupman:
@tonit /home/janusgraph is great. can you squash into 1 commit and sign the commit to pass the DCO check?

tonit:
check

chupman:
@amcp It looks like your comment was resolved. Are you ok with a merge?

janusgraph-dist/janusgraph-dist-hadoop-2/Dockerfile
tonit: You mean the 999? Well i remember it being problematic when mounting the filesystem from another system. If janusgraph user has a different userid on a different system, then you don't have the exact same privs for files created by the "original" janusgraph. IMHO.

FlorianHockmann: OK, thanks for clarifying.

FlorianHockmann:
We can merge this of course from my side as it's definitely an improvement over the existing image.
I'm just wondering what our plan is with this image as we now have a dedicated repo for images that we want to publish.
I think we could use a dedicated image for testing, e.g., of the language drivers and this image seems like a good candidate for that. If we however only want to use this image internally for testing, then we should probably move it to a different location.
Or do you guys see other valid use cases for this `Dockerfile` here?

tonit:
If you have a dedicated repo for images now, then i think having this image does not make much sense. I think it did not exist back when the PR got opened.
If you want to run  integrationtests using docker within the build, then it probably does make sense to have the Dockerfile in here.

chupman:
I'm going to merge this as it is an improvement and we still haven't actually pushed the docker images. In the future if we remove this Dockerfile I assume we'll first compare with the new repo so I see no reason not to make this Dockerfile better.

Issue #1091, opened 2018-05-30
Flaky Solr test


YevIgn:
There appeared several times a failure of the `SolrIndexText#testTTL` in
- https://travis-ci.org/JanusGraph/janusgraph/jobs/373477718 on solr6
- https://travis-ci.org/JanusGraph/janusgraph/jobs/385071158 on solr5
- https://travis-ci.org/JanusGraph/janusgraph/jobs/371272709 on solr5
- https://travis-ci.org/JanusGraph/janusgraph/jobs/368598831 on solr5

With no seemingly related changes that can affect this test case. Both of the failures looked identically:
```
org.janusgraph.diskstorage.solr.SolrIndexTest
testTTL(org.janusgraph.diskstorage.solr.SolrIndexTest)  Time elapsed: 6.187 sec  <<< FAILURE!
java.lang.AssertionError: expected:<2> but was:<4>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.janusgraph.diskstorage.indexing.IndexProviderTest.testTTL(IndexProviderTest.java:772)
        ...
```

twilmes:
Thanks for reporting this @YevIgn. My guess is this is a timing issue and for whatever reason, the 6 second sleep isn't long enough. Perhaps we could refactor some of these that rely on a sleep to instead poll for the condition that is being asserted from up to a max amount of time instead of adding sleeps and relying on the index to converge.

jerryjch:
There seem to other flaky tests in Solr in addition to the testTTL mentioned in the description of this issue.  We will add them here when we see them.

porunov:
Looks like this one is a flaky test: https://travis-ci.org/JanusGraph/janusgraph/jobs/489592012

PR #1327, opened 2018-11-28
Switch testing framework to junit 5

farodin91:
As a plan to streamline tests, I started to switch all tests from junit 4 to junit 5. 

The corresponding mailing list entry can be found here: https://groups.google.com/forum/#!topic/janusgraph-dev/QTAaFLC6aew

Todo:
- [x] test tinkerpop tests
- [x] activated maven enforcer
- [x] graphdb cql test failing

-----

### For all changes:

### For code changes:

### Note:


farodin91:
I could deactivate the speed test for now. In a follow up PR, we can switch to http://openjdk.java.net/projects/code-tools/jmh/ in the future because junit-Benchmarks is deprecated.

farodin91:
@pluradj Do you like to review the PR?

chupman:
Thanks for the submission @farodin91. I'll try to carve out some time next week to start a review. Sorry for the delay, still catching up from the holidays.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallbackTest.java
```diff
@@ -47,17 +43,14 @@
...
+    @BeforeEach
     public void setUp() throws Exception {
-        MockitoAnnotations.initMocks(this);
-
-        when(httpAsyncClientBuilderMock.setSSLContext(anyObject())).thenReturn(httpAsyncClientBuilderMock);
+        when(httpAsyncClientBuilderMock.setSSLContext(any(SSLContext.class))).thenReturn(httpAsyncClientBuilderMock);
 
         doReturn(sslContextMock).when(sslContextBuilderMock).build();
-        doReturn(sslContextBuilderMock).when(sslContextBuilderMock).loadTrustMaterial(
-                any(File.class), any(char[].class), any(TrustStrategy.class));
```
chupman: Why was this doReturn removed?

janusgraph-test/pom.xml
```diff
@@ -79,20 +79,39 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
```
chupman: Just curious, are there still some tests written in the old junit 3 and 4 syntax? If so do you have a rough estimate of how many?

janusgraph-hadoop-parent/pom.xml
```diff
@@ -191,13 +191,6 @@
             <artifactId>logback-classic</artifactId>
         </dependency>
         <!-- End logging backends. -->
-
-        <!-- TESTING -->
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
```
chupman: Is this not needed because janusgraph-test is in the dependency list?

janusgraph-cassandra/pom.xml
```diff
@@ -103,6 +103,10 @@
                     <groupId>jline</groupId>
                     <artifactId>jline</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>junit</groupId>
+                    <artifactId>junit</artifactId>
+                </exclusion>
```
chupman: I see this exclusion in
```
/janusgraph-cassandra/pom.xml
/janusgraph-solr/pom.xml
/janusgraph-hadoop-parent/janusgraph-hadoop-2/pom.xml
/janusgraph-hadoop-parent/janusgraph-hadoop-core/pom.xml
```

Is because mrunit, solr, and cassandra are bringing in conflicting version of junit?

farodin91:
@chupman thank you for reviewing. I‘m currently on vacation. I‘ll go through all comments on Monday or Tuesday.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -31,58 +31,66 @@
...
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
  */
-
-@RunWith(Parameterized.class)
 public class VertexIDAssignerTest {
 
-    final VertexIDAssigner idAssigner;
+    private VertexIDAssigner idAssigner;
```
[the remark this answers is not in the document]
farodin91: They are not initialized before calling a test method.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -31,58 +31,66 @@
...
                 }
             }
         }
 
-        return configurations;
+        return configurations.stream();
     }
 
-    private final long maxIDAssignments;
+    private long maxIDAssignments;
```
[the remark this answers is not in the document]
farodin91: They are not initialized before calling a test method.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -31,58 +31,66 @@
...
 
-    private final long numPartitionsBits;
+    private long numPartitionsBits;
```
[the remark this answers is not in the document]
farodin91: They are not initialized before calling a test method.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/rest/util/BasicAuthHttpClientConfigCallbackTest.java
```diff
@@ -53,7 +50,7 @@ private CredentialsProvider basicAuthTestBase(final String realm) {
 
         cb.customizeHttpClient(httpAsyncClientBuilderMock);
 
-        final ArgumentCaptor<BasicCredentialsProvider> cpCaptor = ArgumentCaptor.forClass(BasicCredentialsProvider.class);
+        final ArgumentCaptor<CredentialsProvider> cpCaptor = ArgumentCaptor.forClass(BasicCredentialsProvider.class);
```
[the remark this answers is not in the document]
farodin91: Mockito requires the actual type, here the interface.

janusgraph-hadoop-parent/pom.xml
farodin91: Correct

janusgraph-cassandra/pom.xml
farodin91: correct conflicting with the version in junit-vintage

janusgraph-test/src/main/java/org/janusgraph/diskstorage/IDAuthorityTest.java
```diff
@@ -100,16 +98,16 @@ public static ModifiableConfiguration getBasicConfig() {
...
-    public final boolean hasEmptyUid;
-    public final long blockSize;
-    public final long idUpperBoundBitWidth;
-    public final long idUpperBound;
+    public int uidBitWidth;
+    public boolean hasFixedUid;
+    public boolean hasEmptyUid;
+    public long blockSize;
+    public long idUpperBoundBitWidth;
+    public long idUpperBound;
```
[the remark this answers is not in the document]
farodin91: They are not initialized before calling a test method.

janusgraph-test/pom.xml
farodin91: Tinkerpop tests, some Benchmark and speed tests.

janusgraph-cassandra/src/test/java/org/janusgraph/graphdb/thrift/ThriftGraphSpeedTest.java
```diff
@@ -1,70 +0,0 @@
-// Copyright 2017 JanusGraph Authors
```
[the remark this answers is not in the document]
farodin91: I created a issue #1395.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -39,79 +37,71 @@
...
-    public void testSerialization() throws Exception {
-        testSerialization(null);
-        testSerialization(makeLine);
-        testSerialization(makePoly);
-        testSerialization(makeMultiPoint);
-        testSerialization(makeMultiLine);
-        testSerialization(makeMultiPolygon);
+    @ParameterizedTest
+    @MethodSource("data")
+    public void testSerialization(String name, Function<Graph, GraphReader> readerMaker, Function<Graph, GraphWriter> writerMaker) throws Exception {
```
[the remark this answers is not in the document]
farodin91: I added the readerMaker as name to assert.

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/ThriftSolrTest.java
```diff
@@ -58,17 +57,6 @@ The following two test cases do not pass for Solr since there is no (efficient)
...
-    public void testDeleteVertexThenAddProperty() throws BackendException {
-
-    }
-
-
-    @Override
-    public void testDeleteVertexThenModifyProperty() throws BackendException {
-
-    }
-
```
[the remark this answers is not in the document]
farodin91: readerMaker

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/rest/util/SSLConfigurationCallbackTest.java
farodin91: Due to changes of stubbing in mockito 2

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/rest/RestClientSetupTest.java
```diff
@@ -329,8 +324,7 @@ public void testCustomAuthenticator() throws Exception {
         assertEquals(1, customAuth.numInitCalls);
 
         // verifying that the custom callback is in the chain
-        final HttpAsyncClientBuilder hacb = PowerMockito.mock(HttpAsyncClientBuilder.class);
-        doReturn(hacb).when(hacb).setDefaultCredentialsProvider(anyObject());
```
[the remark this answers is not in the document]
farodin91:
```
Following stubbings are unnecessary (click to navigate to relevant line of code):
  1. -> at org.janusgraph.diskstorage.es.rest.RestClientSetupTest.testCustomAuthenticator(RestClientSetupTest.java:328)
Please remove unnecessary stubbings or use 'lenient' strictness. More info: javadoc for UnnecessaryStubbingException class.
```

farodin91:
@chupman Do like to review a again? Thanks
* I rebased on top of master
* updated the testing md
* make sure all tests are running with junit5 tags

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphIoTest.java
```diff
@@ -93,15 +93,15 @@ public void setup() {
     @ParameterizedTest
     @MethodSource("data")
     public void testSerialization(String name, Function<Graph, GraphReader> readerMaker, Function<Graph, GraphWriter> writerMaker) throws Exception {
-        testSerializationSpecialized(null, readerMaker, writerMaker);
-        testSerializationSpecialized(makeLine,readerMaker, writerMaker);
-        testSerializationSpecialized(makePoly,readerMaker, writerMaker);
-        testSerializationSpecialized(makeMultiPoint,readerMaker, writerMaker);
-        testSerializationSpecialized(makeMultiLine,readerMaker, writerMaker);
-        testSerializationSpecialized(makeMultiPolygon,readerMaker, writerMaker);
+        testSerializationSpecialized("", null, readerMaker, writerMaker);
```
chupman: Nit. In the event this ever triggered a failure a name specifying it's a null input, like `nullGeoshape` or `nullInput` would be preferable to an empty string.

farodin91:
@chupman renamed

farodin91:
I will push it tomorrow with a correct CLA.

chupman:
LGTM. Unless someone else wants to do a review the clock for lazy consensus can start now.

pom.xml
```diff
@@ -1049,7 +1081,7 @@
             <dependency>
                 <groupId>com.carrotsearch.randomizedtesting</groupId>
                 <artifactId>randomizedtesting-runner</artifactId>
-                <version>2.0.8</version>
+                <version>2.7.1</version>
```
porunov: Why not `2.7.2`?

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrJanusGraphIndexTest.java
```diff
@@ -15,22 +15,22 @@
...
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
  */
 public abstract class SolrJanusGraphIndexTest extends JanusGraphIndexTest {
 
-    @BeforeClass
-    public static void setUpMiniCluster() throws Exception {
+    @BeforeAll
+    private static void setUpMiniCluster() throws Exception {
```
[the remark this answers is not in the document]
farodin91: It has no reason to be public.

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/ThriftSolrTest.java
```diff
@@ -40,8 +39,8 @@ public WriteConfiguration getConfiguration() {
         return config.getConfiguration();
     }
 
-    @BeforeClass
-    public static void beforeClass() {
+    @BeforeAll
+    private static void beforeClass() {
```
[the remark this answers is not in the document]
farodin91: It has no reason to be public.

pom.xml
farodin91: Didn't checked again.

janusgraph-test/src/main/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -69,17 +63,14 @@ public void graphShouldNotOpenWithSameInstanceId() {
...
-        final String err = "A JanusGraph graph with the same instance id [not-unique] is already open. Might required forced shutdown.";
-        thrown.expectMessage(equalTo(err));
-        final StandardJanusGraph graph2 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
-
-        graph1.close();
+        assertThrows(JanusGraphException.class, () -> {
+            final StandardJanusGraph graph1 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
+
+            assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
+            assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
```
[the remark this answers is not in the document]
farodin91: These tests are never executed!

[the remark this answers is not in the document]
farodin91: I will move them from main to test.

janusgraph-test/src/main/java/org/janusgraph/diskstorage/IDAuthorityTest.java
```diff
@@ -222,8 +216,10 @@ private void checkBlock(IDBlock block, LongSet ids) {
 //        }
 //    }
 
-    @Test
-    public void testAuthorityUniqueIDsAreDistinct() {
+    @ParameterizedTest
+    @MethodSource("configs")
+    public void testAuthorityUniqueIDsAreDistinct(WriteConfiguration baseConfig) throws Exception {
+        setUp(baseConfig);
```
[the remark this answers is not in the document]
farodin91: I was able to move these line into a `BeforeEach` method.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphBaseTest.java
```diff
@@ -402,6 +403,9 @@ public static void assertMissing(Transaction g, Object vid) {
         assertFalse(g.vertices(vid).hasNext());
     }
 
+    private static void assertFalse(boolean hasNext) {
```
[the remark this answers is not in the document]
farodin91: Removed.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -31,58 +31,66 @@
...
+
     /**
      *
      * @param numPartitionsBits The number of partitions bits to use. This means there are exactly (1<<numPartitionBits) partitions.
      * @param partitionMax The maximum number of ids that can be allocated per partition. This is artificially constrained by the MockIDAuthority
      * @param localPartitionDef This array contains three integers: 1+2) lower and upper bounds for the local partition range, and
      *                          3) the bit width of the local bounds. The bounds will be bit-shifted forward to consume the bit width
      */
-    public VertexIDAssignerTest(int numPartitionsBits, int partitionMax, int[] localPartitionDef) {
+    public void setup(int numPartitionsBits, int partitionMax, int[] localPartitionDef){
```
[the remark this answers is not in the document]
farodin91: I have no idea to clean up this class.

janusgraph-test/src/main/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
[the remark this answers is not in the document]
farodin91: This test doesn't work with inmemory.

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrJanusGraphIndexTest.java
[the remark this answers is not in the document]
porunov: Possibly I am wrong and missing something but the [documentation](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/BeforeAll.html) says: `@BeforeAll methods must have a void return type, must not be private, and must be static by default.`
I didn't use junit 5 yet, could you explain why this method has no reason to be public?

janusgraph-test/src/test/java/org/janusgraph/graphdb/serializer/SerializerSpeedTest.java
```diff
@@ -14,23 +14,23 @@
...
-@Category({ PerformanceTests.class })
+@Tag(TestCategory.PERFORMANCE_TESTS)
 public class SerializerSpeedTest extends SerializerTestCommon {
 
     @Rule
     public TestRule benchmark = JUnitBenchmarkProvider.get();
 
     @Test
-    public void performanceTestStringSerialization() {
+    public void PERFORMANCE_TESTStringSerialization() {
```
porunov: Change the name of the method back as it was

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -51,35 +45,34 @@ public void graphShouldOpenWithSameInstanceId() {
         final StandardJanusGraph graph1 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
 
         assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
+        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");
```
porunov: I think you accidentally added to `non-unique` strings in a lot of places the suffix `(current)`. Please, check [this amend](https://github.com/JanusGraph/janusgraph/compare/a9d4ad380c399f734d2e8d026f44daf1d4f51eb0..f99c14f5fe55d1b673242ca474ba0ab6d2387f9c) and remove `(current)` suffix from all those strings.

janusgraph-test/src/main/java/org/janusgraph/TestCategory.java
```diff
@@ -12,13 +12,13 @@
...
-/**
- * This is a JUnit category for tests that need to run against Cassandra
- * configured for SSL-based client authentication.
- *
- * If you rename or move this class, then you must also update mentions of it in
- * the Cassandra module's pom.xml.
- */
-public interface CassandraSSLTests { }
+public final class TestCategory {
+    public static final String BRITTLE_TESTS = "BRITTLE_TESTS";
```
porunov: Actually I was talking about the variable names and not values. I.e. values like `BritleTests` are OK. It is more like a preference of choice. As fore me the perfect constant would be:
`public static final String BRITTLE_TESTS = "BritleTests";`
but current constants are OK too. You can use either proposed constants or leave these constants as you wish.

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
farodin91: I didn't work without suffix `(current)`.

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -51,35 +45,34 @@ public void graphShouldOpenWithSameInstanceId() {
...
         assertEquals(graph2.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph2.openManagement().getOpenInstances().toArray()[0], "not-unique");
+        assertEquals(graph2.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");
         graph1.close();
         graph2.close();
     }
 
     @Test
+    @Disabled("This should be run against a backend which support multiple backends.")
```
porunov: Why you disabled this particular test?

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -51,35 +45,34 @@ public void graphShouldOpenWithSameInstanceId() {
...
-        final String err = "A JanusGraph graph with the same instance id [not-unique] is already open. Might required forced shutdown.";
-        thrown.expectMessage(equalTo(err));
-        final StandardJanusGraph graph2 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
-
-        graph1.close();
+        JanusGraphException janusGraphException = assertThrows(JanusGraphException.class, () -> {
+            final StandardJanusGraph graph1 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
+
+            assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
+            assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");
```
porunov: This assert fails because you added `(current)` to `not-unique` string. Remove `(current)` suffix.

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -51,35 +45,34 @@ public void graphShouldOpenWithSameInstanceId() {
...
         final Map<String, Object> map = new HashMap<String, Object>();
         map.put(STORAGE_BACKEND.toStringWithoutRoot(), "inmemory");
         map.put(UNIQUE_INSTANCE_ID.toStringWithoutRoot(), "not-unique");
         final MapConfiguration config = new MapConfiguration(map);
-        final StandardJanusGraph graph1 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
-
-        assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
-
-        thrown.expect(JanusGraphException.class);
```
porunov: Notice that the exception is expected after `graph1` is successfully opened and asserts executed successfully. 
In you code in case when the exception is on opening `graph1` then the unit test won't fail (even so it should fail in such scenario) and asserts won't be executed. Please rewrite you code like this, so that it would be consistent to the old flow:
```
final StandardJanusGraph graph1 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
assertThrows(JanusGraphException.class, () -> {
    final StandardJanusGraph graph2 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
    graph1.close();
}, "A JanusGraph graph with the same instance id [not-unique] is already open. Might required forced shutdown.");
```

janusgraph-test/src/main/java/org/janusgraph/diskstorage/IDAuthorityTest.java
```diff
@@ -101,16 +100,25 @@ public static ModifiableConfiguration getBasicConfig() {
...
-    public final long idUpperBoundBitWidth;
-    public final long idUpperBound;
+    public int uidBitWidth;
+    public boolean hasFixedUid;
+    public boolean hasEmptyUid;
+    public long blockSize;
+    public long idUpperBoundBitWidth;
+    public long idUpperBound;
 
-    public IDAuthorityTest(WriteConfiguration baseConfig) {
```
porunov: As I see this block of code is executed only once before all test cases.
and current implementation execute this block of code individually for each case **after** `setUp`. It means that the current tests flow are different. I don't say that it is broken (didn't check it yet) but it is different. I will take another look on this either tomorrow or early next week. If you manage to fix it earlier it would be grate. Again, I didn't use JUnit 5 yet so I need some time to investigate this problem. 
I don't know if it is possible but can we use something like this?:
```
@BeforeAll
@MethodSource("configs")
public void open(WriteConfiguration baseConfig) throws BackendException {
    Preconditions.checkNotNull(baseConfig);
    TestGraphConfigs.applyOverrides(baseConfig);
    this.baseStoreConfiguration = baseConfig;
    blockSize = config.get(IDS_BLOCK_SIZE);
    idUpperBoundBitWidth = 30;
    idUpperBound = 1L <<idUpperBoundBitWidth;
}
```

Can you check it? If it works then we can remove `open(baseConfig);` from all tests and the flow fill be fixed.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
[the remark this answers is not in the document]
porunov: You code changes the flow because this part of code if executed only once for all tests and you execute this part of code for each individual test. If the solution for [this comment](https://github.com/JanusGraph/janusgraph/pull/1327#discussion_r) works then you can apply the fix here also.

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
porunov: The assert didn't fail because you disabled this test. Please, enable it.

janusgraph-test/src/main/java/org/janusgraph/TestCategory.java
porunov: As I understand (possibly wrong) you use the value of the variable in `pom` right? Or you use the name or the variable? I didn't get why the name of the variable and the value in the variable should be equal.

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
porunov: Please, use another PR to disable the test if you have an issue with it. Don't disable it in this PR because it isn't relevant to the current PR (as I understand). Currently I see that the test is enabled and works in `master`. If you must certainly disable this test when moving from `JUnit 4` to `JUnit 5` then explain why you must disable it for `JUnit 5`.

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
porunov: Please change the line:
`assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");`
to line:
`assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");`

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -51,35 +45,34 @@ public void graphShouldOpenWithSameInstanceId() {
...
 
         assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
+        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");
 
         final StandardJanusGraph graph2 = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
 
         assertEquals(graph1.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");
+        assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");
```
porunov: Please change the line:
`assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");`
to
`assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");`

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -51,35 +45,34 @@ public void graphShouldOpenWithSameInstanceId() {
...
         assertEquals(graph2.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph2.openManagement().getOpenInstances().toArray()[0], "not-unique");
+        assertEquals(graph2.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");
```
porunov: Please change the line:
`assertEquals(graph2.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");`
to
`assertEquals(graph2.openManagement().getOpenInstances().toArray()[0], "not-unique");`

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
porunov: Please change the line:
`assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique(current)");`
to
`assertEquals(graph1.openManagement().getOpenInstances().toArray()[0], "not-unique");`

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -90,7 +83,7 @@ public void instanceIdShouldEqualHostname() throws UnknownHostException {
         final MapConfiguration config = new MapConfiguration(map);
         final StandardJanusGraph graph = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
         assertEquals(graph.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName());
+        assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName() + "(current)");
```
porunov: Please change the line
`assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName() + "(current)");`
to
`assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName());`

janusgraph-test/src/test/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -103,7 +96,7 @@ public void instanceIdShouldEqualHostnamePlusSuffix() throws UnknownHostExceptio
 				final MapConfiguration config = new MapConfiguration(map);
         final StandardJanusGraph graph = new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(new CommonsConfiguration(config)));
         assertEquals(graph.openManagement().getOpenInstances().size(), 1);
-        assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName() + "1");
+        assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName() + "1(current)");
```
porunov: Please change the line 
`assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName() + "1(current)");`
to
`assertEquals(graph.openManagement().getOpenInstances().toArray()[0], Inet4Address.getLocalHost().getHostName() + "1");`

janusgraph-test/src/main/java/org/janusgraph/TestCategory.java
farodin91: Content of the variable is the tag and not the variable.

janusgraph-test/src/main/java/org/janusgraph/diskstorage/IDAuthorityTest.java
farodin91: With Junit 5.5, We can do class wide parametrization.
https://github.com/junit-team/junit5/issues/878#issuecomment-354544841

porunov:
@farodin91 Also, I see that you use in some `assertThrows` as third a message as expected one but I just checked the documentation and it looks like it isn't an expected message but an assertion message, so you should change all your `assertThrows` with 3 params from:
`assertThrows(Exception.class, ()->{...}, "expected error message");`
to
```
Exception exception = assertThrows(Exception.class, ()->{...});
assertEquals("expected error message", exception.getMessage());
```
Maybe there is a more elegant solution exists but I don't know any.

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
```diff
@@ -158,15 +156,18 @@ public void testSupport() {
...
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void testMapKey2Field_IllegalCharacter() {
-        KeyInformation keyInfo = new StandardKeyInformation(Boolean.class, Cardinality.SINGLE);
-        index.mapKey2Field("here is an illegal character: •", keyInfo);
+        assertThrows(IllegalArgumentException.class, () -> {
+
+            KeyInformation keyInfo = new StandardKeyInformation(Boolean.class, Cardinality.SINGLE);
+            index.mapKey2Field("here is an illegal character: •", keyInfo);
+        });
```
chupman: After reading @porunov 's comments and rereading the docs and [user guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests-assertions) I think this should be rewritten like this:
```
Throwable exception = assertThrows(IllegalArgumentException.class, () -> {
    KeyInformation keyInfo = new StandardKeyInformation(Boolean.class, Cardinality.SINGLE);
        index.mapKey2Field("here is an illegal character: •", keyInfo);
    });
    assertEquals("here is an illegal character: •", exception.getMessage());
```

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/LuceneIndexTest.java
```diff
@@ -135,16 +128,12 @@ public void testSupport() {
...
-//    }
-
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void testMapKey2Field_IllegalCharacter() {
-        index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
+        assertThrows(IllegalArgumentException.class, () ->{
+            index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
+        });
```
chupman: I haven't actually tested this, but I think if we want to capture that the error message is as expected we can update this to:
```
Throwable exception = assertThrows(IllegalArgumentException.class, () ->{
    index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
});
assertEquals("here is an illegal character: " + REPLACEMENT_CHAR, exception.getMessage());
```

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
```diff
@@ -254,9 +252,11 @@ public void testIngestPipeline() throws Exception {
         }
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void testMapKey2Field_IllegalCharacter() {
-        index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
+        assertThrows(IllegalArgumentException.class, () -> {
+            index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
+        });
```
chupman: I haven't actually tested this, but I think if we want to capture that the error message is as expected we can update this to:
```
Throwable exception = assertThrows(IllegalArgumentException.class, () ->{
    index.mapKey2Field("here is an illegal character: " + REPLACEMENT_CHAR, null);
});
assertEquals("here is an illegal character: " + REPLACEMENT_CHAR, exception.getMessage());
```

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
farodin91: I can add a assert to check the exception.This check didn't exists before. I want to change the flow  test as less as possible.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
farodin91: Same here. I don't think `"here is an illegal character: " + REPLACEMENT_CHAR`is the exception message.

janusgraph-solr/src/test/java/org/janusgraph/diskstorage/solr/SolrIndexTest.java
porunov: @chupman , I agree here with @farodin91 . I think we shouldn't change the flow of tests in this PR.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/LuceneIndexTest.java
porunov: Currently this test doesn't check the error message, so we shouldn't change the flow of the test in this PR. We can make other PRs for test improvements if needed.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticSearchIndexTest.java
porunov: Agree here with @farodin91 also.

porunov:
@chupman Yes there are next issues which should be addressed:
https://github.com/JanusGraph/janusgraph/pull/1327#discussion_r253273529
https://github.com/JanusGraph/janusgraph/pull/1327#discussion_r253274195
https://github.com/JanusGraph/janusgraph/pull/1327#issuecomment-460004469

I couldn't get time to research for possible solutions today. I will try to research for possible solutions tomorrow. Right now the flow of some tests is broken (because the constructor was removed in some tests). That is why we should rewrite those tests a little bit to manage to follow the old flow of the tests.
I think it is possible to use either `@BeforeAll` annotation or class wide parametrization as @farodin91 mentioned [here](https://github.com/JanusGraph/janusgraph/pull/1327#discussion_r253280245) bit I didn't check it yet.

farodin91:
I changed these both test classes `IDAuthorityTest` and `VertexIDAssignerTest` in way which only increase runtime. In normal case all tests should be independent. Therefore initialization everything in the test instead of doing it globally.

@porunov For `GraphDatabaseConfigurationInstanceIdTest`, I changed it as you want. Besides, this test class has the same problem as i mentioned in https://github.com/JanusGraph/janusgraph/pull/1403#discussion_r253641336.

porunov:
@farodin91 As I mentioned, this PR should not change the flow of the tests (even if it needs to be changed). If everything were initialized before the tests are executed, then after upgrading to JUnit 5 we should see the same flow. We should use other PRs to improve tests, change their flows and so on.

porunov:
@farodin91 what do you think about moving configuration parameter from `configs()` into `ParameterResolver` and use `@BeforeAll` with parameter?
I didn't try it yet but looks like [@BeforeAll](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/BeforeAll.html) supports parameters resolved by [ParameterResolver](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/BeforeAll.html).

I think it could look like:
```
@ExtendWith(BaseConfigurationResolver.class)
public abstract class IDAuthorityTest {

    @BeforeAll 
    public static void presetUp(WriteConfiguration baseConfig){
        Preconditions.checkNotNull(baseConfig);
        TestGraphConfigs.applyOverrides(baseConfig);
        ...
        idUpperBound = 1L <<idUpperBoundBitWidth;
    }

    @BeforeEach
    public static void setUp(){
        StoreManager m = openStorageManager();
        m.clearStorage();
        m.close();
    }

}
```

porunov:
@FlorianHockmann , yes. The code which was in the constructor of `IDAuthorityTest` was executed per class and now it is executed per method.

farodin91:
@porunov I think we shouldn't implement parametrization for classes by our self. Beside resolving a parameter for `@BeforeAll`, you need to executed test class for all configs.

I don't see a problem with the current implementation. Now, the tests are more independent, as everything is initialized for each test in combination with each parametrization.

porunov:
@farodin91 , you can open a separate PR for changing the flow of current tests, merge that PR and then rebase the commit `Switch testing framework to junit 5`. In such way your commit won't change the flow of the tests. I don't tell that your current code is wrong, I am just telling that your commit which is called `Switch testing framework to junit 5` actually making more then just `Switch testing framework to junit 5`. If you don't want to open a separate PR you can make two commits:

1st commit - `make tests separate from each other in IDAuthorityTest and VertexIDAssignerTest`
2nd commit - `Switch testing framework to junit 5`
then you should rename your current PR to `Make tests separate from each other in IDAuthorityTest and VertexIDAssignerTest and switch testing framework to junit 5`. 
I don't actually suggest to use several commits in the PR because the reviewer of the PR should review 2 commits and the reviewer may be agree with one commit and disagree (or just couldn't review) with another commit that is why it will delay merging.

Also, notice that this PR can be merged even if I don't approve it. My requests are considered as suggestions. If the community don't agree with me then it is OK and the PR can be merged by JanusGraph members.

farodin91:
@porunov With junit4, you are unable to do parametrization of a single test. You are only able to parametrization a test classes.

porunov:
@farodin91 I have checked the code once again. Everything looks good, but I am not sure about `IDAuthorityTest` and `VertexIDAssignerTest`. I am not JUnit guru, so I need more time to research if your implementation is OK or not. Right now I see that the flow of tests isn't the same as it was. That's all. I can't tell if it is right or not.
Current implementation executes the code which was in the constructor each time for each test. Knowing that, we need to know if tests may run in parallel. Can tests with different configurations run in parallel? If yes, is there a problem with race conditions? I.e. we are executing `setUp` method with different configurations, so there is possibility that some class variables will be assigned with values from different configurations. For example, variable `uidBitWidth` will be assigned with value from `configuration1` and variable `blockSize` will be assigned with value from `configuration2`.

farodin91:
>  Can tests with different configurations run in parallel? If yes, is there a problem with race conditions?

@porunov All tests in janusgraph are executed sequential.

janusgraph-test/src/main/java/org/janusgraph/diskstorage/IDAuthorityTest.java
```diff
@@ -101,16 +99,16 @@ public static ModifiableConfiguration getBasicConfig() {
...
-    public final boolean hasEmptyUid;
-    public final long blockSize;
-    public final long idUpperBoundBitWidth;
-    public final long idUpperBound;
+    public int uidBitWidth;
+    public boolean hasFixedUid;
+    public boolean hasEmptyUid;
+    public long blockSize;
+    public long idUpperBoundBitWidth;
+    public long idUpperBound;
```
porunov: Make all variables as `private` and the last two as:
```
private final long idUpperBoundBitWidth = 30;
private final long idUpperBound = 1L <<idUpperBoundBitWidth;
```

janusgraph-test/src/main/java/org/janusgraph/diskstorage/IDAuthorityTest.java
```diff
@@ -52,43 +51,42 @@
...
     public static final int CONCURRENCY = 8;
     public static final int MAX_NUM_PARTITIONS = 4;
     public static final String DB_NAME = "test";
 
     public static final Duration GET_ID_BLOCK_TIMEOUT = Duration.ofMillis(300000L);
 
-    @Parameterized.Parameters
-    public static Collection<Object[]> configs() {
-        final List<Object[]> configurations = new ArrayList<>();
+    public static Stream<Arguments> configs() {
```
porunov: Move first two lines from `setUp` method to a separate method, so that it would look something like:
```
public static Stream<Arguments> configs() {
    final List<Arguments> configurations = new ArrayList<>();

    ModifiableConfiguration c = getBasicConfig();
    configurations.add(arguments(checkAndReturnWriteConfiguration(c)));

    c = getBasicConfig();
    c.set(IDAUTHORITY_CAV_BITS,9);
    c.set(IDAUTHORITY_CAV_TAG,511);
    configurations.add(arguments(checkAndReturnWriteConfiguration(c)));

    c = getBasicConfig();
    c.set(IDAUTHORITY_CAV_RETRIES,10);
    c.set(IDAUTHORITY_WAIT, Duration.ofMillis(10L));
    c.set(IDAUTHORITY_CAV_BITS,7);
    //c.set(IDAUTHORITY_RANDOMIZE_UNIQUEID,true);
    c.set(IDAUTHORITY_CONFLICT_AVOIDANCE, ConflictAvoidanceMode.GLOBAL_AUTO);
    configurations.add(arguments(checkAndReturnWriteConfiguration(c)));

    return configurations.stream();
}

private WriteConfiguration checkAndReturnWriteConfiguration(ModifiableConfiguration baseConfig){
    WriteConfiguration writeConfiguration = baseConfig.getConfiguration();
    Preconditions.checkNotNull(writeConfiguration);
    TestGraphConfigs.applyOverrides(writeConfiguration);
    return writeConfiguration;
}

private void setUp(WriteConfiguration baseConfig) throws Exception {
    this.baseStoreConfiguration = baseConfig;
    Configuration config = StorageSetup.getConfig(baseConfig);
    uidBitWidth = config.get(IDAUTHORITY_CAV_BITS);
    //hasFixedUid = !config.get(IDAUTHORITY_RANDOMIZE_UNIQUEID);
    hasFixedUid = !ConflictAvoidanceMode.GLOBAL_AUTO.equals(config.get(IDAUTHORITY_CONFLICT_AVOIDANCE));
    hasEmptyUid = uidBitWidth==0;
    blockSize = config.get(IDS_BLOCK_SIZE);
    StoreManager m = openStorageManager();
    m.clearStorage();
    m.close();
    open();
}
```

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -31,58 +31,66 @@
...
-@RunWith(Parameterized.class)
 public class VertexIDAssignerTest {
 
-    final VertexIDAssigner idAssigner;
+    private VertexIDAssigner idAssigner;
 
-    @Parameterized.Parameters
-    public static Collection<Object[]> configs() {
-        final List<Object[]> configurations = new ArrayList<>();
+    public static Stream<Arguments> configs() {
```
porunov: If it is OK I would rewrite it to something like this (not tested):
```
public static Stream<Arguments> configs() {
    final List<Arguments> configurations = new ArrayList<>();

    for (int maxPerPartition : new int[]{Integer.MAX_VALUE, 100, 300}) {
        for (int numPartitions : new int[]{2, 4, 10}) {
            for (int[] local : new int[][]{null, {0,2, numPartitions}, {235,234,8}, {1,1,2}, {0,1<<(numPartitions-1),numPartitions}}) {
                configurations.add(Arguments.arguments(generateConfigurationArguments(numPartitions, maxPerPartition, local)));
            }
        }
    }

    return configurations.stream();
}

private static Arguments generateConfigurationArguments(int numPartitionsBits, int partitionMax, int[] localPartitionDef){
    MockIDAuthority idAuthority = new MockIDAuthority(11, partitionMax);

    StandardStoreFeatures.Builder fb = new StandardStoreFeatures.Builder();

    if (null != localPartitionDef) {
        fb.localKeyPartition(true);
        idAuthority.setLocalPartition(PartitionIDRangeTest.convert(localPartitionDef[0],localPartitionDef[1],localPartitionDef[2]));
    }
    StoreFeatures features = fb.build();

    ModifiableConfiguration config = GraphDatabaseConfiguration.buildGraphConfiguration();
    config.set(GraphDatabaseConfiguration.CLUSTER_MAX_PARTITIONS,1<<numPartitionsBits);
    idAssigner = new VertexIDAssigner(config, idAuthority, features);
    System.out.println(String.format("Configuration [%s|%s|%s]",numPartitionsBits,partitionMax,Arrays.toString(localPartitionDef)));

    long maxIDAssignments;
    if (localPartitionDef!=null && localPartitionDef[0]<localPartitionDef[1] && localPartitionDef[2]<=numPartitionsBits) {
        maxIDAssignments = ((localPartitionDef[1]-localPartitionDef[0])<<(numPartitionsBits-localPartitionDef[2]))*((long)partitionMax);
    } else {
        maxIDAssignments = (1<<numPartitionsBits)*((long)partitionMax);
    }

    return Arguments.arguments(idAssigner, maxIDAssignments, numPartitionsBits);
}

private void setup(VertexIDAssigner idAssigner, long maxIDAssignments, int numPartitionsBits){
    this.idAssigner = idAssigner;
    this.maxIDAssignments = maxIDAssignments;
    this.numPartitionsBits = numPartitionsBits;
}
```

In such case your `setup` method will be lightweight. Even so, we can completely remove `setup` method and use method variables in tests instead of class variables.

porunov: Tests will look something like this if we use class variables:
```
@ParameterizedTest
@MethodSource("configs")
public void testIDAssignment(VertexIDAssigner idAssigner, long maxIDAssignments, int numPartitionsBits) {
    setup(idAssigner, maxIDAssignments, numPartitionsBits);
   // Actual test below
}
```
or something like this if we use method variables:
```
@ParameterizedTest
@MethodSource("configs")
public void testIDAssignment(VertexIDAssigner idAssigner, long maxIDAssignments, int numPartitionsBits) {
   // Actual test below
}
```

farodin91:
@porunov These changes working perfect. I will amend these changes.
@chupman Do you check both files?

chupman:
Yup. I moved it back to approved. If @porunov approves and the CI passes I think we can still merge today.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -31,58 +31,38 @@
...
-
-    private enum CustomIdStrategy {
-        LOW,
-        HIGH
+        return configurations.stream();
     }
 
-    /**
-     *
-     * @param numPartitionsBits The number of partitions bits to use. This means there are exactly (1<<numPartitionBits) partitions.
```
porunov: Don't remove this method's comments. Move this comment above `generateConfigurationArguments`.

porunov:
@farodin91 Looks like CI failed because of a flaky test. I've reported it here #1091 .

chupman:
Thanks @farodin91 for the patch and @porunov for the additional review. I'm looking forward to working with junit 5 going forward.

Issue #1314, opened 2018-11-03
g.E().hasId() not working


FlorianHockmann:
`g.E().hasId()` seems to be not working:

```groovy
gremlin> edgeId = g.V(4184).bothE().id().next()
==>362-380-9hx-388
gremlin> g.E().hasId(edgeId)
gremlin> g.E(edgeId) // this works as expected
==>e[362-380-9hx-388][4176-battled->4184]
```

hasId() works correctly though when it's used mid-traversal:

```groovy
gremlin> g.V(4184).bothE().hasId(edgeId)
==>e[362-380-9hx-388][4176-battled->4184]
```

This problem was [reported on janusgraph-users by Thanos Giannakis](https://groups.google.com/d/msg/janusgraph-users/czc_ik0SpkY/UvrrGN3SCgAJ).

RaulGuo:
I tried to fix this. waiting to be merged. [pr](https://github.com/JanusGraph/janusgraph/pull/1317)

RaulGuo:
Hi Hockmann, I think I made a mistake, should build the pr based on some other branchs other than the master? It is blocked, does that mean that I should build another pr based on other branches? This is my first time to do so, please help to give some instruction, thanks.

porunov:
Fixed in #2849

PR #1343, opened 2018-12-10
fix issue that g.E().hasId() not returning right result(issue 1314)

RaulGuo:
Issue fix [1314](https://github.com/JanusGraph/janusgraph/issues/1314) that g.E().hasId() didn't return the right result. There is not judge whether hasId is trying to return a vertex or edge, this result in the issue.
-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
@@ -20,6 +20,8 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.janusgraph.core.*;
 import org.janusgraph.core.attribute.Cmp;
 import org.janusgraph.core.schema.ConsistencyModifier;
@@ -42,6 +44,7 @@ import static org.junit.Assert.*;
 import org.janusgraph.graphdb.internal.ElementCategory;
 import org.janusgraph.graphdb.internal.InternalRelationType;
 import org.janusgraph.graphdb.internal.InternalVertexLabel;
+import org.janusgraph.graphdb.relations.RelationIdentifier;
 import org.janusgraph.graphdb.types.CompositeIndexType;
 import org.janusgraph.graphdb.types.IndexType;
 import org.janusgraph.testcategory.SerialTests;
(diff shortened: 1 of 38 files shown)


porunov:
@RaulGuo Have you signed your commit? If no then run next commands in your branch:
```
git config user.name "Raul Guo"
git config user.email "zaixiaguozhen@163.com"
git commit --amend -s --reset-author
git push -f
```
More details here: https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#commit-changes-and-sign-the-developer-certificate-of-origin

RaulGuo:
@porunov Thank you for your notice, I am sorry, you are right I forgot to reset it. I have reset and commit it now.

janusgraph-test/src/main/java/org/janusgraph/graphdb/GetEdgeByHasIdTest.java
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017 JanusGraph Authors
```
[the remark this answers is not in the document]
RaulGuo: Oh yes, you are right, I didn't think about those questions you mentioned, your remind and suggest is really helpful  for me.  Thank you. I will fix those problems.

mad:
Can we support this case?

```
g.E().id().is("6c8-3b4-cb9x-394");
```

Because cypher-for-gremlin used this construction for translating cypher like that `MATCH ()-[r]-() WHERE id(r)= "6c8-3b4-cb9x-394" RETURN r`

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
```diff
@@ -20,6 +20,8 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
```
porunov: Move tinkerpop dependencies to the line 52 (i.e. where all tinkerpop dependencies are listed)

porunov:
@RaulGuo  don't forget to squash your commits into a single one

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
```diff
@@ -643,4 +646,25 @@ private double testAllVertices(long vid, int numV) {
         return ((System.nanoTime()-start)/1000000.0);
     }
 
+
+    @Test
+    public void getEdgeByHasId() {
```
FlorianHockmann: I just executed this test with the Thrift and also the Berkeley backend and it passed even when I removed your changes to `JanusGraphStep`. I guess we first need to find a way to reproduce the problem with a test case. Although, I can't quite see why this suddenly works in the test case, but didn't when I tried it from the console when I created the issue.

mad: For reproduce change 659 and 660 lines to use `eid.toString()`

FlorianHockmann: Ok, but why should it work when you provide a string instead of a `RelationIdentifier`? That's not really a supported feature and also not what the issue was about.

mad: Yes, right, this is another case related with https://github.com/JanusGraph/janusgraph/pull/1343#issuecomment-447857246

May be `.profile()` will help

porunov:
Thanks @RaulGuo for the contribution. The issue was now fixed in #2849

Issue #1163, opened 2018-07-10
Document, and possibly improve, how to list existing indexes


chupman:
I wasn't able to find how to list existing indexes (or indices) in the docs. 

The command is currently mgmt.getGraphIndexes(Vertex.class) although I'm not sure if vertex-centric would show up under mgmt.getGraphIndexes(Edge.class) instead. 

In addition to documentation I think it would be great if we created a getGraphIndexes() method that took no parameters and gave a list of all the indexes and their current state.

chupman:
Thanks for the link to @robertdale 's [schema describe](https://gist.github.com/robertdale/ad4c63910009dd1118abe67b33ce41e1) script. My real goal was easier to use administrative commands to gather schema information. I just started on a getGraphSchema() method for the management API using schema describe as a reference. 

Also you're 100% correct about the other approaches being more work so I'm ultimately taking the path of least resistance 💯

chupman:
I still need to add tests and do some work to make each section formatted correctly when separately called, but I finished a first pass. Right now I'm working with a width of 100 characters. I'd really like to get it to 80, but it's a bit of a challenge. If anyone has a really strong feeling about it I can shrink the first column from 30 to 15 and abbreviate some headers to get the other 5.

```
gremlin> mgmt.printGraphSchema()
==>------------------------------------------------------------------------------------------------
Vertex Label Name              | Partitioned | Static                                             |
---------------------------------------------------------------------------------------------------
titan                          | false       | false                                              |
location                       | false       | false                                              |
god                            | false       | false                                              |
demigod                        | false       | false                                              |
human                          | false       | false                                              |
monster                        | false       | false                                              |
---------------------------------------------------------------------------------------------------
Edge Label Name                | Directed    | Unidirected | Multiplicity                         |
---------------------------------------------------------------------------------------------------
brother                        | true        | false       | MULTI                                |
father                         | true        | false       | MANY2ONE                             |
mother                         | true        | false       | MANY2ONE                             |
battled                        | true        | false       | MULTI                                |
lives                          | true        | false       | MULTI                                |
pet                            | true        | false       | MULTI                                |
---------------------------------------------------------------------------------------------------
Property Key Name              | Cardinality | Data Type                                          |
---------------------------------------------------------------------------------------------------
name                           | SINGLE      | class java.lang.String                             |
age                            | SINGLE      | class java.lang.Integer                            |
time                           | SINGLE      | class java.lang.Integer                            |
[...]

PR #1369, opened 2019-01-08
Created commands to print various schema elements within mgmt api

chupman:
Signed-off-by: Chris Hupman <chupman@us.ibm.com>

I didn't add the new commands to the docs, but I want to take a look at the indexing docs to include information from the wiki as well as new commands in the next week or two. 

Width is currently 100 characters and does not trim fields so long names can affect the table structure. Personally if I had multiple indexes with matching lengthy prefixes I would rather still see the information and have an ugly table. I may also feel that anyone with index names over 30 characters long deserves an ugly table, but it's easy to change if anyone has a strong opinion about it.

Partially addresses #1163 
-----

### For all changes:

### For code changes:
- [X] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/schema/JanusGraphManagement.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/schema/JanusGraphManagement.java
@@ -401,4 +401,43 @@ public interface JanusGraphManagement extends JanusGraphConfiguration, SchemaMan
      */
     public void rollback();
 
+    /*
+    ##################### PRINT SCHEMA ELEMENTS ##########################
+     */
+
+    /**
+     * Prints out schema information related to vertex and edge labels, indexes, and property keys.
+     *
+     * @return a collection of all the mgmt API schema printing methods
+     */
+    public String printSchema();
+
+    /**
+     * Prints out schema information related to vertex labels.
+     *
+     * @return String with vertex label schema information
+     */
+    public String printVertexLabels();
+
+    /**
+     * Prints out schema infomration related to edge labels.
+     *
+     * @return String with edge label schema information
+     */
+    public String printEdgeLabels();
+
+    /**
+     * Prints out schema information related to property keys.
+     *
+     * @return String with property key schema information
+     */
+    public String printPropertyKeys();
+
+    /**
+     * Prints out schema information related to indexes
+     *
+     * @return String with graph index information
+     */
+    public String printIndexes();
+
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
@@ -22,57 +22,58 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import org.janusgraph.core.Cardinality;
 import org.janusgraph.core.EdgeLabel;
+import org.janusgraph.core.JanusGraph;
+import org.janusgraph.core.JanusGraphEdge;
+import org.janusgraph.core.JanusGraphException;
+import org.janusgraph.core.JanusGraphVertex;
+import org.janusgraph.core.JanusGraphVertexProperty;
 import org.janusgraph.core.Multiplicity;
 import org.janusgraph.core.PropertyKey;
 import org.janusgraph.core.RelationType;
-import org.janusgraph.core.JanusGraphEdge;
-import org.janusgraph.core.JanusGraphException;
-import org.janusgraph.core.JanusGraph;
-import org.janusgraph.core.JanusGraphVertex;
-import org.janusgraph.core.JanusGraphVertexProperty;
-import org.janusgraph.core.VertexLabel;
 import org.janusgraph.core.schema.ConsistencyModifier;
 import org.janusgraph.core.schema.EdgeLabelMaker;
+import org.janusgraph.core.schema.Index;
+import org.janusgraph.core.schema.JanusGraphConfiguration;
+import org.janusgraph.core.schema.JanusGraphIndex;
+import org.janusgraph.core.schema.JanusGraphManagement;
+import org.janusgraph.core.schema.JanusGraphSchemaElement;
+import org.janusgraph.core.schema.JanusGraphSchemaType;
 import org.janusgraph.core.schema.JobStatus;
 import org.janusgraph.core.schema.Parameter;
 import org.janusgraph.core.schema.PropertyKeyMaker;
 import org.janusgraph.core.schema.RelationTypeIndex;
 import org.janusgraph.core.schema.SchemaAction;
 import org.janusgraph.core.schema.SchemaStatus;
-import org.janusgraph.core.schema.JanusGraphConfiguration;
-import org.janusgraph.core.schema.JanusGraphIndex;
-import org.janusgraph.core.schema.Index;
-import org.janusgraph.core.schema.JanusGraphManagement;
-import org.janusgraph.core.schema.JanusGraphSchemaElement;
-import org.janusgraph.core.schema.JanusGraphSchemaType;
 import org.janusgraph.core.schema.VertexLabelMaker;
+import org.janusgraph.core.VertexLabel;
 import org.janusgraph.diskstorage.BackendException;
+import org.janusgraph.diskstorage.configuration.backend.KCVSConfiguration;
 import org.janusgraph.diskstorage.configuration.BasicConfiguration;
 import org.janusgraph.diskstorage.configuration.ConfigOption;
 import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
 import org.janusgraph.diskstorage.configuration.TransactionalConfiguration;
 import org.janusgraph.diskstorage.configuration.UserModifiableConfiguration;
-import org.janusgraph.diskstorage.configuration.backend.KCVSConfiguration;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.ScanMetrics;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.StandardScanner;
 import org.janusgraph.diskstorage.log.Log;
-import org.janusgraph.graphdb.database.IndexSerializer;
-import org.janusgraph.graphdb.database.StandardJanusGraph;
 import org.janusgraph.graphdb.database.cache.SchemaCache;
+import org.janusgraph.graphdb.database.IndexSerializer;
 import org.janusgraph.graphdb.database.serialize.DataOutput;
+import org.janusgraph.graphdb.database.StandardJanusGraph;
 import org.janusgraph.graphdb.internal.ElementCategory;
 import org.janusgraph.graphdb.internal.InternalRelationType;
-import org.janusgraph.graphdb.internal.Order;
 import org.janusgraph.graphdb.internal.JanusGraphSchemaCategory;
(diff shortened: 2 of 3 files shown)


chupman:
Thanks for the feedback @porunov.

Anyone have any clue why travis didn't run on this PR? It ran on my fork, albeit with a solr flaky test failure.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -148,6 +152,9 @@
     private boolean graphShutdownRequired;
     private boolean isOpen;
 
+    private final static String firstDash = "------------------------------------------------------------------------------------------------\n";
+    private final static String dashBreak = "---------------------------------------------------------------------------------------------------\n";
```
porunov: As these variables are constants now, please rename them with capital letters (I.e. like `FIRST_DASH` and `DASH_BREAK`)

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -450,6 +457,172 @@ public JanusGraphIndex apply(@Nullable IndexType indexType) {
...
+        Iterable<VertexLabel> labels = getVertexLabels();
+        boolean hasResults = false;
+        sb.append(FIRSTDASH);
+        sb.append(String.format(pattern, "Vertex Label Name", "Partitioned", "Static"));
+        sb.append(DASHBREAK);
+        for(VertexLabel label: labels) {
+            hasResults = true;
+            sb.append(String.format(pattern, label.name(), label.isPartitioned(), label.isStatic()));
+        }
+        if (hasResults && calledDirectly) sb.append(DASHBREAK);
```
porunov: Add braces to this `if` statement

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -450,6 +457,172 @@ public JanusGraphIndex apply(@Nullable IndexType indexType) {
...
+        } else {
+            sb.append(DASHBREAK);
+        }
+        sb.append(String.format(pattern, "Edge Label Name", "Directed", "Unidirected", "Multiplicity"));
+        sb.append(DASHBREAK);
+        for (EdgeLabel label: labels) {
+            hasResults = true;
+            sb.append(String.format(pattern, label.name(), label.isDirected() ,label.isUnidirected(), label.multiplicity()));
+        }
+        if (hasResults && calledDirectly) sb.append(DASHBREAK);
```
porunov: Add braces to this `if` statement

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -450,6 +457,172 @@ public JanusGraphIndex apply(@Nullable IndexType indexType) {
...
+        } else {
+            sb.append(DASHBREAK);
+        }
+        sb.append(String.format(pattern, "Property Key Name", "Cardinality", "Data Type"));
+        sb.append(DASHBREAK);
+        for (PropertyKey key: keys) {
+            hasResults = true;
+            sb.append(String.format(pattern, key.name(), key.cardinality(), key.dataType()));
+        }
+        if (hasResults && calledDirectly) sb.append(DASHBREAK);
```
porunov: Add braces to this `if` statement

chupman:
Travis tests still aren't picking up this PR. The tests are running on [my fork ](https://travis-ci.org/chupman/janusgraph/builds/477954571) so I can confirm it's passing, but not sure if I should resubmit the PR to trigger the tests in the JanusGraph repo or if we'd be ok with the results from my fork.

porunov:
@chupman I don't know why Travis doesn't run this PR. I am OK with the results from your fork (when it is passed). To merge this PR we need to wait when one of JanusGraph members approve this PR because my approval is non-binding.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
```diff
@@ -57,4 +62,36 @@ public boolean supportsWildcardQuery() {
     protected boolean supportsCollections() {
         return false;
     }
+
+    @Test
+    public void testPrintSchemaElements() {
```
FlorianHockmann: I know JanusGraph has lots of big tests that often test all aspects of a certain functionality in a single test, but I think it's more helpful to have smaller tests where each one only tests one aspect. That makes it easier to understand what's wrong as the failed test won't be named `testPrintSchemaElements` but something like `testPrintVertexLabels` (assuming that each assert here gets its own test) or even `printVertexLabelsShouldOnlyIncludeHeadlineIfNoVertexLabelsDefined`.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
```diff
@@ -57,4 +62,36 @@ public boolean supportsWildcardQuery() {
...
+        GraphOfTheGodsFactory.load(graph);
+        mgmt = graph.openManagement();
+
+        String expected = "99EB9D82CE85D15A9120E31969B3F861";
+        String output = mgmt.printSchema();
+        String outputHash = DigestUtils.md5Hex(output).toUpperCase();
+        assertEquals(outputHash, expected);
```
FlorianHockmann: Would it be too much work to do more fine-grained asserts? Right now, we'll only get the result that two hashes aren't equal if anything doesn't work as expected which could mean that just the formatting changed a bit or even that the output is completely empty. Maybe you could for example check these aspects with relatively low effort:
* The number of lines should be the number of schema elements + head line + dash lines
* The names of all schema elements should be included in the string.
Any way, that's just a suggestion. You can also keep it as is if you prefer it that way.

janusgraph-core/src/main/java/org/janusgraph/core/schema/JanusGraphManagement.java
```diff
@@ -401,4 +401,43 @@
...
+
+    /**
+     * Prints out schema infomation related to property keys.
+     *
+     * @return String with property key schema information
+     */
+    public String printPropertyKeys();
+
+    /**
+     * Prints out schema infomation related to Composite, Mixed, Vertex, Edge, and relation indexes
```
FlorianHockmann: (nitpick) I would not enumerate all kinds of indexes JanusGraph supports here and instead just say that it returns information about index as we probably won't remember that we need to update this comment in case we add another type of index or remove one.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
chupman: I had it split out initially, but wanted to only call GraphOfTheGodsFactory.load() once for all the tests. I was thinking about using `@Before` for the graph setup, but didn't want to affect the other tests in the super class. I'll take another look to see If I can group the setup without affecting the other tests.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -450,6 +457,179 @@ public JanusGraphIndex apply(@Nullable IndexType indexType) {
...
+            if (keyStatus.length > 1) {
+                for (int i = 1; i < keyStatus.length; i++) {
+                    sb.append(String.format(pattern, "", "", "", "", keyStatus[i][0] + ":", keyStatus[i][1]));
+                }
+            }
+        }
+        
+        sb.append(DASHBREAK);
+        sb.append(String.format(pattern, "Edge Index (VCI) Name", "Type", "Unique", "Backing", "Key:", "Status"));
+        sb.append(DASHBREAK);
```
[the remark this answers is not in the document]
chupman: Thanks for catching that. I meant to go back and fix that after I split them out into separate headings.

mikehowey1:
Nice work, might be nice to have the option to return the result similar to printSchema as json  instead of formatted string. I've a script that is very similar to this that our developers are asking for json out instead of a string with the extra formatting for a console.

chupman:
@mikehowey1 I totally agree. The problem I hit is there's really no precedent fort returning JSON that I could find. I can return a string that's formatted as json, but normally objects are returned and I would have to create a new "schema" object that contained all the other objects to conform normally. I'll start a thread on the dev list and see if anyone has any opinions on the matter.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
chupman: A `@Before` statement with conditionals to specify which test to run against will get evaluated by ~2500 lines of inherited tests, due to the way it's structured right now. The other option that now comes to me after reviewing 2000 lines of junit5 updates is to set it up as a Parameterized test.  It'll have timing implications with merging @farodin91's PR and I'll have to rewrite it when I merge to master, but I'll take a crack at switching it over.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
chupman: Number of lines would be really easy to add and I can also pass an array of expected strings I expect output to contain. I still want to keep the hash compare though. We never change graph of gods and if did all it would take is copy the new hash from test failure into the test to update the "mock". I'll add both of these checks when I update to a parameterized test. Thanks for the input.

chupman: @FlorianHockmann I actually can't apply my proposed fix this until we switch to junit 5 and that won't be coming to the 0.2 branch. I just opened up issue #1393 for myself to update the test after both PRs merged. I think I'm happy enough with this for 0.2 and would like to merge it if you agree.

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
FlorianHockmann: Ah, I didn't think of the high setup costs here. Then it makes of course more sense to use just one test here.
Wouldn't a parameterized test have the same problem?

chupman: In junit5 you don't have to parameterize the entire class you can just specify `@ParametrizedTest` instead of `@Test`. I can initialize the graph in the method source for the test and then add the output Strings to the argument, and then do a single cleanup. It should end up a lot cleaner and easier to manage as I add new info, like schema constraints.

FlorianHockmann: OK, thanks for the explanation @chupman! That sounds really like a good solution then.

PR #1392, opened 2019-01-25
Fix tx leaks in ConfigurationManagementGraph

mad:
https://github.com/JanusGraph/janusgraph/issues/1401
-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
@@ -89,6 +89,11 @@ public class ConfigurationManagementGraph {
         instance = this;
     }
 
+    // To be used for testing purposes
+    protected static void shutdownConfigurationManagementGraph() {
+        instance = null;
+    }
+
     /**
      * If one of your "graphs" key was equivalent to "ConfigurationManagementGraph" in your
      * YAML file supplied at server start, then we return the ConfigurationManagementGraph
@@ -303,6 +308,9 @@ public class ConfigurationManagementGraph {
             }
             management.commit();
             graph.tx().commit();
+        } else {
+            management.rollback();
+            graph.tx().rollback();
         }
     }
 
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/management/ConfigurationManagementGraphTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/management/ConfigurationManagementGraphTest.java
@@ -15,32 +15,36 @@
 package org.janusgraph.graphdb.management;
 
 import org.janusgraph.graphdb.configuration.builder.GraphDatabaseConfigurationBuilder;
-import org.janusgraph.graphdb.management.JanusGraphManager;
 import org.janusgraph.core.schema.JanusGraphManagement;
 import static org.janusgraph.core.schema.SchemaStatus.ENABLED;
 import org.janusgraph.core.schema.JanusGraphIndex;
+import org.janusgraph.core.JanusGraphFactory;
 import org.janusgraph.core.PropertyKey;
 
-import org.janusgraph.graphdb.management.ConfigurationManagementGraph;
 import org.janusgraph.diskstorage.configuration.backend.CommonsConfiguration;
 import org.janusgraph.graphdb.database.StandardJanusGraph;
-import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_BACKEND;
 
-import org.apache.tinkerpop.gremlin.server.Settings;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.commons.configuration.MapConfiguration;
 
 import java.util.Map;
 import java.util.HashMap;
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import static org.junit.jupiter.api.Assertions.*;
 
 public class ConfigurationManagementGraphTest {
 
+    @AfterEach
+    public void removeStaticSingletonAfterTest() {
+        JanusGraphManager.shutdownJanusGraphManager();
+        ConfigurationManagementGraph.shutdownConfigurationManagementGraph();
+    }
+
     @Test
-    public void shouldReindexIfPropertyKeyExists() throws Exception {
-        final JanusGraphManager gm = new JanusGraphManager(new Settings());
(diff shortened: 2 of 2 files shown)


chupman:
Hello. Thanks for taking the time to contribute. There are a few things on the checklist that are required to merge. 

1.  An issue opened describing the problem with steps to reproduce if applicable
2. A unit test that would fail without the fix and will pass once it's applied
3. A signed CLA ( individual or corporate) instructions can be found in [janusgraph-legal](https://github.com/JanusGraph/legal)

If you have questions or need help with any of these steps please let us know in this issue or you can email janusgraph-cla@googlegroups.com if it's a CLA related question.

mad:
I'm done

janusgraph-test/src/test/java/org/janusgraph/graphdb/management/ConfigurationManagementGraphTest.java
```diff
@@ -27,16 +28,34 @@
...
+            gmInstance.set(null, null);
+
+            final Field cmgInstance = ConfigurationManagementGraph.class.getDeclaredField("instance");
+            cmgInstance.setAccessible(true);
+            cmgInstance.set(null, null);
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
```
chupman: Nice work.
Only suggestion I have is that JanusGraphManager actually has a method that already does this named `shutdownJanusGraphManager()`

If you added a similiar method to `/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java`
like:
``` 
    // To be used for testing purposes
    protected static void shutdownConfigurationManagementGraph() {
        instance = null;
    }
```

you could simplify the try block to 
```
JanusGraphManager.shutdownJanusGraphManager();
ConfigurationManagementGraph.shutdownConfigurationManagementGraph();
```

How's it going with the CLA?

mad: Done
CLA was sent on Jan 29

porunov:
@mad Please sign your commit with `git commit --amend -s`

mbrukman:
@mad — your CLA has been approved and merged, but you're not using your full name (as specified on the CLA) in your commit, just your first name: see https://github.com/JanusGraph/janusgraph/pull/1392.patch for details.

Please updated your commit as per https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md#configure-your-repo-to-match-the-cla and the bot should pick up the changes automatically.

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
```diff
@@ -303,6 +308,9 @@ private void createIndexIfDoesNotExist(String indexName, String propertyKeyName,
             }
             management.commit();
             graph.tx().commit();
+        } else {
```
FlorianHockmann: So, `management.rollback()` and `graph.tx().rollback()` are now executed for both branches of the `if` statement and additionally at the end of the `catch` block. How about just moving these two lines to the end of the method? Then it could be removed from the three places it is currently in.

janusgraph-test/src/test/java/org/janusgraph/graphdb/management/ConfigurationManagementGraphTest.java
```diff
@@ -64,4 +72,22 @@ public void shouldReindexIfPropertyKeyExists() throws Exception {
         assertEquals(ENABLED, index.getIndexStatus(propertyKey));
         management.commit();
     }
+
+    @Test
+    public void shouldCloseAllTxsIfIndexExists() {
+        final JanusGraphManager gm = new JanusGraphManager(new Settings());
```
FlorianHockmann: (nitpick) You aren't using `gm` anywhere. So, you just remove this line.

FlorianHockmann:
Thanks for your first contribution, @mad!

The change looks really good. I only added two nitpick comments.

Since this is also a bug fix, could you please change the target branch to `0.3` so the fix also goes into the 0.3.2 release?

janusgraph-test/src/test/java/org/janusgraph/graphdb/management/ConfigurationManagementGraphTest.java
mad: Ok

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
mad: Ok, I wrap potentially exception block in try/finllay

FlorianHockmann: Wait, why do you need to change the scope of the try block for that and why is `finally` needed? All I meant is that you can move the two lines out of the different branches to the end of the method.
You had it like this before:

```java
                ); // log.warn()
                management.rollback();
                graph.tx().rollback();
            } // catch
            management.commit();
            graph.tx().commit();
        } else {
            management.rollback();
            graph.tx().rollback();
        }
```

and my suggestion was to do it like this:

```java
        ); // log.warn()
    } // catch
} // if
management.rollback();
graph.tx().rollback();
```

This has exactly the same execution flow as in your previous version, just with less complexity.

With your recent change, you increased the scope of the `try` block a lot which shouldn't be necessary.

mad: We need commit block for initial case when property/index firstly created, but we can't commit if any statements failed or index already exists

Increase try/finally make method more robust, because any exceptions may occur in strange case ex. when property exists with same name but different types/cardinality and indexBuild failed

FlorianHockmann: Wow, I'm just realizing that I overlooked completely that in one case `rollback` is used here and `commit` in the other. So, my suggestion didn't make any sense at all here. Sorry about that!

> any exceptions may occur in strange case ex. when property exists with same name but different types/cardinality

But I'm not sure whether it's a good idea to just catch an exception like this and only log a warning. This sounds more like an error case that really has to be handled by the user.

So, I would suggest that you just change it back to what you had in your first version here. Sorry again for the inconvenience! I really should have given this a more detailed look.

mad: Ok, I'm revert

PR #1400, opened 2019-01-29
Upgrade Cassandra to version 2.2 #1398

### LINKED ISSUE #1398: Upgrade Cassandra to 2.2
JanusGraph still includes Cassandra 2.1 in its distribution which [only gets critical fixes by now](http://cassandra.apache.org/download/) whereas Cassandra 2.2 will be supported until the release of Cassandra 4.0.
Cassandra 2.1 also still uses Thrift in its CQL input format for Spark which means that Spark cannot be used with JanusGraph without having Thrift enabled in Cassandra. This remaining Thrift usage [was removed in Cassandra 2.2](https://issues.apache.org/jira/browse/CASSANDRA-8358). So, the upgrade to Cassandra 2.2 also makes it possible to implement an input format for Spark that only uses CQL (#985).

FlorianHockmann:
This PR includes:
* Upgrade Cassandra from 2.1.20 to 2.2.13.
* Update Astyanax to 3.10.2.
* Include JNA as a test dependency for projects depending on `janusgraph-cassandra`. This was necessary as Cassandra did not start any more without it.
* Updated from `com.codahale.metrics` to `io.dropwizard.metrics` as Cassandra didn't work with the older version from `com.codahale.metrics` any more (the `test.log` contained a lot of errors about a missing method in that package).

This fixes #1398.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,10 +74,9 @@
         <junit.version>5.3.2</junit.version>
         <mrunit.version>1.1.0</mrunit.version>
         <mockito.version>2.23.0</mockito.version>
-        <cassandra.version>2.1.20</cassandra.version>
+        <cassandra.version>2.2.13</cassandra.version>
         <jamm.version>0.3.0</jamm.version>
-        <metrics2.version>2.1.2</metrics2.version>
-        <metrics3.version>3.0.1</metrics3.version>
+        <metrics.version>3.2.2</metrics.version>
         <sesame.version>2.7.10</sesame.version>
         <slf4j.version>1.7.12</slf4j.version>
         <httpcomponents.version>4.4.1</httpcomponents.version>
@@ -711,7 +710,7 @@
             <dependency>
                 <groupId>ch.qos.logback</groupId>
                 <artifactId>logback-classic</artifactId>
-                <version>1.1.2</version>
+                <version>1.1.3</version>
             </dependency>
             <dependency>
                 <groupId>org.codehaus.jettison</groupId>
@@ -1049,24 +1048,24 @@
                 <version>0.7.9</version>
             </dependency>
             <dependency>
-                <groupId>com.yammer.metrics</groupId>
+                <groupId>io.dropwizard.metrics</groupId>
                 <artifactId>metrics-core</artifactId>
-                <version>${metrics2.version}</version>
+                <version>${metrics.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.codahale.metrics</groupId>
-                <artifactId>metrics-core</artifactId>
-                <version>${metrics3.version}</version>
+                <groupId>io.dropwizard.metrics</groupId>
+                <artifactId>metrics-jvm</artifactId>
+                <version>${metrics.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.codahale.metrics</groupId>
+                <groupId>io.dropwizard.metrics</groupId>
                 <artifactId>metrics-ganglia</artifactId>
-                <version>${metrics3.version}</version>
+                <version>${metrics.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.codahale.metrics</groupId>
+                <groupId>io.dropwizard.metrics</groupId>
                 <artifactId>metrics-graphite</artifactId>
-                <version>${metrics3.version}</version>
+                <version>${metrics.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.github.stephenc.high-scale-lib</groupId>
--- a/docs/upgrade.adoc
+++ b/docs/upgrade.adoc
@@ -5,6 +5,15 @@
 Please follow these instructions when upgrading from Titan or an older
 JanusGraph release.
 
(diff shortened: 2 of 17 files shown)


FlorianHockmann:
I just squashed the commits and rebased on master as there were merge conflicts. So, this is now ready for review, especially since we now have a 0.3 branch. I also added @farodin91 as a co-author as he helped me with the error that resulted in the metrics update.

pom.xml
```diff
@@ -662,7 +661,7 @@
             <dependency>
                 <groupId>ch.qos.logback</groupId>
                 <artifactId>logback-classic</artifactId>
-                <version>1.1.2</version>
+                <version>1.1.3</version>
```
porunov: Why do we need `logback-classic` be upgraded specifically to `1.1.3`?

pom.xml
```diff
@@ -74,10 +74,9 @@
         <junit.version>5.3.2</junit.version>
         <mrunit.version>1.1.0</mrunit.version>
         <mockito.version>2.23.0</mockito.version>
-        <cassandra.version>2.1.20</cassandra.version>
+        <cassandra.version>2.2.13</cassandra.version>
         <jamm.version>0.3.0</jamm.version>
-        <metrics2.version>2.1.2</metrics2.version>
-        <metrics3.version>3.0.1</metrics3.version>
+        <metrics.version>3.2.2</metrics.version>
```
porunov: Can we use a newer version? Something like `3.2.6` or `4.0.5`?

pom.xml
FlorianHockmann: `cassandra-all` has this dependency with version `1.1.3`. So, the alternative would be to exclude the dependency from `cassandra-all`, but I figured that it makes more sense to just update the patch version here from `1.1.2` to `1.1.3` to avoid this exclusion.

pom.xml
FlorianHockmann: We used this version, because it's already used in `janusgraph-solr` and the DataStax Cassandra driver also used this version (there could be more dependencies that I don't remember right now).
A 4.* version will probably not work as the other dependencies expect a 3.2.* version which probably leads to problems due to breaking changes introduced in the 4.* versions. I could try if it works with 3.2.6, but then I'll have to add more dependency exclusions to avoid the dependency conflicts.

porunov: @FlorianHockmann if so then it makes sense to use `3.2.2` also.

FlorianHockmann: After your question I looked again where exactly the dependency is used in the different modules and found 2 places where it could be simplified: one now unnecessary dependency exclusion and one dependency that isn't needed anymore. I pushed this as a separate commit in case you want to review that so you don't have to review everything again, but I'll squash these two commits before merging.

porunov: Thanks for the commit. Changes looks good.

sjudeng:
It looks like this was merged with an in progress Travis build and then the branch was deleted, which caused the build to fail (event captured in [commit log](https://github.com/JanusGraph/janusgraph/commits/master)). I thought if possible we want to wait for checks to pass before merging. But glancing through developer docs I don't see this documented anywhere so maybe I'm wrong.

farodin91:
@sjudeng The merge commit seem to run perfectly. https://travis-ci.org/JanusGraph/janusgraph/builds/495967326

FlorianHockmann:
@sjudeng I simply squashed two commits that were already reviewed and built successfully before as they were individual commits on the branch that I merged.
Since I didn't change anything about the content and only the commit history I thought it was OK to just merge directly. Tell me if you disagree and I will do squash and merge as two separate operations in the future and wait for the build of the squashed commits before I merge.

sjudeng:
It's the build for the commit here (https://github.com/JanusGraph/janusgraph/pull/1400/commits/e03d8ae7e3f6054616f9a8b3c4ee05bc9d768055) that I notice the failure on: https://travis-ci.org/JanusGraph/janusgraph/builds/495967033. It's mostly a process question.

FlorianHockmann:
Yes, that was the squashed commit for this feature branch. This commit consisted of two commits before that lived in the feature branch for ~10 days. I just squashed them as it's our usual process to only have one commit per contribution.
The build for this commit then failed because @farodin91 deleted the branch after I merged it on `master` as we don't need it anymore now. That seems to trigger Travis to abort the build.

sjudeng:
In the absence of documentation on the question I was just looking over the master commit log: https://github.com/JanusGraph/janusgraph/commits/master and I don't see other examples of this. So I think we've been waiting for the (albeit often redundant and always slow) checks before merging. But again I don't see this in our developer docs. So I can bring it up on the dev list sometime.

FlorianHockmann:
OK, but I will keep this in mind and wait for the build next time in cases like this. It certainly doesn't look good to have failed builds on `master` even if they only failed because Travis aborted the build.

sjudeng:
Agreed. Though it's really a small price to pay for the heroic work you and @farodin91 are doing to improve testing and documentation in the project.

Issue #1364, opened 2019-01-06
Update TinkerPop to 3.4.0


porunov:
TinkerPop 3.4.0 is released. I think it would be good to update it  in JanusGraph. 
There are a bunch of fixes and enhancements: 
https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asciidoc#release-3-4-0

FlorianHockmann:
I just gave this a first try and noticed that the upgrade to TinkerPop 3.4.0 requires work in a few different areas. Therefore, I suggest that we split this up into separate PRs so we don't have one big PR that is hard to review and to allow different contributors to collaborate on this.
Based on my first try and after briefly studying the [TinkerPop upgrade docs](http://tinkerpop.apache.org/docs/3.4.0/upgrade/) I identified the following tasks for this upgrade, but this list might not be complete.

Mandatory changes:
- [ ] [Credential DSL](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_credential_dsl_changes): Credential graph was removed, instead the Credential DSL introduced in TinkerPop 3.3.3 should be used.
- [ ] [Removal of Giraph Support](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_deprecation_and_removal)
- [ ] [gremlin-server.sh -i Removal](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_deprecation_and_removal): We should probably compare our version of `gremlin-server.sh` with that in TinkerPop
- [ ] [Many other deprecated classes, methods and fields were removed](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_deprecation_and_removal)
- [ ] [Deprecated GraphSONMessageSerializerGremlinV2d0](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_deprecation_and_removal): We need to remove this serializer from our Gremlin Server config files.

Optional / might be done with a separate issue:
- [ ] [Support GraphBinary serialization format](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_graphbinary)
- [ ] [Support TinkerPop Text Predicates for indices](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_text_predicates)
- [ ] [Upsert Graph Feature](tinkerpop.apache.org/docs/3.4.0/upgrade/#_graph_database_providers)
[...]

PR #1413, opened 2019-02-11
Stop using deprecated TinkerPop methods for upgrade to TinkerPop 3.4.0 #1364

FlorianHockmann:
The following methods were deprecated in TinkerPop 3.3.z and removed in TinkerPop 3.4.0:

- `org.apache.tinkerpop.gremlin.structure.Element.Exceptions#elementAlreadyRemoved(Class, Object)`
- `org.apache.tinkerpop.gremlin.process.traversal.step.util.HasContainer#makeHasContainers(String, P)`

I just inlined these methods in the two places where they were called.

There were two more deprecated methods we're using that were removed in TinkerPop 3.4.0, but we can only remove them together with the upgrade to 3.4.0 as they're part of the `GraphManager` interface until 3.4.0, so we still have to implement them until 3.4.0:

- `org.apache.tinkerpop.gremlin.server.GraphManager#getGraphs()`
- `org.apache.tinkerpop.gremlin.server.GraphManager#getTraversalSources()`

Part of #1364

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
@@ -88,9 +88,9 @@ public class AdjacentVertexFilterOptimizerStrategy extends AbstractTraversalStra
                                 && (direction == Direction.BOTH || direction.equals(vertexStep.getDirection().opposite()))) {
                             //Now replace the step with a has condition
                             TraversalHelper.replaceStep(originalStep,
-                                    new HasStep(traversal,
-                                            HasContainer.makeHasContainers(ImplicitKey.ADJACENT_ID.name(), P.eq(vertex))),
-                                    traversal);
+                                new HasStep(traversal,
+                                    new HasContainer(ImplicitKey.ADJACENT_ID.name(), P.eq(vertex))),
+                                traversal);
                         }
                     }
 
--- a/janusgraph-core/src/main/java/org/janusgraph/core/InvalidElementException.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/InvalidElementException.java
@@ -53,9 +53,9 @@ public class InvalidElementException extends JanusGraphException {
     }
 
     public static IllegalStateException removedException(JanusGraphElement element) {
-        Class elementClass = Vertex.class.isAssignableFrom(element.getClass())?Vertex.class:
-                (Edge.class.isAssignableFrom(element.getClass())?Edge.class:VertexProperty.class);
-        return Element.Exceptions.elementAlreadyRemoved(elementClass, element.id());
+        Class elementClass = Vertex.class.isAssignableFrom(element.getClass()) ? Vertex.class :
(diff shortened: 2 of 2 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
```diff
@@ -88,9 +88,9 @@ public void apply(final Traversal.Admin<?, ?> traversal) {
                                 && (direction == Direction.BOTH || direction.equals(vertexStep.getDirection().opposite()))) {
                             //Now replace the step with a has condition
                             TraversalHelper.replaceStep(originalStep,
-                                    new HasStep(traversal,
-                                            HasContainer.makeHasContainers(ImplicitKey.ADJACENT_ID.name(), P.eq(vertex))),
```
chupman: I didn't do enough digging so this is just a question, `.makeHasContainers`  has a return type of  HasContainer[] whereas now I think we'll be passing a single instance. The message `@deprecated As of release 3.2.4. Providers should handle composite {@link P#and} predicates and not rely on splitting.` pretty much answers my question, but I figured I would verify that we're not losing any details.

FlorianHockmann: You're right, `makeHasContainers` [consists of this](https://github.com/apache/tinkerpop/blob/3.3.5/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/HasContainer.java#L191):

```java
if (predicate instanceof AndP) {
    // not important, as our predicate is P.eq() and therefore not AndP
} else {
    return new HasContainer[]{new HasContainer(key, predicate)};
}
```

So, it just returns a new `HasContainer` wrapped in an array. This array is then provided as a 2nd argument to:

```java
HasStep(final Traversal.Admin traversal, final HasContainer... hasContainers)
```

where it's [used like this](https://github.com/apache/tinkerpop/blob/3.3.5/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/HasStep.java#L45):

```java
Collections.addAll(this.hasContainers, hasContainers);
```

With my change, `hasContainers` will be just one container, but that shouldn't change anything for `Collections.addAll()`. At least, that's my understanding of how varargs should work in Java (but I'm also not really a Java dev).

porunov: Yes, 
```java
new HasContainer[]{new HasContainer(key, predicate)}
``` 
and 
```java
new HasContainer(key, predicate)
```
will both be the same array in varargs

FlorianHockmann: @chupman: Are you ok with these answers?

chupman: yes

PR #1414, opened 2019-02-12
Replace assert with Preconditions

### LINKED ISSUE #770: replace all assert statements with Preconditions.checkState
I was bitten by an ugly bug in socialsensor/graphdb-benchmark because Titan/JG was failing silently when checking the validity of a manually-created vertex id. Actually throwing the exception will help us catch the poison pills and increase code quality.

farodin91:
Fixes #770

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/utils/CassandraHelper.java
+++ b/janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/utils/CassandraHelper.java
@@ -80,7 +80,7 @@ public class CassandraHelper {
 
         @Override
         public boolean apply(@Nullable E e) {
-            assert e!=null;
+            Preconditions.checkNotNull(e);
             if (count>=limit || BufferUtil.equals(lastColumn, getter.getColumn(e))) return false;
             count++;
             return true;
@@ -110,7 +110,7 @@ public class CassandraHelper {
             throw new UnsupportedOperationException();
 
         // if left part is BytesToken, right part should be too, otherwise there is no sense in the ring
-        assert rightKeyInclusive instanceof BytesToken;
+        Preconditions.checkArgument(rightKeyInclusive instanceof BytesToken);
 
         // l is exclusive, r is inclusive
         BytesToken l = (BytesToken) leftKeyExclusive;
--- a/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
@@ -79,13 +79,13 @@ public class ConfiguredGraphFactory {
         final Map<String, Object> graphConfigMap = configManagementGraph.getConfiguration(graphName);
         Preconditions.checkState(null == graphConfigMap, String.format("Configuration for graph %s already exists.", graphName));
         final Map<String, Object> templateConfigMap = configManagementGraph.getTemplateConfiguration();
-        Preconditions.checkState(null != templateConfigMap,
+        Preconditions.checkNotNull(templateConfigMap,
                                 "Please create a template Configuration using the ConfigurationManagementGraph#createTemplateConfiguration API.");
         templateConfigMap.put(ConfigurationManagementGraph.PROPERTY_GRAPH_NAME, graphName);
         templateConfigMap.put(ConfigurationManagementGraph.PROPERTY_CREATED_USING_TEMPLATE, true);
 
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
-        Preconditions.checkState(jgm != null, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
+        Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
         final CommonsConfiguration config = new CommonsConfiguration(new MapConfiguration(templateConfigMap));
         final JanusGraph g = (JanusGraph) jgm.openGraph(graphName, (String gName) -> new StandardJanusGraph(new GraphDatabaseConfigurationBuilder().build(config)));
         configManagementGraph.createConfiguration(new MapConfiguration(templateConfigMap));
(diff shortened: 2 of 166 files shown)


farodin91:
I tried to want over the most checkArgument and convert to checkNotNull if possible.

farodin91:
Janusgraph has a lot of a Preconditions checks :see_no_evil:

chupman:
@farodin91 at least now it won't have a lot of asserts as well. I'll try to look at the changes in the near term to start the countdown clock.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/condition/DirectionCondition.java
```diff
@@ -32,9 +33,8 @@
     private final Direction direction;
 
     public DirectionCondition(JanusGraphVertex vertex, Direction dir) {
-        assert vertex != null && dir != null;
```
chupman: This changes the flow a bit. previously it checked that both vertex and dir were not null. Now it checks vertex, assigned it to `this.baseVertex`, and then checks dir and assigns it to `this.direction`. In the event that vertex is populated, but dir is null we'll get different behavior. Either we need 2 separate null checks at the top, or 1 null check and one in line null check like:

```
Preconditions.checkNotNull(dir);
this.baseVertex = Preconditions.checkNotNull(vertex);
this.direction = dir;
```

farodin91: If a single check fails preconditions throws an exception, so the constructor will fail. The assigned is dir null case only performance change, but exception handling take in the most languages much longer than writing a variable. 

One case i would be if baseVertex is static but this is not the case.

porunov:
Hmm. I do understand the original issue but I am not sure if it is wise to replace all `assert` with `Preconditions`. At least if we use `Preconditions` it means that we add new conditions into production code. Assertions by default are disabled and never triggered. I don't think that all assertions should be replaced with `Preconditions`. I think we should really investigate each assertion and decide if we should turn it into a new condition or not. 
Possibly I am wrong but I would like to listen to other opinions before we go farther.

janusgraph-cassandra/src/main/java/org/janusgraph/diskstorage/cassandra/utils/CassandraHelper.java
```diff
@@ -80,7 +80,7 @@ private FilterResultColumns(StaticBuffer lastColumn, int limit,
 
         @Override
         public boolean apply(@Nullable E e) {
```
farodin91: We could use the https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html from jetbrains. Here we write that is nullable but make a check that is not null.

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
```diff
@@ -79,13 +79,13 @@ public static synchronized JanusGraph create(final String graphName) {
...
         Preconditions.checkState(null == graphConfigMap, String.format("Configuration for graph %s already exists.", graphName));
         final Map<String, Object> templateConfigMap = configManagementGraph.getTemplateConfiguration();
-        Preconditions.checkState(null != templateConfigMap,
+        Preconditions.checkNotNull(templateConfigMap,
                                 "Please create a template Configuration using the ConfigurationManagementGraph#createTemplateConfiguration API.");
         templateConfigMap.put(ConfigurationManagementGraph.PROPERTY_GRAPH_NAME, graphName);
         templateConfigMap.put(ConfigurationManagementGraph.PROPERTY_CREATED_USING_TEMPLATE, true);
 
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
-        Preconditions.checkState(jgm != null, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
```
farodin91: Often is checkState used in case of a null check.

farodin91:
We can split the PR into multiple small.
* Replace all not checks which are done with a `checkState` or `checkArguement` with `checkNotNull`.
* Use `checkNotNull` run value as assigment for variable, for example in constructors
* Remove all assert by replacing them with a precondition or remove it.

@porunov What do you think?

porunov:
@farodin91 I definitely think that multiple small PR are better.

-----

> * Replace all not checks which are done with a `checkState` or `checkArguement` with `checkNotNull`.

Agree here.

-----

> * Use `checkNotNull` run value as assigment for variable, for example in constructors

Not sure here. It depends on the code. If it should be checked then sure it is OK to use `checkNotNull`. If it should not be checked but there is an `assert` which checks it for testing purposes then we should not use `checkNotNull`.

-----

> * Remove all assert by replacing them with a precondition or remove it.

I think there are not many places where we should replace `assert` with `Precondition` but I am not sure (didn't check the code). So the more general question is: do we need to remove asserts or not. I think that we should leave them. I think they might be helpful to provide more information about some methods.

farodin91:
> Not sure here. It depends on the code. If it should be checked then sure it is OK to use `checkNotNull`. If it should not be checked but there is an `assert` which checks it for testing purposes then we should not use `checkNotNull`.

Sure, we should check all places.

---
> I think there are not many places where we should replace `assert` with `Precondition` but I am not sure (didn't check the code). So the more general question is: do we need to remove asserts or not. I think that we should leave them. I think they might be helpful to provide more information about some methods.

For me assert are more confusing, because they are default disable. I prefer replacement or removing.
Replacement could an annotation to function parameters or using preconditions.

porunov:
> For me assert are more confusing, because they are default disable. I prefer replacement or removing.
> Replacement could an annotation to function parameters or using preconditions.

Annotations are good. And I do think that using annotations will better explain the code but they are used more like documentation. `assert` are used more like for logic bugs finding or something like that. I.e. like when you know that no `null` value can arrive in the method so you even don't contract a checking condition because you know that `null` value is impossible here. But if you are not completely sure, you add `assert` to check your code and if it fails then you still don't add a condition but rather search for the bug and fix it in another place.
I don't think that removing all asserts is a good idea. Yes, they are disabled by default because they are meant to be used only for testing and not in production. When you want to manually test janusgraph you can enable them and try to fail some assert. If you fail any assert, then it is good because you found a bug which we can resolve later.
Right now I think:
- Adding annotations is a good idea
- Removing all asserts is a bad idea

farodin91:
We should move this design decision to the mailing list to allow to get more publicity. 
> And I do think that using annotations will better explain the code but they are used more like documentation. `assert` are used more like for logic bugs finding or something like that. I.e. like when you know that no `null` value can arrive in the method so you even don't contract a checking condition because you know that `null` value is impossible here. But if you are not completely sure, you add `assert` to check your code and if it fails then you still don't add a condition but rather search for the bug and fix it in another place.

For Null checks, i not prefer `assert` as your idea for documentation. If you use intellij idea to debug the tests which contains null annotation they are converted into assert during compilation. So for local tests they would fail.

For more complex, I would i agree on leaving them in the code.

porunov:
@farodin91 Agree. We should move this topic into `janusgraph-dev` group.

farodin91:
I will close this PR and open a new PR for:
* Replace all null checks which are done with a `checkState` or `checkArguement` with `checkNotNull`.
* Use `checkNotNull` run value as assigment for variable, for example in constructors

PR #1427, opened 2019-02-19
Use new Credential DSL for authentication #1364

FlorianHockmann:
TinkerPop [deprecated](http://tinkerpop.apache.org/docs/3.4.0/upgrade/#_credential_dsl_changes) the old `CredentialGraph` with version 3.3.3 and it will be removed completely in version 3.4.0. Instead, the new Credential traversal DSL should be used.
So, this is a precondition for the upgrade to TinkerPop 3.4.0.

The tests made extensive use of mocking of the `CredentialGraph` which is now not possible any more. Since they were quite complex in general and contained a lot of duplicated code, I decided to refactor them completely instead of trying to fix them as that would have been probably even more work.
During the refactoring of the tests I also noticed some minor issues in the code which I also fixed.

I tested manually that authentication with the `JanusGraphSimpleAuthenticator` still works and that it can also use credentials stored in Cassandra that were created with JanusGraph 0.3.1.

Part of #1364.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
+++ b/janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
@@ -20,7 +20,8 @@ import static org.apache.tinkerpop.gremlin.server.auth.SimpleAuthenticator.CONFI
 import java.net.InetAddress;
 import java.util.Map;
 
-import org.apache.tinkerpop.gremlin.groovy.jsr223.dsl.credential.CredentialGraph;
+import org.apache.tinkerpop.gremlin.groovy.jsr223.dsl.credential.CredentialTraversalSource;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
 import org.apache.tinkerpop.gremlin.server.auth.Authenticator;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.janusgraph.core.Cardinality;
@@ -53,10 +54,12 @@ abstract public class JanusGraphAbstractAuthenticator implements Authenticator {
      */
     public static final String CONFIG_DEFAULT_PASSWORD = "defaultPassword";
 
-    private static final String USERNAME_INDEX_NAME = "byUsername";
+    /**
+     * Name of the index that enables an efficient retrievals by usernames.
+     */
+    public static final String USERNAME_INDEX_NAME = "byUsername";
 
-    protected CredentialGraph credentialStore;
-    protected JanusGraph backingGraph;
+    private CredentialTraversalSource credentials;
 
     @Override
     public boolean requireAuthentication() {
@@ -66,8 +69,8 @@ abstract public class JanusGraphAbstractAuthenticator implements Authenticator {
     @Override
     abstract public SaslNegotiator newSaslNegotiator(final InetAddress remoteAddress);
 
-    public CredentialGraph createCredentialGraph(JanusGraph graph) {
-        return new CredentialGraph(graph);
+    public CredentialTraversalSource createCredentials(JanusGraph graph) {
+        return graph.traversal(CredentialTraversalSource.class);
     }
 
     public JanusGraph openGraph(String conf) {
@@ -80,18 +83,16 @@ abstract public class JanusGraphAbstractAuthenticator implements Authenticator {
         Preconditions.checkArgument(config != null, String.format(
                     "Could not configure a %s - provide a 'config' in the 'authentication' settings",
                     this.getClass().getName()));
-        
 
         Preconditions.checkState(config.containsKey(CONFIG_CREDENTIALS_DB), String.format(
-                    "Credential configuration missing the %s key that points to a graph config file or graph name", CONFIG_CREDENTIALS_DB));
-
-        if (!config.containsKey(CONFIG_DEFAULT_USER) || !config.containsKey(CONFIG_DEFAULT_PASSWORD)) {
-            logger.warn(String.format("%s and %s should be defined to bootstrap authentication", CONFIG_DEFAULT_USER, CONFIG_DEFAULT_PASSWORD));
-        }
+            "Credential configuration missing the %s key that points to a graph config file or graph name", CONFIG_CREDENTIALS_DB));
+        Preconditions.checkState(config.containsKey(CONFIG_DEFAULT_USER), String.format(
+            "Credential configuration missing the %s key for the default user", CONFIG_DEFAULT_USER));
+        Preconditions.checkState(config.containsKey(CONFIG_DEFAULT_PASSWORD), String.format(
+            "Credential configuration missing the %s key for the default password", CONFIG_DEFAULT_PASSWORD));
 
         final JanusGraph graph = openGraph(config.get(CONFIG_CREDENTIALS_DB).toString());
-        backingGraph = graph;
-        credentialStore = createCredentialGraph(graph);
+        credentials = createCredentials(graph);
         graph.tx().rollback();
         ManagementSystem mgmt = (ManagementSystem) graph.openManagement();
         if (!mgmt.containsGraphIndex(USERNAME_INDEX_NAME)) {
@@ -114,8 +115,32 @@ abstract public class JanusGraphAbstractAuthenticator implements Authenticator {
         }
 
         final String defaultUser = config.get(CONFIG_DEFAULT_USER).toString();
-        if (credentialStore.findUser(defaultUser) == null) {
-            credentialStore.createUser(defaultUser, config.get(CONFIG_DEFAULT_PASSWORD).toString());
+        if (!userExists(defaultUser)) {
+            createUser(defaultUser, config.get(CONFIG_DEFAULT_PASSWORD).toString());
+        }
+    }
+
+    protected Vertex findUser(final String username) {
+        credentials.tx().rollback();
+        final GraphTraversal<Vertex,Vertex> t = credentials.users(username);
+        final Vertex v = t.hasNext() ? t.next() : null;
+        if (t.hasNext()) throw new IllegalStateException(String.format("Multiple users with username %s", username));
+        return v;
+    }
+
+    private boolean userExists(final String username) {
+        return findUser(username) != null;
+    }
+
+    private void createUser(final String username, final String password) {
+        credentials.tx().rollback();
+
+        try {
+            credentials.user(username,password).iterate();
+            credentials.tx().commit();
+        } catch (Exception ex) {
+            credentials.tx().rollback();
+            throw new RuntimeException(ex);
         }
     }
 }
--- a/janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/HMACAuthenticator.java
+++ b/janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/HMACAuthenticator.java
@@ -126,7 +126,8 @@ public class HMACAuthenticator extends JanusGraphAbstractAuthenticator {
             return user;
         } else if (credentials.get(PROPERTY_TOKEN) != null) {
             if (validateToken(credentials)) {
-                return new AuthenticatedUser(credentials.get(PROPERTY_USERNAME));
+                final Map<String, String> tokenMap = parseToken(credentials.get(PROPERTY_TOKEN));
+                return new AuthenticatedUser(tokenMap.get(PROPERTY_USERNAME));
             } else {
                 throw new AuthenticationException("Invalid token");
             }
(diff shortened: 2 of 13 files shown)


janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
```diff
@@ -114,8 +115,33 @@ public void setup(final Map<String,Object> config) {
...
         final String defaultUser = config.get(CONFIG_DEFAULT_USER).toString();
-        if (credentialStore.findUser(defaultUser) == null) {
-            credentialStore.createUser(defaultUser, config.get(CONFIG_DEFAULT_PASSWORD).toString());
+        if (!userExists(defaultUser)) {
+            createUser(defaultUser, config.get(CONFIG_DEFAULT_PASSWORD).toString());
+        }
+    }
+
+    protected Vertex findUser(final String username) {
+        credentials.tx().rollback();
```
farodin91: Why do you call rollback?

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
```diff
@@ -114,8 +115,33 @@ public void setup(final Map<String,Object> config) {
...
+        final GraphTraversal<Vertex,Vertex> t = credentials.users(username);
+        final Vertex v = t.hasNext() ? t.next() : null;
+        if (t.hasNext()) throw new IllegalStateException(String.format("Multiple users with username %s", username));
+        return v;
+    }
+
+    private boolean userExists(final String username) {
+        credentials.tx().rollback();
```
farodin91: Same here

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
```diff
@@ -114,8 +115,33 @@ public void setup(final Map<String,Object> config) {
...
+        return credentials.users(username).hasNext();
+    }
+
+    private void createUser(final String username, final String password) {
+        credentials.tx().rollback();
```
farodin91: Same here

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
FlorianHockmann: I basically just copied these methods over from TinkerPop's `CredentialGraph` and didn't want to change anything about their functionality. `CredentialGraph.findUser` [looks like this](https://github.com/apache/tinkerpop/blob/ea1e5a323a57d265bc3e588c6e40a30dfc2e8c42/gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/dsl/credential/CredentialGraph.java#L56):

```java
public Vertex findUser(final String username) {
        if (supportsTransactions) g.tx().rollback();
        final GraphTraversal<Vertex,Vertex> t = g.V().has(CredentialGraphTokens.PROPERTY_USERNAME, username);
        final Vertex v = t.hasNext() ? t.next() : null;
        if (t.hasNext()) throw new IllegalStateException(String.format("Multiple users with username %s", username));
        return v;
}
```

Since JanusGraph supports transactions, the first line does the rollback.

I assume that this rollback is done to prevent a situation where credentials that were not committed at all are used to authenticate a user. That could be very confusing when the authentication doesn't work any more later one if these credentials were never committed.

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
FlorianHockmann: Same answer as above basically. Instead of `userExists()` the code previously did this to perform the existence check:
```java
credentialStore.findUser(defaultUser) == null
```
Since `findUser()` had the rollback, I also added it here.

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
FlorianHockmann: Again, same answer. [Here](https://github.com/apache/tinkerpop/blob/ea1e5a323a57d265bc3e588c6e40a30dfc2e8c42/gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/dsl/credential/CredentialGraph.java#L69) you can find `CredentialGraph.createUser()` from TinkerPop 3.3.z for reference.

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
farodin91: Make sense.

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
porunov: @FlorianHockmann should we throw an exception in this method if multiple users exist?

FlorianHockmann: `userExists()` is currently only called in line 119 for the `defaultUser`:

```java
final String defaultUser = config.get(CONFIG_DEFAULT_USER).toString();
if (!userExists(defaultUser)) {
    createUser(defaultUser, config.get(CONFIG_DEFAULT_PASSWORD).toString());
}
```

This code will be executed on every start of JanusGraph and it's expected that the `defaultUser` exists after the first start of JanusGraph. So, at least in this case we should not throw an exception.

porunov: @FlorianHockmann I am just interested in the flow when there multiple default users exist. Is it a normal case when multiple same users exist? I didn't check but I assume that this is a possible scenario when we create multiple default users. Should we remove unnecessary users or they will not disturb us?

FlorianHockmann: I'm not sure how multiple default users should exist. You specify the default user in the config like this:
```
config: {
     defaultUsername: user,
     defaultPassword: password,
     credentialsDb: conf/janusgraph-credentials-server.properties
}
```

You could of course change the `defaultUsername` and start JanusGraph again which would then create another _default_ user, but I don't see how we could prevent that as JanusGraph won't know any more that the already existing user _user_ was the default user before. We would have to either remove **all** existing users (sounds like a very bad idea to me) or mark the default user somehow so we can check whether such a default user already exists and then delete all old ones. If we want to follow that path, then we should do it in a separate PR in my opinion.
I'm however not sure in general how the credential graph should be managed right now. We should probably at something for that to our docs.

porunov: @FlorianHockmann sure, I don't tell to do it in this PR. Possibly we should discuss it somewhere else. I just thought about consistency and race conditions when we creating a default user.
1. We could use eventual consistent storage backend. That is why `userExists(defaultUser)` doesn't guarantee that the user exists or not exists. For example, if we use Cassandra with small consistency level or with multi data center setup and we start multiple instances together then this method isn't sufficient (unless it uses some locking mechanism).
2. Again, even if we use one DC with big consistency level there is a small time gap between `userExists(...)` and `createUser(...)` which may be enough to execute `userExists(...)` two times and return `false`.

FlorianHockmann: Ah, now I see what you mean. I really didn't think about race conditions due to eventual consistency at all here :-/

I think an issue makes sense for this case. Do you want to create one?

porunov: Sure. I've created it here #1434

janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java
```diff
@@ -114,8 +115,33 @@ public void setup(final Map<String,Object> config) {
         }
 
         final String defaultUser = config.get(CONFIG_DEFAULT_USER).toString();
-        if (credentialStore.findUser(defaultUser) == null) {
-            credentialStore.createUser(defaultUser, config.get(CONFIG_DEFAULT_PASSWORD).toString());
+        if (!userExists(defaultUser)) {
```
porunov: We used to use a check like `credentialStore.findUser(defaultUser) == null` here which throws an exception if multiple users exist. Now we don't throw this exception if multiple users exist because we use `userExists(defaultUser)`. Is it OK?

FlorianHockmann: Good find! I'll modify `userExists()` to also throw an exception in that case to not change the behaviour here.

FlorianHockmann: After some more investigation I noticed that it's not possible to get duplicate users because we're using a uniqueness constraint on the `username` property. I explained this a bit further in [your issue about the race condition](https://github.com/JanusGraph/janusgraph/issues/1434#issuecomment-465906853).

So, I actually went the opposite way and removed the duplicate check in `findUser()` instead.

I pushed that with a separate commit so you can easier review it and I will squash them when you're done with the review.

porunov: @FlorianHockmann I answered about commented in the issue that I am not sure how it will work with eventual consistency because that code doesn't use locks. I.e. I think we should use something like:
```
mgmt.setConsistency(name, ConsistencyModifier.LOCK)
mgmt.setConsistency(index, ConsistencyModifier.LOCK)
```
but I am not sure. We need to test it better. 

As this isn't related to the current PR, I think right now it will be better to not change a flow (I.e. throw exception in `findUser` and `userExists` methods) and then after some investigation open a new PR (if needed) to remove duplicate checks.

What do you think?

FlorianHockmann: I'm not sure what you mean with _that code doesn't use locks_. Sure, the code itself doesn't set any locks, but JanusGraph will acquire a lock in the backend to modify anything that is guarded by a uniqueness constraint. This is basically distributed locking backed by the storage backend.
So, you won't be able to add two vertices with the same `username` property, even if you try it in parallel from different threads / JanusGraph instances.

porunov: In the issue you said that we use `uniqueness constraint`. And you pointed to this part of code:
https://github.com/JanusGraph/janusgraph/blob/3812806a98598965950c366f1613210091330b64/janusgraph-server/src/main/java/org/janusgraph/graphdb/tinkerpop/gremlin/server/auth/JanusGraphAbstractAuthenticator.java#L99

I checked that code and see that we don't use locks there. I.e. we don't use `mgmt.setConsistency(name, ConsistencyModifier.LOCK)`.

Yes, we use `uniqueness constraint` but I pointed to [this documentation](https://docs.janusgraph.org/latest/eventual-consistency.html) which explains that we must obtain locks to ensure consistency. As I said, I don't know how `uniqueness constraint` works in eventual consistent backend without locks. If you know that with `unique` constraint we won't be able to create duplicate user without `ConsistencyModifier.LOCK` in eventual consistent backend then I am OK with the current change. If you are not sure then I think we should not remove duplicate checks in your PR.

FlorianHockmann: Ah, now I see what you mean. My understanding was always that making an index `unique` is already enough to enforce uniqueness of its indexed properties. But it looks like I was wrong about that and you actually have to enable locking explicitly for eventually consistent storage backends like it's shown in the example you linked to in the docs.

This is also explained[ here](https://docs.janusgraph.org/0.3.1/indexes.html#_composite_index) in the docs:

> To enforce uniqueness against an eventually consistent storage backend, the consistency of the index must be explicitly set to enabling locking.

So, I agree with you that I should not remove the existence checks with this PR. I'll revert my last commit here and instead add another check in `userExists()` as you suggested initially.
I suggest that we enable locking explicitly for eventually consistent storage backends in the issue you created.

FlorianHockmann: Done

FlorianHockmann:
Thanks for your review, @porunov. I squashed the two commits.

PR #1442, opened 2019-02-25
Allow marking of tests as flaky #1016

### LINKED ISSUE #1016: Only retry only failing tests instead of running the complete test project again
I saw only a single failing tests in Travis CI run and the complete test project is ran again up to 3 times.

Idea to implement:
* Mark tests with an attribute and use a maven test runner for the retry policy.

farodin91:
Fixes #1016 

Based on stackoverflow post: https://stackoverflow.com/a/46207476/2978513

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTest.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTest.java
@@ -1,0 +1,35 @@
+// Copyright 2019 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Based on Stackoverflow post: https://stackoverflow.com/a/46207476/2978513
+// License: https://choosealicense.com/licenses/unlicense/
+
+package org.janusgraph.testutil;
+
+import org.junit.jupiter.api.TestTemplate;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@TestTemplate
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+@ExtendWith(FlakyTestExtension.class)
+public @interface FlakyTest {
+    int invocationCount() default 4;
+    int minSuccess() default 2;
+}
--- a/janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTestExtension.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTestExtension.java
@@ -1,0 +1,58 @@
+// Copyright 2019 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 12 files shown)


janusgraph-test/src/main/java/org/janusgraph/testutil/RetryTemplateContext.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public String getDisplayName(int invocationIndex) {
+        return "Invocation number " + invocationIndex + " (requires " + minSuccess + " success)";
+    }
+
+    @Override
+    public List<Extension> getAdditionalExtensions() {
+        List<Extension> objects = new java.util.ArrayList<>();
+        objects.add(new RetryingTestExecutionExtension(invocation, maxInvocations, minSuccess));
+        return objects;
```
porunov: `return Collections.singletonList(new RetryingTestExecutionExtension(invocation, maxInvocations, minSuccess))`

janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@TestTemplate
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+@ExtendWith(FlakyTestExtension.class)
+public @interface FlakyTest {
+    int invocationCount();
```
porunov: (not necessary)
Maybe add default values? I.e. `int invocationCount() default 8` and `int minSuccess() default 4`?

farodin91:
The `testIndexReplay` fails often.

janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTestExtension.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.stream.Stream;
+
+public class FlakyTestExtension implements TestTemplateInvocationContextProvider {
+
+    @Override
+    public boolean supportsTestTemplate(ExtensionContext context) {
+        return AnnotationUtils.isAnnotated(context.getTestMethod(), FlakyTest.class);
+    }
+
+    public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context) {
```
porunov: Add `@Overwrite` annotation

janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTestExtension.java
```diff
@@ -0,0 +1,57 @@
...
+        FlakyTest annotation = context.getTestMethod()
+            .flatMap(testMethods -> AnnotationSupport.findAnnotation(testMethods, FlakyTest.class))
+            .orElseThrow(() -> new ExtensionContextException("The extension should not be executed "
```
porunov: I didn't check but is it possible that we get a method here which isn't annotated with `@FlakyTest`?
I assume that JUnit checks `supportsTestTemplate` before calling `provideTestTemplateInvocationContexts`. If so, I think we can change `orElseThrow` to `get`. What do you think?

farodin91: This shouldn't happen. All example, i saw throw an explicit exception.

farodin91:
@porunov I squashed everything and rebase it on the top of 0.3.

janusgraph-test/src/main/java/org/janusgraph/testutil/FlakyTestExtension.java
porunov: I just thought that it is an unnecessary check if it isn't possible to get a method without `FlakyTest` here. If it is possible then sure we must throw `ExtensionContextException`.

farodin91: The extension can be used using `@ExtendWith`, If i'm right.

Issue #984, opened 2018-04-02
Document replacement for deprecated StandardJanusGraphTx#multiQuery() method


mbrukman:
See [this thread][1] for details:

> As remarked earlier here, my guess is that the MultiQuery API is deprecated because of replacement by:
>
> https://static.javadoc.io/org.janusgraph/janusgraph-core/0.2.0/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.html 
>
> It would be useful, though, to have:
> - a section in the JanusGraph ref docs about the MultiVertexCentricQuery
> - some pointers in the deprecated API towards the new one

[1]: https://groups.google.com/d/msg/janusgraph-users/PBRZhQNLbxw/235QNyQzAAAJ

twilmes:
I can only guess what drove the deprecation decision originally, but my bet is that it had something to do with deprecating it for external use. Historically, prior to Titan 1.0, the only way to parallelize communication with the storage layer was via direct use of the Titan multiquery calls. Gremlin traversals would not make use of them and you were limited to lazy loading from the storage layer unless you used this API directly. Titan 1.0 came with TinkerPop 3 support which in turn gave the graph providers the ability to apply their own optimization strategies to Gremlin traversals. With this, Titan 1.0 came with a `query.batch` setting that lives on, to this day, in Janus. If you flip that on, Gremlin traversals will use multiquery behind the scenes and you'll see the benefits of reduced latencies. So, long story short, multiquery is currently used and internally, and we'll be into the foreseeable future. I believe the deprecation was probably meant more as a marker to outside consumers not to use this API directly, and to instead stick accessing it indirectly, through Gremlin. The documentation used to contain a section on using multiquery, but it was removed.

At this point, I'm tempted to remove deprecation because it seems inconsistent with how the rest of the "internal" Janus API is treated. Technically a user could use the other methods we expose, like the regular old `query`, but I'd never recommend they do that, and those other methods are not deprecated. I think the docs could benefit from a more opinionated section near the beginning driving users towards using the Gremlin process API and steering them away from the TinkerPop structure API (`graph.addVertex`, `graph.addEdge`, etc.) and the Janus APIs (`query`, `multiquery`, etc.).

smcquillan:
[...]

Issue #1459, opened 2019-03-01
Flaky Berkeley test: JanusGraphOperationCountingTest.testIdCounts


FlorianHockmann:
```java
[ERROR] Failures: 
[ERROR]   BerkeleyOperationCountingTest>JanusGraphOperationCountingTest.testIdCounts:109->JanusGraphOperationCountingTest.verifyStoreMetrics:410 org.janusgraph.sys.janusgraph_ids.getSlice.calls ==> expected: <6> but was: <5>
```

Job was:
```
MODULE='berkeleyje'
```

FlorianHockmann:
Confirming that it's independent of the backend as this test just also failed for CQL for a PR that shouldn't affect this test: #3239.

```
Error:  Failures: 
Error:    CQLOperationCountingTest>JanusGraphOperationCountingTest.testIdCounts:137->JanusGraphOperationCountingTest.verifyStoreMetrics:438 org.janusgraph.sys.janusgraph_ids.getSlice.calls ==> expected: <6> but was: <5>
```

PR #1437, opened 2019-02-21
Modernize Documentation using mkdocs

farodin91:
* Automate travis ci build and release
* Use mkdocs-material as design for the documentation
* Convert all asciidocs into markdown

Part of #1420

Signed-off-by: Jan Jansen <jan.jansen@gdata.de>

Important part to review
* multi version support
* rewrite of the TravisCI file

DEMO:  https://farodin91.github.io/janusgraph/

Need to be done by a TSC member:
* [ ] allowing TravisCI to push the documentation
   * https://travis-ci.com/farodin91/janusgraph/builds/101604015
   * We can use the `gh-pages` branch or a different repo
* [ ] release branch needed to be started with v.., for example v0.2, v0.3

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there an issue associated with this PR? Is it referenced in the commit message?
- [x] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered?
- [x] If this PR is a documentation-only change, have you added a `[skip ci]`
  tag to the first line of your commit message to avoid spending CPU cycles in
  Travis CI when no code, tests, or build configuration are modified?

### Note:
Please ensure that once the PR is submitted, you check Travis CI for build issues and submit an update to your PR as soon as possible.

Issue #1044, opened 2018-05-01
Enable multiquery optimization for RepeatStep, MatchStep, BranchStep, and OptionalStep


twilmes:
`RepeatStep`, `MatchStep`, `BranchStep`, and `OptionalStep` are currently incompatible with the JanusVertexStep multiquery optimization. Currently, if a `JanusGraphVertexStep` or `JanusGraphPropertiesStep` is nested within one of these steps, multiquery is disabled. This does not effect other candidate Janus specific steps in other spots in the query. #71 added tests in for all of the storage backends with multiquery enabled and also added this multiquery incompatible step list & logic to unbreak the `query.batch` config option. For this issue, we should make the necessary updates to allow the multiquery optimization to work with all steps.

smcquillan:
@twilmes @FlorianHockmann I've been digging into the reasons behind the NPE reported in #71 which prevented the mulitQuery being used against these types of parent steps. And found that problem occurs due to the way that the parent traversal feeds its starts into its `JanusGraphVertexStep` child.

The parent drip feeds the steps one at a time into its child traversals. However, the `JanusGraphVertexStep` child only does its initialisation on receipt of the fist start vertex. So when a subsequent start is fed in it won't reinitialise and this results in the NPE.

I've been working towards a way to feed all the relevant start vertices into the child steps and will be raising a pull request shortly.

PR #1460, opened 2019-03-01
Un-deprecate multiQuery #984 and pre-fetch properties to cache

smcquillan:

Pull request  to un-deprecates multiQuery for issue #984 

In addition it adds an optimisation to use multiQuery to pre-fetch properties for multiple vertices into the cache if the next step of the query is a `has` step. For queries following large number of edges this eliminates many network trips to the storage backend for the vertex properties and can give significant performance improvements, see the issue for an example.

-----

### For all changes:
- [x ] Is your initial contribution a single, squashed commit?

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
@@ -264,6 +264,12 @@ public class GraphDatabaseConfiguration {
                     "performance improvement if there is a non-trivial latency to the backend.",
             ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<Boolean> BATCH_PROPERTY_PREFETCHING = new ConfigOption<>(QUERY_NS,"batch-property-prefetch",
+            "Whether to do a batched pre-fetch of all properties on adjacent vertices against the storage backend prior to evaluating a has condition against those vertices. " +
+                    "Because these vertex properties will be loaded into the transaction-level cache of recently-used vertices when the condition is evaluated this can " +
+                    "lead to significant performance improvement if there are many edges to adjacent vertices and there is a non-trivial latency to the backend.",
+            ConfigOption.Type.MASKABLE, false);
+
     // ################ SCHEMA #######################
     // ################################################
 
@@ -1214,6 +1220,7 @@ public class GraphDatabaseConfiguration {
     private Boolean propertyPrefetching;
     private boolean adjustQueryLimit;
     private Boolean useMultiQuery;
+    private Boolean batchPropertyPrefetching;
     private boolean allowVertexIdSetting;
     private boolean logTransactions;
     private String metricsPrefix;
@@ -1298,6 +1305,10 @@ public class GraphDatabaseConfiguration {
 
     public boolean useMultiQuery() {
         return useMultiQuery;
+    }
+
+    public boolean batchPropertyPrefetching() {
+        return batchPropertyPrefetching;
     }
 
     public boolean adjustQueryLimit() {
@@ -1410,6 +1421,7 @@ public class GraphDatabaseConfiguration {
 
         propertyPrefetching = configuration.get(PROPERTY_PREFETCHING);
         useMultiQuery = configuration.get(USE_MULTIQUERY);
+        batchPropertyPrefetching = configuration.get(BATCH_PROPERTY_PREFETCHING);
         adjustQueryLimit = configuration.get(ADJUST_LIMIT);
         allowVertexIdSetting = configuration.get(ALLOW_SETTING_VERTEX_ID);
         logTransactions = configuration.get(SYSTEM_LOG_TRANSACTIONS);
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
@@ -141,6 +141,12 @@ public class MultiVertexCentricQueryBuilder extends BasicVertexCentricQueryBuild
     }
 
     @Override
+    public void preFetch() {
+        profiler.setAnnotation(QueryProfiler.MULTIPREFETCH_ANNOTATION, true);
+        properties();
+    }
+
+    @Override
     public Map<JanusGraphVertex, Iterable<JanusGraphRelation>> relations() {
         return (Map)(isImplicitKeyQuery(RelationCategory.RELATION)?
                 executeImplicitKeyQuery():
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
@@ -16,6 +16,7 @@ package org.janusgraph.graphdb.tinkerpop.optimize;
 
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.lambda.ElementValueTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.step.map.FlatMapStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.GraphStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.NoOpBarrierStep;
 import org.apache.tinkerpop.gremlin.process.traversal.util.AndP;
@@ -29,7 +30,6 @@ import org.apache.tinkerpop.gremlin.process.traversal.Order;
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.step.HasContainerHolder;
-import org.apache.tinkerpop.gremlin.process.traversal.step.Ranging;
 import org.apache.tinkerpop.gremlin.process.traversal.step.filter.HasStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.filter.OrStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.filter.RangeGlobalStep;
@@ -42,9 +42,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.util.HasContainer;
 import org.javatuples.Pair;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Comparator;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -328,5 +326,35 @@ public interface HasStepFolder<S, E> extends Step<S, E> {
                 traversal.removeStep(nextStep);
             }
         }
+    }
+
+    /**
+     * @param janusgraphStep The step to test
+     * @return True if there are 'has' steps following this step and no subsequent range limit step
+     */
+    static boolean foldableHasContainerNoLimit(FlatMapStep<?, ?> janusgraphStep) {
+        boolean foldableHasContainerNoLimit = false;
+        Step<?, ?> currentStep = janusgraphStep.getNextStep();
+        while (true) {
+            if (currentStep instanceof OrStep) {
+                for (final Traversal.Admin<?, ?> child : ((OrStep<?>) currentStep).getLocalChildren()) {
+                    if (!validFoldInHasContainer(child.getStartStep(), false)){
+                        return false;
+                    }
+                }
+                foldableHasContainerNoLimit =  true;
+            } else if (currentStep instanceof HasContainerHolder) {
+                if  (validFoldInHasContainer(currentStep, true)) {
+                    foldableHasContainerNoLimit =  true;
+                }
+            } else if (currentStep instanceof RangeGlobalStep) {
+                return false;
+            } else if (!(currentStep instanceof IdentityStep) && !(currentStep instanceof NoOpBarrierStep)) {
+                break;
+            }
+            currentStep = currentStep.getNextStep();
+        }
+
+        return foldableHasContainerNoLimit;
     }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphEdgeVertexStep.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphEdgeVertexStep.java
@@ -1,0 +1,111 @@
+// Copyright 2019 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.tinkerpop.optimize;
+
+import org.janusgraph.core.JanusGraphMultiVertexQuery;
+import org.janusgraph.graphdb.query.profile.QueryProfiler;
+import org.janusgraph.graphdb.query.vertex.BasicVertexCentricQueryBuilder;
+import org.janusgraph.graphdb.tinkerpop.profile.TP3ProfileWrapper;
+
+import com.google.common.collect.Sets;
+
+import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
+import org.apache.tinkerpop.gremlin.process.traversal.step.Profiling;
+import org.apache.tinkerpop.gremlin.process.traversal.step.map.EdgeVertexStep;
+import org.apache.tinkerpop.gremlin.process.traversal.util.FastNoSuchElementException;
+import org.apache.tinkerpop.gremlin.process.traversal.util.MutableMetrics;
+import org.apache.tinkerpop.gremlin.structure.Direction;
+import org.apache.tinkerpop.gremlin.structure.Edge;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A JanusGraphEdgeVertexStep is identical to a {@link EdgeVertexStep}. The only difference
+ * being that it can use multiQuery to pre-fetch the vertex properties prior to the execution
+ * of any subsequent has steps and so eliminate the need for a network trip for each vertex.
+ * It implements the optimisation enabled via the query.batch-property-prefetch config option.
+ */
+public class JanusGraphEdgeVertexStep extends EdgeVertexStep implements Profiling {
+
+    private boolean initialized = false;
+    private QueryProfiler queryProfiler = QueryProfiler.NO_OP;
+    private final int txVertexCacheSize;
+
+    public JanusGraphEdgeVertexStep(EdgeVertexStep originalStep, int txVertexCacheSize) {
+        super(originalStep.getTraversal(), originalStep.getDirection());
+        originalStep.getLabels().forEach(this::addLabel);
+        this.txVertexCacheSize = txVertexCacheSize;
+    }
+
+    private void initialize() {
+        assert !initialized;
+        initialized = true;
+
+        if (!starts.hasNext()) {
+            throw FastNoSuchElementException.instance();
+        }
+        List<Traverser.Admin<Edge>> edges = new ArrayList<>();
+        Set<Vertex> vertices = Sets.newHashSet();
+        starts.forEachRemaining(e -> {
+            edges.add(e);
+
+            if (vertices.size() < txVertexCacheSize) {
+                if (Direction.IN.equals(direction) || Direction.BOTH.equals(direction)) {
+                    vertices.add(e.get().inVertex());
+                }
+                if (Direction.OUT.equals(direction) || Direction.BOTH.equals(direction)) {
+                    vertices.add(e.get().outVertex());
+                }
+            }
+        });
+
+        // If there are multiple vertices then fetch the properties for all of them in a single multiQuery to
+        // populate the vertex cache so subsequent queries of properties don't have to go to the storage back end
+        if (vertices.size() > 1) {
+            JanusGraphMultiVertexQuery multiQuery = JanusGraphTraversalUtil.getTx(traversal).multiQuery();
+            ((BasicVertexCentricQueryBuilder) multiQuery).profiler(queryProfiler);
+            multiQuery.addAllVertices(vertices).preFetch();
+        }
+
+        starts.add(edges.iterator());
+    }
+
+    @Override
+    protected Traverser.Admin<Vertex> processNextStart() {
+        if (!initialized) initialize();
+        return super.processNextStart();
+    }
+
+    @Override
+    public void reset() {
+        super.reset();
+        this.initialized = false;
+    }
+
+    @Override
+    public JanusGraphEdgeVertexStep clone() {
+        final JanusGraphEdgeVertexStep clone = (JanusGraphEdgeVertexStep) super.clone();
+        clone.initialized = false;
+        return clone;
+    }
+
+    @Override
+    public void setMetrics(MutableMetrics metrics) {
+        queryProfiler = new TP3ProfileWrapper(metrics);
+    }
+}
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphLocalQueryOptimizerStrategy.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphLocalQueryOptimizerStrategy.java
@@ -14,18 +14,18 @@
 
 package org.janusgraph.graphdb.tinkerpop.optimize;
 
-import org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent;
 import org.apache.tinkerpop.gremlin.process.traversal.step.branch.BranchStep;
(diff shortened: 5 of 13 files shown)


FlorianHockmann:
Thanks for the contribution. The avoided networks trips definitely sound nice.
I will try to review this in the next few days.

Travis failed here because of a known flaky test (#1459). I just restarted the job.
Regarding the comments from Codacy: We just activated that app and it definitely generates some false positives. I suggest that you just check its comments to see which are valid complaints so you can fix them and ignore the comments that don't really make sense.

smcquillan:
The comments from Codacy were all pretty minor and many were about existing issues. I've just amended the commit to move the constructor to below the field declarations and taken out the unused imports which looks like it'll be enough to keep Codacy happy.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -4023,6 +4021,8 @@ public void testTinkerPopOptimizationStrategies() {
         assertNumStep(1, 1, gts.V().has("id", numV / 5).has("weight", (numV / 5) % 5).as("x"), JanusGraphStep.class);
 
 
+        assertNumStep(superV * (numV / 5 * 2), 2, gts.V().has("id", sid).outE("knows").has("weight", P.between(1, 3)).inV().has("weight", P.between(1, 3)), JanusGraphStep.class, JanusGraphVertexStep.class);
```
FlorianHockmann: How can these tests verify your changes? I just removed all code changes and only kept these tests here to verify that they fail without your changes but the test method is still passing.
Unfortunately, I don't know how these changes could be tested else. Would it maybe be possible to add tests where a traversal pre-optimization is given and the optimized version is asserted [like it's done in `JanusGraphStepStrategyTest`](https://github.com/JanusGraph/janusgraph/blob/2fb971dd69abcd49e44dea33ccbe0d54a5df5357/janusgraph-test/src/test/java/org/janusgraph/blueprints/process/traversal/strategy/optimization/JanusGraphStepStrategyTest.java#L132)?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/HasStepFolder.java
```diff
@@ -329,4 +331,26 @@ static void foldInRange(final HasStepFolder janusgraphStep, final Step<?, ?>  ti
             }
         }
     }
+
+	static boolean foldableHasContainer(FlatMapStep<?, ?> janusgraphStep, Admin<?, ?> traversal, Admin<?, ?> traversal2) {
```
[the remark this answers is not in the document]
smcquillan: Well spotted, I'd used the logic in the `foldInHasContainer` method as a starting point and it had those extra two parameters but they're are not needed in `foldableHasContainer` so I'll remove them.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphLocalQueryOptimizerStrategy.java
```diff
@@ -169,6 +173,43 @@ private static void unfoldLocalTraversal(final Traversal.Admin<?, ?> traversal,
...
+        if (hasPropertyPrefetching) {
+	        final int txVertexCacheSize = janusGraph.getConfiguration().getTxVertexCacheSize();
+	        
+		    if (Vertex.class.isAssignableFrom(vertexStep.getReturnClass())) {
+		        if (HasStepFolder.foldableHasContainer(vertexStep, traversal, traversal)) {
+		            vertexStep.setPreFetchProperties(hasPropertyPrefetching);
+		            vertexStep.setTxVertexCacheSize(txVertexCacheSize);
+		        }
+		    }
+		    else if (nextStep instanceof EdgeVertexStep) {
```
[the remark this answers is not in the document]
smcquillan: There are two different flavours of the optimisation here, hence the condition:

- The first branch handles the `outE().inV().has()` flavour and from within the existing JanusGraphVertexStep class which will, in this case have a return type `Vertex` for `JanusGraphVertexStep.getReturnClass()`

- The second branch handles that case where there is another step like your `outE().has('weight', 0.5).inV().has()` example. And in this case the return type is `Edge` for `JanusGraphVertexStep.getReturnClass()`. It's for this flavour that I added the new `JanusGraphEdgeVertexStep` class which is doing the `.inV().has()` part of the query and loading the properties for those vertices into the cache.

I've not tested what happens if other step is an `as` step like `outE().as('edge').inV()` so will investigate that one further.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphEdgeVertexStep.java
```diff
@@ -0,0 +1,106 @@
...
+        if (useMultiQuery) {
+            if (!starts.hasNext()) {
+                throw FastNoSuchElementException.instance();
+            }
+            List<Traverser.Admin<Edge>> edges = new ArrayList<>();
+            Set<Vertex> vertices = Sets.newHashSet();
+            starts.forEachRemaining(e -> {
+                edges.add(e);
+
+                if (vertices.size() < txVertexCacheSize) {
```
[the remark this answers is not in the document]
smcquillan: That's right, and following on from your earlier comment about What happens to a traversal that has a limit() it might be possible to limit it further based on that. I'll investigate.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphEdgeVertexStep.java
```diff
@@ -0,0 +1,106 @@
...
+                        vertices.add(e.get().outVertex());
+                    }
+                }
+            });
+
+            // If there are multiple vertices then fetch the properties for all of them in a single multiQuery to
+            // populate the vertex cache so subsequent queries of properties don't have to go to the storage back end
+            if (vertices.size() > 1) {
+                JanusGraphMultiVertexQuery multiQuery = JanusGraphTraversalUtil.getTx(traversal).multiQuery();
+                multiQuery.addAllVertices(vertices).properties();
```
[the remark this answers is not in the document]
smcquillan: This is triggered for queries like `outE().has('weight', 0.5).inV().has()` where return type is Edge for JanusGraphVertexStep.getReturnClass(). In the situation I think there are two calls to `multiQuery` being made. 

- The first in `JanusGraphVertexStep` to get the edge properties and filter on those.
- The second one `JanusGraphEdgeVertexStep` to get the vertex properties after following the edges from the first.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphVertexStep.java
```diff
@@ -109,7 +120,26 @@ private void initialize() {
     protected Iterator<E> flatMap(final Traverser.Admin<Vertex> traverser) {
         if (useMultiQuery) {
             assert multiQueryResults != null;
-            return (Iterator<E>) multiQueryResults.get(traverser.get()).iterator();
+            
+            Iterable<? extends JanusGraphElement> result = multiQueryResults.get(traverser.get());
```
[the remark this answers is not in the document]
smcquillan: No. This is triggered for the `outE().inV().has()` flavour and from within the existing JanusGraphVertexStep class which will, in this case have a return type Vertex for JanusGraphVertexStep.getReturnClass(). So its skipping right over the edges and returning the vertices at the far end of those edges, it's those vertices we're doing the multiQuery on.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
smcquillan: I too was unsure how best to test this given that if it's working then it should return exactly the same results but faster. So we'd not expect the tests to fail without the changes, just to be slower when there are a large number of edges.

I verified that the two tests call the new code both by printing out temporary debug to stdout and by placing breakpoints in eclipse and stopping the debugger in the new methods.

I could/should look to add a check for the flavour of optimisation which uses the new  `JanusGraphEdgeVertexStep` class I think the profiler would pick that up. But the profile for the other flavour won't look any different.

smcquillan:
Like you I haven't worked on these parts of the code base before though I have been digging through them with the help the debugger in eclipse for some weeks now. I agree that it would be good if someone familiar with this area of the code could also review this.

Wanted to respond to the comments questions straight away. But I've taken a few actions items such as following up on the use of `limit()` and `as()`. And also looking to see if, at the very least, I can add some tests that use the profiler to verify that the optimisation has been triggered.

smcquillan:
@FlorianHockmann I've looked at what happens if we have an intermediate `as` in the query like `outE().as('edge').inV()` and the good news there is that the optimisation does still kick in.

With respect so something like the use of the use of limit in a traversal like the following `g.V().has('name', 'hercules').out().has('age', 5).limit(1)` we've a chicken and egg situation as the limit is applied **after** the has step checks if the age is 5 and the optimisation needs to kick in **before** the has step so not sure there's much we can do there.

Better news on the use of profiling in the tests though as I worked out how to explicitly profile the multiQuery which is fetching the properties and was able to tag it with a new annotation in the query profile, which I've named `_multiPreFetch`. I was able to alter the test to ensure that this annotation was set but better still the profiling data is now much more informative. 

For example this is a query profile from version the  current JanusGraph version `0.3.1`:
```
gremlin> g.V(1050048).outE('aggregation').inV().has('uniqueId', 'interface831').profile()
==>Traversal Metrics
GraphStep(vertex,[1050048])                                            1           1          37.725     1.14
JanusGraphVertexStep(OUT,[aggregation],vertex)                      1001        1001         158.576     4.78
    \_condition=type[aggregation]
    \_isFitted=true
    \_vertices=1
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
    \_orders=[]
    \_isOrdered=true
    \_multi=true
  optimization                                                                                84.298
  backend-query                                                     1001                      20.898
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
HasStep([uniqueId.eq(interface831)])                                   1           1        3123.755    94.09
                                            >TOTAL                     -           -        3320.057        -
```
And here's the same query run with the property pre-fetch optimisation 
```
gremlin> g.V(1050048).outE('aggregation').inV().has('uniqueId', 'interface831').profile()
==>Traversal Metrics
GraphStep(vertex,[1050048])                                            1           1          37.031     6.60
JanusGraphVertexStep(OUT,[aggregation],vertex)                      1001        1001         404.666    72.17
    \_condition=(PROPERTY AND visibility:normal)
    \_isFitted=true
    \_vertices=1001
    \_multiPreFetch=true
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8019d62e
    \_orders=[]
    \_isOrdered=true
    \_multi=true
  optimization                                                                                80.020
  backend-query                                                     1001                      15.789
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
  optimization                                                                                 0.676
  backend-query                                                    13013                     245.383
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8019d62e
HasStep([uniqueId.eq(interface831)])                                   1           1         119.047    21.23
                                            >TOTAL                     -           -         560.744        -
```
Points to note:
- It's much faster at just over half a second vs. over 3 seconds before. 
- It's tagged with `\_multiPreFetch=true` so we know the optimisation kicked in
- Explicitly shows backend query to get the 13013 properties (1001 vertices with 13 properties each)
- Run a second time and the profile shows it **doesn't** re-fetch, the now cached, 13013 properties

And for reference here's the profiles for the second flavour of query. First for version 0.3.1 
```
gremlin> g.V(1050048).outE('aggregation').has('edgeType', 'contains').inV().has('uniqueId', 'interface831').profile()
==>Traversal Metrics
GraphStep(vertex,[1050048])                                            1           1          37.682     1.17
JanusGraphVertexStep([edgeType.eq(contains)])                       1001        1001         219.878     6.81
    \_condition=(edgeType = contains AND type[aggregation])
    \_isFitted=false
    \_vertices=1
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
    \_orders=[]
    \_isOrdered=true
    \_multi=true
  optimization                                                                                88.495
  backend-query                                                     1001                      15.346
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
EdgeVertexStep(IN)                                                  1001        1001           7.951     0.25
HasStep([uniqueId.eq(interface831)])                                   1           1        2964.526    91.78
                                            >TOTAL                     -           -        3230.039        -
```
And now with the with the property pre-fetch optimisation:
```
gremlin> g.V(1050048).outE('aggregation').has('edgeType', 'contains').inV().has('uniqueId', 'interface831').profile()
==>Traversal Metrics
GraphStep(vertex,[1050048])                                            1           1          37.789     6.14
JanusGraphVertexStep([edgeType.eq(contains)])                       1001        1001         214.317    34.80
    \_condition=(edgeType = contains AND type[aggregation])
    \_isFitted=false
    \_vertices=1
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
    \_orders=[]
    \_isOrdered=true
    \_multi=true
  optimization                                                                                84.072
  backend-query                                                     1001                      14.929
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81bcaf2b
JanusGraphEdgeVertexStep(IN)                                        1001        1001         247.477    40.19
    \_condition=(PROPERTY AND visibility:normal)
    \_isFitted=true
    \_vertices=1001
    \_multiPreFetch=true
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8019d62e
    \_orders=[]
    \_isOrdered=true
    \_multi=true
  optimization                                                                                 0.907
  backend-query                                                    13013                     216.843
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8019d62e
HasStep([uniqueId.eq(interface831)])                                   1           1         116.257    18.88
                                            >TOTAL                     -           -         615.842        -
```

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
smcquillan: There's now a check to ensure that the profiling shows that `multiPreFetch=true` meaning the optimisation did kick in.

smcquillan:
Currently this optimisation would be automatically triggered if both of the following are true:
```
query.batch=true
query.fast-property=true
```
Was wondering if we ought to give more control by adding a new configuration option along the lines of:
```
query.batch-property-prefetch=true
```
or something along those lines.

FlorianHockmann:
Thanks for the explanations and for trying out the different traversals with your optimization.

> use of limit in a traversal like the following g.V().has('name', 'hercules').out().has('age', 5).limit(1) we've a chicken and egg situation as the limit is applied after the has step checks if the age is 5 and the optimisation needs to kick in before the has step so not sure there's much we can do there.

Wouldn't it be possible to check whether `has` is followed by `limit` and in that case not apply the optimization?
I really like the optimization you developed, but it leads to a performance hit for a traversal that uses `limit` in such a way which means that it should probably only be activated if users know that they won't use such traversals. If we can restrict the optimization to not kick in for such a traversal, then the optimization would really have no negative implications (at least I don't see any).

> Was wondering if we ought to give more control by adding a new configuration option

Good question. I'd say that it makes sense to add a new configuration option in either of these conditions:

1. Users won't expect such an optimization to kick in based on the docs of the existing options.
2. There are valid reasons for users to not want this new optimization if they have the other two options activated OR only to use this optimization, but not necessarily the other two.

If performance for traversals that use `limit()` like discussed above suffer from this optimization, then condition 2 warrants a new config option in my opinion.

smcquillan:
@FlorianHockmann I've modified things such that if there is a limit step following the has steps the optimisation will not kick in. And I've also added a new `query.batch-property-prefetch=true` configuration option allowing the new optimisation to be turned on or off an defaulted this to `false` so that existing users won't get any surprises.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -4023,12 +4021,36 @@ public void testTinkerPopOptimizationStrategies() {
...
         assertNumStep(superV * (numV / 5 * 2), 2, gts.V().has("id", sid).outE("knows").has("weight", P.between(1, 3)), JanusGraphStep.class, JanusGraphVertexStep.class);
         assertNumStep(superV * 10, 2, gts.V().has("id", sid).local(__.outE("knows").has("weight", P.gte(1)).has("weight", P.lt(3)).limit(10)), JanusGraphStep.class, JanusGraphVertexStep.class);
         assertNumStep(superV * 10, 1, gts.V().has("id", sid).local(__.outE("knows").has("weight", P.between(1, 3)).order().by("weight", decr).limit(10)), JanusGraphStep.class);
         assertNumStep(superV * 10, 0, gts.V().has("id", sid).local(__.outE("knows").has("weight", P.between(1, 3)).order().by("weight", decr).limit(10)), LocalStep.class);
+
+        clopen(option(BATCH_PROPERTY_PREFETCHING), true);
+        gts = graph.traversal();
+
+        // This tests an edge property before inV and will trigger the multiQuery property pre-fetch optimisation in JanusGraphEdgeVertexStep
+        t = gts.V().has("id", sid).outE("knows").has("weight", P.between(1, 3)).inV().has("weight", P.between(1, 3)).profile("~metrics");
```
FlorianHockmann: Can't you omit the `has` after the `inV` here as it's the `has` before the `inV` you want to test here?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -4023,12 +4021,36 @@ public void testTinkerPopOptimizationStrategies() {
...
+        assertNumStep(superV * (numV / 5 * 2), 2, (GraphTraversal)t, JanusGraphStep.class, JanusGraphVertexStep.class);
+        metrics = t.asAdmin().getSideEffects().get("~metrics");
```
FlorianHockmann: (nitpick) How about moving this and the next line into a method like:
```java
assertMultiPrefetchUsed(Traversal t)
```

The test below that checks that it's not using multi prefetching could then use a method like `assertMultiPrefetchNotUsed`.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -4023,12 +4021,36 @@ public void testTinkerPopOptimizationStrategies() {
         assertNumStep(1, 1, gts.V().has("id", numV / 5).has("weight", (numV / 5) % 5).as("x"), JanusGraphStep.class);
 
 
+        assertNumStep(superV * (numV / 5 * 2), 2, gts.V().has("id", sid).outE("knows").has("weight", P.between(1, 3)).inV().has("weight", P.between(1, 3)), JanusGraphStep.class, JanusGraphVertexStep.class);
+        assertNumStep(superV * (numV / 5 * 2), 2, gts.V().has("id", sid).outE("knows").inV().has("weight", P.between(1, 3)), JanusGraphStep.class, JanusGraphVertexStep.class);
```
FlorianHockmann: I think you can remove this and the assert above now as you have included these now already below where it's also easier to understand them as you've added comments there.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -4023,12 +4021,36 @@ public void testTinkerPopOptimizationStrategies() {
...
         assertNumStep(superV * 10, 0, gts.V().has("id", sid).local(__.outE("knows").has("weight", P.between(1, 3)).order().by("weight", decr).limit(10)), LocalStep.class);
+
+        clopen(option(BATCH_PROPERTY_PREFETCHING), true);
+        gts = graph.traversal();
+
+        // This tests an edge property before inV and will trigger the multiQuery property pre-fetch optimisation in JanusGraphEdgeVertexStep
+        t = gts.V().has("id", sid).outE("knows").has("weight", P.between(1, 3)).inV().has("weight", P.between(1, 3)).profile("~metrics");
+        assertNumStep(superV * (numV / 5 * 2), 2, (GraphTraversal)t, JanusGraphStep.class, JanusGraphVertexStep.class);
+        metrics = t.asAdmin().getSideEffects().get("~metrics");
+        assertTrue(metrics.toString().contains(QueryProfiler.MULTIPREFETCH_ANNOTATION + "=true"));
```
FlorianHockmann: Shouldn't we also test that this optimization won't be used if the `BATCH_PROPERTY_PREFETCHING` is not set?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
smcquillan: Actually we can't the `has` after the `inV` is the one that the optimisation is fetching the properties for. The first `has` is also needed as it's what sends us down the path which needs the `JanusGraphEdgeVertexStep`

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphLocalQueryOptimizerStrategy.java
```diff
@@ -92,8 +88,22 @@ public void apply(final Traversal.Admin<?, ?> traversal) {
 
             if (useMultiQuery && !(isChildOf(vertexStep, MULTIQUERY_INCOMPATIBLE_STEPS))) {
                 vertexStep.setUseMultiQuery(true);
+            }
 
-                preFetchPropertiesWithMultiQuery(janusGraph, vertexStep, traversal, nextStep);
+            if (janusGraph.getConfiguration().batchPropertyPrefetching()) {
```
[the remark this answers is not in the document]
smcquillan: good idea have moved it

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
smcquillan: added an extra test for that

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
smcquillan: removed them

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
smcquillan: Liked the idea of moving those lines into a method, but went for something a bit more generic so added `queryProfilerAnnotationIsPresent()` so I could also use it for testing the multiQuery annotation was being set as previously that wasn't being tested.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
```diff
@@ -140,6 +140,12 @@ public JanusGraphMultiVertexQuery addAllVertices(Collection<? extends Vertex> ve
                 execute(RelationCategory.PROPERTY, new RelationConstructor()));
     }
 
+    @Override
+    public void preFetch() {
+        profiler.setAnnotation(QueryProfiler.MULTIPREFETCH_ANNOTATION,true);
```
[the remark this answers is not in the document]
smcquillan: done

smcquillan:
@FlorianHockmann suggested changes all looked good so have implemented them. Have also squashed everything into a single commit and rebased against master because, as you say, this is new functionality.

Incidentally do you know roughly when the next release might be?  I ask because for us to really benefit from multiQuery I think we'll need to address issue #1044 as well and if I've time I may look into what can be done there.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
FlorianHockmann: That's even better 👍

FlorianHockmann:
> Incidentally do you know roughly when the next release might be? 

The [v0.4.0 milestone](https://github.com/JanusGraph/janusgraph/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Release+v0.4.0%22) summarizes some issues that we want to fix for the 0.4.0 release. Most of them are just about dropping support for outdated backend versions which shouldn't require that much work. A bigger issue is the TinkerPop upgrade to 3.4.0 as that resulted in a lot of failed `GraphComputer` tests which probably requires some work on the `FulgoraGraphComputer`, an area I haven't worked on yet.
Then we also have a big PR currently open that changes the whole documentation generation process and which therefore requires a thorough review: #1437.
Therefore I'd say that some time in April sounds realistic for the 0.4.0 release.
So, you probably have enough time to tackle that issue.

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -1214,6 +1220,7 @@ public boolean apply(@Nullable String s) {
     private Boolean propertyPrefetching;
     private boolean adjustQueryLimit;
     private Boolean useMultiQuery;
+    private Boolean batchPropertyPrefetching;
```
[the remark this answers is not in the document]
smcquillan: Thanks for taking the time to review. And while I expect `query.batch-property-prefetch` probably could use `boolean` I followed the existing convention used by the `query.batch` (_useMultiQuery_) and `query.fast-property` (_propertyPrefetching_) both of which use the `Boolean`.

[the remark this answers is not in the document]
porunov: @smcquillan I am OK here with both `Boolean` and `boolean` types.

FlorianHockmann:
@twilmes you seem to be the only one who has already worked on this part of the code base, so if you're available to review this in the next few weeks, then we can leave the PR open.
Otherwise, I'd say that we merge this with @porunov's and my approval.

PR #1517, opened 2019-03-29
Enable multiQuery optimization from within parent steps #1044

smcquillan:
This pull request seeks to address issue #1044 by enabling the multiQuery optimization for JanusGraphVertexStep which are children of the RepeatStep, MatchStep, BranchStep, and OptionalStep and more.

As I mentioned in the issue, the parent step drip feeds the starts one at a time into its child traversals. Re-initialising for each new start is one option, but that means a trip to the storage backend for each new start and defeats the point of the multiQuery optimisation. The simplest way to get around this would be for the JanusGraphVertexStep to request all the starts from its parent and use those when doing its initialisation, but there's no public method which allows us do do this.

To get around the fact that tinkerpop doesn't expose the starts I've created a new step named `JanusGraphMultiQueryStep` which, like the profiling steps, is just a simple passthrough. It caches the starts sent to the parent step and later allows a child `JanusGraphVertexStep` to request those starts and fully initialise itself with a single multiQuery to the storage backend. 

I've been running custom builds with these changes against our application which makes heavy use of repeat, union, choose, filer and optional steps and am seeing significant performance improvements.

-----

### For all changes:

### For code changes:

### Note:


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphMultiQueryStep.java
```diff
@@ -0,0 +1,75 @@
...
+        if (!cachedStarts.contains(start))
+        {
+            if (cachedStartsAccessed == true) {
+                cachedStarts.clear();
+                cachedStartsAccessed = false;
+            }
+            final List<Traverser.Admin<Vertex>> newStarters = new ArrayList<>();
+            starts.forEachRemaining(s -> { newStarters.add(s); });
+            starts.add(newStarters.iterator());
+            cachedStarts.addAll(newStarters);
```
porunov: I just think that it is an unnecessary loop. It isn't critical but I think we can omit it like here:
```
final List<Traverser.Admin<Vertex>> newStarters = new ArrayList<>();
while (starts.hasNext()){
    Admin<Vertex> vertexAdmin = starts.next();
    newStarters.add(vertexAdmin);
    cachedStarts.add(vertexAdmin);
}
starts.add(newStarters.iterator());
cachedStarts.add(start);
```
Also, could you explain the line `starts.add(newStarters.iterator());`? I am not sure why do we need it. I initially sought that you want to expend the `starts` iterator with same elements. Then I checked that the iterator uses `TraverserSet` for its elements. So, when you actually call `starts.add(newStarters.iterator());` it actually executes the code like `existing.merge(traverser);` because all elements from `newStarters` already exists in `starts`. So we actually can replace the code to be like:
```
while (starts.hasNext()){
    Admin<Vertex> vertexAdmin = starts.next();
    cachedStarts.add(vertexAdmin);
    vertexAdmin.merge(vertexAdmin);
}
cachedStarts.add(start);
```
Am I missing something?

porunov: Also, I didn't check but I assume that `merge` method just rewrites all changes into the existing object. So, in current situation it does nothing (need to check). If so, the code can be simplified like:
```
while (starts.hasNext()){
    cachedStarts.add(starts.next());
}
cachedStarts.add(start);
```

porunov: Sorry, I missed that `starts` iterator won't be reusable after that

porunov: I think then the next code can be used:
```
final List<Traverser.Admin<Vertex>> newStarters = new ArrayList<>();
starts.forEachRemaining(s -> {
    newStarters.add(s);
    cachedStarts.add(s);
});
starts.add(newStarters.iterator());
cachedStarts.add(start);
```

smcquillan: I followed the patter used in [JanusGraphVertexStep. initialize()](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphVertexStep.java#L94) and it confused me at first too but as I think you've just spotted the [starts.next()](https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/ExpandableStepIterator.java#L53) is actually removing the start. So after we've looked at all the starts we need to call `starts.add(newStarters.iterator());` to put them all back again. 

Have just pushed up a change for the other comments, let me know if you think we need changes here too.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphMultiQueryStep.java
```diff
@@ -0,0 +1,75 @@
...
+    public JanusGraphMultiQueryStep(Step<Vertex,?> originalStep) {
+        super(originalStep.getTraversal());
+    }
+
+    @Override
+    protected Admin<Vertex> processNextStart() throws NoSuchElementException {
+        Admin<Vertex> start = this.starts.next();
+        if (!cachedStarts.contains(start))
+        {
+            if (cachedStartsAccessed == true) {
```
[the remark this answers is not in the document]
smcquillan: fixed and codacy-bot bot picked me up on this one too.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphMultiQueryStep.java
smcquillan: Think our comments crossed each other - your last one looks good will change it to do that.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphVertexStep.java
```diff
@@ -85,20 +89,68 @@ private void initialize() {
...
+        vertices.forEach(v -> multiQuery.addVertex(v.get()));
+        parentStarts.forEach(v -> multiQuery.addVertex(v.get()));
+        makeQuery(multiQuery);
+
+        Map<JanusGraphVertex, Iterable<? extends JanusGraphElement>> results = (Vertex.class.isAssignableFrom(getReturnClass())) ? multiQuery.vertices() : multiQuery.edges();
+        if (multiQueryResults == null) {
+            multiQueryResults = results;
+        } else {
+            multiQueryResults.putAll(results);
+            results.clear();
```
porunov: Same here. Just would like to know why do we call `clear` here.

smcquillan: Good Question, and given that the results goes out of scope shortly afterwards we don't need it, I've removed it.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphPropertiesStep.java
```diff
@@ -97,12 +95,32 @@ private void initialize() {
...
+
+        Map<JanusGraphVertex, Iterable<? extends JanusGraphProperty>> results = multiQuery.properties();
+        if (multiQueryResults == null)
+        {
+            multiQueryResults = results;
+        }
+        else
+        {
+            multiQueryResults.putAll(results);
+            results.clear();
```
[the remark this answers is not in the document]
smcquillan: Good Question, and given that the results goes out of scope shortly afterwards we don't need it, I've removed it.

FlorianHockmann:
@smcquillan I didn't find the time to review this yet, but I think that I found a way how we can test these kind of optimizations easier. Since these optimizations are applied as a traversal strategy that changes the traversal, it should be possible to compare an unoptimized traversal after the strategies were applied with an optimized traversal as we expect it. So, something like this:

```java
GraphTraversal t = g.V().repeat(out().hasId(3)).times(3);

t.asAdmin().applyStrategies();

GraphTraversal optimizedTraversal = // construct the traversal as you expect it after optimization
assertEquals(optimizedTraversal , t);
```

It may be necessary to construct the optimized traversal with something like `t.asAdmin().addStep()` as we probably can't add JanusGraph specific steps otherwise.

That's also how strategies are [tested in TinkerPop](https://github.com/apache/tinkerpop/blob/667134d608ca29828457cd3eebe365ee1fd330ce/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/EarlyLimitStrategyTest.java#L69). After looking for where we could add such a test in JanusGraph I realized that we already have that kind of tests in [JanusGraphStepStrategyTest](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-test/src/test/java/org/janusgraph/blueprints/process/traversal/strategy/optimization/JanusGraphStepStrategyTest.java).

Maybe this can help you here (and maybe in the other PR) to better test this, especially to verify in which scenarios the optimization kicks in.

smcquillan:
@FlorianHockmann thanks for the tip, I think that'll be a great help in verifying that the new step is appearing as expected in the correct parts of the traversal. I'll look to add some additional tests using this approach. I've also had some success in extending this approach to the **repeat** and **match** steps so will include tests for them too.

janusgraph-test/src/test/java/org/janusgraph/blueprints/process/traversal/strategy/optimization/JanusGraphStepStrategyTest.java
```diff
@@ -191,4 +261,52 @@ public void doTest(Traversal original, Traversal optimized, Collection<Traversal
                 g_V(__.or(g_V("name", eq("marko")), g_V("lang", eq("java"), new HasStepFolder.OrderEntry("name", Order.decr))), new HasStepFolder.OrderEntry("lang", Order.incr)), Collections.emptyList())
         });
     }
+    
+    private static Stream<Arguments> generateMultiQueryTestParameters() {
```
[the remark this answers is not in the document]
smcquillan: This method is used but via an annotation but it seems the codacy-bot wasn't able to pick that it.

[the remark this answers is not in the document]
smcquillan: This method is used via an annotation but it appears that the codacy-bot can't detect that.

smcquillan:
@FlorianHockmann I’ve pushed up some additional changes in a new commit so people can see the differences to original pull request as reviewed by @porunov. In summary the additional changes are:

- JanusGraphMultiQueryStep now stores the class name of the parent traversal step it supports, which makes the profiling data clearer
- JanusGraphVertexStep now caches the JanusGraphMultiQueryStep (if found) so it doesn’t need to look it up again each time
- Side effect type steps are no longer counted when checking to see if the first step of a child traversal is a JanusGraphVertexStep
- RepeatStep is now supported, but unlike the other steps the JanusGraphMultiQueryStep goes at the end so it can feed its starts into the next iteration

In addition, as suggested, I’ve added a set of tests for this new optimisation strategy into JanusGraphStepStrategyTest. This was a worthwhile exercise as it led to my realisation that a simple `.as("name")` side effect at the start of a child traversal would have prevented the optimisation kicking in so fixed that and added a test for it. 

With regards to the support for MatchStep I’ve given up on that one. The reason being that while we could feed the starts into all its child traversals it’s not appropriate for all of them given that some children are actually fed their starts in from other children of the MatchStep. On the bright side however, this pull request does now allow the use of multiQuery within the child traversals of the match step, so it’s still an improvement over what we had before. 

I should also add that I’ve not been testing these changes in isolation. I’ve also been creating custom builds of the application I work on which include these changes and have seen the query times for some traversals drop by half.

smcquillan:
Amended the commit to make some minor changes to comments and white space and remove an unnecessary condition.

smcquillan:
@FlorianHockmann FYI I made a some minor changes today to fix some white space and comments, plus remove one redundant check. But the build seems to have failed, I assume for spurious reasons as I don't see how the changes I made could have caused that.

porunov:
@smcquillan It is my mistake caused by #1506 . I will merge #1523 now with CTR to fix the problem. You will have to just rebase against the latest commit

porunov:
@smcquillan I've merged the fix. Now you can rebase against the latest commit in master and the build should pass.

janusgraph-test/src/test/java/org/janusgraph/blueprints/process/traversal/strategy/optimization/JanusGraphStepStrategyTest.java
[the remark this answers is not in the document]
smcquillan: Method is used, but via an annotation and codacy-bot can't detect that.

smcquillan:
As I was expecting, got some conflicts now that https://github.com/JanusGraph/janusgraph/pull/1460 has been merged, will resolve them an push up a fix shortly.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphTraversalUtil.java
```diff
@@ -85,4 +108,45 @@ public static JanusGraphTransaction getTx(Traversal.Admin<?, ?> traversal) {
...
+                        // Want the RepeatEndStep so the start of one iteration can feed into the next 
+                        steps.remove(step);
+                        steps.add(repeatEndSteps.get(0));
+                    }
+                }
+            }
+        }
+        return Lists.newArrayList(steps);
+    }
+    private static void getMultiQueryCompatibleStepsFromChildTraversal(Traversal.Admin<?,?> traversal, Step step, Set<Step> steps) {
```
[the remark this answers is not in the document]
smcquillan: Fair nitpick, have renamed these parameters to make things clearer

smcquillan:
Have rebased against develop and squashed things into a single commit

PR #1518, opened 2019-03-30
Value deduplication for fields with SET cardinality for Elasticsearch

fufler:
Current implementation makes no difference between `SET` and `LIST` cardinality when using ES as indexing backend: no deduplication is done thus making ES contain non-unique elements even if SET cardinality specified.

This PR resolves the issue by adding deduplication step to addition script. I'm not sure if it's the right choice to use `distinct` from Stream API, so please suggest any replacement if such approach is not acceptable for some reason.

I've also provided two integration tests to demonstrate the issue with invalid handling of cardinality. Quite probably these tests should be moved from ElastisearchIndexTest to somewhere else (IndexProviderTest?) since they are backend independent. But as far as I know there is no complete support for these cardinalities across different indexing backends, that's why I've put them into ES-specific test suite.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -778,25 +778,35 @@ public class ElasticSearchIndex implements IndexProvider {
     private String getAdditionScript(KeyInformation.IndexRetriever information, String storeName,
                                      IndexMutation mutation) throws PermanentBackendException {
         final StringBuilder script = new StringBuilder();
+
         for (final IndexEntry e : mutation.getAdditions()) {
             final KeyInformation keyInformation = information.get(storeName).get(e.field);
-            switch (keyInformation.getCardinality()) {
-                case SET:
-                case LIST:
-                    script.append("if(ctx._source[\"").append(e.field).append("\"] == null) ctx._source[\"").append(e.field).append("\"] = [];");
-                    script.append("ctx._source[\"").append(e.field).append("\"].add(").append(convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation))).append(");");
-                    if (hasDualStringMapping(keyInformation)) {
-                        script.append("if(ctx._source[\"").append(getDualMappingName(e.field)).append("\"] == null) ctx._source[\"").append(getDualMappingName(e.field)).append("\"] = [];");
-                        script.append("ctx._source[\"").append(getDualMappingName(e.field)).append("\"].add(").append(convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation))).append(");");
-                    }
-                    break;
-                default:
-                    break;
+            final Cardinality cardinality = keyInformation.getCardinality();
 
+            if (cardinality != Cardinality.SET && cardinality != Cardinality.LIST) {
+                continue;
             }
 
+            String value = convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation));
+
+            appendAdditionScript(script, value, e.field, cardinality == Cardinality.SET);
+
+            if (hasDualStringMapping(keyInformation)) {
+                appendAdditionScript(script, value, getDualMappingName(e.field), cardinality == Cardinality.SET);
+            }
         }
+
         return script.toString();
+    }
+
+    private void appendAdditionScript(StringBuilder script, String value, String fieldName, Boolean distinct) {
+        script.append("if (ctx._source[\"").append(fieldName).append("\"] == null) ctx._source[\"").append(fieldName).append("\"] = [];");
+
+        if (distinct) {
+            script.append("if (ctx._source[\"").append(fieldName).append("\"].indexOf(").append(value).append(") == -1) ");
+        }
+
+        script.append("ctx._source[\"").append(fieldName).append("\"].add(").append(value).append(");");
     }
 
     private Map<String,Object> getAdditionDoc(KeyInformation.IndexRetriever information,
--- a/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
+++ b/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
@@ -16,6 +16,7 @@ package org.janusgraph.diskstorage.es;
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Multimap;
 
 import org.apache.commons.configuration.BaseConfiguration;
@@ -49,6 +50,8 @@ import org.json.simple.parser.ParseException;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
 
@@ -57,6 +60,7 @@ import java.net.InetAddress;
 import java.nio.charset.Charset;
 import java.util.Date;
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.*;
 
@@ -321,6 +325,46 @@ public class ElasticsearchIndexTest extends IndexProviderTest {
         assertEquals(parameterValue.toString(), returnedProperty);
 
         IOUtils.closeQuietly(response);
+    }
+
+    public static Stream<String> cardinalityTestCollectionNameParams() {
+        return ImmutableList.of(PHONE_SET, PHONE_LIST).stream();
+    }
+
+    @ParameterizedTest
+    @MethodSource("cardinalityTestCollectionNameParams")
+    public void testCollectionCardinality(String collectionName) throws Exception {
+        initialize("vertex");
+
+        Multimap<String, Object> initialDoc = HashMultimap.create();
+        initialDoc.put(collectionName, "12345");
+
+        add("vertex", "test", initialDoc, true);
+
+        clopen();
+
+        Multimap<String, Object> updateDoc = HashMultimap.create();
+        updateDoc.put(collectionName, "123456");
+
+        add("vertex", "test", updateDoc, false);
+
+        clopen();
+
+        add("vertex", "test", initialDoc, false);
+
+        clopen();
+
+        tx.delete("vertex", "test", collectionName, "12345", false);
+
+        clopen();
+
+        assertEquals("test", tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(collectionName, Cmp.EQUAL, "123456"))).toArray()[0]);
+
+        if(PHONE_SET.equals(collectionName)){
+            assertEquals(0, tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_SET, Cmp.EQUAL, "12345"))).count());
+        } else {
+            assertEquals("test", tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_LIST, Cmp.EQUAL, "12345"))).toArray()[0]);
+        }
     }
 
     private CloseableHttpResponse getESMapping(String indexName, String mappingTypeName) throws IOException {


porunov:
@fufler Thank you for the contribution!

There is a requirement that each contributor must sign a CLA. It isn't automate right now but JanusGraph members are working on it.

To contribute you should:
1. Sign a CLA. Instructions are here: https://github.com/JanusGraph/legal/blob/master/README.md
2. Follow this instructions: https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md

Will you be able to sing a CLA?
Please feel free to ask any questions if you have some.

fufler:
@porunov Done. Just sent it via email.

porunov:
@fufler Thank you! JanusGraph PMC members should check your CLA and then add you into the list.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -780,19 +780,25 @@ private String getAdditionScript(KeyInformation.IndexRetriever information, Stri
...
-                        script.append("if(ctx._source[\"").append(getDualMappingName(e.field)).append("\"] == null) ctx._source[\"").append(getDualMappingName(e.field)).append("\"] = [];");
-                        script.append("ctx._source[\"").append(getDualMappingName(e.field)).append("\"].add(").append(convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation))).append(");");
-                    }
-                    break;
-                default:
-                    break;
+            final Cardinality cardinality = keyInformation.getCardinality();
+
+            if (cardinality != Cardinality.SET && cardinality != Cardinality.LIST)
+                continue;
```
porunov: Please add braces `{}`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -780,19 +780,25 @@ private String getAdditionScript(KeyInformation.IndexRetriever information, Stri
...
+
+            script.append("if(ctx._source[\"").append(e.field).append("\"] == null) ctx._source[\"").append(e.field).append("\"] = [];");
+            script.append("ctx._source[\"").append(e.field).append("\"].add(").append(convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation))).append(");");
+
+            if (cardinality == Cardinality.SET)
```
porunov: Please add braces `{}`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -780,19 +780,25 @@ private String getAdditionScript(KeyInformation.IndexRetriever information, Stri
...
+                script.append("ctx._source[\"").append(e.field).append("\"] = ").append("ctx._source[\"").append(e.field).append("\"].stream().distinct().collect(Collectors.toList());");
+
+            if (hasDualStringMapping(keyInformation)) {
+                final String dualFieldName = getDualMappingName(e.field);
+
+                script.append("if(ctx._source[\"").append(dualFieldName).append("\"] == null) ctx._source[\"").append(dualFieldName).append("\"] = [];");
+                script.append("ctx._source[\"").append(dualFieldName).append("\"].add(").append(convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation))).append(");");
 
+                if (cardinality == Cardinality.SET)
```
porunov: Please add braces `{}`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -780,19 +780,25 @@ private String getAdditionScript(KeyInformation.IndexRetriever information, Stri
...
+                    script.append("ctx._source[\"").append(dualFieldName).append("\"] = ").append("ctx._source[\"").append(dualFieldName).append("\"].stream().distinct().collect(Collectors.toList());");
```
porunov: This like looks like a duplicate like of 792 line. Maybe it is better to create a separate method? Something like:
```
private void appendDistinct(StringBuilder script, String fieldName){
    script.append("ctx._source[\"").append(fieldName).append("\"] = ").append("ctx._source[\"").append(fieldName).append("\"].stream().distinct().collect(Collectors.toList());");
}
```

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
```diff
@@ -323,6 +323,66 @@ public void testCustomMappingProperty() throws BackendException, IOException, Pa
...
+        tx.delete("vertex", "test", PHONE_SET, "12345", false);
+
+        clopen();
+
+        assertEquals("test", tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_SET, Cmp.EQUAL, "123456"))).toArray()[0]);
+        assertEquals(0, tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_SET, Cmp.EQUAL, "12345"))).count());
+    }
+
+    @Test
+    public void testListCardinality() throws Exception {
```
porunov: This method looks identical to `testSetCardinality` except the last line and `PHONE_SET`. Can move duplicate logic into a separate method? Something like:
```
public void testSetCardinality() throws Exception {
    testCollectionCardinality(PHONE_SET);
    assertEquals(0, tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_SET, Cmp.EQUAL, "12345"))).count());
}

public void testListCardinality() throws Exception {
    testCollectionCardinality(PHONE_LIST);
    assertEquals("test", tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_LIST, Cmp.EQUAL, "12345"))).toArray()[0]);
}

private void testCollectionCardinality(String collectionName) throws Exception{
    ...
}
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
fufler: Done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
fufler: Done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
fufler: Done

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
fufler: Done. But I think that with previous version was it was a little bit simpler to understand the idea of test, because it was easier to keep then «context» when all assertions were in single function.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
fufler: Done. Initially I've thought about creating a new function but decided not to add it just for performance reasons. But actually I do not think it does really matter. 

If your point here is code readability then I'd like to suggest to rewrite it as follows
```java
        script.append(
            String.format(
                "ctx._source[\"%1$s\"] = ctx._source[\"%1$s\"].stream().distinct().collect(Collectors.toList());",
                fieldName
            )
        );
```

What do you think about it?

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
porunov: You can use parametrized test if you want to keep all association in the single function.
```
public static Stream<String> collectionNameParams() {
    return ImmutableList.of(PHONE_SET, PHONE_LIST).stream();
}

@ParametrizedTest
@MethodSource("collectionNameParams")
public void testCardinality(String collectionName){

   // you code test code...

  if(PHONE_SET.equals(collectionName)){
    assertEquals(0, tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_SET, Cmp.EQUAL, "12345"))).count());
  } else {
    assertEquals("test", tx.queryStream(new IndexQuery("vertex", PredicateCondition.of(PHONE_LIST, Cmp.EQUAL, "12345"))).toArray()[0]);
  }
}
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
porunov: > Initially I've thought about creating a new function but decided not to add it just for performance reasons

What do you mean by that? I thought Java compilers and JVM optimize method calls. Am I wrong?

> If your point here is code readability then I'd like to suggest to rewrite it as follows

I don't think we should sacrifice performance for code readability. I wouldn't use regular expression here because it is often too slow.

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
```diff
@@ -323,6 +323,50 @@ public void testCustomMappingProperty() throws BackendException, IOException, Pa
         IOUtils.closeQuietly(response);
     }
 
+    /**
+     * Helper function used by testSetCardinality and testListCardinality.
+     * Performs initialization and assertion checks that are common for both tests.
+     */
+    private void testCollectionCardinality(String collectionName) throws Exception {
```
porunov: Not critical but could you please move this method below `testListCardinality` method?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
fufler: > What do you mean by that? I thought Java compilers and JVM optimize method calls. Am I wrong?

You're right, JIT does method inlining. My bad.

> I don't think we should sacrifice performance for code readability. I wouldn't use regular expression here because it is often too slow.

I agree. But to tell the truth  I'm not sure if we really sacrifice here, since we're building a string with script that will be executed by ES. I assume executing script in ES is much slower then calling String.format. But ok, let's keep it as is.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -775,24 +775,38 @@ private String getDeletionScript(KeyInformation.IndexRetriever information, Stri
         return script.toString();
     }
 
+    private void appendDistinct(StringBuilder script, String fieldName) {
```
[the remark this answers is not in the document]
fufler: Done

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
fufler: Good idea. Rewritten tests using this approach .

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java
fufler: Removed this method in favour of parametrized test as suggested above.

fufler:
@porunov I've realized that calling `distinct` to remove duplicates is an overkill. It would be enough to lookup an element by call to `indexOf` just before insertion. If element is not found then insert it, otherwise — do nothing. Should I update my PR?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -778,27 +778,42 @@ private String getDeletionScript(KeyInformation.IndexRetriever information, Stri
...
         }
+
         return script.toString();
     }
 
+    private void appendAdditionScript(StringBuilder script, String value, String fieldName, Boolean distinct) {
+        script.append("if (ctx._source[\"").append(fieldName).append("\"] == null) ctx._source[\"").append(fieldName).append("\"] = [];");
+
+        if (distinct) {
+            script.append("def index = ctx._source[\"").append(fieldName).append("\"].indexOf(").append(value).append(");");
```
porunov: Maybe I am missing something but isn't `_source[fieldName]` returns a collection? If so, how can we use `indexOf` here? Maybe we should use `.contains(value)`?
P.S. Maybe I am missing something.

fufler: AFAIK it returns ArrayList. Maybe it's some kind of hack and by contract it returns collection, but it works. Take a look at `getDeletionScript` that uses `indexOf` call to remove elements. I've just reused the same approach.

porunov: My bad. I see ArrayList has `indexOf` method. Also I see that method `contains` actually uses `indexOf` in it's ArrayList implementation.

porunov:
Merging this PR by following a lazy consensus.

porunov:
Thank you @fufler for the contribution!

Issue #1306, opened 2018-10-25
Security vulnerabilities in lucene-solr/solr 7.0.0 up to 7.4.0


bleemr:
janusgraph-solr is pulling in lucene-solr 7.0.0, which lists the following vulnerabilities.

7.4.0 fixes SOLR-12450: Don't allow referal to external resources in various config files (CVE-2018-8026).
7.3.1 fixes SOLR-12316: Do not allow to use absolute URIs for including other files in solrconfig.xml and schema parsing (CVE-2018-8010).
7.3.1 fixes SOLR-12204: Upgrade commons-fileupload dependency to 1.3.3 to address CVE-2016-1000031.
7.3.0 fixes SOLR-11971: Don't allow referal to external resources in DataImportHandler's dataConfig request parameter (CVE-2018-1308).

I expect these vulnerabilities only affect solr core, and the server, but they way they're being reported makes it unclear, and janusgraph is pulling it all in anyway.

https://github.com/apache/lucene-solr/blob/master/solr/CHANGES.txt

pluradj:
The `janusgraph-solr` component only pulls in [`solr-solrj`](https://github.com/JanusGraph/janusgraph/blob/v0.3.1/janusgraph-solr/pom.xml#L69-L71) so as @bleemr noted, it seems that the impact of those vulnerabilities is on the Solr core/server. I reviewed the patches for the JIRA tickets associated with the CVEs listed above:

* CVE-2018-8026: [SOLR-12450](https://issues.apache.org/jira/browse/SOLR-12450): JanusGraph does not package `solr-core-7.0.0.jar`
* CVE-2018-8010: [SOLR-12316](https://issues.apache.org/jira/browse/SOLR-12316): JanusGraph does not package `solr-cell-7.0.0.jar` or `solr-core-7.0.0.jar`
* CVE-2016-1000031: [SOLR-12204](https://issues.apache.org/jira/browse/SOLR-12204): JanusGraph does not package `commons-fileupload-1.3.2.jar`
* CVE-2018-1308: [SOLR-11971](https://issues.apache.org/jira/browse/SOLR-11971): JanusGraph does not package `solr-dataimporthandler-7.0.0.jar`

It'd be worthwhile to test JanusGraph against Solr 7.5.0 to verify compatibility.

FlorianHockmann:
[...]

PR #1567, opened 2019-05-09
Drop support for Solr < 7.0.0

### LINKED ISSUE #1491: Remove support for Solr versions < 6.4
The lowest still supported Solr version is 6.4.x ([Source](http://lucene.apache.org/solr/community.html#about-versions)). [We decided](https://groups.google.com/forum/#!topic/janusgraph-dev/MA69slAbNvc) that we want to drop support for older versions that already reached EOL.
We should remove everything specific to older Solr versions, including source code, configuration, mentions in the docs and the Travis CI jobs.

FlorianHockmann:
Fixes #1491

-----

### For all changes:

### For code changes:
- [-] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?

### Note:


janusgraph-solr/src/test/resources/docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
       - "3888:3888"
 
   solr:
-    image: solr:5.5.4
+    image: solr:7.7
```
wojciechwojcik: @FlorianHockmann the Solr 7.x generates a lot of warnings when run with current 0.3.1 Janus example schema:
https://github.com/JanusGraph/janusgraph/blob/274eefcb0d10d7ac2a2d96f9c803f31819e51805/janusgraph-dist/src/assembly/static/conf/solr/schema.xml

```
Solr loaded a deprecated plugin/analysis class [solr.TrieIntField]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.TrieFloatField]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.TrieLongField]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.TrieDoubleField]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.TrieDateField]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.SynonymFilterFactory]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.WordDelimiterFilterFactory]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.LatLonType]. Please consult documentation how to replace it accordingly.
Solr loaded a deprecated plugin/analysis class [solr.CurrencyField]. Please consult documentation how to replace it accordingly.
Replace 'com.spatial4j.core' with 'org.locationtech.spatial4j' in your schema.
```

wojciechwojcik: The current 0.3.1 Janus example solrconfig.xml still uses Lucene match version 5.0.0
https://github.com/JanusGraph/janusgraph/blob/274eefcb0d10d7ac2a2d96f9c803f31819e51805/janusgraph-dist/src/assembly/static/conf/solr/solrconfig.xml

It should instead reflect the Solr release level e.g. 7.7.1

wojciechwojcik: There is Solr 7.7.1 release available already.

FlorianHockmann: Yes, but 7.7 points to the same image as 7.7.1 right now. If however 7.7.2 gets released, then 7.7 will point to that which means that it will be more up to date.

wojciechwojcik: Ok

FlorianHockmann: Thanks for noting, I will have to look into that.

Regarding the `luceneMatchVersion`, shouldn't that be already a problem with the current version in JanusGraph as [that is already 7.0.0](https://github.com/JanusGraph/janusgraph/blob/b6ed365ffe3a9c07110ae8829acb6f9e3aa2f66e/pom.xml#L92)?
But yes, it looks like that should also be changed to a higher version. Right now, that should probably also be 7.0.0 as that seems to be the version we are using by default, but since that is also already EOL, we should probably update in general to 7.7.1.
It would be good if someone with more knowledge about Solr/Lucene could say something on that matter, @sjudeng maybe?

wojciechwojcik: Solr will work with any lucene match version but it prodcues warnings during startup if there is even a slight mismatch between versions e.g. `7.7.1` vs. 7.7.0`, not mentioning major version mismatch.

There is a noticable drop in performance in such scenario esecially during mass reindex operations (e.g. when you have 10mln documents), although everything still "works" ok during reading. That is probably why it was not noticed earlier.

wojciechwojcik: Btw Solr provides even an upgrade tool for migrating collections:
https://lucene.apache.org/solr/guide/7_7/indexupgrader-tool.html

FlorianHockmann: I just looked into the deprecation warnings and it seems non-trivial to resolve them as not all of them have a simple replacement, e.g., [the `SynonymFilterFactory` was replaced by the `SynonymGraphFilterFactory` but it's recommended to additionally use the `FlattenGraphFilterFactory` at index time](https://lucene.apache.org/core/7_0_0/analyzers-common/org/apache/lucene/analysis/synonym/SynonymFilterFactory.html).
Since Solr 7 was also already the default before this PR, I don't see a reason to address the deprecation warnings in this PR. All that changed here is that these warnings now also occur in tests that use the test compose file.
I suggest that we address these warnings in a follow-up PR as they are not really related to this one, the latest point would be when we support Solr 8 (#1569) as that removes most of the deprecated fields.

Regarding the `luceneMatchVersion`: The comment mentions that re-indexing is recommended and the `IndexUpgrader Tool` is only able to upgrade from the latest major version which is 6 in this case. I'm wondering whether we should really bump two major versions here directly as that means that users have to manually upgrade their indices first to version 6 and then again from there to 7. Should we maybe only bump it to `6.0.0` for the `0.4.0` release and then to `7.0.0` in `0.5.0`?

wojciechwojcik: As I understand the janus distribution only provides an example solr config for creating new collections, that a user must explicitly copy to solr (standalone) or upload to zookeeper (cloud mode). 
So xxisting users will already have their collections created using older config and changing the example file is not going to affect them at all.

IMO the example should provide most up-to-date config version for new users.

wojciechwojcik: As for warnings I'm ok to deal with them a seperate PR as long as there is an issue open for that, so it is not lost. As you said, these changes are not trivial and are pretty important ones if we want truly to support Solr 7/8, not only in backward compatibility mode.

FlorianHockmann: > So existing users will already have their collections created using older config and changing the example file is not going to affect them at all.

That's probably correct so I just pushed an updated commit where this value is set to `7.0.0`, the default Solr version JanusGraph uses. I also added a note to the upgrade docs so users are aware of this fact in case they want to use the new config file together with an existing Solr installation.

> As for warnings I'm ok to deal with them a seperate PR as long as there is an issue open for that, so it is not lost. 

True, I created an issue to resolve those deprecation warnings (#1584) and we already have an issue to update Solr to a more recent 7.y version (#1306)

PR #1574, opened 2019-05-13
external-locking configuration parameter for CQL #1636

### LINKED ISSUE #1636: Allow usage of external locking mechanism for CQL
In cases where an external locking mechanism is used, it makes sense to disable JanusGraph's internal locking mechanism. This allows to improve performance of the CQL storage as it doesn't make it perform redundant read-before-write operation for each column participated in transaction.

ph1lm:
…able build-in locking mechanism in JanusGraph for the case when locking is already provided by the application.

This allows to improve performance of underlying storage (Cassandra in our case) and doesn't make it perform redundant read-before-write operation for each column participated in transaction.

Fixes #1636

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
@@ -65,6 +65,19 @@ public interface CQLConfigOptions {
                 "and always use LOCAL_QUORUM instead",
             ConfigOption.Type.MASKABLE, false);
 
+    ConfigOption<Boolean> USE_EXTERNAL_LOCKING = new ConfigOption<>(
+            CQL_NS,
+            "use-external-locking",
+            "True to prevent JanusGraph from using its own locking mechanism. Setting this to true eliminates " +
+            "redundant checks when using an external locking mechanism outside of JanusGraph. Be aware that " +
+            "when use-external-locking is set to true, that failure to employ a locking algorithm which locks " +
+            "all columns that participate in a transaction upfront and unlocks them when the transaction ends, " +
+            "will result in a 'read uncommitted' transaction isolation level guarantee. If set to true without " +
+            "an appropriate external locking mechanism in place side effects such as " +
+            "dirty/non-repeatable/phantom reads should be expected.",
+            ConfigOption.Type.MASKABLE,
+            false);
+
     // The number of statements in a batch
     ConfigOption<Integer> BATCH_STATEMENT_SIZE = new ConfigOption<>(
             CQL_NS,
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
@@ -365,7 +365,10 @@ public class CQLKeyColumnValueStore implements KeyColumnValueStore {
 
     @Override
     public void acquireLock(final StaticBuffer key, final StaticBuffer column, final StaticBuffer expectedValue, final StoreTransaction txh) throws BackendException {
-        throw new UnsupportedOperationException();
+        final boolean hasLocking = this.storeManager.getFeatures().hasLocking();
+        if (!hasLocking) {
+            throw new UnsupportedOperationException(String.format("%s doesn't support locking", getClass()));
+        }
     }
 
     @Override
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
@@ -42,6 +42,7 @@ import static org.janusgraph.diskstorage.cql.CQLConfigOptions.SSL_ENABLED;
 import static org.janusgraph.diskstorage.cql.CQLConfigOptions.SSL_TRUSTSTORE_LOCATION;
 import static org.janusgraph.diskstorage.cql.CQLConfigOptions.SSL_TRUSTSTORE_PASSWORD;
 import static org.janusgraph.diskstorage.cql.CQLConfigOptions.WRITE_CONSISTENCY;
+import static org.janusgraph.diskstorage.cql.CQLConfigOptions.USE_EXTERNAL_LOCKING;
 import static org.janusgraph.diskstorage.cql.CQLKeyColumnValueStore.EXCEPTION_MAPPER;
 import static org.janusgraph.diskstorage.cql.CQLTransaction.getTransaction;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.AUTH_PASSWORD;
@@ -185,11 +186,14 @@ public class CQLStoreManager extends DistributedStoreManager implements KeyColum
 
         final Boolean onlyUseLocalConsistency = configuration.get(ONLY_USE_LOCAL_CONSISTENCY_FOR_SYSTEM_OPERATIONS);
 
+        final Boolean useExternalLocking = configuration.get(USE_EXTERNAL_LOCKING);
+
         final StandardStoreFeatures.Builder fb = new StandardStoreFeatures.Builder();
 
         fb.batchMutation(true).distributed(true);
         fb.timestamps(true).cellTTL(true);
         fb.keyConsistent((onlyUseLocalConsistency ? local : global), local);
+        fb.locking(useExternalLocking);
         fb.optimisticLocking(true);
         fb.multiQuery(false);
 
--- a/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLStoreTest.java
+++ b/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLStoreTest.java
@@ -107,6 +107,13 @@ public class CQLStoreTest extends KeyColumnValueStoreTest {
     }
 
     @Test
+    public void testExternalLocking() throws BackendException {
+        assertFalse(this.manager.getFeatures().hasLocking());
+        assertTrue(openStorageManager(getBaseStorageConfiguration()
+                .set(USE_EXTERNAL_LOCKING, true)).getFeatures().hasLocking());
+    }
+
+    @Test
     public void testDefaultCompactStorage() throws BackendException {
         final String cf = TEST_CF_NAME + "_defaultcompact";
 


mbrukman:
@ph1lm — note that https://github.com/JanusGraph/legal/pull/137 is still not yet merged, so that's one issue causing the `[cla: no]` label, but I also think that your email address is not part of your GitHub profile, which will prevent it from being recognized even after the PR is merged.

Note that your email address doesn't have to be _public_, it just has to be added to your GitHub profile and verified for it to connect your username with your email address.

ph1lm:
@mbrukman I've just updated my profile with correct work email.
Thanks.

FlorianHockmann:
@ph1lm Thanks for your contribution! I just have two general remarks right now as I haven't looked (yet)  at the implementation here.
Please change the target branch to `master` here as this is actually a new feature. We have basically closed the `0.2` by now for new features and `0.3` is also only used for bug fixes. New features should go directly into `master`.
Apart from that, it would be good if you could create issues for the features you want to add / bugs you want to fix and then reference them in the pull request. That makes it easier to understand what a pull request should achieve and if you create the issue before you start working on the implementation, then it's possible to discuss about the idea first in general.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -65,6 +65,11 @@
                 "and always use LOCAL_QUORUM instead",
             ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<Boolean> EXTERNAL_LOCKING =
+            new ConfigOption<Boolean>(CQL_NS, "external-locking",
```
chupman: remove 4 spaces of indentation

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -65,6 +65,11 @@
...
+                    "True to use external locking mechanism instead mechanism provided by JanusGraph",
+                    ConfigOption.Type.MASKABLE, false);
```
chupman: remove 8 spaces of indentation on lines 70 and 71

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -185,6 +186,13 @@ public CQLStoreManager(final Configuration configuration) throws BackendExceptio
         fb.optimisticLocking(true);
         fb.multiQuery(false);
 
+        Boolean externalLocking = configuration.get(EXTERNAL_LOCKING);
```
chupman: make final and change name to `hasExternalLocking`

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -185,6 +186,13 @@ public CQLStoreManager(final Configuration configuration) throws BackendExceptio
...
+        LOGGER.info("External locking: {}", externalLocking);
```
chupman: I don't think this log is necessary, please remove.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -185,6 +186,13 @@ public CQLStoreManager(final Configuration configuration) throws BackendExceptio
...
+        if (externalLocking) {
+            // let JanusGraph core know that we'll handle lock on our own
+            fb.locking(true);
```
chupman: please place with the other `fb.` setters and use `hasExternalLocking` instead of `true`

```
fb.locking(hasExternalLocking);
```

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -366,7 +366,6 @@ public void mutate(final StaticBuffer key, final List<Entry> additions, final Li
 
     @Override
     public void acquireLock(final StaticBuffer key, final StaticBuffer column, final StaticBuffer expectedValue, final StoreTransaction txh) throws BackendException {
-        throw new UnsupportedOperationException();
```
chupman: Rather than remove this can you you wrap it in a conditional statement based off of `EXTERNAL_LOCKING`

porunov:
@FlorianHockmann Looks like @ph1lm wants to add these changes specifically to the `0.2` branch from this discussion https://groups.google.com/forum/#!topic/janusgraph-dev/4K_rpmBWQMI . I know that `0.2` branch is a pain to maintain and I was voting to drop the support for `0.2` ASAP, but if @ph1lm or someone else will merge `0.2` branch to `0.3` and then to `master` then I am OK with it.
I think you should decide if we should accept these changes in `0.2` because you are a PM.

chupman:
@porunov and @FlorianHockmann I've already handled a few in the past and I don't mind doing the merges from 0.2. That being said I don't think all 4 will be making it into 0.2.0 since I don't intend to push back on the May 22nd release date. I'll update the janusgraph-dev conversation today or tomorrow depending on how fast I can finish up my remaining reviews.

FlorianHockmann:
> Looks like @ph1lm wants to add these changes specifically to the 0.2 branch from this discussion

Yeah, I only read that discussion after seeing this PR. If @chupman as the release manager for 0.2.3 is OK to let these PRs (or some of them) target `0.2`, then I'm fine with that.
My intention was in general mainly to point @ph1lm at our conventions for contributions and our workflow.

I also just created one issue for one of these PRs and linked an existing issue for another PR. For the other 2 PRs, I just didn't understand them good enough from their description to create a meaningful issue.

ph1lm:
Thanks for the comments. I'll address all of them asap.

chupman:
@ph1lm please rebase and re-target at master.

ph1lm:
> @ph1lm please rebase and re-target at master.

roger that

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
ph1lm: Done

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
ph1lm: Done

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
ph1lm: Done

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
ph1lm: Done

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
ph1lm: Done

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
ph1lm: Done

ph1lm:
@chupman All comments are addressed and target branch changed to master

FlorianHockmann:
@ph1lm In order to review this PR, some knowledge about how JanusGraph uses locking is required. Since you apparently already spent the time to look into this, could you please give a short description of this read-before-write mechanism or maybe a pointer to where in the code this happens and how your change avoids this mechanism from kicking in?
(I know this might seem strange to ask from a contributor, but most of this internal code of JanusGraph was written when the project was still Titan and those contributors aren't active here any more. Most current contributors don't typically have intimate knowledge of this part of the code base which means that reviews take some time.)
I'm also wondering whether this config option should really be backend specific. Wouldn't this also be helpful to users of other backends?

ph1lm:
@FlorianHockmann 

> @ph1lm In order to review this PR, some knowledge about how JanusGraph uses locking is required. Since you apparently already spent the time to look into this, could you please give a short description of this read-before-write mechanism or maybe a pointer to where in the code this happens and how your change avoids this mechanism from kicking in?

Sure.

This flag controls whether `KeyColumnValueStoreManager` is wrapped by special `ExpectedValueCheckingStoreManager` implementation ([here](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java#L169)) which uses special `ExpectedValueCheckingStore` implementation to read and track old values for columns that participate in transaction. Essentially it's needed to have own [snapshot isolation](https://en.wikipedia.org/wiki/Snapshot_isolation) mechanism as following:
- read all values for the columns that participate in transaction 
- lock those columns if their values have not been changed 
- perform transaction
- unlock columns

This mechanism is needed for backends that don't support [isolation](https://en.wikipedia.org/wiki/Isolation_(database_systems)) natively and `org.janusgraph.diskstorage.keycolumnvalue.StoreFeatures#hasLocking` flag is used to inform JanusGraph core to use its own (described above) mechanism. 

As for Cassandra, such Read-Before-Write pattern is one of the anti-pattern which results in doubled number of roundtrips 'cause every read (as well as write) requires some number of replicas to confirm value in order to treat it as consistent.

For some projects it would be better to have opportunity to turn this mechanism off, for example, if they use [shared-nothing approach](https://en.wikipedia.org/wiki/Shared-nothing_architecture) or if they prefer (like we) to have external pessimistic [coarse grain locking](https://en.wikipedia.org/wiki/Lock_(computer_science)#Granularity).

> I'm also wondering whether this config option should really be backend specific. Wouldn't this also be helpful to users of other backends?

It might be restricted for some backends to be used without aforementioned "isolation mechanism" and Cassandra backend is good example of that. I'm afraid that we may end up with two configuration parameters (on both: core and backend levels) if we want to be backward compatible.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -65,6 +65,13 @@
                 "and always use LOCAL_QUORUM instead",
             ConfigOption.Type.MASKABLE, false);
 
+    ConfigOption<Boolean> EXTERNAL_LOCKING = new ConfigOption<>(
+            CQL_NS,
+            "external-locking",
+            "True to use external locking mechanism instead mechanism provided by JanusGraph",
```
FlorianHockmann: Just a nitpick basically, but when just reading this, I'm not sure whether it activates some other external locking mechanism directly or whether it just tells JanusGraph to not use its locking mechanism because I already have my own mechanism.

How about something like: _True to prevent JanusGraph from using its own locking mechanisms because an external locking mechanism is used outside of JanusGraph._ ?

I would probably also add a sentence that makes it clear that users have to employ their own locking mechanism in this case or get inconsistent data.

ph1lm:
@FlorianHockmann Thanks for your remarks. Answers are inlined below.

> shouldn't the config option then be named _use-internal-locking_ instead of _external-locking_?

Yeah I had the same dilemma.
`external` makes more sense in fully qualified format:
```
storage.cql.use-external-locking=true
```
otherwise in a case of `internal` it can be treated like 'CQL takes care of that' which is not true:
```
storage.cql.use-internal-locking=true
```
But I agree that `use-` prefix should be used.
 
> > if we want to be backward compatible.
> 
> First of all, we don't have to be backward compatible for version 0.4.0 as we can introduce breaking changes in that version. But wouldn't it just be the same config option you just added for CQL, but for all backends?

Currently [core decides to use or not to use internal locking mechanism](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java#L169) using `StoreFeatures` which are [constructed on backend level ](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java#L188). Lets image we have backend that has 2 options for locking which are configured by `storage.name.use-locking-option=1 (or 2)` for example. If we had core-level configuration parameters for internal/external locking then we would need to set two configuration parameters to configure locking:
```
storage.use-internal-locking=false
storage.name.use-locking-option=1 
```
when just one `storage.name.use-locking-option=1 ` configuration parameters would be more than enough 'cause using of any `locking-option` should automatically set [hasLocking(true)](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java#L60) 
That's what meant by saying:
> I'm afraid that we may end up with two configuration parameters (on both: core and backend levels)

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
ph1lm: I could not agree with you more.

FlorianHockmann:
OK, from my point we can use this configuration parameter now as is in this PR for CQL. I just added a nitpick comment about two linguistic issues. The PR is good to go in general from my side and I vote already to get the 7 day period started after which we can merge this with lazy consensus in case no one else reviews this until then.

I also added an issue for this PR for our changelog and linked it in the description. If you change the language thing, then you could also include the issue number in the commit message to link those two together.

VOTE +1

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -65,6 +65,18 @@
                 "and always use LOCAL_QUORUM instead",
             ConfigOption.Type.MASKABLE, false);
 
+    ConfigOption<Boolean> USE_EXTERNAL_LOCKING = new ConfigOption<>(
+            CQL_NS,
+            "use-external-locking",
+            "True to prevent JanusGraph from using its own locking mechanism because an external locking " +
+            "mechanism is used outside of JanusGraph. Be aware that in case of using external locking " +
+            "mechanism you should employ your own locking algorithm which locks all columns that participate " +
+            "in transaction upfront and unlocks them when transaction ends, otherwise you can rely on " +
```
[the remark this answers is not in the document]
ph1lm: Fixed.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -65,6 +65,18 @@
                 "and always use LOCAL_QUORUM instead",
             ConfigOption.Type.MASKABLE, false);
 
+    ConfigOption<Boolean> USE_EXTERNAL_LOCKING = new ConfigOption<>(
+            CQL_NS,
+            "use-external-locking",
+            "True to prevent JanusGraph from using its own locking mechanism because an external locking " +
```
chupman: I took another glance and have some recommendations for the config option description. 
```
                "True to prevent JanusGraph from using its own locking mechanism. Setting this to true eliminates " +
                "redundant checks when using an external locking mechanism outside of JanusGraph. Be aware that " + 
                "when use-external-locking is set to true, that failure to employ a locking algorithm which locks " +
                "all columns that participate in a transaction upfront and unlocks them when the transaction ends, " +
                "will result in a 'read uncommitted' transaction isolation level guarantee. If set to true without " + 
                "an appropriate external locking mechanism in place side effects such as " +
                "dirty/non-repeatable/phantom reads should be expected.",
```

ph1lm: Fixed.

PR #1575, opened 2019-05-13
Fix the problem of index update for the columns that have Cardinality.SET

### LINKED ISSUE #1583: Index update not working for Cardinality.SET in single transaction
If we have a mixed index on a property with `Cardinality.SET` and perform remove and add value for this property in one transaction then remove is swallowed by the consolidation mechanism which assumes that there will be no need for deletion since old value will be replaced by a new one, which is not true for cardinality != `SINGLE`.

ph1lm:
Fixes #1583

-----

### For all changes:

### For code changes:
- [-] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
@@ -16,11 +16,13 @@ package org.janusgraph.diskstorage.indexing;
 
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import org.janusgraph.core.Cardinality;
 import org.janusgraph.diskstorage.EntryMetaData;
 import org.janusgraph.diskstorage.Mutation;
 
-import javax.annotation.Nullable;
+import java.util.AbstractMap;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * An index mutation contains the field updates (additions and deletions) for a particular index entry.
@@ -33,19 +35,25 @@ import java.util.List;
 
 public class IndexMutation extends Mutation<IndexEntry,IndexEntry> {
 
+    private final KeyInformation.StoreRetriever storeRetriever;
     private final boolean isNew;
     private boolean isDeleted;
 
-    public IndexMutation(List<IndexEntry> additions, List<IndexEntry> deletions, boolean isNew, boolean isDeleted) {
+    public IndexMutation(KeyInformation.StoreRetriever storeRetriever,
+                         List<IndexEntry> additions, List<IndexEntry> deletions,
+                         boolean isNew, boolean isDeleted) {
         super(additions, deletions);
         Preconditions.checkArgument(!(isNew && isDeleted),"Invalid status");
+        this.storeRetriever = storeRetriever;
         this.isNew = isNew;
         this.isDeleted = isDeleted;
     }
 
-    public IndexMutation(boolean isNew, boolean isDeleted) {
+    public IndexMutation(KeyInformation.StoreRetriever storeRetriever,
+                         boolean isNew, boolean isDeleted) {
         super();
         Preconditions.checkArgument(!(isNew && isDeleted),"Invalid status");
+        this.storeRetriever = storeRetriever;
         this.isNew = isNew;
         this.isDeleted = isDeleted;
     }
@@ -68,22 +76,26 @@ public class IndexMutation extends Mutation<IndexEntry,IndexEntry> {
         isDeleted=false;
     }
 
-    public static final Function<IndexEntry,String> ENTRY2FIELD_FCT = new Function<IndexEntry, String>() {
-        @Nullable
-        @Override
-        public String apply(final IndexEntry indexEntry) {
-            return indexEntry.field;
-        }
-    };
+    private boolean isCollection(String field) {
+        KeyInformation keyInformation = storeRetriever.get(field);
+        return keyInformation != null && keyInformation.getCardinality() != Cardinality.SINGLE;
+    }
+
+    private static final Object DUMMY = new Object();
+
+    private final Function<IndexEntry, AbstractMap.SimpleEntry<String, Object>> entryConversionFunction =
+            indexEntry -> isCollection(Objects.requireNonNull(indexEntry).field) ?
+                    new AbstractMap.SimpleEntry<>(indexEntry.field, indexEntry.value) :
+                    new AbstractMap.SimpleEntry<>(indexEntry.field, DUMMY);
 
     @Override
     public void consolidate() {
-        super.consolidate(ENTRY2FIELD_FCT,ENTRY2FIELD_FCT);
+        super.consolidate(entryConversionFunction, entryConversionFunction);
     }
 
     @Override
     public boolean isConsolidated() {
-        return super.isConsolidated(ENTRY2FIELD_FCT,ENTRY2FIELD_FCT);
+        return super.isConsolidated(entryConversionFunction, entryConversionFunction);
     }
 
     public int determineTTL() {
--- a/docs/basics.adoc
+++ b/docs/basics.adoc
@@ -202,6 +202,15 @@ The schema type - i.e. edge label, property key, or vertex label - is assigned t
 
 Beyond the schema definition options explained in this section, schema types provide performance tuning options that are discussed in <<advanced-schema>>.
 
+=== Displaying Schema Information
+There are methods to view specific elements of the graph schema within the management API.
+These methods are `mgmt.printIndexes()`, `mgmt.printPropertyKeys()`, `mgmt.printVertexLabels()`, and `mgmt.printEdgeLabels()`.
+There is also a method that displays all the combined output named `printSchema()`.
+
+[source, gremlin]
+mgmt = graph.openManagement()
+mgmt.printSchema()
+
 === Defining Edge Labels
 
 Each edge connecting two vertices has a label which defines the semantics of the relationship. For instance, an edge labeled `friend` between vertices A and B encodes a friendship between the two individuals.
@@ -603,7 +612,7 @@ gremlin> g.V().values('name')
 
 === Cleaning up after the Pre-Packaged Distribution
 
-If you want to start fresh and remove the database and logs you can use the clean command with `janusgraph.sh`. The server should be stopped before running the clean operation. 
+If you want to start fresh and remove the database and logs you can use the clean command with `janusgraph.sh`. The server should be stopped before running the clean operation.
 [source, text]
 ----
 $ cd /Path/to/janusgraph/janusgraph-0.2.0-hadoop2/
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexTransaction.java
@@ -79,7 +79,7 @@ public class IndexTransaction implements BaseTransaction, LoggableTransaction {
         final Map<String, IndexMutation> storeMutations = mutations.computeIfAbsent(store, k -> new HashMap<>(DEFAULT_INNER_MAP_SIZE));
         IndexMutation m = storeMutations.get(documentId);
         if (m==null) {
-            m = new IndexMutation(isNew,isDeleted);
+            m = new IndexMutation(keyInformation.get(store), isNew, isDeleted);
             storeMutations.put(documentId, m);
         } else {
             //IndexMutation already exists => if we deleted and re-created it we need to remove the deleted flag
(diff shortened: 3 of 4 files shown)


FlorianHockmann:
@ph1lm I created an issue for this PR and linked it here. Please use the checklist that is included in every PR description in the future to make sure that it adheres to our guideline for contributions.
It would be good if you could update the commit message as a last step to reference the new issue: #1583 that this PR should fix (that's the first point of the checklist). You can for example do this with:

```bash
git commit --amend
// edit the commit message to include '#1583', ideally in subject line (which is the first one)
git push -f
```

ph1lm:
Thanks @FlorianHockmann 
I will do it asap.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
```diff
@@ -68,11 +75,16 @@ public void resetDelete() {
...
+        KeyInformation keyInformation = storeRetriever.get(field);
+        return keyInformation != null && keyInformation.getCardinality() != Cardinality.SINGLE;
+    }
+
+    private final Function<IndexEntry,String> ENTRY2FIELD_FCT = new Function<IndexEntry, String>() {
         @Nullable
         @Override
         public String apply(final IndexEntry indexEntry) {
-            return indexEntry.field;
+            return isCollection(indexEntry.field) ? indexEntry.field + "." + indexEntry.value : indexEntry.field;
```
porunov: As @chupman said here https://github.com/JanusGraph/janusgraph/pull/1547#issuecomment-493235533 this won't help if we have duplicate values (for example for `Cardinality.LIST`).

chupman: Are you referring the very specific use case of a user with duplicate values like:
```
b = { VertexLabel : user,
  name: Robert,
  alias: [bob,bob,jimbob,bobO,bob]
}
```
and then had a transaction that added a 4th bob, but removed 2 others that those removals would get wiped out?

I would agree that this won't cover that edge case, but If you look at the test attached it will properly retain deletions if a property of LIST cardinialty is removed entirely and then readded.

porunov: @chupman Yes, I am referring this use case. I thought about it. Maybe, if we change `consolidate` method to work with duplicates then this fix would be good. I could propose to change `consolidate` method in `Mutation` class with something like this:

```
public<V> void consolidate(Function<E,V> convertAdditions, Function<K,V> convertDeletions) {
    if (hasDeletions() && hasAdditions()) {
        Map<V, Integer> addsMap = new HashMap<>();
        for(E add : additions){
            V name = convertAdditions.apply(add);
            Integer currentSimilarAdditions = addsMap.get(name);
            if(currentSimilarAdditions == null){
                currentSimilarAdditions = 1;
            } else {
                ++currentSimilarAdditions;
            }
            addsMap.put(name, currentSimilarAdditions);
        }

        deletions.removeIf(k -> {

            V name = convertDeletions.apply(k);

            Integer currentSimilarAdditions = addsMap.get(name);

            if(currentSimilarAdditions == null){
                return false;
            }

            --currentSimilarAdditions;
            if(currentSimilarAdditions==0){
                addsMap.remove(name);
            } else {
                addsMap.put(name, currentSimilarAdditions);
            }

            return true;
        });
    }
}
```

I didn't test the above logic yet. Maybe it contains some mistakes. I just proposing to rewrite this logic so that it would be aware of duplicates. I am not yet sure, maybe it is a stupid idea. I just need some time to test it but not sure I will be able till Monday.

chupman: I think a new, very specific issue and PR would make sense for that.

If you agree with it being a separate PR, is the only remaining comment to address making `ENTRY2FIELD_FCT` static again? Or, if there's a good reason for it not being static, having it renamed something like `entry2field_fct` so it's not confused as being static?

porunov: `ENTRY2FIELD_FCT` should be static.
Also, I want to mention that this solution won't work if the value is an object with non overwritten `toString` method.
I agree we should create a separate issue for the discussed problem.

porunov: The field name may contain dots, so the solution is error prone right now.
For example:
field name `hello.world`, value `!` will conflict with field name `hello`, value `world.!`.

porunov: We could return a custom object with overwritten `equals` and `hashCode`.
The idea is to use not a string but a custom object if the field is collection. I.e.:
```
public static final Function<IndexEntry,Object> ENTRY2FIELD_FCT = new Function<IndexEntry, Object>() {
    @Nullable
    @Override
    public Object apply(final IndexEntry indexEntry) {

        if(isCollection(indexEntry.field)){
            return new FieldDefinition(indexEntry.field, indexEntry.value);
        }

        return indexEntry.field;
    }
};
```

FieldDefinition could be something like:
```
class FieldDefinition {
    private String field;
    private Object value;

    public FieldDefinition(String field, Object value) {
        this.field = field;
        this.value = value;
    }

    @Override
    public boolean equals(Object o){
        if(o instanceof FieldDefinition){
            return Objects.equals(field, ((FieldDefinition) o).field) && 
                Objects.equals(value, ((FieldDefinition) o).value);
        }
        return false;
    }
    
    @Override
    public int hashCode(){
        return Objects.hash(field, value);
    }
}
```

In such situation we won't have above problems with field name and custom objects don't need to overwrite `toString` method.

chupman: @porunov I think this would work if `ENTRY2FIELD_FCT` returned the `hashCode()` instead of the FieldDefinition  object. I might be missing it, but I don't think we even need the equals method since the check is `contains`, not `equal`. Nice write-up! I was still thinking about this and really like your approach.

Separately how do you want to handle this PR? Right now my thoughts are `ENTRY2FIELD_FCT ` needs to be made static and it needs to be rebased onto 0.3. It doesn't create any new issues. It just doesn't completely fix every use case. I also lays good groundwork to fully fix the problem. My suggestion would be to finish cleaning this PR up and do the other fixes in a separate follow-up PR. 

This bug and underlying concept is at least moderately complicated and I think it's better to go with small focused commits as much as possible. That being said it's totally your call.

porunov: > I think this would work if ENTRY2FIELD_FCT returned the hashCode() instead of the FieldDefinition object. I might be missing it, but I don't think we even need the equals method since the check is contains, not equal. Nice write-up! I was still thinking about this and really like your approach.

It isn't quite right to use `hashCode` only because `hashCode` is used in hash table (aka HashSet) only for bucket identification. If the collision acquires then `equals` method is executed to add or overwrite the object. Notice, `hashCode` is `int`, that is why it is easy to get collisions with `long` data type. So, we have to overwrite both `hashCode` and `equals` methods. Otherwise our `Set` will not work properly for `add` operation.

> Separately how do you want to handle this PR? Right now my thoughts are ENTRY2FIELD_FCT needs to be made static and it needs to be rebased onto 0.3.

I would suggest do next things:
1) Use `FieldDefinition` and rewrite `ENTRY2FIELD_FCT` as I shown above.
2) Make `ENTRY2FIELD_FCT` static.
3) Rebase against `0.3` branch.

I am open for suggestions and criticism of this solutions.

chupman: https://github.com/JanusGraph/janusgraph/blob/ff4fe7148a5197eed65bf0b90f2b86c480522f51/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java#L151-L157

From what I understand the `ENTRY2FIELD_FCT` function being passed to `consolidate` doesn't affect that values in the `additions` and `deletions` Lists for the mutation. When consolidate is called right before a commit `adds` is populated based of logic in `ENTRY2FIELD_FCT`. `adds` is being populated with `IndexEntry.field`, which is a string, not the value so there should not be any collision concerns. 

I like hashing better than key, value concatenation because on large strings we'd be eating up memory on creating huge strings that won't be interned for comparison.

porunov: > From what I understand the ENTRY2FIELD_FCT function being passed to consolidate doesn't affect that values in the additions and deletions Lists for the mutation. When consolidate is called right before a commit adds is populated based of logic in ENTRY2FIELD_FCT. adds is being populated with IndexEntry.field, which is a string, not the value so there should not be any collision concerns.

`ENTRY2FIELD_FCT` function decides what objects will be included in `adds` HashSet. Then it removes similar adds from `deletions`. Right now all objects in `adds` are `String` with `IndexEntry.field`.
If we want to fix problems for `Cardinality.SET` we should include in `adds` not only objects that are represented by a specific `field` but also `value`.
As we know, the value is `Object` and it can be anything. My proposal is to return not a `String` in `ENTRY2FIELD_FCT` but an `Object` which represents both `field` and `value`.
I am not sure if we are talking about same collision.
Do you propose to use something like?:
```
public static final Function<IndexEntry,Object> ENTRY2FIELD_FCT = new Function<IndexEntry, Object>() {
    @Nullable
    @Override
    public Object apply(final IndexEntry indexEntry) {

        if(isCollection(indexEntry.field)){
            return Objects.hash(indexEntry.field, indexEntry.value);
        }

        return indexEntry.field;
    }
};
```

If yes, then it is a wrong solution because of hashCode collision. 
You said that if we have a unique `indexEntry.field` then we won't get a collision. It is wrong. We can get a collision here. Here are facts:
- `int` is `32` bits. `indexEntry.field` can be more then `32` bits. (Can get collision based only on this fact)
- `value` is anything. For example, value can be `long` which is `64` bits. (Can get collision based only on this fact)
- unique value doesn't guarantee a unique hashCode. (Can get collision based only on this fact)

We are calculating one `hashCode` for `indexEntry.field` and `indexEntry.value`. That is why the probability of collision even higher.

We must use a custom object like `FieldDefinition` (shown here https://github.com/JanusGraph/janusgraph/pull/1575#discussion_r286517381) with overwritten `hashCode` **and** `equals` to be added into `adds`. In such situation if the collision acquires when we calculate `hashCode` then HashSet will resolve this collision automatically by executing `equals` method. In your situation you would pass `Integer` objects of calculated hash which would be overwritten in a case of collision.

Based on these facts I am proposing the solution described here: https://github.com/JanusGraph/janusgraph/pull/1575#discussion_r286517381

I am still open for discussion. I could miss something also.

chupman: So 3 things I suppose:
1. I was narrowly (and incorrectly) focused on a collision in respect to the aforementioned issue where a `key` containing a `.` could collide with a different `key + . + value`.
2. After brushing up on the contract between equals() and hashCode() I agree with your approach.and reasoning.
3. Thanks for the extra explanation. I'll look at this again after I do some more reading.

porunov:
@ph1lm Will you be able to provide fixes explained here https://github.com/JanusGraph/janusgraph/pull/1575#discussion_r286517381 and rebase from `0.3` branch?

ph1lm:
@porunov Sorry for delay
_We had a really hard 2-weeks long deployment on prod which made our knowledge of JG much better._
I'll do my best to address all comments in next 1-2 days.

ph1lm:
> @ph1lm Will you be able to provide fixes explained here [#1575 (comment)](https://github.com/JanusGraph/janusgraph/pull/1575#discussion_r286517381) and rebase from `0.3` branch?

Is it not too late to merge it to 0.3? Or maybe I should rebase it from master?

chupman:
@ph1lm @porunov is the release manager for 0.3.2 so you should be safe targeting the 0.3 branch. It's a bug fix rather than a new feature so it's better to target 0.3 so 0.3.2 will get it as well.

porunov:
> Is it not too late to merge it to 0.3? Or maybe I should rebase it from master?

It isn't too late for `0.3`. I am going to approve it and merge when the above comments are resolved and the PR is tergeted to `0.3` branch. The fix will automatically be applied to `master` branch.

porunov:
@ph1lm Will you be able to provide a fix this week? If you are busy I can create a separate PR and fix it.

ph1lm:
@porunov @chupman I'm really sorry, I realized that I rebased and pushed changes before I changed target branch and it made @codacy-bot crazy. I'm closing this PR and open a new one from the new branch.

PR #1576, opened 2019-05-13
Improved support of ElasticSearch 6 by refactoring usage of painless scripts functionality

### LINKED ISSUE #1702: Use parameterized scripts instead of hardcoded for ElasticSearch
If the script isn't parametrized, than queries which use similar script will most likely miss scripts cache on ElasticSearch side which will cause script compilation for each (or almost each) query.
Script compilation is a heavy process.
If the script is parametrized, than it will be compiled and put into cache. As our all next queries will use the same script, the ElasticSearch will get the compiled version of the script from cache instead of compiling it again. That is why the overall performance will increase if we use parametrized scripts.

ph1lm:
Now painless scripts are parametrized and won't be compiled on every update and as a result load on ElasticSearch will be lower.

Fixes #1702 

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchConstants.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchConstants.java
@@ -24,10 +24,11 @@ public class ElasticSearchConstants {
     public static final String ES_DOC_KEY = "doc";
     public static final String ES_UPSERT_KEY = "upsert";
     public static final String ES_SCRIPT_KEY = "script";
-    public static final String ES_INLINE_KEY = "inline";
+    public static final String ES_SOURCE_KEY = "source";
+    public static final String ES_PARAMS_KEY = "params";
+    public static final String ES_PARAMS_FIELDS_KEY = "fields";
     public static final String ES_LANG_KEY = "lang";
     public static final String ES_TYPE_KEY = "type";
-    public static final String ES_INDEX_KEY = "index";
     public static final String ES_ANALYZER = "analyzer";
     public static final String ES_GEO_COORDS_KEY = "coordinates";
     public static final String CUSTOM_ALL_FIELD = "all";
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -27,13 +27,11 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
 import org.janusgraph.diskstorage.es.compat.ES6Compat;
 import org.janusgraph.diskstorage.es.compat.ES7Compat;
 import org.janusgraph.diskstorage.es.rest.util.HttpAuthTypes;
 import org.locationtech.spatial4j.shape.Rectangle;
-import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
-import org.apache.tinkerpop.shaded.jackson.databind.ObjectWriter;
-import org.apache.tinkerpop.shaded.jackson.databind.SerializationFeature;
 import org.janusgraph.core.Cardinality;
 import org.janusgraph.core.JanusGraphException;
 import org.janusgraph.core.attribute.Cmp;
@@ -85,6 +83,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.UUID;
@@ -259,15 +258,30 @@ public class ElasticSearchIndex implements IndexProvider {
      */
     public static final double DEFAULT_GEO_DIST_ERROR_PCT = 0.025;
 
+    private static final String PARAMETERIZED_DELETION_SCRIPT = String.join(" ", "",
+            "for (field in params.fields) {",
+            "    if (field.cardinality == 'SINGLE') {",
+            "        ctx._source.remove(field.name);",
+            "    } else if (ctx._source.containsKey(field.name)) {",
+            "        def fieldIndex = ctx._source[field.name].indexOf(field.value);",
+            "        if (fieldIndex >= 0 && fieldIndex < ctx._source[field.name].size()) {",
+            "            ctx._source[field.name].remove(fieldIndex);",
+            "        }",
+            "    }",
+            "}");
+
+    private static final String PARAMETERIZED_ADDITION_SCRIPT = String.join(" ", "",
+            "for (field in params.fields) {",
+            "    if (ctx._source[field.name] == null) {",
+            "        ctx._source[field.name] = [];",
+            "    }",
+            "    if (field.cardinality != 'SET' || ctx._source[field.name].indexOf(field.value) == -1) {",
+            "        ctx._source[field.name].add(field.value);",
+            "    }",
+            "}");
+
     private static final String MAX_OPEN_SCROLL_CONTEXT_PARAMETER = "search.max_open_scroll_context";
     private static final Map<String, Object> MAX_RESULT_WINDOW = ImmutableMap.of("index.max_result_window", Integer.MAX_VALUE);
-
-    private static final ObjectWriter mapWriter;
-    static {
-        final ObjectMapper mapper = new ObjectMapper();
-        mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
-        mapWriter = mapper.writerWithView(Map.class);
-    }
 
     private static final Parameter[] NULL_PARAMETERS = null;
 
@@ -682,11 +696,12 @@ public class ElasticSearchIndex implements IndexProvider {
                             requestByStore.add(ElasticSearchMutation.createDeleteRequest(indexStoreName, storeName,
                                     documentId));
                         } else {
-                            final String script = getDeletionScript(information, storeName, mutation);
-                            final Map<String,Object> doc = compat.prepareScript(script).build();
+                            List<Map<String, Object>> params = getParameters(information.get(storeName),
+                                mutation.getDeletions(), true);
+                            Map doc = compat.prepareScript(PARAMETERIZED_DELETION_SCRIPT, params).build();
+                            log.trace("Deletion script {} with params {}", PARAMETERIZED_DELETION_SCRIPT, params);
                             requestByStore.add(ElasticSearchMutation.createUpdateRequest(indexStoreName, storeName,
-                                    documentId, doc));
-                            log.trace("Adding script {}", script);
+                                documentId, doc));
                         }
                     }
                     if (mutation.hasAdditions()) {
@@ -704,12 +719,13 @@ public class ElasticSearchIndex implements IndexProvider {
                                 upsert = null;
                             }
 
-                            final String inline = getAdditionScript(information, storeName, mutation);
-                            if (!inline.isEmpty()) {
-                                final ImmutableMap.Builder builder = compat.prepareScript(inline);
+                            List<Map<String, Object>> params = getParameters(information.get(storeName),
+                                    mutation.getAdditions(), false, Cardinality.SINGLE);
+                            if (!params.isEmpty()) {
+                                ImmutableMap.Builder builder = compat.prepareScript(PARAMETERIZED_ADDITION_SCRIPT, params);
                                 requestByStore.add(ElasticSearchMutation.createUpdateRequest(indexStoreName, storeName,
                                         documentId, builder, upsert));
-                                log.trace("Adding script {}", inline);
+                                log.trace("Adding script {} with params {}", PARAMETERIZED_ADDITION_SCRIPT, params);
                             }
 
                             final Map<String, Object> doc = getAdditionDoc(information, storeName, mutation);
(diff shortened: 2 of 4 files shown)


chupman:
This is failing on Elasticsearch versions 1 and 2. I'm going to hold off on reviewing until it passes in travis. Without looking into why it's failing you would probably have better luck targeting the master/0.4 branch since support for ES versions 1 and 2 will be removed.

ph1lm:
Thanks for the comments. I'll address all of them asap.

chupman:
@ph1lm please rebase and re-target at master.

ph1lm:
> @ph1lm please rebase and re-target at master.

roger that

FlorianHockmann:
@sjudeng and @davidclement90 You seem to have the best knowledge of the Elasticsearch backend, would you mind reviewing this PR?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -726,6 +768,26 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
         }
     }
 
+    private List<Map<String, Object>> getParameters(KeyInformation.StoreRetriever storeRetriever,
+                                                    List<IndexEntry> entries,
+                                                    Predicate<Cardinality> predicate) {
```
FlorianHockmann: (nitpick) I think this `predicate` is a bit confusing. It's not clear from it's name what it should actually test and it seems to be only necessary for the addition case.
Would it make sense to create two different methods for additions and deletions here? If you want to share the code between both, then you could add a third method with the actual implementation, but then I would probably use a boolean flag instead that has a name like `skipSingleCardinality` to get rid of the predicate.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -264,6 +265,29 @@
      */
     public static final double DEFAULT_GEO_DIST_ERROR_PCT = 0.025;
 
+    private static final String DELETION_SCRIPT = String.join(" ", "",
+            "for (field in params.fields) {",
+            "    if (field.cardinality == 'SINGLE') {",
+            "        ctx._source.remove(field.name);",
+            "    } else if (ctx._source.containsKey(field.name)) {",
+            "        def fieldIndex = ctx._source[field.name].indexOf(field.value);",
```
FlorianHockmann: I just compared this string to the one created in `getDeletionScript` and noticed that `field.value` is converted there with `convertToJsType` for this `indexOf` operation. Any reason why this isn't necessary any more?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -264,6 +265,29 @@
      */
     public static final double DEFAULT_GEO_DIST_ERROR_PCT = 0.025;
 
+    private static final String DELETION_SCRIPT = String.join(" ", "",
```
FlorianHockmann: (nitpick) This script can only be used if parameters can be used. So, shouldn't we name it `PARAMETERIZED_DELETION_SCRIPT`? (same for `ADDITION_SCRIPT` below)

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -693,12 +724,23 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
...
 
-                            final String inline = getAdditionScript(information, storeName, mutation);
-                            if (!inline.isEmpty()) {
-                                final ImmutableMap.Builder builder = compat.prepareScript(inline);
-                                requestByStore.add(ElasticSearchMutation.createUpdateRequest(indexStoreName, storeName,
-                                        documentId, builder, upsert));
-                                log.trace("Adding script {}", inline);
+                            if (compat.supportsParams()) {
+                                List<Map<String, Object>> params = getParameters(information.get(storeName), mutation.getAdditions(),
+                                        c -> c == Cardinality.SET || c == Cardinality.LIST);
```
FlorianHockmann: I basically don't know anything about script parameterization with ES, so this might be a stupid question, but why does this only work for cardinality `SET` or `LIST` and how about indexed properties with `Cardinality.SINGLE`? How are they added?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/ES6Compat.java
```diff
@@ -25,9 +34,38 @@
...
+     * See https://github.com/elastic/elasticsearch/issues/26823
+     */
+    @Override
+    public ImmutableMap.Builder prepareScript(String inline) {
+        Map<String, String> script = ImmutableMap.of(ES_SOURCE_KEY, inline, ES_LANG_KEY, scriptLang());
+        return ImmutableMap.<String, Object>builder().put(ES_SCRIPT_KEY, script);
+    }
+
+    /**
+     * Enable params for last version of ES.
```
FlorianHockmann: (nitpick) I think this comment doesn't really add anything that isn't clear from the code so I would just remove it, but if you want to keep it, then I would at least remove the mention of this ES version as _last_ as that changes over time and there is in fact already Elasticsearch 7 so it isn't even true right now.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -745,29 +807,54 @@ private String getDeletionScript(KeyInformation.IndexRetriever information, Stri
                 case LIST:
                     final String jsValue = convertToJsType(deletion.value, compat.scriptLang(), Mapping.getMapping(keyInformation));
                     String index = INDEX_NAME + i++;
-                    script.append("def ")
+                    script.append("if (ctx._source.containsKey(\"")
```
FlorianHockmann: Could you briefly explain the changes to this method? I'm not sure why they are necessary to support parametrized scripts as this method is only used for scripts that cannot be parametrized.

FlorianHockmann:
@ph1lm Are you planning to come back to this shortly? I'm of course not trying to rush you, but since this is basically the only PR left open for the 0.4.0 release (#1637 is also open but ready to be merged), I wonder whether we should just move this to the milestone for 0.4.1 to not delay the 0.4.0 release.

ph1lm:
@FlorianHockmann I was planning to do that tomorrow.  How much time we have? Till the end of the week?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -745,29 +807,54 @@ private String getDeletionScript(KeyInformation.IndexRetriever information, Stri
...
+                        .append(deletion.field)
+                        .append("\")) {")
+                        .append("def ")
```
porunov: Please, combine all predefined strings into a single string where possible. For example, `.append("\")) {").append("def ")` should be `.append("\")) {def ")`

FlorianHockmann:
>  How much time we have? Till the end of the week?

End of the week would be good. I would like to start release preparations either end of the week or on Monday otherwise.

FlorianHockmann:
I moved this to the _v0.4.x_ milestone to not block the 0.4.0 release.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: @FlorianHockmann `convertToJsType` is used to serialize value to `String` which is going to be done by ES client itself since we don't generate dynamic scripts anymore. But I've re-used `convertToEsType` only (which is called inside `convertToJsType`) for backward compatibility.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: What do you think if we had `cardinalityToSkip` vararg where you can enumerate cardinality?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: Here I just added few guards for NPEs that we faced in production. And yes you're right this code won't be used in parametrized scripts but I added them just in case if we decided to switch parameterization off someday.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: Done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -671,11 +695,18 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
                             requestByStore.add(ElasticSearchMutation.createDeleteRequest(indexStoreName, storeName,
                                     documentId));
                         } else {
-                            final String script = getDeletionScript(information, storeName, mutation);
-                            final Map<String,Object> doc = compat.prepareScript(script).build();
+                            Map doc;
+                            if (compat.supportsParams()) {
+                                List<Map<String, Object>> params = getParameters(information.get(storeName), mutation.getDeletions(), c -> true);
+                                doc = compat.prepareScript(DELETION_SCRIPT, params).build();
+                                log.trace("Deletion script {} with params {}", DELETION_SCRIPT, params);
```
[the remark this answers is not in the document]
ph1lm: Done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: Done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/ES6Compat.java
ph1lm: Removed.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: `SINGLE` will just replace the existent value and there is no need in scripts since we're not merging anything.

ph1lm:
@FlorianHockmann @porunov All comments are addressed. 
_I apologize for the delay I was temporarily moved to another project_

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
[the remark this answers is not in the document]
FlorianHockmann: > Done

Isn't the log message still `Deletion script {} with params {}`?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
FlorianHockmann: Did you add any tests here that produce the NPEs without your changes? If not, I can't really verify the changes.
I would suggest that you exclude these changes and create a separate issue for them where you describe the problem you saw in production. That will make it easier to address the problem in a separate PR.

ph1lm: Yes, I've added tests, they are:
- `testListDelete`
- `testRemoveNonExistentProperty`
I'll file them as a separate PR.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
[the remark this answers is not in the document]
ph1lm: yes, please, correct me if I don't understand, but IMO it should be like (pseudocode):
```
if (mutation.hasDeletions()) {
    if (mutation.isDeleted()) {
        log.trace("Deleting entire document");
    } else {
        if (compat.supportsParams()) {
            log.trace("Deletion script with params");
        } else {
            log.trace("Deletion script");
        }
    }
}
if (mutation.hasAdditions()) {
    if (mutation.isNew()) { 
        log.trace("Adding entire document");
    } else {
        if (compat.supportsParams()) {
                log.trace("Adding script with params");
        } else {
                log.trace("Adding script");
        }
    }
}
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -760,6 +804,31 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
         }
     }
 
+    private List<Map<String, Object>> getParameters(KeyInformation.StoreRetriever storeRetriever,
+                                                    List<IndexEntry> entries,
+                                                    boolean deletion,
+                                                    Cardinality... cardinalityToSkip) {
```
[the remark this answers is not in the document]
ph1lm: Done

ph1lm:
@FlorianHockmann All comments are addressed.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/ES6Compat.java
```diff
@@ -25,9 +34,35 @@
 
     private static final IndexFeatures FEATURES = coreFeatures().setWildcardField(CUSTOM_ALL_FIELD).supportsGeoContains().build();
 
+    /**
+     * Script {@code inline} was renamed to {@code source}.
```
porunov: We do not support ElasticSearch version less than 6 anymore in master branch. Currently only ES 6 and ES 7 is supported in master branch. I believe that you can move the below methods directly into `AbstractESCompat` and `supportsParams` methods won't be needed in such case

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -705,11 +730,19 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
...
-                            final String script = getDeletionScript(information, storeName, mutation);
-                            final Map<String,Object> doc = compat.prepareScript(script).build();
+                            Map doc;
+                            if (compat.supportsParams()) {
+                                List<Map<String, Object>> params = getParameters(information.get(storeName),
+                                        mutation.getDeletions(), true);
+                                doc = compat.prepareScript(PARAMETERIZED_DELETION_SCRIPT, params).build();
+                                log.trace("Deletion script {} with params {}", PARAMETERIZED_DELETION_SCRIPT, params);
+                            } else {
+                                final String script = getDeletionScript(information, storeName, mutation);
```
porunov: If you move the logic from `ES6Compat` into `AbstractESCompat` than `supportsParams` parameter isn't needed (because it will be always `true`). That is why, the logic in `else` statement won't be needed also.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -727,12 +760,23 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
...
+                                List<Map<String, Object>> params = getParameters(information.get(storeName),
+                                        mutation.getAdditions(), false, Cardinality.SINGLE);
+                                if (!params.isEmpty()) {
+                                    ImmutableMap.Builder builder = compat.prepareScript(PARAMETERIZED_ADDITION_SCRIPT, params);
+                                    requestByStore.add(ElasticSearchMutation.createUpdateRequest(indexStoreName, storeName,
+                                            documentId, builder, upsert));
+                                    log.trace("Adding script {} with params {}", PARAMETERIZED_ADDITION_SCRIPT, params);
+                                }
+                            } else {
+                                final String inline = getAdditionScript(information, storeName, mutation);
```
porunov: Same here. We can just drop the whole logic inside `else`. We don't need to support ES version less than 6 anymore

porunov:
@ph1lm I've added an issue for your PR. #1702

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/ES6Compat.java
ph1lm: Great, that will simplify things a lot...

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: done

ph1lm:
@porunov I addressed your comments.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -259,16 +258,32 @@
...
+            "    }",
+            "}");
+
+    private static final String PARAMETERIZED_ADDITION_SCRIPT = String.join(" ", "",
+            "for (field in params.fields) {",
+            "    if (ctx._source[field.name] == null) {",
+            "        ctx._source[field.name] = [];",
+            "    }",
+            "    if (field.cardinality == 'SET' && ctx._source[field.name].indexOf(field.value) != -1) {",
+            "        break;",
```
porunov: Can `params.fields` contain fields with different cardinality?
If so, I am not sure it is right to break the loop here. Maybe you thought about `continue;`?
I think we can even invert it to be like:
```
if (field.cardinality != 'SET' || ctx._source[field.name].indexOf(field.value) == -1) {
    ctx._source[field.name].add(field.value);
}
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -737,87 +754,29 @@ public void mutate(Map<String, Map<String, IndexMutation>> mutations, KeyInforma
...
             }
-
-            String value = convertToJsType(e.value, compat.scriptLang(), Mapping.getMapping(keyInformation));
-
-            appendAdditionScript(script, value, e.field, cardinality == Cardinality.SET);
-
-            if (hasDualStringMapping(keyInformation)) {
-                appendAdditionScript(script, value, getDualMappingName(e.field), cardinality == Cardinality.SET);
+            Object jsValue = deletion && info.getCardinality() == Cardinality.SINGLE ?
+                    entry.value : convertToEsType(entry.value, Mapping.getMapping(info));
```
porunov: This is not necessary, but if `cardinality` == SINGLE for deletion, we don't need `entry.value` at all. So we could just make:
```
Object jsValue = deletion && info.getCardinality() == Cardinality.SINGLE ?
                    "" : convertToEsType(entry.value, Mapping.getMapping(info));
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchConstants.java
```diff
@@ -25,6 +25,9 @@
     public static final String ES_UPSERT_KEY = "upsert";
     public static final String ES_SCRIPT_KEY = "script";
     public static final String ES_INLINE_KEY = "inline";
```
porunov: I would also remove `ES_INLINE_KEY` and `ES_INDEX_KEY` because they are not needed anymore

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/AbstractESCompat.java
```diff
@@ -72,9 +74,16 @@ public String scriptLang() {
         return "painless";
     }
 
+    public ImmutableMap.Builder prepareScript(String inline, List<Map<String, Object>> fields) {
```
porunov: I think we can rename `inline` argument to `source`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/AbstractESCompat.java
```diff
@@ -72,9 +74,16 @@ public String scriptLang() {
...
+        Map<String, Object> script = ImmutableMap.of(ES_SOURCE_KEY, inline,
+            ES_PARAMS_KEY, ImmutableMap.of(ES_PARAMS_FIELDS_KEY, fields),
+            ES_LANG_KEY, scriptLang());
+        return ImmutableMap.<String, Object>builder().put(ES_SCRIPT_KEY, script);
+    }
+
     public ImmutableMap.Builder prepareScript(String inline) {
```
porunov: Here also, I think we can rename `inline` argument to `source`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: OMG  I meant `continue` there. Good catch.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
ph1lm: done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchConstants.java
ph1lm: done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/AbstractESCompat.java
ph1lm: done

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/compat/AbstractESCompat.java
ph1lm: done

ph1lm:
@porunov thank you 🇺🇦

PR #1606, opened 2019-05-28
Fix the problem of index update for the columns that have collection cardinality.

### LINKED ISSUE #1583: Index update not working for Cardinality.SET in single transaction

ph1lm:
Fixes #1583

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
@@ -16,10 +16,11 @@ package org.janusgraph.diskstorage.indexing;
 
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import org.janusgraph.core.Cardinality;
 import org.janusgraph.diskstorage.EntryMetaData;
 import org.janusgraph.diskstorage.Mutation;
 
-import javax.annotation.Nullable;
+import java.util.AbstractMap;
 import java.util.List;
 
 /**
@@ -33,19 +34,30 @@ import java.util.List;
 
 public class IndexMutation extends Mutation<IndexEntry,IndexEntry> {
 
+    private final KeyInformation.StoreRetriever storeRetriever;
     private final boolean isNew;
     private boolean isDeleted;
 
-    public IndexMutation(List<IndexEntry> additions, List<IndexEntry> deletions, boolean isNew, boolean isDeleted) {
+    private final Function<IndexEntry, Object> entryConversionFunction =
+            indexEntry -> isCollection(indexEntry.field) ?
+                    new AbstractMap.SimpleEntry<>(indexEntry.field, indexEntry.value) :
+                    indexEntry.field;
+
+    public IndexMutation(KeyInformation.StoreRetriever storeRetriever,
+                         List<IndexEntry> additions, List<IndexEntry> deletions,
+                         boolean isNew, boolean isDeleted) {
         super(additions, deletions);
         Preconditions.checkArgument(!(isNew && isDeleted),"Invalid status");
+        this.storeRetriever = storeRetriever;
         this.isNew = isNew;
         this.isDeleted = isDeleted;
     }
 
-    public IndexMutation(boolean isNew, boolean isDeleted) {
+    public IndexMutation(KeyInformation.StoreRetriever storeRetriever,
+                         boolean isNew, boolean isDeleted) {
         super();
         Preconditions.checkArgument(!(isNew && isDeleted),"Invalid status");
+        this.storeRetriever = storeRetriever;
         this.isNew = isNew;
         this.isDeleted = isDeleted;
     }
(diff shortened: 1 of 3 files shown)


janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
```diff
@@ -68,22 +78,24 @@ public void resetDelete() {
...
-    };
+    private boolean isCollection(String field) {
+        KeyInformation keyInformation = storeRetriever.get(field);
+        return keyInformation != null && keyInformation.getCardinality() != Cardinality.SINGLE;
+    }
+
+    private final Function<IndexEntry, AbstractMap.SimpleEntry<String, Object>> entryConversionFunction =
+            indexEntry -> isCollection(Objects.requireNonNull(indexEntry).field) ?
+                    new AbstractMap.SimpleEntry<>(indexEntry.field, indexEntry.value) :
+                    new AbstractMap.SimpleEntry<>(indexEntry.field, DUMMY);
```
porunov: I think we don't need to wrap `indexEntry.field` into `AbstractMap.SimpleEntry`. I.e. we can run like this:
```
private final Function<IndexEntry, Object> entryConversionFunction =
    indexEntry -> isCollection(Objects.requireNonNull(indexEntry).field) ?
        new AbstractMap.SimpleEntry<>(indexEntry.field, indexEntry.value) :
        indexEntry.field;
```

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
```diff
@@ -68,22 +78,24 @@ public void resetDelete() {
...
-            return indexEntry.field;
-        }
-    };
+    private boolean isCollection(String field) {
+        KeyInformation keyInformation = storeRetriever.get(field);
+        return keyInformation != null && keyInformation.getCardinality() != Cardinality.SINGLE;
+    }
+
+    private final Function<IndexEntry, AbstractMap.SimpleEntry<String, Object>> entryConversionFunction =
+            indexEntry -> isCollection(Objects.requireNonNull(indexEntry).field) ?
```
porunov: Also, could you explain why are you executing `Objects.requireNonNull(indexEntry).field` ?
As for me it is unnecessary because both `Objects.requireNonNull(indexEntry).field` and `indexEntry.field` produces the same `NullPointerException`.

ph1lm:
@porunov 
> I think we don't need to wrap indexEntry.field into AbstractMap.SimpleEntry
I just wouldn't feel comfortable if function returns `Object` but you're right it's not needed here.

> Also, could you explain why are you executing Objects.requireNonNull(indexEntry).field
OMG it was auto-generated while converting to lambda-style function...removed.
 
Done with your comments, squashed and re-pushed again.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/indexing/IndexMutation.java
```diff
@@ -33,19 +34,30 @@
 
 public class IndexMutation extends Mutation<IndexEntry,IndexEntry> {
 
+    private final KeyInformation.StoreRetriever storeRetriever;
     private final boolean isNew;
     private boolean isDeleted;
 
-    public IndexMutation(List<IndexEntry> additions, List<IndexEntry> deletions, boolean isNew, boolean isDeleted) {
+    private final Function<IndexEntry, Object> entryConversionFunction =
```
porunov: Sorry I forgot also to mention that this function was `static`. I think it should be `private static final Function<IndexEntry, Object>`.

porunov: I see that is uses `isCollection` method, so we can make both `entryConversionFunction` and `isCollection` to be `static`.

porunov: My bad. I see that `isCollection` is uses `storeRetriever` which isn't static.

chupman: Where is the apply method within this function? Consolidate still expects it to be present. If this doesn't cause a test failure we have a coverage issue we'll need to look into. 

Here's consolidate for reference:
https://github.com/JanusGraph/janusgraph/blob/1e64abc197f35292f0ba55eaa72a492208af3ad3/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java#L150-L153

ph1lm: It's a syntax sugar of java 8 which allows to create anonymous single-method-class using lambda expression.

porunov: > Where is the apply method within this function?

It is represented as lambda function. 

Looks like those are flaky tests. I've restarted them

porunov: @chupman Tests passed

chupman: Sorry for the delay, but I had to do some more testing and reading before I was comfortable approving.

PR #1644, opened 2019-06-28
Fix resource leaks in LuceneIndex.restore

### LINKED ISSUE #1643: Resource leaks during REINDEX LuceneIndex
[LuceneIndex.restore](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java#L276) does not close IndexReader

mad:
Wrap IndexReader on try-resource

Fix #1643

### DIFF
--- a/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
+++ b/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
@@ -275,30 +275,35 @@ public class LuceneIndex implements IndexProvider {
         writerLock.lock();
         try {
             for (final Map.Entry<String, Map<String, List<IndexEntry>>> stores : documents.entrySet()) {
-                final String store = stores.getKey();
-                final IndexWriter writer = getWriter(store, information);
-                final IndexReader reader = DirectoryReader.open(writer, true, true);
-                final IndexSearcher searcher = new IndexSearcher(reader);
+                IndexReader reader = null;
+                try {
+                    final String store = stores.getKey();
+                    final IndexWriter writer = getWriter(store, information);
+                    reader = DirectoryReader.open(writer, true, true);
+                    final IndexSearcher searcher = new IndexSearcher(reader);
 
-                for (final Map.Entry<String, List<IndexEntry>> entry : stores.getValue().entrySet()) {
-                    final String docID = entry.getKey();
-                    final List<IndexEntry> content = entry.getValue();
+                    for (final Map.Entry<String, List<IndexEntry>> entry : stores.getValue().entrySet()) {
+                        final String docID = entry.getKey();
+                        final List<IndexEntry> content = entry.getValue();
 
-                    if (content == null || content.isEmpty()) {
-                        if (log.isTraceEnabled())
-                            log.trace("Deleting document [{}]", docID);
+                        if (content == null || content.isEmpty()) {
+                            if (log.isTraceEnabled())
+                                log.trace("Deleting document [{}]", docID);
 
-                        writer.deleteDocuments(new Term(DOCID, docID));
-                        continue;
+                            writer.deleteDocuments(new Term(DOCID, docID));
+                            continue;
+                        }
+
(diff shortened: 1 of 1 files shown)


porunov:
Thank you @mad ! Could you please retarget this to `0.3` branch?

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -265,8 +263,6 @@ private void mutateStores(Map.Entry<String, Map<String, IndexMutation>> stores,
                 writer.updateDocument(new Term(DOCID, documentId), doc);
             }
             writer.commit();
-        } finally {
-            IOUtils.closeQuietly(reader);
```
porunov: Do you know why this was `closeQuietly`? As I understand it was used to not throw an exception on failed close which is strange but still. Is it OK that `mutateStores` will throw an exception on failed reader closing?

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -277,28 +273,29 @@ public void restore(Map<String, Map<String, List<IndexEntry>>> documents, KeyInf
             for (final Map.Entry<String, Map<String, List<IndexEntry>>> stores : documents.entrySet()) {
                 final String store = stores.getKey();
                 final IndexWriter writer = getWriter(store, information);
```
porunov: Not related to this PR but I would like to ask a question. Should we wrap writer into `try catch` and close it if it is opened on failed commit?
I.e. something like:
```
final IndexWriter writer = getWriter(store, information);
try(IndexReader reader = DirectoryReader.open(writer, true, true)){
    // other logic
} finally {
    if (writer.isOpen()){
        writer.close();
    }
}
```

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
mad: I don't know, but if close() throw any exception then full tx will be rolled back. From one hand this is strange, because any write operation done without exception, but another IndexReader [instantiated](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java#L227) from IndexWriter, and any exceptions may mean inconsistent state of IndexWriter too

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
mad: Writer is heavy resource and [reused](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java#L145) for any write operations

If exception occured writer.rollback() will be ok

porunov: Thank you for the explanation. I missed that it is reused.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
porunov: I didn't research much but I see that `reader.close()` may produce an exception in some cases. I am just thinking if there is any logic behind swallowing an exception here. Maybe we should always close the reader quietly here and in `restore` method? I just don't know if it is OK to produce an exception on failed reader closing.

mad: Ok, I change try-resource to try-finally-closeQuietly

porunov:
Merging by following a lazy consensus

Issue #1481, opened 2019-03-07
Close transaction during execution lead to excetpion on berkeleyje backend


mad:
I need to interrupt long running queries, and if I just rollback tx then getting exception:

```
Caused by: java.lang.IllegalStateException: Transaction 481 detected open cursors while aborting
	at com.sleepycat.je.txn.Txn.abortInternal(Txn.java:1256)
	at com.sleepycat.je.txn.Txn.abort(Txn.java:1160)
	at com.sleepycat.je.txn.Txn.abort(Txn.java:1133)
	at com.sleepycat.je.Transaction.abort(Transaction.java:220)
	at org.janusgraph.diskstorage.berkeleyje.BerkeleyJETx.rollback(BerkeleyJETx.java:69)
	at org.janusgraph.diskstorage.keycolumnvalue.cache.CacheTransaction.rollback(CacheTransaction.java:203)
	at org.janusgraph.diskstorage.BackendTransaction.rollback(BackendTransaction.java:170)
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.rollback(StandardJanusGraphTx.java:1407)
	... 48 more
```

I think root cause is that BerkeleyJETx/BerkeleyJEKeyValueStore not proper handle openCursors fields, that field not updated instead cursors directly created in BerkeleyJEKeyValueStore

Sample of code to reproduce it
```
graph = JanusGraphFactory.open("berkeleyje:/tmp/jg")

mgmt = graph.openManagement()
a = mgmt.makePropertyKey('a').cardinality(Cardinality.SINGLE).dataType(String.class).make()
mgmt.buildIndex("composite", Vertex.class).addKey(a).buildCompositeIndex()
mgmt.commit()

g = graph.traversal()
clock(1000000) { g.addV().property("a", "2").next() }
g.tx().commit()

g = graph.tx().createThreadedTx().traversal()
Thread.start { g.V().has("a", "2").toList().size() }
sleep(100)
g.tx().rollback()
```

Result
```

         \,,,/
         (o o)
-----oOOo-(3)-oOOo-----
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/pavele/work/janusgraph-0.3.1-hadoop2/lib/slf4j-log4j12-1.7.12.jar!/org/slf4j/impl/StaticLoggerBinder.class]
[...]

PR #1676, opened 2019-07-19
Fix rollback long running tx in bdb

mad:
Fix for #1481

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-berkeleyje/src/main/java/org/janusgraph/diskstorage/berkeleyje/BerkeleyJETx.java
+++ b/janusgraph-berkeleyje/src/main/java/org/janusgraph/diskstorage/berkeleyje/BerkeleyJETx.java
@@ -16,6 +16,7 @@ package org.janusgraph.diskstorage.berkeleyje;
 
 import com.google.common.base.Preconditions;
 import com.sleepycat.je.Cursor;
+import com.sleepycat.je.Database;
 import com.sleepycat.je.DatabaseException;
 import com.sleepycat.je.LockMode;
 import com.sleepycat.je.Transaction;
@@ -35,6 +36,7 @@ public class BerkeleyJETx extends AbstractStoreTransaction {
     private static final Logger log = LoggerFactory.getLogger(BerkeleyJETx.class);
 
     private volatile Transaction tx;
+    private volatile boolean isOpen;
     private final List<Cursor> openCursors = new ArrayList<>();
     private final LockMode lm;
 
@@ -42,6 +44,7 @@ public class BerkeleyJETx extends AbstractStoreTransaction {
         super(config);
         tx = t;
         lm = lockMode;
+        isOpen = true;
         // tx may be null
         Preconditions.checkNotNull(lm);
     }
@@ -50,8 +53,28 @@ public class BerkeleyJETx extends AbstractStoreTransaction {
         return tx;
     }
 
-    private void closeOpenIterators() {
-        openCursors.forEach(Cursor::close);
+    Cursor openCursor(Database db) throws BackendException {
+        synchronized (openCursors) {
+            if (!isOpen) {
+                throw new PermanentBackendException("Transaction already closed");
+            }
+            Cursor cursor = db.openCursor(tx, null);
+            openCursors.add(cursor);
+            return cursor;
+        }
+    }
+
+    void closeCursor(Cursor cursor) {
+        synchronized (openCursors) {
+            cursor.close();
+            openCursors.remove(cursor);
+        }
+    }
+
+    private void closeOpenCursors() {
+        synchronized (openCursors) {
+            openCursors.forEach(Cursor::close);
+        }
     }
 
     LockMode getLockMode() {
@@ -65,7 +88,8 @@ public class BerkeleyJETx extends AbstractStoreTransaction {
         if (log.isTraceEnabled())
             log.trace("{} rolled back", this.toString(), new TransactionClose(this.toString()));
         try {
-            closeOpenIterators();
+            isOpen = false;
+            closeOpenCursors();
             tx.abort();
             tx = null;
         } catch (DatabaseException e) {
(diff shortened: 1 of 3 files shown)


janusgraph-berkeleyje/src/main/java/org/janusgraph/diskstorage/berkeleyje/BerkeleyJETx.java
```diff
@@ -50,8 +53,28 @@ public Transaction getTransaction() {
         return tx;
     }
 
-    private void closeOpenIterators() {
-        openCursors.forEach(Cursor::close);
+    Cursor openCursor(Database db) throws BackendException {
+        synchronized (openCursors) {
+            if (!isOpen) {
+                throw new PermanentBackendException("Transaction is not valid");
```
FlorianHockmann: (nitpick) I think it would be helpful for users if the message would include the reason why the transaction is not valid which is that it's already closed in this case.

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
```diff
@@ -0,0 +1,52 @@
...
+        // Needs for ignore assert when cursor is closed during scan
+        ClassLoader.getSystemClassLoader().setClassAssertionStatus(CursorImpl.class.getName(), false);
+    }
+
+    @Test
+    void longRunningTxShouldBeRolledBacked(@TempDir File dir) throws InterruptedException {
+        JanusGraph graph = JanusGraphFactory.open("berkeleyje:" + dir.getAbsolutePath());
+
+        GraphTraversalSource traversal = graph.traversal();
+        for (int i = 0; i < 100_000; i++) {
```
FlorianHockmann: Do we really need to add 100,000 vertices? That takes ~15 sec on my machine, but the test failed for me without your changes even if just one vertex is added here.

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.concurrent.ExecutionException;
+
+class BerkeleyTransactionTest {
+    static {
+        // Needs for ignore assert when cursor is closed during scan
+        ClassLoader.getSystemClassLoader().setClassAssertionStatus(CursorImpl.class.getName(), false);
+    }
+
+    @Test
+    void longRunningTxShouldBeRolledBacked(@TempDir File dir) throws InterruptedException {
```
FlorianHockmann: (nitpick) _ShouldBeRolledBacked_ -> _ShouldBeRolledBack_

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.sleepycat.je.dbi.CursorImpl;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
+import org.janusgraph.core.JanusGraph;
+import org.janusgraph.core.JanusGraphFactory;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.*;
```
FlorianHockmann: Please use single class imports instead. It's a convention we have for JanusGraph so we should stick to it. (You can also configure your IDE to handle that automatically for you.)

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
mad: Potentially cursor scan may complete before we rollback tx and then future do not throw exception. I ran 10 tests with one vertex 7 failed, 1k vertices 6 failed, but 10k 0 failed

I mark it `@FlakyTest(invocationCount = 5, minSuccess = 1)` and change to 1k vertices

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
mad: Done

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
mad: Done

janusgraph-berkeleyje/src/main/java/org/janusgraph/diskstorage/berkeleyje/BerkeleyJETx.java
mad: Done

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/BerkeleyTransactionTest.java
FlorianHockmann: If 10k are enough to reliably reproduce the problem, then I would prefer that as a value here or maybe 20k which should already reduce the runtime of this test to something in the range of 1-2 seconds which is acceptable in my opinion. (Use a value where you're relatively sure that the tests works deterministically.)
We should only use `FlakyTest` as a workaround for legacy tests that are for some reason flaky where we don't know yet why they are flaky and how we can change them to work deterministically and not as a feature for new tests.

mad: Ok, I set up 15k

PR #1695, opened 2019-07-25
Optimize indexStore names generation in ElasticSearchIndex by caching it

### LINKED ISSUE #1694: Optimize indexStore names generation in ElasticSearchIndex by caching it
We execute `getIndexStoreName` for each search with same or similar `store`. And the next logic is executed during the index store name retrieval: `indexName + "_" + store.toLowerCase();` which creates unnecessary objects and call `.toLowerCase()` which takes time. Because `store` is heavily reused, I am proposing to cache the names for optimization.

porunov:
The optimization is tested on a laptop. Tested with unique index stores count from 1 to 1000000 with `^10` gap. Each gap is tested with parallel threads from 1 to 100. The amount of calls in each thread were from `100000` to `10000000`. 
The threads amount and method call didn't influence differences in speedup much.
The amount of unique index stores amount influence the results.
If unique index stores amount is from 1 to 1000 the result speed up is about 3.4 - 3.8 times faster.
If unique index stores amount is from 1000 to 10000 the result speed up is about 2 times faster.
If unique index stores amount is from 10000 to 50000 the result speed up is about 20-30 % faster.
If unique index stores amount is from 50000 to 100000 the result speed up is about 5 % faster.
If unique index stores amount is from 100000 to 1000000 the result speed up is about 30 % **slower**.

For security measure I setup the cache limit of `50000` elasticsearch indexes' names. If there are more than `50000` ES indexes which are used by JanusGraph, the cache will be disabled and cleared and from that time the current indexStoreName generation logic will be used. 

I don't imagine a project with more than `50000` ElasticSearch indexes but in theory there could be such amount of indexes. So, the PR increases performance of index names generation if there are less than 50000 indexes used. Otherwise the performance will not be changed.

Fixes #1694

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -88,6 +88,7 @@ import java.util.Objects;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


porunov:

Issue #1700, opened 2019-07-25
Improve AddedRelationsContainer


mad:
Currently these containers not efficient in some cases, like multiple property/edge changes in same transaction. Conccurent version used lock but may be switched to non blocking analogue

For single switchet to HPPC conatiner
For concurrent switched to high-scale continer

mad:
Fixed here #1701

PR #1701, opened 2019-07-25
Improve added relations containers #1700

mad:
Fix for #1700 

-----

### For all changes:

### For code changes:

### Note:


porunov:
@mad There are also Travis failures

mad:
Yes, implementation is wrong

AddedRelationsContainer must contain all InternalRelations, not unique by it's id

mad:
I change set to multiset -- now it consume more memory, but more efficient for multiple remove operations

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
```diff
@@ -14,46 +14,109 @@
...
+
+import java.util.*;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
  */
 
-public class SimpleAddedRelations extends ArrayList<InternalRelation> implements AddedRelationsContainer {
-
+public class SimpleAddedRelations implements AddedRelationsContainer {
```
porunov: I didn't check the logic yet but I think this class could be simplified (if I am not missing something). What do you think about this implementation?
```
public class SimpleAddedRelations implements AddedRelationsContainer {
    private static final int INITIAL_ADDED_SIZE = 10;
    private final LongObjectMap<ObjectArrayList<InternalRelation>> container;

    public SimpleAddedRelations() {
        this.container = new LongObjectHashMap<>(INITIAL_ADDED_SIZE);
    }

    @Override
    public boolean add(InternalRelation relation) {
        ObjectArrayList<InternalRelation> relations = container.get(relation.longId());
        if (relations == null) {
            relations = new ObjectArrayList<>(1);
            container.put(relation.longId(), relations);
        }
        relations.add(relation);
        return true;
    }

    @Override
    public boolean remove(InternalRelation relation) {
        ObjectArrayList<InternalRelation> relations = container.get(relation.longId());
        if (relations != null) {
            relations.removeFirst(relation);
            if(relations.size()==0){
                container.remove(relation.longId());
            }
        }
        return true;
    }

    @Override
    public Iterable<InternalRelation> getView(Predicate<InternalRelation> filter) {
        return Iterables.filter(this::iterator, filter);
    }

    @Override
    public boolean isEmpty() {
        return container.size()==0;
    }

    @Override
    public Collection<InternalRelation> getAll() {
        return Collections.unmodifiableCollection(new AbstractCollection<InternalRelation>() {
            @Override
            @Nonnull
            public Iterator<InternalRelation> iterator() {
                return SimpleAddedRelations.this.iterator();
            }

            @Override
            public int size() {
                return Iterators.size(SimpleAddedRelations.this.iterator());
            }
        });
    }

    private Iterator<InternalRelation> iterator() {
        return new Iterator<InternalRelation>() {
            private final Iterator<ObjectCursor<ObjectArrayList<InternalRelation>>> iterator = container.values().iterator();
            private Iterator<ObjectCursor<InternalRelation>> child = Collections.emptyIterator();

            @Override
            public boolean hasNext() {
                return child.hasNext() || iterator.hasNext();
            }

            @Override
            public InternalRelation next() {
                if(!child.hasNext()){
                    child = iterator.next().value.iterator();
                }
                return child.next().value;
            }
        };
    }
}
```

porunov: Also, I didn't check neither the above class nor below, but I think we could optimize memory consumption by using counters. I.e. something like:
```
public class SimpleAddedRelations implements AddedRelationsContainer {
    private static final int INITIAL_ADDED_SIZE = 10;
    private final LongObjectMap<RelationWithCounter> container;

    public SimpleAddedRelations() {
        this.container = new LongObjectHashMap<>(INITIAL_ADDED_SIZE);
    }

    @Override
    public boolean add(InternalRelation relation) {
        RelationWithCounter relations = container.get(relation.longId());
        if (relations == null) {
            container.put(relation.longId(), new RelationWithCounter(relation));
        } else {
            ++relations.amount;
        }
        return true;
    }

    @Override
    public boolean remove(InternalRelation relation) {
        RelationWithCounter relations = container.get(relation.longId());
        if (relations != null) {
            if(--relations.amount <= 0){
                container.remove(relation.longId());
            }
            return true;
        }
        return false;
    }

    @Override
    public Iterable<InternalRelation> getView(Predicate<InternalRelation> filter) {
        return Iterables.filter(this::iterator, filter);
    }

    @Override
    public boolean isEmpty() {
        return container.size()==0;
    }

    @Override
    public Collection<InternalRelation> getAll() {
        return Collections.unmodifiableCollection(new AbstractCollection<InternalRelation>() {
            @Override
            @Nonnull
            public Iterator<InternalRelation> iterator() {
                return SimpleAddedRelations.this.iterator();
            }

            @Override
            public int size() {
                return Iterators.size(SimpleAddedRelations.this.iterator());
            }
        });
    }

    private Iterator<InternalRelation> iterator() {
        return new Iterator<InternalRelation>() {
            private final Iterator<ObjectCursor<RelationWithCounter>> iterator = container.values().iterator();
            private InternalRelation child;
            private int childAmount;

            @Override
            public boolean hasNext() {
                return childAmount > 0 || iterator.hasNext();
            }

            @Override
            public InternalRelation next() {
                if(childAmount == 0){
                    RelationWithCounter relationWithCounter = iterator.next().value;
                    child = relationWithCounter.relation;
                    childAmount = relationWithCounter.amount;
                }
                --childAmount;
                return child;
            }
        };
    }

    private static class RelationWithCounter{
        private final InternalRelation relation;
        private int amount = 1;
        private RelationWithCounter(InternalRelation relation){
            this.relation = relation;
        }
    }
}
```

porunov: @mad I also thinking, what if we just change the old implementation to `HashSet<InternalRelation>` instead of `ArrayList<InternalRelation>`. Is there any reason why we are using a list for it? If we switch to `Set` implementation then we don't need to build additional logic because same relations will be overridden during addition.

mad:
@porunov container should contains all of relations even with same id, if not some tests will fail, see previous travis builds. Unfortunately, version with counter and hashset will not pass some tests

porunov:
> container should contains all of relations even with same id, if not some tests will fail, see previous travis builds. Unfortunately, version with counter and hashset will not pass some tests

@mad This is interesting. We should investigate it. I am just thinking, during the remove operation we remove any relation which has the same id. It is logical because their `equal` method compares only ids. 
I am thinking, if we create 2 relations with the same id, lets call them A and B and then we would like to remove B, then instead of B we will remove A relation. 
So, if it doesn't matter which relation we remove, does it matter which relation we return?

I am just thinking. Either there is a bug in `remove` method right now or there is a bug in tests. Or am I missing something?
Can the relation be added in the array during the same transaction?

mad:
Here all failed tests for Set implementation, from build https://travis-ci.org/JanusGraph/janusgraph/builds/563754398
```
Failures
[ERROR]   JanusGraphIndexTest.testSimpleUpdate:239 expected: <1> but was: <0>
[ERROR]   JanusGraphTest.simpleLogTest:4305 expected: <4> but was: <3>
[ERROR]   JanusGraphTest.simpleLogTest:4452 expected: <4> but was: <0>
[ERROR]   JanusGraphTest.testSelfLoop:1941 expected: <2> but was: <1>
[ERROR]   JanusGraphPartitionGraphTest.testVertexPartitioning:131 expected: <8> but was: <1>
[ERROR]   JanusGraphPartitionGraphTest.testVLabelOnOrderedStorage:450 expected: <true> but was: <false>

Errors
[ERROR]   JanusGraphTest.testConsistencyEnforcement:2518 » NoSuchElement
[ERROR]   JanusGraphTest.testSupportChangeNameOfEdgeAndUpdateConnections:3712 » IllegalArgument
[ERROR]   JanusGraphPartitionGraphTest.testKeyBasedGraphPartitioning:462->JanusGraphPartitionGraphTest.setupGroupClusters:315 » IllegalArgument
[ERROR]   JanusGraphPartitionGraphTest.testPartitionSpreadNoFlushBatch:342->JanusGraphPartitionGraphTest.testPartitionSpread:359->JanusGraphBaseTest.getOnlyVertex:363->JanusGraphBaseTest.getOnlyElement:371->JanusGraphBaseTest.getOnlyElement:375 » NoSuchElement
[ERROR]   JanusGraphPartitionGraphTest.testPartitionSpreadNoFlushNoBatch:347->JanusGraphPartitionGraphTest.testPartitionSpread:355->JanusGraphPartitionGraphTest.setupGroupClusters:315 » IllegalArgument
[ERROR]   JanusGraphPartitionGraphTest.testVertexPartitionOlapBatch:385->JanusGraphPartitionGraphTest.testVertexPartitionOlap:409->JanusGraphBaseTest.getOnlyVertex:363->JanusGraphBaseTest.getOnlyElement:371->JanusGraphBaseTest.getOnlyElement:375 » NoSuchElement
[ERROR]   JanusGraphPartitionGraphTest.testVertexPartitionOlapCluster:390->JanusGraphPartitionGraphTest.testVertexPartitionOlap:409->JanusGraphBaseTest.getOnlyVertex:363->JanusGraphBaseTest.getOnlyElement:371->JanusGraphBaseTest.getOnlyElement:375 » NoSuchElement
[ERROR]   JanusGraphPartitionGraphTest.testVertexPartitionOlapIndividual:395->JanusGraphPartitionGraphTest.testVertexPartitionOlap:405->JanusGraphPartitionGraphTest.setupGroupClusters:315 » IllegalArgument
[ERROR]   JanusGraphEventualGraphTest.testConsistencyModifier:319->JanusGraphBaseTest.getOnlyElement:375 » NoSuchElement
[ERROR]   JanusGraphEventualGraphTest.testTimestampedEdgeUpdates:207 » NoSuchElement
```

All of that errors can be described as 'Container may contain PropertyKey many times where property is same by id/name'

Consider the following cases where multiple property keys with same id/name

1. Add vertex property and drop edge property 
1. Add self loop - one entry for IN second for OUT
1. Remove property from vertex and from edge 
1. Add mutliple cardinality properties

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleBufferAddedRelations.java
```diff
@@ -1,86 +0,0 @@
...
-        return added.isEmpty();
-    }
-
-    private void cleanup() {
-        if (deleted==null || deleted.isEmpty()) return;
-        final Set<InternalRelation> deletedSet = new HashSet<>(deleted);
-        deleted=null;
-        final List<InternalRelation> newlyAdded = new ArrayList<>(added.size()-deletedSet.size()/2);
-        for (InternalRelation r : added) {
-            if (!deletedSet.contains(r)) newlyAdded.add(r);
```
porunov: Looks like it removes all elements with the same `id` during the remove operation.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
```diff
@@ -14,46 +14,109 @@
...
+            relations = new ObjectArrayList<>(1);
+            container.put(relation.longId(), relations);
+        }
+        relations.add(relation);
+        return true;
     }
 
     @Override
     public boolean remove(InternalRelation relation) {
-        return super.remove(relation);
```
porunov: Here it removes only the first element which has the same `id`

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -219,12 +219,12 @@ public void close() {
 
         int concurrencyLevel;
         if (config.isSingleThreaded()) {
-            addedRelations = new SimpleBufferAddedRelations();
```
porunov: `SimpleBufferAddedRelations` and `SimpleAddedRelations` have different implementation of `remove` operation.
`SimpleBufferAddedRelations` - removes all elements with the same id during the remove operation.
`SimpleAddedRelations` - removes only the first one with the same id during the remove operation.

mad:
I checked failed tests with Set implementation only one test failed `testSelfLoop`

Probably only one case needs for array container - self loop. I rewrite QueryExecutor.getNew for handling loop

mad:
Flaky test testReadWideVertexWithManyProperties?

FlorianHockmann:
@mad You should be able to manually restart a job in Travis if you think that it failed because of a flaky test. If it passes the second run, it would be good if you could create an issue for the flaky test so we can work on the flaky test afterwards.

mad:
@FlorianHockmann This test failed in second run. 
```
Caused by: com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.exceptions.BusyPoolException: [localhost/127.0.0.1] Pool is busy (no available connection and the queue has reached its max size 256)))
```
I think it may relate with CQLStoreManager. Current container do not create addition list for iteration and iterate slightly faster that old implementation. That's leads to overflow cql driver queue, default is 256 entries. CQLStoreManager try to execute multiple [async](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java#L499) request without care of size of queue. Probably, it's rare case in production where multiple host used and queries executed faster

I'am not sure how to fix it. Some thoughts

- increase `poolTimeoutMillis` (default 5 sec) and `maxQueueSize` for tests
- just retry if `BusyPoolException` exception occurred

FlorianHockmann:
> Probably, it's rare case in production where multiple host used and queries executed faster

Are you sure about that? We should ensure that we don't make the CQL backend more instable through this change before we adapt the tests to work around this. Maybe the `CQLStoreManager` also needs to be changed?
I haven't really looked into the code here, so I can't assess whether that's the case or not. If this is really only a problem for the tests, then I'd prefer this option:

> increase poolTimeoutMillis (default 5 sec) and maxQueueSize for tests

as it's only a config change and doesn't require a special retry logic.

porunov:
@mad The test is flaky. It passed on the third run.

mad:
@FlorianHockmann Potentially this exception may produced without these changes. Currently no config for change poolTimeoutMills/maxQueueSize. Can I add these options in this issue or make separate?

@porunov test failed so frequently, i think we need add options for cql driver

mad:
Set based container consume more memory about four times more than List based, but avoid linear scan when change operations produced in same tx. May be make this container implementation configurable?

FlorianHockmann:
> Potentially this exception may produced without these changes. Currently no config for change poolTimeoutMills/maxQueueSize. Can I add these options in this issue or make separate?

@farodin91 mentioned in a conversation we had offline that he ran into a similar problem and was already working on a config change to increase the queue size. @farodin91: Could you check whether it's really the same problem and then coordinate where this change should be made?
From my side, it can be made in this PR or with a separate one that we merge first. Whatever is easier for you.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
```diff
@@ -14,46 +14,74 @@
...
     public SimpleAddedRelations() {
-        super(INITIAL_ADDED_SIZE);
+        this.container = new ObjectHashSet<>(INITIAL_ADDED_SIZE);
     }
 
     @Override
     public boolean add(InternalRelation relation) {
-        return super.add(relation);
+        container.add(relation);
+        return true;
```
porunov: Why it returns always `true`?
Should it be `return container.add(relation);`?

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
```diff
@@ -14,46 +14,74 @@
...
     }
 
     @Override
     public boolean remove(InternalRelation relation) {
-        return super.remove(relation);
+        container.removeAll(relation);
+        return true;
```
porunov: Why `removeAll` instead of `remove`?
Why it returns `true`? Should it be `return remove(relation)`?

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
```diff
@@ -14,46 +14,74 @@
...
 import java.util.Collection;
-import java.util.List;
+import java.util.Collections;
+import java.util.Iterator;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
  */
 
-public class SimpleAddedRelations extends ArrayList<InternalRelation> implements AddedRelationsContainer {
```
porunov: I didn't check the logic much but it looks like it can be simplified to:
```
public class SimpleAddedRelations extends ObjectHashSet<InternalRelation> implements AddedRelationsContainer {
    private static final int INITIAL_ADDED_SIZE = 10;

    public SimpleAddedRelations() {
        super(INITIAL_ADDED_SIZE);
    }

    @Override
    public Iterable<InternalRelation> getView(Predicate<InternalRelation> filter) {
        return Iterables.filter(this::unwrappedIterator, filter);
    }

    @Override
    public Collection<InternalRelation> getAll() {
        return Collections.unmodifiableCollection(new AbstractCollection<InternalRelation>() {
            @Override
            @Nonnull
            public Iterator<InternalRelation> iterator() {
                return SimpleAddedRelations.this.unwrappedIterator();
            }

            @Override
            public int size() {
                return SimpleAddedRelations.this.size();
            }
        });
    }

    private Iterator<InternalRelation> unwrappedIterator() {
        
        return Iterators.transform(iterator(), e -> {
            assert e != null;
            return e.value;
        });
    }
}
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
mad: This is set based container, and if element already exists it return false but client interpret it as addition failed

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
mad: `removeAll` is alias for `remove` for set based, so I fix it

porunov: Thanks for explanation. I see it. Not necessary but you can extend `ObjectHashSet` class and overwrite necessary methods instead of making a proxy for `ObjectSet` (something like here https://github.com/JanusGraph/janusgraph/pull/1701#discussion_r323361562). Then you will have `remove` method and you won't need to overwrite `isEmpty` method.
For example:
```
@Override
public boolean remove(InternalRelation relation) {
  super.remove(relation);
  return true;
}
```

This is up to you. You may leave it as a proxy or make it as an extended class.

porunov: I now think that proxy might be a better option. Right now we can safely extend `ObjectHashSet` because it uses `this.remove(..)` inside `removeAll(..)` method but if it changes the implementation for some reason to `remove(..)` instead of `this.remove(..)` we might have problems with such implementation because we overwrite `remove(..)` method of `ObjectHashSet`.
So, that said, I think your current implementation is good because it decreases potential future bugs. I now think that we should not extend `ObjectHashSet` as I suggested earlier.

farodin91:
> @farodin91 mentioned in a conversation we had offline that he ran into a similar problem and was already working on a config change to increase the queue size. @farodin91: Could you check whether it's really the same problem and then coordinate where this change should be made?

My problem seems to be a different problem.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/SimpleAddedRelations.java
```diff
@@ -14,46 +14,74 @@
...
-
+public class SimpleAddedRelations implements AddedRelationsContainer {
     private static final int INITIAL_ADDED_SIZE = 10;
+    private final ObjectSet<InternalRelation> container;
```
porunov: We can use `ObjectHashSet` here instead of `ObjectSet`. Then we would have a direct access to `remove` method instead of `removeAll`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1124,24 +1124,50 @@ public void executeMultiQuery(final Collection<InternalVertex> vertices, final S
...
+                        if (loop != null) {
+                            InternalRelation loop = this.loop;
+                            this.loop = null;
+                            return loop;
+                        }
+                        while (iterator.hasNext()) {
+                            InternalRelation next = iterator.next();
+                            if (query.matches(next)) {
+                                if (query.getDirection() == Direction.BOTH && next instanceof JanusGraphEdge && next.isLoop()) {
+                                    loop = next;
```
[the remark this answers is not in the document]
mad: Old container had two entities for loop case, so just skip one if query request IN/OUT (not BOTH)
New container has only one entities for loop case, so just duplicate if query request BOTH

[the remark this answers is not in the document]
porunov: Thank you @mad for the explanation!

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1124,24 +1124,50 @@ public void executeMultiQuery(final Collection<InternalVertex> vertices, final S
...
-                    public boolean apply(final InternalRelation relation) {
-                        if ((relation instanceof JanusGraphEdge) && relation.isLoop()
-                                && query.getDirection() != Direction.BOTH) {
-                            if (relation.equals(previous))
-                                return false;
+                    public boolean hasNext() {
+                        if (current == null) {
+                            current = nextRelation();
+                        }
+                        return current != null;
```
porunov: (nitpick) we need to check for `current != null` only if it was `null` before calling this method. I.e. I would rewrite this method as:
```
@Override
public boolean hasNext() {
    if (current == null) {
        current = nextRelation();
        return current != null;
    }
    return true;
}
```

porunov:
Tested this PR with personal integration tests. All tests passed.

porunov:
Tested all TinkerPop tests on current master branch with applying this PR on it. All tests passed (using `mvn clean install -Dtest.skip.tp=false -DskipTests=true`).

@mad I think this PR is ready to be merged. What do you think? Will you merge it?

PR #1704, opened 2019-07-26
Make mapping support configurable for ES7 and drop it for future ES v…

### LINKED ISSUE #1659: Do not use mapping types for future ElasticSearch versions
Mapping is deprecated and disabled by default in ElasticSearch 7. Mapping support will be completely removed in ES 8. We need to prepare possibility to work without mappings for ES 8.

If we drop mapping types for ES 7 then it will be easier for users to upgrade their JanusGraph indexes from ES 7 to ES 8 but will require reindexing from ES 6 to ES 7. If we drop support for mapping types in ES 8 but save them in ES 7 then it will be easier for users to upgrade their JanusGraph indexes from ES 6 to ES 7 but will require reindexing from ES 7 to ES 8.

Maybe it is better to add an optional parameter to enable / disable mappings for ES 7 in JanusGraph.

porunov:
…ersions

- Added possibility to configure mapping types usage on ElasticSearch 7
- Dropped mapping types usage for ElasticSearch > 7
- Small performance improvement in bulk api

Fixes #1659

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/docs/upgrade.adoc
+++ b/docs/upgrade.adoc
@@ -10,6 +10,7 @@ JanusGraph release.
 ==== ElasticSearch: Upgrade from 6.6.0 to 7.2.0 and drop support for 5.x version
 The ElasticSearch version has been changed to 7.2.0 which removes support for `max-retry-timeout` option. That is why this option no longer available in JanusGraph.
 Users should be aware that by default JanusGraph setups maximum open scroll contexts to maximum value of `2147483647` with the parameter `setup-max-open-scroll-contexts` for ElasticSearch 7.y. This option can be disabled and updated manually in ElasticSearch but you should be aware that ElasticSearch starting from version 7 has a default limit of 500 opened contexts which most likely be reached by the normal usage of JanusGraph with ElasticSearch.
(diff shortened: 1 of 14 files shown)


porunov:
@FlorianHockmann could you please disable Codacy `Use explicit scoping instead of the default package private level` check? It isn't right. Package private level is stickier than `protected`.

porunov:
@FlorianHockmann I have disabled `Use explicit scoping instead of the default package private level` check in Codacy. Also, I don't know why, but we are using `PMD (Legacy) 5.8.1` instead of `PMD 6.14.0` in Codacy patterns. Is there any reason why we use the older version?

FlorianHockmann:
> Also, I don't know why, but we are using PMD (Legacy) 5.8.1 instead of PMD 6.14.0 in Codacy patterns. Is there any reason why we use the older version?

I haven't configured anything about that so it must be the default setting in Codacy. If you think that the newer version makes more sense, just go ahead and change that.

porunov:

Issue #1249, opened 2018-09-07
Improve CQL driver performance


chinhuang007:
In our performance tests for insert vertices/edges, and queries with Cassandra v3.11.3, the CQL driver is slower than the deprecated astyanax/thrift driver. It would be great to look into ways to improve CQL performance since it seems the recommended driver moving forward.

marco-scoppetta:
Hey @chinhuang007 any updates on this statement? What's the volume of your benchmark (roughly how many nodes and vertices), are all the queries consistently slower?

It would be nice to have more info about this as we're also planning to move from Astyanax to CQL.

Thanks

PR #1708, opened 2019-07-29
Optimise open and close graph using CQL driver

marco-scoppetta:
When switching from Thrift to CQL driver we noticed how all our tests started timing out,
after some digging we noticed that `storeManager` (which physically connects to the Cassandra backend) is initialised twice every time a new graph is created.

In this PR we change the behaviour of instantiation of a graph: store manager only gets created once and used to read global configuration and then used to create a `Backend` which is in turn used by the new Graph.

List of changes contained in this PR:

- `ConfiguredGraphFactory` and `JanusGraphFactory` now instantiate both `StoreManager` and `Backend` which are then injected in the components that need them.

- `static JanusGraph open(ReadConfiguration configuration, String backupName)` is now a *private* method as the associated Javadoc states: "This method shouldn't be called by end users;"

- `GraphDatabaseConfiguration` is no longer used as a `Backend` factory, instead the Graph object is used to fetch the associated Backend

- `GraphDatabaseConfigurationBuilder` is now a static class, as it does not hold any state

- `CQLStoreManager` uses the native `closeAsync` instead of the synchronous `close`, this saves at least 2 seconds every time a graph gets closed

- updated tests

Possibly related to this: https://github.com/JanusGraph/janusgraph/issues/1249

Signed-off-by: Marco Scoppetta <marco@grakn.ai>

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/berkeleyje/BerkeleyGraphTest.java
+++ b/janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/berkeleyje/BerkeleyGraphTest.java
@@ -66,7 +66,7 @@ public class BerkeleyGraphTest extends JanusGraphTest {
         config.set(ConfigElement.getPath(GraphDatabaseConfiguration.DROP_ON_CLEAR), true);
         Backend backend = getBackend(config, false);
         assertTrue(backend.getStoreManager().exists(), "graph should exist before clearing storage");
-        clearGraph(config);
+        backend.clearStorage();
         backend.close();
         backend = getBackend(config, false);
         assertFalse(backend.getStoreManager().exists(), "graph should not exist after clearing storage");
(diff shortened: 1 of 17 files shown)


farodin91:
I had fight against this problem during switching all cql tests to testcontainers. I fixed it with following code: https://github.com/JanusGraph/janusgraph/blob/b878f94a5a1c4a6fb4bf7acbe97d7fd837f19bcc/janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java#L135 which was already part of old logic.

marco-scoppetta:
Hey @farodin91,

I am not sure what you mean, how does `setWrapperStoreManager` relate to this PR?
I probably don't have enough context

farodin91:
@marco-scoppetta These two seconds stop time can be solved in tests with the following class https://github.com/JanusGraph/janusgraph/blob/b878f94a5a1c4a6fb4bf7acbe97d7fd837f19bcc/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CachingCQLStoreManager.java

PR #1710, opened 2019-07-29
Improve vertex label lookup

mad:

-----

### For all changes:

### For code changes:

### Note:


mad:
Root cause is it added relations container

We use multiples EventListener in our traversal that do some work on newly added/changed vertices. When we add multuple vertices with properties and trying to access to label in same tx from listeners it lead to sequntial scan added relations container this apect is inefficient.

Test `JanusGraphPerformanceMemoryTest.labelRetrivial` demonstrte this case

We see same issue when we want to rewrite properties in same tx. Rewrite properties required remove from added relation container is it sequntial scan operations, see https://github.com/JanusGraph/janusgraph/pull/1701

For avoid this issues we may avoid this pattern of use OR add label cache, use more complex structure for added relations container

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
```diff
@@ -85,9 +86,12 @@
     private boolean restrict2Partitions = true;
 
 
+    private final BaseVertexCentricQuery queryForLabel;
+
     public BasicVertexCentricQueryBuilder(final StandardJanusGraphTx tx) {
         super(tx);
         this.tx = Preconditions.checkNotNull(tx);
+        this.queryForLabel = queryForLabel();
```
FlorianHockmann: Do we really want to construct this query already in the constructor and not simply when it's used for the first time?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
```diff
@@ -18,12 +18,16 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import java.util.Random;
+import java.util.*;
```
FlorianHockmann: Please don't use wildcard imports (per convention).

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
```diff
@@ -18,12 +18,16 @@
...
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.commons.math.stat.descriptive.SummaryStatistics;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+import org.janusgraph.core.EdgeLabel;
```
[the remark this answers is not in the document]
FlorianHockmann: As Codacy already says, these imports are unused.

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
```diff
@@ -45,6 +50,35 @@
...
+        // Try access to label in same tx
+        List<Vertex> vertices = graph.traversal()
+                                     .V()
+                                     .toList();
+        long start = System.currentTimeMillis();
+        for (int i = 0; i < 100; i++) {
+            vertices.forEach(v -> v.label());
+        }
+        graph.tx().rollback();
+        System.out.println(System.currentTimeMillis() - start);
```
FlorianHockmann: Why a `println` in a test?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
```diff
@@ -45,6 +50,35 @@
     @Rule
     public TestRule benchmark = JUnitBenchmarkProvider.get();
 
+    @Test
+    void labelRetrivial() {
```
FlorianHockmann: What is this test testing? It has no asserts  and it probably also passes without your changes. Did you only use it during development?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
```diff
@@ -123,20 +158,20 @@ public void testTransactionalMemory() throws Exception {
...
+                    JanusGraphVertex v = getVertex(tx, "uid", random.nextInt(maxUID) + 1);
                     assertCount(2, v.properties());
                     int count = 0;
                     for (Object e : v.query().direction(Direction.BOTH).edges()) {
                         count++;
                         assertTrue(((JanusGraphEdge) e).<Integer>value("time") >= 0);
                     }
                     assertTrue(count <= 2);
-//                        if (t%(trials/10)==0) System.out.println(t);
+                    //                        if (t%(trials/10)==0) System.out.println(t);
```
FlorianHockmann: (nitpick) Why these added spaces?

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
mad: This test demonstrate performance improvement

How these type of test use assertions? One of the way it performance tests must executed in less time than referenced values for latest released version for same env

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
mad: Another tests in this class used `println` for writing some info

Moreover this test has `MEMORY_TESTS` category that activated by memory-test profile, but this profile not activated in travis, so it may be outdated test. I think JG need to rework performance/memory tests for proper checking regression in this space

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
mad: No, not needed

mad:
I think current approach may be simplified, for avoid redundant operation we need use `SimpleVertexQueryProcessor` for this label query. I will try this and test

And  remove new field from `AbstractVertex` that increase vertex size to 8 byte

janusgraph-test/src/main/java/org/janusgraph/graphdb/JanusGraphPerformanceMemoryTest.java
FlorianHockmann: @farodin91 mentioned some time ago that we could use a dedicated library for these kind of performance tests, but I think that we don't have anything like that right now.

I'm not sure what we want to do until then with tests like these, but since there are already similar tests in this class as you pointed out, we can also add this from my side. Then we can later hopefully easily port it to a framework where a certain performance threshold is enforced.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
```diff
@@ -414,10 +417,25 @@ protected BaseVertexCentricQuery constructQuery(RelationCategory returnType) {
         return query;
     }
 
+    protected BaseVertexCentricQuery queryForLabel() {
+        if (queryForLabel != null) {
```
porunov: When `queryForLabel` will not be `null`?

janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/AbstractVertex.java
```diff
@@ -122,14 +123,18 @@ public String label() {
...
 
     @Override
     public VertexLabel vertexLabel() {
-        Vertex label = getVertexLabelInternal();
-        if (label==null) return BaseVertexLabel.DEFAULT_VERTEXLABEL;
-        else return (VertexLabelVertex)label;
+        if (vertexLabel == null) {
+            vertexLabel = getVertexLabelInternal();
+        }
+        if (vertexLabel == null) return BaseVertexLabel.DEFAULT_VERTEXLABEL;
```
porunov: I think we could just use the next logic:
```
if (vertexLabel == null) {
  vertexLabel = (VertexLabelVertex) getVertexLabelInternal();
  if (vertexLabel == null){
    vertexLabel = BaseVertexLabel.DEFAULT_VERTEXLABEL;
  }
}
return vertexLabel;
```

mad:
Current approach introduced new field to `AbstractVerex`, so vertex size is increased. I don't think it will good for some use cases. 

Need more investigation. May be some query cache will help with this type of issue https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java#L306

li-boxuan:
@mad Is this ready for review? If so, can you rebase it? Otherwise, we can convert it to a draft.

Issue #1719, opened 2019-08-05
Use EntryArrayList instead of StaticArrayEntryList


kptfh:
Consider use EntryArrayList instead of StaticArrayEntryList to avoid data buffers copying in SliceQuery.getSubset().
StaticArrayEntryList copies all data buffers from elements to internal data buffer. EntryArrayList instead keeps elements as they are.

mad:
Hi @kptfh 

But this approach may consume more memory because StaticArrayEntryList backed by primitive arrays, but EntryArrayList backed by object array, sometimes it may critical. Approx 16 times more

porunov:
Fixed in #2084

PR #1720, opened 2019-08-05
EntryArrayList instead of StaticArrayEntryList

kptfh:
https://github.com/JanusGraph/janusgraph/issues/1719


FlorianHockmann:
This seems to introduce a problem as [a CQL job is now timing out on Travis](https://travis-ci.org/JanusGraph/janusgraph/jobs/568299150).

PR #1722, opened 2019-08-07
Use stored ES scripts instead of inline scripts

### LINKED ISSUE #1703: Use stored scripts for ElasticSearch instead of source scripts
I am proposing to use stored scripts instead of `source` scripts. We will be able to pass a small `id` instead of a big `source` for script usage.
It will decrease bandwidth and possibly performance on ES side.

I didn't think how to make it yet. The thing is that the script may change depending of JanusGraph version. Currently I think of the next possible solutions:
1. We could use unique id which increments each time the script is updated. When the script is updated we will need to increment this id. When JanusGraph instance is started with ElasticSearch indexes, we are checking if the script exists with the necessary id. If it isn't, we store the script. Pros: we update the script only when it is needed to be updated. Cons: the programmer may forgot to update the script version.
2. We can use JanusGraph version for scripts versions. The pros: we don't need to manually update the version for script. Cons: even if the script is not changed in a newer JanusGraph version, we will still store a new script for each JanusGraph version. 
3. We may use a single unique id for the script and when JanusGraph instance is up we will check if the script is the same as the script in JanusGraph. If they differ, we will update the script. Pros: we always store only 1 stored script for a necessary operation. Cons: it may not work properly if ElasticSearch is used by multiple JanusGraph instances with different versions.

I think we could use option 3 but use index prefix for the script id also which will eliminate its cons. Something like `<index_prefix>+"-s-"+"script_unique_id"`. So by default we would use script ids something like: `janusgraph-s-1` for addition script, `janusgraph-s-2` for deletion script.

porunov:
This is generally very small improvement but it is still an improvement.

Instead of passing the next 257 symbols:
```
"lang":"painless","source":"for (field in params.fields) {if (ctx._source[field.name] == null) {ctx._source[field.name] = [];}if (field.cardinality != 'SET' || ctx._source[field.name].indexOf(field.value) == -1) {ctx._source[field.name].add(field.value);}}"
```
We will pass 21 symbols by default during the addition operation:
```
"id":"janusgraph-add"
```

Instead of passing the next 352 symbols:
```
"lang":"painless","source":"for (field in params.fields) {if (field.cardinality == 'SINGLE') {ctx._source.remove(field.name);} else if (ctx._source.containsKey(field.name)) {def fieldIndex = ctx._source[field.name].indexOf(field.value);if (fieldIndex >= 0 && fieldIndex < ctx._source[field.name].size()) {ctx._source[field.name].remove(fieldIndex);}}}"
```
We will pass 21 symbols by default during the deletion operation:
```
"id":"janusgraph-del"
```

Internally tested with ES 6.8 and ES 7.2

Fixes #1703

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchClient.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchClient.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.janusgraph.diskstorage.es.mapping.IndexMapping;
+import org.janusgraph.diskstorage.es.script.ESScriptResponse;
 
 public interface ElasticSearchClient extends Closeable {
 
@@ -32,6 +33,10 @@ public interface ElasticSearchClient extends Closeable {
     boolean isIndex(String indexName);
 
     boolean isAlias(String aliasName);
+
+    void createStoredScript(String scriptName, Map<String,Object> script) throws IOException;
+
+    ESScriptResponse getStoredScript(String scriptName) throws IOException;
 
     void createIndex(String indexName, Map<String,Object> settings) throws IOException;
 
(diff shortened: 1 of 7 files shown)


porunov:

PR #1725, opened 2019-08-07
Close async CQLStoreManager

marco-scoppetta:
When switching from Thrift to CQL driver we noticed how all our tests started timing out,
after some digging we found the reason to be `manager.close()` which is invoked both when building [GraphDatabaseConfiguration](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/backend/builder/KCVSConfigurationBuilder.java#L46) and when [closing a graph](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java#L520).

When using CQL driver, therefore using CQLStoreManager, the time needed to close the cluster is around 2 seconds, whereas using the native closeAsync we can close the store in a few milliseconds.

Graph creation time without async: ~ 4 seconds
Graph creation time with async: usually less than 2 seconds

This might also mean that there is no more need to use `CachingCQLStoreManager` in tests I guess.

Signed-off-by: Marco Scoppetta <marco@grakn.ai>

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
@@ -374,7 +374,7 @@ public class CQLStoreManager extends DistributedStoreManager implements KeyColum
             this.session.close();
         } finally {
             try {
-                this.cluster.close();
+                this.cluster.closeAsync();
             } finally {
                 this.executorService.shutdownNow();
             }


farodin91:
@marco-scoppetta I think that could problematic, if the async closing fails, we didn't get an exception.

marco-scoppetta:
@farodin91 I agree and I am also not a big fan of this approach.
But this seems to me the only viable solution if we don't want to rewrite a fairly big piece of the architecture or put some runtime overriding hack in place.

As I mentioned before, without this change all of our tests (at Grakn) will timeout and when using our application the impact is pretty noticeable as we often open and close graphs.

Therefore missing a possible `RuntimeException` at closing time seems the lesser of all evils, because if the Cluster is corrupted and for some reason fails the closure it will very likely fail all the other operations.

I would like to find another solution but I have no other ideas, any suggestions?

PR #1731, opened 2019-08-15
Fix GhostVertexRemover

mad:
Fix for #1749

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
@@ -110,9 +110,16 @@ public class GhostVertexRemover extends VertexJobConverter {
             iterator.remove();
             removedRelations++;
         }
+        v.remove();
         //There should be no more system relations to remove
         metrics.incrementCustom(REMOVED_VERTEX_COUNT);
         metrics.incrementCustom(REMOVED_RELATION_COUNT,removedRelations);
+    }
+
+    @Override
+    public void workerIterationEnd(final ScanMetrics metrics) {
+        tx.commit();
+        super.workerIterationEnd(metrics);
     }
 
     @Override
--- a/janusgraph-test/src/main/java/org/janusgraph/olap/OLAPTest.java
+++ b/janusgraph-test/src/main/java/org/janusgraph/olap/OLAPTest.java
@@ -195,9 +195,15 @@ public abstract class OLAPTest extends JanusGraphBaseTest {
         mgmt.commit();
 
         ScanMetrics result = executeScanJob(new GhostVertexRemover(graph));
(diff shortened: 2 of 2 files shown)


porunov:
@mad Due to the problem with test containers, I had to revert the latest commit. Please exclude 2 additional commits from your PR

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
```diff
@@ -110,11 +110,18 @@ public void process(StaticBuffer key, Map<SliceQuery, EntryList> entries, ScanMe
...
         }
+        v.remove();
         //There should be no more system relations to remove
         metrics.incrementCustom(REMOVED_VERTEX_COUNT);
         metrics.incrementCustom(REMOVED_RELATION_COUNT,removedRelations);
     }
 
+    @Override
+    public void workerIterationEnd(final ScanMetrics metrics) {
+        tx.commit();
```
FlorianHockmann: Wouldn't it be better to commit in the `process` method? That way, the commit is closer to the mutation and I think that it also makes more sense to have those in the same method.
The documentation for the base method also says:

> Can be used to close any resources held by this job.

which means that I wouldn't expect a commit here.

janusgraph-test/src/main/java/org/janusgraph/olap/OLAPTest.java
```diff
@@ -198,6 +198,12 @@ public void removeGhostVertices() throws Exception {
         assertEquals(1,result.getCustom(GhostVertexRemover.REMOVED_VERTEX_COUNT));
         assertEquals(2,result.getCustom(GhostVertexRemover.REMOVED_RELATION_COUNT));
         assertEquals(0,result.getCustom(GhostVertexRemover.SKIPPED_GHOST_LIMIT_COUNT));
+
+        // Second scan should not find any ghost vertices
+        result = executeScanJob(new GhostVertexRemover(graph));
+        assertEquals(0,result.getCustom(GhostVertexRemover.REMOVED_VERTEX_COUNT));
```
FlorianHockmann: (nitpick) Please add a space after the `,`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
mad: Process invoked multiple times for same tx so if we close transaction in intermediate state second process failed.

In the same doc:

> This method is semantically aligned with {@link org.apache.tinkerpop.gremlin.process.computer.VertexProgram#workerIterationEnd(Memory)}

And `workIterationEnd` has implementation that commit tx `BulkLoaderVertexProgram`

PR #1739, opened 2019-08-16
Replace thrift index tests with CQL

### LINKED ISSUE #1736: Index should run against CQL instead of Cassandra-Thrift
In order to remove Cassandra, we have to replace all places which uses thrift with CQL.

Therefor, solr+thrift and elasticsearch+thrift need to be replaced with a CQL combination.

Part of #1697

farodin91:
Fixes #1736

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/CQLElasticsearchTest.java
+++ b/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/CQLElasticsearchTest.java
@@ -14,33 +14,31 @@
 
 package org.janusgraph.diskstorage.es;
 
-import org.janusgraph.CassandraStorageSetup;
+import org.janusgraph.JanusGraphCassandraContainer;
 import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
 import org.janusgraph.diskstorage.configuration.WriteConfiguration;
 import org.janusgraph.graphdb.JanusGraphIndexTest;
 import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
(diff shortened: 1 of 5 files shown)


farodin91:
I'm okay with activate the full build, but it won't make any since we are not including solr and es test into our full build. 

We can do it in the future. For example during the step of dropping thrift support.

porunov:
@farodin91 Due to the problem with test containers, I had to revert the latest commit. Please exclude additional commits from your PR

farodin91:
@porunov Do like to review this pr?

janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/CQLElasticsearchTest.java
```diff
@@ -59,4 +57,8 @@ protected boolean supportsCollections() {
         return true;
     }
 
+    @Override
+    @Disabled("CQL seems to not clear storage correctly")
```
porunov: Do we have a bug with CQL so that it cannot clear storage right now or is it a normal behavior? Should we open an issue to track it?

farodin91: It seems to be an issue which only exists in combination with elasticsearch. Could be a timing problem of the test? We should open an issue. I think clearing should not be a blocker which basically only used in tests.

porunov:
The full build passed for this PR: https://travis-ci.org/porunov/janusgraph/builds/574742051
I think it can be merged

PR #1784, opened 2019-09-04
Add index lifecycle documentation [doc only]

### LINKED ISSUE #1780: Add documentation page on index states
The are often troubles with creating indexes by new JanusGraph users. Often they skip `REINDEX` or `ENABLE` step and facing the problem that the index doesn't work.

I would suggest to move [Index states page](https://github.com/JanusGraph/janusgraph/wiki/Indexing) created by @robertdale to our current documentation.

porunov:
Fixes #1780

-----

### For all changes:

### For code changes:

### Note:


docs/basics/index-lifecycle.md
```diff
@@ -0,0 +1,39 @@
...
+
+JanusGraph uses only indexes which have status `ENABLED`. 
+When the index is created it will not be used by JanusGraph until it is enabled. 
+After the index is build you should wait until it is registered (i.e. available) by JanusGraph:
+```java
+//Wait for the index to become available (i.e. wait for status REGISTERED)
+ManagementSystem.awaitGraphIndexStatus(graph, "nameAndAgeIndex").call();
+```
+
+After the index is registered we should either enable the index
```
FlorianHockmann: Wouldn't it also be good to show how the index can be enabled without reindexing the data?

docs/basics/index-lifecycle.md
```diff
@@ -0,0 +1,39 @@
...
+mgmt.updateIndex(mgmt.getGraphIndex("nameAndAgeIndex"), SchemaAction.REINDEX).get();
+mgmt.commit();
+```
+
+Index states and transitions
+-----------
+
+![States and transitions](index-lifecycle.png)
+
+## States (SchemaStatus)
```
FlorianHockmann: I think it would be good to have a short sentence as an introduction for this enumeration, maybe something like: _An index can be in one of the following states:_

docs/basics/index-lifecycle.md
```diff
@@ -0,0 +1,39 @@
...
+* **INSTALLED** The index is installed in the system but not yet registered with all instances in the cluster
+* **REGISTERED** The index is registered with all instances in the cluster but not (yet) enabled
+* **ENABLED** The index is enabled and in use
+* **DISABLED** The index is disabled and no longer in use
+
+## Actions (SchemaAction)
+* **REGISTER_INDEX** Registers the index with all instances in the graph cluster. After an index is installed, it must be registered with all graph instances
```
FlorianHockmann: Same as above, I would add a short sentence like _The following actions can be performed on an index to change its state via `mgmt.updateIndex()`:_

porunov:
Thank you @FlorianHockmann , I've added those notes and an example

FlorianHockmann:
Really nice to see that the documentation is directly live now thanks to the new docs system: https://docs.janusgraph.org/basics/index-lifecycle/

One note though: I think it's a bit confusing that the two example operations - enabling of the index and reindexing - are shown in the same listing, although users will only want to do one of them. I usually interpret a listing as something that should be executed together as a single unit.

porunov:
@FlorianHockmann I also like the new documentation system very much.

>> One note though: I think it's a bit confusing that the two example operations - enabling of the index and reindexing - are shown in the same listing, although users will only want to do one of them. I usually interpret a listing as something that should be executed together as a single unit.

I didn't think about it. You are right. Could you provide a commit to split those examples? Or tell how they should look. Should it be something like the following example?

---

After the index is registered we should either enable the index 
(if we are sure that the current data should not be indexed by the newly created index) or we should reindex current data so that it would be available in the newly created index.

Reindex and automatically enable the index example:
```java
mgmt = graph.openManagement();
//Reindex the existing data and enable the index
mgmt.updateIndex(mgmt.getGraphIndex("myIndex"), SchemaAction.REINDEX).get();
mgmt.commit();
```
Enable the index without reindexing existing data example:
```java
mgmt = graph.openManagement();
//Enable the index without reindexing existing data 
mgmt.updateIndex(mgmt.getGraphIndex("myAnotherIndex"), SchemaAction.ENABLE_INDEX).get();
mgmt.commit();
```

FlorianHockmann:
> Should it be something like the following example?

Yes, that's how I would do it (only without the code comments as they are unnecessary with the added sentence before the listing).

porunov:
Okay, I will push commit with CTR

Issue #460, opened 2017-08-13
Improve Index Documentation


robertdale:
- explain different states
- explain that there are states on each key of an index and the index as a whole
- explain state transitions (how to get from one to the next)
- explain index requirements
- troubleshooting

dpitera:
For troubleshooting, ensure to look at [this SO post](https://stackoverflow.com/questions/40585417/titan-db-ignoring-index/40591478#40591478) where I've made notes of all potential issues I've run into for having indexes get stuck in the `installed` or `registered` states.

robertdale:
**States:**
- INSTALLED - The index is installed in the system but not yet registered with all instances in the cluster
- REGISTERED - The index is registered with all instances in the cluster but not (yet) enabled
- ENABLED - The index is enabled and in use
- DISABLED - The index is disabled and no longer in use

**Actions:**
- REGISTER_INDEX - Registers the index with all instances in the graph cluster. After an index is installed, it must be registered with all graph instances
- REINDEX - Re-builds the index from the graph
- ENABLE_INDEX - Enables the index so that it can be used by the query processing engine. An index must be registered before it can be enabled.
- DISABLE_INDEX - Disables the index in the graph so that it is no longer used.
- REMOVE_INDEX - Removes the index from the graph (optional operation). Only on composite index.

dpitera:
Also note that a REINDEX automatically ENABLES an index.

robertdale:
I made a diagram. Not sure where to put it.

sjudeng:
Wiki? https://github.com/JanusGraph/janusgraph/wiki

robertdale:
I still had to upload the image to somewhere other than here.  Anyway, you can see it here - https://github.com/JanusGraph/janusgraph/wiki/Indexes

pluradj:
I came looking for this issue because of the nice writeup from @robertdale on [janusgraph-users](https://groups.google.com/d/msg/janusgraph-users/KOn6hso1hJM/pQV5xzYxBAAJ), which would be a nice addition to the indexing docs.

I didn't know some docs were going up on a wiki. Why not add that image to the [asciidoc](https://github.com/JanusGraph/janusgraph/blob/v0.1.1/docs/basics.txt#L738)?

farodin91:
Link #1780

PR #1801, opened 2019-09-19
Make index management more reachable [doc only]

farodin91:
Part of #460

-----

### For all changes:

### Note:


farodin91:
Preview https://farodin91.github.io/janusgraph/

docs/index-management/index-reindexing.md
```diff
@@ -0,0 +1,215 @@
+# Index Reindexing
```
FlorianHockmann: _Index Reindexing_ sounds strange. Why not just _Reindexing_?

docs/index-management/index-common-problems.md
```diff
@@ -0,0 +1,102 @@
+# Common Problems with Index Management
+
+## IllegalArgumentException when starting job
+
+When a reindexing job is started shortly after a the index has been
```
FlorianHockmann: These problems are specific to reindexing and index removal, but they now stand independent of that which is confusing in my opinion.
I'm not sure if it's really a good idea for that reason to split the current _Index Management_ page into three different pages as they are about the same topic, namely managing already existing indices.

docs/index-management/index-reindexing.md
farodin91: If you only write reindexing, what do we want to reindex, e.g. a file system.

docs/index-management/index-common-problems.md
farodin91: It currently only contains reindex and removal part. I see a lot of question to all of these sections.

The hole section is called Index Management. Index management should also contain creation.

PR #1804, opened 2019-09-19
retry_on_conflict for mixed index

### LINKED ISSUE #1797: retry_on_conflict for mixed index
retry_on_conflict for mixed index
i have already fixed by overriding RestElasticSearchClient, 

can can find more details here :  https://groups.google.com/forum/#!searchin/janusgraph-users/retry_on_conflict%7Csort:date/janusgraph-users/ixIRJVQTqto/erbBXbUTAwAJ

nareshbabuy:
fixes #1797 

added retry_on_conflict in bulkRequest to avoid version conflict error while updating Elastic search document


farodin91:
@nareshbabuy You have to sign you all commits using git commit -s to pass the DCO check. Besides, you have to sign our CLA. It seems so that some tests are failling now.

nareshbabuy:
Thanks for suggestion, all steps completed, but still tests are failing. 
test are passing for -Pelasticsearch7, and -Pelasticsearch6 but failing for -Pelasticsearch60 profile
any suggestion on this?

nareshbabuy:
Done. Thanks

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -356,17 +358,23 @@ public void deleteIndex(String indexName) throws IOException {
...
             } else {
-                requestData = ImmutableMap.of(
+                requestData.putAll(ImmutableMap.of(
                     "_index", request.getIndex(),
-                    "_id", request.getId());
+                    "_id", request.getId()));
+            }
+
+            if (request.getRequestType() == ElasticSearchMutation.RequestType.UPDATE) {
+                final String retryOnConflictKey = majorVersion.getValue() >= 7 ? "retry_on_conflict" : "_retry_on_conflict";
```
porunov: Why you use `_retry_on_conflict` parameter for elasticsearch 6 here? Shouldn't it be `retry_on_conflict` (i.e. same as for ES 7 version)?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -245,6 +245,10 @@
                 "this parameter to true but we do recommend to reindex your indexes and don't use this parameter.",
             ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<String> RETRY_ON_CONFLICT =
```
porunov: I would use `Integer` here also because no other value is possible.

Also, not necessary but I think maybe it is better to have no default value here? I.e. use ES default value?
I would use `null` as a default value here and then would use `retry_on_conflict` in the `RestElasticSearchClient` only if it isn't null. For example:
```
if (retryOnConflict != null && request.getRequestType() == ElasticSearchMutation.RequestType.UPDATE) {
  requestData.putAll(ImmutableMap.of(retryOnConflictKey, retryOnConflict));
}
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
porunov: Also, I suggest to make `retryOnConflictKey` as a global variable and assign value to it in the constructor if there is a different `retryOnConflictKey` between different ES versions.

porunov:
@farodin91 Could you please check why documentation check fails? https://travis-ci.org/JanusGraph/janusgraph/jobs/588523268
I cannot figure it out.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -108,6 +108,8 @@
 
     private final boolean esVersion7;
 
+    private String retryOnConflict;
```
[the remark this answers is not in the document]
nareshbabuy: updated

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
nareshbabuy: ES 6.0 supports only _retry_on_conflict but ES7 has blocking change it will support only retry_on_conflict

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
nareshbabuy: cannot keep null as default value, as we have null check in ConfigOption.java, hence lets use 0 as default value. any other suggestions are welcome.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
nareshbabuy: Agreed, changes done accordingly.

nareshbabuy:
> @farodin91 Could you please check why documentation check fails? https://travis-ci.org/JanusGraph/janusgraph/jobs/588523268
> I cannot figure it out.

i think i need to update config(cfg file) for retry_on_conflict. update now.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -356,17 +361,22 @@ public void deleteIndex(String indexName) throws IOException {
...
             } else {
-                requestData = ImmutableMap.of(
+                requestData.putAll(ImmutableMap.of(
                     "_index", request.getIndex(),
-                    "_id", request.getId());
+                    "_id", request.getId()));
             }
+
+            if (retryOnConflict != null && request.getRequestType() == ElasticSearchMutation.RequestType.UPDATE) {
+                requestData.putAll(ImmutableMap.of(retryOnConflictKey, retryOnConflict));
```
porunov: We can ommit additional map creation here by using 'put' method instead of putAll.
'requestData.put(retryOnConflictKey, retryOnConflict);'

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -108,12 +108,17 @@
...
+
+    private final String retryOnConflictKey;
+
     public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean useMappingTypesForES7) {
         this.delegate = delegate;
         majorVersion = getMajorVersion();
         this.scrollKeepAlive = scrollKeepAlive+"s";
         esVersion7 = ElasticMajorVersion.SEVEN.equals(majorVersion);
         useMappingTypes = majorVersion.getValue() < 7 || (useMappingTypesForES7 && esVersion7);
+        retryOnConflictKey = majorVersion.getValue() >= 7 ? "retry_on_conflict" : "_retry_on_conflict";
```
porunov: I don't see that underscore is needed for elasticsearch 6. Could you point me to ES documentation which shows that? Maybe I am missing something.
Below both documentations shows that the correct parameter is `retry_on_conflict` and not `_retry_on_conflict`.
https://www.elastic.co/guide/en/elasticsearch/reference/6.2/docs-bulk.html
https://www.elastic.co/guide/en/elasticsearch/reference/6.4/docs-update.html

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -245,6 +245,10 @@
                 "this parameter to true but we do recommend to reindex your indexes and don't use this parameter.",
             ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<Integer> RETRY_ON_CONFLICT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry_on_conflict",
+            "Specify how many times should the operation be retried when a conflict occurs. Default: 0.", ConfigOption.Type.MASKABLE, 0);
```
porunov: As I understand if the user don't specift this parameter then the default value will not be used. So, I would remove `Default: 0` from documentation.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -356,17 +361,22 @@ public void deleteIndex(String indexName) throws IOException {
     public void bulkRequest(List<ElasticSearchMutation> requests, String ingestPipeline) throws IOException {
         final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         for (final ElasticSearchMutation request : requests) {
-            final Map<String, String> requestData;
-            if(useMappingTypes){
-                requestData = ImmutableMap.of(
+            Map<String, Object> requestData = new HashMap<>();
+            if (useMappingTypes) {
+                requestData.putAll(ImmutableMap.of(
```
porunov: You create additional Immutable map which is redundant.
You can just use `put` method for each entry instead of creating additional map and then using putAll.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -356,17 +361,22 @@ public void deleteIndex(String indexName) throws IOException {
...
                     "_index", request.getIndex(),
                     "_type", request.getType(),
-                    "_id", request.getId());
+                    "_id", request.getId()));
             } else {
-                requestData = ImmutableMap.of(
+                requestData.putAll(ImmutableMap.of(
```
porunov: Same here. Change `putAll` to sequance of `put` calls.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
farodin91: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/docs-bulk.html

nareshbabuy:
@farodin91 , any idea why build is failing for documentation?

farodin91:
@nareshbabuy You have to execute ` mvn --quiet clean install -DskipTests=true -pl janusgraph-doc -am` and check in the changes.

nareshbabuy:
@farodin91 we have it already. i can see logs it executed and exited with 0. Correct me if am wrong

farodin91:
On travis ci the git diff checks fails. It seems files have a different order.

docs/basics/janusgraph-cfg.md
```diff
@@ -116,6 +116,7 @@ Elasticsearch index configuration
 | index.[X].elasticsearch.setup-max-open-scroll-contexts | Whether JanusGraph should setup max_open_scroll_context to maximum value for the cluster or not. | Boolean | true | MASKABLE |
 | index.[X].elasticsearch.use-all-field | Whether JanusGraph should add an "all" field mapping. When enabled field mappings will include a "copy_to" parameter referencing the "all" field. This is supported since Elasticsearch 6.x  and is required when using wildcard fields starting in Elasticsearch 6.x. | Boolean | true | GLOBAL_OFFLINE |
 | index.[X].elasticsearch.use-mapping-for-es7 | Mapping types are deprecated in ElasticSearch 7 and JanusGraph will not use mapping types by default for ElasticSearch 7 but if you want to preserve mapping types, you can setup this parameter to true. If you are updating ElasticSearch from 6 to 7 and you don't want to reindex your indexes, you may setup this parameter to true but we do recommend to reindex your indexes and don't use this parameter. | Boolean | false | MASKABLE |
+| index.[X].elasticsearch.retry_on_conflict | Specify how many times should the operation be retried when a conflict occurs. Default: 0. | Integer | 0 | MASKABLE |
```
farodin91: This line need to before `.elasticsearch.scroll-keep-alive`

nareshbabuy: Done, but still build is failing

nareshbabuy:
> On travis ci the git diff checks fails. It seems files have a different order.

What is the solution? Any idea

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
nareshbabuy: updated

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
nareshbabuy: removed "Default: 0"

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
porunov: That is interesting. In version `6.0` the parameter is `_retry_on_conflict`, in version `6.1` and above the parameter is `retry_on_conflict`. As I understand it is breaking change (which is really strange). I will try `_retry_on_conflict` parameter with ES version 6.x later today. Maybe it works also. 
The thing is that we are checking only major version here. If we should use `_retry_on_conflict` for `6.0` and `retry_on_conflict` for `6.1` then we would need to get and compare `feature version` also.

porunov: OK. I've checked this parameter. In the versions `6.x` bigger than 6.0 both `_retry_on_conflict` and `retry_on_conflict` works identically, so this isn't a breaking change.

nareshbabuy:
> LGTM. Thank you @nareshbabuy for this feature!
> Waiting for the Travis to complete.

Travis failed, is it possible to trigger again?

PR #1882, opened 2019-11-05
Minor code cleanup (NPEs etc)

micpod:
Signed-off-by: Michal Podstawski <mpodstawski@gmail.com>

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/TransactionalConfiguration.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/TransactionalConfiguration.java
@@ -68,7 +68,7 @@ public class TransactionalConfiguration implements WriteConfiguration {
     @Override
     public Iterable<String> getKeys(final String prefix) {
         return Iterables.concat(
-        Iterables.filter(writtenValues.keySet(), s -> s.startsWith(prefix)),
+        Iterables.filter(writtenValues.keySet(), s -> s != null && s.startsWith(prefix)),
         Iterables.filter(config.getKeys(prefix), s -> !writtenValues.containsKey(s)));
     }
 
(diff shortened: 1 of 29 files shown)


janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/TransactionalConfiguration.java
```diff
@@ -68,7 +68,7 @@ public WriteConfiguration copy() {
     @Override
     public Iterable<String> getKeys(final String prefix) {
         return Iterables.concat(
-        Iterables.filter(writtenValues.keySet(), s -> s.startsWith(prefix)),
+        Iterables.filter(writtenValues.keySet(), s -> s != null && s.startsWith(prefix)),
```
micpod: Hello @porunov , thank you for reviewing. Sorry that I removed previous PR but I ran into problems with CLA. Should be ok now.

It seems rather not usual to have null value here. But as far as we expose public method that allows appending elements to this HashTable I think it is slightly better to check it explicitly. If you not agree, we can remove this one.

porunov: @micpod That's fine. I don't mind additional check because it might prevent future bugs.

About CLA problems. I see previous PRs had problems with CLA because some PRs included old commits or commits from non-cla signers. Generally when it happens you can just push with amend and the CLA will automatically be updated for the PR. So, you don't need to close a wrong PR each time, you can amend it.
Something like:
```
git reset --hard your-correct-branch
git push -f
```
or something like:
```
git reset --hard origin/master
git rebase your-feature-branch
git push -f
```

porunov:

PR #1886, opened 2019-11-08
Optimize single index query

mad:
Pass limit for single index query

More info https://groups.google.com/forum/m/?utm_medium=email&utm_source=footer#!msg/janusgraph-users/cnVF3hC5OEY/vqq3Vc4HBgAJ

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/JointIndexQuery.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/JointIndexQuery.java
@@ -100,9 +100,11 @@ public class JointIndexQuery extends BaseQuery implements BackendQuery<JointInde
 
     @Override
     public JointIndexQuery updateLimit(int newLimit) {
-        JointIndexQuery ji = new JointIndexQuery(Lists.newArrayList(queries));
-        ji.setLimit(newLimit);
-        return ji;
+        List<Subquery> subqueries = (queries.size() == 1) ? Lists.newArrayList(queries.get(0).updateLimit(newLimit)) :
+            Lists.newArrayList(queries);
+        JointIndexQuery jointIndexQuery = new JointIndexQuery(subqueries);
+        jointIndexQuery.setLimit(newLimit);
+        return jointIndexQuery;
     }
 
     public static class Subquery implements BackendQuery<Subquery>, ProfileObservable {
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
@@ -359,8 +359,7 @@ public class GraphCentricQueryBuilder implements JanusGraphQuery<GraphCentricQue
             }
             indexLimit = Math.min(HARD_MAX_LIMIT,
                 QueryUtil.adjustLimitForTxModifications(tx, coveredClauses.size(), indexLimit));
-            jointQuery.setLimit(indexLimit);
-            query = new BackendQueryHolder<>(jointQuery,
+            query = new BackendQueryHolder<>(jointQuery.updateLimit(indexLimit),
                     coveredClauses.size() == conditions.numChildren(), isSorted);
         } else {
             query = new BackendQueryHolder<>(new JointIndexQuery(), false, isSorted);


janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/JointIndexQuery.java
```diff
@@ -100,9 +100,14 @@ public String toString() {
 
     @Override
     public JointIndexQuery updateLimit(int newLimit) {
-        JointIndexQuery ji = new JointIndexQuery(Lists.newArrayList(queries));
-        ji.setLimit(newLimit);
-        return ji;
+        JointIndexQuery queryWithNewLimit;
```
micpod: IMO looks strange. Consider if doing it just like:
```suggestion
        _if (queries.size() == 1) queries.get(0).updateLimit(newLimit);_
        JointIndexQuery ji = new JointIndexQuery(Lists.newArrayList(queries));
        ji.setLimit(newLimit);
        return ji;
```
before isn't a better option. Moreover, it may be more readable to add dedicated method to that handles this line.

micpod: Consider if sth like this wouldn't be more readable:
```suggestion
                List<Subquery> subqueries = (queries.size() == 1) ? Lists.newArrayList(queries.get(0).updateLimit(newLimit)) :
            Lists.newArrayList(queries);
        JointIndexQuery queryWithNewLimit = new JointIndexQuery(subqueries);
        queryWithNewLimit.setLimit(newLimit);
        return queryWithNewLimit;
```
Btw I also wondering what happens if queries count changes after you set this limit. From the class construction it seems it is possible to add additional subquery after setting this limit, wouldn't it affect this process?

mad: Currently JointIndexQuery instantiated in one place where no additions queries added after limit update. But it is not clean, so may be move this type of heuristics into https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java#L355

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/JointIndexQuery.java
```diff
@@ -100,9 +100,11 @@ public String toString() {
...
     @Override
     public JointIndexQuery updateLimit(int newLimit) {
-        JointIndexQuery ji = new JointIndexQuery(Lists.newArrayList(queries));
-        ji.setLimit(newLimit);
-        return ji;
+        List<Subquery> subqueries = (queries.size() == 1) ? Lists.newArrayList(queries.get(0).updateLimit(newLimit)) :
+            Lists.newArrayList(queries);
+        JointIndexQuery jointIndexQuery = new JointIndexQuery(subqueries);
+        jointIndexQuery.setLimit(newLimit);
+        return new JointIndexQuery(subqueries);
```
porunov: I am confused here.
Why do we create `jointIndexQuery` and then don't return it? Why do we need to return `new JointIndexQuery(subqueries)`?

mad: Yes, this is typo

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/JointIndexQuery.java
```diff
@@ -100,9 +100,11 @@ public String toString() {
 
     @Override
     public JointIndexQuery updateLimit(int newLimit) {
-        JointIndexQuery ji = new JointIndexQuery(Lists.newArrayList(queries));
-        ji.setLimit(newLimit);
-        return ji;
+        List<Subquery> subqueries = (queries.size() == 1) ? Lists.newArrayList(queries.get(0).updateLimit(newLimit)) :
+            Lists.newArrayList(queries);
+        JointIndexQuery jointIndexQuery = new JointIndexQuery(subqueries);
```
li-boxuan: Hi @mad!  could you explain why we need:
1. Copy `queries` into a new ArrayList,
and
2. Create a new JointIndexQuery object?

Does an in-place update not work? Thanks!

mad: Old implementation does same things but without update single query limit

And doc for `BackendQuery.updateLimit` sad
```
Creates a new query identical to the current one but with the specified limit.
```

So this method create new one query and update limit. Some `BaseQuery` implementation does not support limit update

li-boxuan: Thanks @mad ! I didn't notice the contract written in `BackendQuery.updateLimit`. I took a quick look and it seems all current implementations support limit update. IMO we could change the contract and always use an in-place update, which could save a little bit of object copy time. I would be happy to contribute a patch if applicable.

li-boxuan: I took a closer look, and you're right, there are some implementations that don't support in-place update. And there are some use cases where a new query is preferable. So I guess we shall leave them as they are.

Issue #1911, opened 2019-12-24
When the adjacent vertex query is not set label,all edges are returned.


492066199:
When the adjacent vertex query is not set label,all edges are returned.

- Version: 6929fc7540c86e06768781b31efc75548827d2ca（current）
- Storage Backend: hbase
- Mixed Index Backend: no
- Steps to Reproduce: 
```
janusGraph.multiQuery(vertex).adjacent(vertex1).edges()

If you query like this, all edges will be returned, not only the edges of adjacent vertex that have been specified，So it will lose the ability to quickly judge whether there is an edge between two vertices in super node vertex.

**Comment by li-boxuan** [comment]
Just want to add more context:
Seems this bug only appears when direction is Direction.BOTH, and labels are null.

If a vertex 'left' has k edges, but vertex 'thirdParty' is not its neighbor, then we can observe the following:
```java
left.query().adjacent(thirdParty).edgeCount(); // returns k, incorrect
left.query().adjacent(thirdParty).direction(Direction.BOTH).edgeCount(); // returns k, incorrect
left.query().adjacent(thirdParty).direction(Direction.IN).edgeCount(); // returns 0, correct
left.query().adjacent(thirdParty).direction(Direction.OUT).edgeCount(); // returns 0, correct
```

PR #1913, opened 2019-12-24
fix:  When the adjacent vertex query is not set label,all edges are r…

492066199:
This PR is related to #1911
It currently contains the following changes:
1.add check in constructQueryWithoutProfile() method in BasicVertexCentricQueryBuilder.java.
2.The check() method will check whether the adjacent query has set the label. If not, all the labels will be obtained and set.

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
@@ -29,10 +29,12 @@ import org.janusgraph.graphdb.query.profile.QueryProfiler;
 import org.janusgraph.graphdb.relations.StandardVertexProperty;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.core.schema.SchemaStatus;
+import org.janusgraph.graphdb.types.system.BaseKey;
 import org.janusgraph.graphdb.types.system.ImplicitKey;
 import org.janusgraph.graphdb.types.system.SystemRelationType;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.janusgraph.util.datastructures.Interval;
+import org.janusgraph.util.datastructures.IterablesUtil;
 import org.janusgraph.util.datastructures.PointInterval;
 import org.janusgraph.util.datastructures.RangeInterval;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


porunov:
@492066199 Thank you for the contribution! Would you be able to add a test for this PR?

492066199:
ok，I have added relevant tests. The changes are related to the query of the graph, so the test cases is added to the JanusGraphTest. and These test cases can cover my changes.

492066199:
@porunov ok，I have added relevant tests. The changes are related to the query of the graph, so the test cases is added to the JanusGraphTest. and These test cases can cover my changes and the bug.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
```diff
@@ -130,6 +130,7 @@ public JanusGraphMultiVertexQuery addAllVertices(Collection<? extends Vertex> ve
 
     @Override
     public Map<JanusGraphVertex, Iterable<JanusGraphEdge>> edges() {
+        check();
```
farodin91: I would like to check here the execution order. 
I’m not able to it before 03.01.2020

492066199: Hello, thank you for reviewing my code. I don't think this check() method is elegant here.and  I changed the location of this method. I don't quite understand what you mean.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
```diff
@@ -418,6 +420,20 @@ protected BaseVertexCentricQuery constructQueryWithoutProfile(RelationCategory r
...
+            Iterable<JanusGraphVertex> allTypes = QueryUtil.getVertices(this.tx,
+                    BaseKey.SchemaCategory, JanusGraphSchemaCategory.EDGELABEL);
+            int length = IterablesUtil.size(allTypes);
+            int index = 0;
+            String[] labels = new String[length];
+            for(JanusGraphVertex labelVertex : allTypes){
+                labels[index] = ((EdgeLabel)labelVertex).name();
+                index++;
+            }
+            labels(labels);
```
li-boxuan: This seems to be a working fix, but it's more like a quick workaround. I think we need to find the root cause first.

What I have observed is, when Direction is BOTH, the bug appears, but when Direction is IN or OUT, the bug does not appear. So the root cause might be related to that.

porunov:
@492066199 thank you for finding the big and for your contribution! It helped very much. The issue was resolved by @li-boxuan in #1993 . I am closing this PR. If you think that the issue isn't resolved or you observe some problems with provided fix, please feel free to reopen the PR.

PR #1915, opened 2019-12-28
Update to TinkerPop 3.4.4

### LINKED ISSUE #1617: Update to TinkerPop 3.4.4
Tinkerpop 3.4.4 is released

porunov:
This update makes a very small breaking change (if it can be counted as a breaking change). If `Automatic Schema Maker` is enabled and we make a query where we create the property (for the first time) and cardinality is `SET` or `LIST`, then we should explicitly define it for each added value (only for the first query). This change can be seen in modified tests. I didn't find any mentioned changes to cardinalities from `3.4.1` to `3.4.4`, so I can't say what causes this change.

Even so, the breaking change is very tiny and most likely nobody will be affected (because our documentation suggests to disable `Automatic Schema Maker`), I decided to make this TinkerPop update only for `master` branch, and not for `0.4` branch.

Testcontainers update is also required due to some conflicting versions of used libreries with TinkerPop.

TinkerPop tests are tested with `mvn clean install -Dtest.skip.tp=false -DskipTests=true`. All TinkerPop tests passed.
TravisCI tests passed also: https://travis-ci.org/porunov/janusgraph/builds/630220943

Fixes #1617

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -26,6 +26,45 @@ use the latest versions of the software.
 
 ## Release Notes
 
+### Version 0.4.1 (Release Date: January ?, 2020)
+
+```xml tab='Maven'
+<dependency>
+    <groupId>org.janusgraph</groupId>
+    <artifactId>janusgraph-core</artifactId>
+    <version>0.4.1</version>
+</dependency>
+```
+
+```groovy tab='Gradle'
+compile "org.janusgraph:janusgraph-core:0.4.1"
+```
+
+**Tested Compatibility:**
+
+-   Apache Cassandra 2.2.10, 3.0.14, 3.11.0
+-   Apache HBase 1.2.6, 1.3.1, 1.4.10, 2.1.5
+-   Google Bigtable 1.3.0, 1.4.0, 1.5.0, 1.6.0, 1.7.0, 1.8.0, 1.9.0, 1.10.0, 1.11.0
+-   Oracle BerkeleyJE 7.5.11
+-   Elasticsearch 5.6.14, 6.0.1, 6.6.0
+-   Apache Lucene 7.0.0
+-   Apache Solr 7.0.0
+-   Apache TinkerPop 3.4.4
+-   Java 1.8
+
+For more information on features and bug fixes in 0.4.1, see the GitHub milestone:
+
+-   <https://github.com/JanusGraph/janusgraph/milestone/15?closed=1>
+
+#### Upgrade Instructions
+
+##### TinkerPop: Upgrade from 3.4.1 to 3.4.4
+
+Adding multiple values in the same query to a new vertex property without explicitly defined type 
+(i.e. using `Automatic Schema Maker` to create a property type) requires explicit usage of `VertexProperty.Cardinality` 
+for each call (only for the first query which defines a property) if the `VertexProperty.Cardinality` is different than 
+`VertexProperty.Cardinality.single`.
+
 ### Version 0.4.0 (Release Date: July 1, 2019)
 Legacy documentation: <https://old-docs.janusgraph.org/0.4.0/index.html>
 
--- a/docs/basics/schema.md
+++ b/docs/basics/schema.md
@@ -221,6 +221,10 @@ implicitly created property keys have cardinality SINGLE and data type
 `Object.class`. Users can control automatic schema element creation by
 implementing and registering their own `DefaultSchemaMaker`.
 
+When defining a cardinality for a vertex property which differs from SINGLE, 
+the cardinality should be used for all values of the vertex property in the 
+first query (i.e. the query which defines a new vertex property key).
+
 It is strongly encouraged to explicitly define all schema elements and
 to disable automatic schema creation by setting `schema.default=none` in
 the JanusGraph graph configuration.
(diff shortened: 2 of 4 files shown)


farodin91:
Could you increase the version in the mkdocs file?

porunov:
@farodin91 Amended the commit with tinkerpop version update in `mkdocs.yml`

farodin91:
@porunov Before, we merge i would like to know why we have a breaking changes.

porunov:
@farodin91 @FlorianHockmann In 3.4.4 order of processing properties is vice versa. The first property to be added is actually the last one and the last property to be added is the first one. Not sure why it happens

porunov:
Something in traversal is changed but it is hard to tell what logic exactly is changed and why we process added vertex properties vice versa in 3.4.4.
@FlorianHockmann @farodin91 are you able to check that? I am not very familiar with that part of the code.

FlorianHockmann:
> Something in traversal is changed but it is hard to tell what logic exactly is changed and why we process added vertex properties vice versa in 3.4.4.
@FlorianHockmann @farodin91 are you able to check that? I am not very familiar with that part of the code.

OK, I just looked into this and I think I found the cause for this change*. apache/tinkerpop#1076 (which fixes [TINKERPOP-2112](https://issues.apache.org/jira/browse/TINKERPOP-2112)) folds `property` steps into graph mutation steps like `addV`. This fold won't be done if the `property` step contains a `Cardinality` (and in some other cases, which is commented in the changed code). So, our `property` steps without a `Cardinality` get folded into the `addV` and therefore executed first. Afterwards, the `property` step with a `Cardinality` is executed which fails as the `Cardinality` is already set to the default value of `SINGLE` by the folded in `property` step that didn't have a `Cardinality`.

So, I see two options for us:

1. We accept the new behaviour and tell users that they need to specify the cardinality on all `property` steps in a traversal where the property key is used the first time when they want to use the automatic schema maker.
1. We implement a traversal strategy that tries to find cardinality information in a `property` step and provides that to the graph mutation step (like `addV`) in which `property` steps were folded into.

I'm in favour of option 1 as we discourage the use of the automatic schema maker in production use cases and because it doesn't seem problematic to me that users have to specify the cardinality on all `property` steps if they want to use a non-default cardinality with the automatic schema maker in the first traversal a new property key is used in.

We can now also decide whether we consider this really a breaking change and therefore only merge it into `master` or whether we also want to have this in the next `0.4` release. I would prefer the latter, again because the automatic schema maker is discouraged and because this problem doesn't seem to be a big issue to me, but if you see it differently, then I'm of course also OK with merging this only into `master`.

\* I simply tested different TinkerPop versions (3.4.2, 3.4.3, and 3.4.4) to find out that the change already occurred in 3.4.2. [The TinkerPop changelog for that version](https://github.com/apache/tinkerpop/blob/3.4.2/CHANGELOG.asciidoc#release-3-3-7) (or to be more specific for the included version 3.3.7) included this which hints at this behaviour:

> Improved folding of property() step into related mutating steps.

The linked ticket is [TINKERPOP-2112](https://issues.apache.org/jira/browse/TINKERPOP-2112).

porunov:
@FlorianHockmann Thank you very much for this investigation!
I see it as a really small breaking change which still should be described in changelog in case someone uses automatic schema maker in their use cases. I couldn't decide which version we should target this PR, so I decided to target a master branch but I am OK to retarget this PR into `0.4.1` version. All we need is to add a breaking change explanation in `Upgrade Instructions` of `0.4.1` version. We have similar situation in `0.2.1` version. 
Are you OK if we add `Upgrade Instructions` in `0.4.1` or should we leave this PR in `master`?

FlorianHockmann:
> Are you OK if we add Upgrade Instructions in 0.4.1

Yes, that would be my preference :)

porunov:
@FlorianHockmann I have retargeted this PR to `v0.4` branch.

Tests `testAutoSchemaMakerAllowsToSetCardinalityList` and `testAutoSchemaMakerAllowsToSetCardinalitySet` were added in `master` branch (https://github.com/JanusGraph/janusgraph/commit/cda90513c31886ab5d78e6be446fcb83498f3adc#diff-31f73d4ab9ff6c3484aa6baecd8b215f), so I will resolve them during the merge into `master` branch.

docs/changelog.md
```diff
@@ -26,6 +26,45 @@ use the latest versions of the software.
...
+
+For more information on features and bug fixes in 0.4.1, see the GitHub milestone:
+
+-   <https://github.com/JanusGraph/janusgraph/milestone/15?closed=1>
+
+#### Upgrade Instructions
+
+##### TinkerPop: Upgrade from 3.4.1 to 3.4.4
+
+Adding multiple values in the same query to a new property without explicitly defined type
```
FlorianHockmann: (nitpick) _to a new **vertex** property_ as this should be specific to vertex properties as edge properties don't have a cardinality

porunov:
Thank you @FlorianHockmann @farodin91 for your reviews. I've amended the commit to include `vertex` word in docs. Also, TinkerPop tests for this PR (rebased on `0.4` branch now) just finished and all TP tests passed. I will wait till Travis tests are finished and merge the PR.

PR #1951, opened 2020-02-08
Fix missing plugins javadoc and maven plugins for driver module

### LINKED ISSUE #1950: Java docs generation error for janusgraph-driver during a build process.
Java docs error is received for `janusgraph-driver` during a build process. The folder `janusgraph-driver/target/apidocs` is empty after the build is successfully finished.

Expected behaviour:
Java docs are generated for `janusgraph-driver` module and populated in corresponding `apidocs` folder after the build is finished.

- Version: 0.5.0-SNAPSHOT
- Steps to Reproduce: `mvn clean install -DskipTests=true`

### Stack Trace (if you have one)
```
...
[INFO] --- maven-javadoc-plugin:2.10.4:jar (attach-javadocs) @ janusgraph-core ---
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:2.10.4:javadoc' has not been previously called for the module: 'org.janusgraph:janusgraph-driver:jar:0.5.0-SNAPSHOT'. Trying to invoke it...
[WARNING] Creating fake javadoc directory to prevent repeated invocations: /home/ws/janusgraph/janusgraph-driver/target/apidocs
[ERROR] Error fetching link: /home/ws/janusgraph/janusgraph-driver/target/apidocs/package-list. Ignored it.
[INFO] 
100 warnings
[WARNING] Javadoc Warnings
...
```

porunov:
Fixes #1950

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-all/pom.xml
+++ b/janusgraph-all/pom.xml
@@ -25,6 +25,11 @@
     <dependencies>
         <dependency>
             <groupId>org.janusgraph</groupId>
+            <artifactId>janusgraph-driver</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.janusgraph</groupId>
             <artifactId>janusgraph-core</artifactId>
             <version>${project.version}</version>
         </dependency>
--- a/janusgraph-core/pom.xml
+++ b/janusgraph-core/pom.xml
@@ -23,16 +23,6 @@
         <!-- TinkerPop -->
         <dependency>
             <groupId>org.apache.tinkerpop</groupId>
-            <artifactId>gremlin-core</artifactId>
-            <exclusions>
-                <exclusion>
-                    <artifactId>gossip</artifactId>
-                    <groupId>org.sonatype.gossip</groupId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.tinkerpop</groupId>
             <artifactId>gremlin-groovy</artifactId>
         </dependency>
         <dependency>
@@ -118,24 +108,6 @@
         </testResources>
         <plugins>
             <plugin>
-                <groupId>org.codehaus.gmavenplus</groupId>
-                <artifactId>gmavenplus-plugin</artifactId>
-                <version>${gmavenplus.version}</version>
-                <executions>
-                    <execution>
-                        <goals>
-                            <goal>generateStubs</goal>
-                            <goal>compile</goal>
-                            <goal>testGenerateStubs</goal>
-                            <goal>testCompile</goal>
-                        </goals>
-                        <configuration>
-                            <providerSelection>1.8</providerSelection>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
-            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <configuration>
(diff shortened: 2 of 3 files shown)


porunov:
@farodin91 @FlorianHockmann @pluradj this is a blocking issue for the `0.5.0` release. Reviews would be very helpful for this PR because we won't be able to make `0.5.0` release without fixing #1950

FlorianHockmann:
Could you give a quick explanation of why these dependencies had to be moved from janusgraph-core to janusgraph-driver? Was this simply forgotten when janusgraph-driver was created as its own module?

porunov:
> Could you give a quick explanation of why these dependencies had to be moved from janusgraph-core to janusgraph-driver? Was this simply forgotten when janusgraph-driver was created as its own module?

@FlorianHockmann Sure. Below I provide reasons for each step which I decided to make when was fixing the dependency problems. 

**General intro and moving `maven-javadoc-plugin` into driver module**

To generate docs we should use `maven-javadoc-plugin` which was previously in `janusgraph-core`. Because all other modules were depending on `janusgraph-core`, the plugin was populated in all modules. In the commit https://github.com/JanusGraph/janusgraph/commit/b96aa6d26f74d3dc8f7404212bddeefe2d0790b4 we extracted JanusGraph Gremlin driver requirements in `janusgraph-driver` which doesn't depend on any modules but `janusgraph-core` now depends on `janusgraph-driver` module.
When the `janusgraph-driver` module was created, we forgot to move plugins which should be populated to all modules from `janusgraph-core` into `janusgraph-driver` (because now we can assume that `janusgraph-driver` is the root module because `janusgraph-core` directly depends on `janusgraph-driver` and all other modules indirectly depend on `janusgraph-driver`).
So, as I said, we should just move all necessary modules from old `"main" module (i.e. janusgraph-core)` into a new `"main" module (i.e. janusgraph-driver)`.

**Removing `gremlin-core` from driver and excluding `org.sonatype.gossip` it driver**

While moving those plugins I found that the dependency `gremlin-core` already exists in `janusgraph-driver`, so we don't need to write it in `janusgraph-driver` (so I removed it). You may notice that it excludes `org.sonatype.gossip` (https://github.com/jdillon/gossip) in `janusgraph-core` but doesn't excludes in `janusgraph-driver`, so I excluded it in `janusgraph-driver` because we are using `logback` for logging instead of `gossip`. Because we need only a single slf4j provider, we exclude `gossip`.

**Moving `org.codehaus.gmavenplus` from core into driver**

`org.codehaus.gmavenplus` (https://github.com/groovy/GMavenPlus) is just a plugin that allows you to integrate Groovy into your Maven projects.
For some reason it was added as a `core` dependency (Maybe because it has `gremlin-groovy` dependency) but nevertheless groovy was available in all other dependant modules. To not break this contract I just moved it into `janusgraph-driver` module.
This isn't necessary, but I thought that we may need it everywhere, so I moved it to `driver` module. I can move it back if you think that we don't need to populate this dependency into `driver` module.

**Removing `janusgraph-release` profile from `janusgraph-core`**

You may notice that `janusgraph-release` already exists in `janusgraph-driver`. That is why I removed it from `core`. But the profile in `driver` module didn't have some `maven-shade-plugin` plugin in `build` so I added it into `driver` module. Also, `maven-gpg-plugin` had `sign-artifacts` exclusion, so I added this exclusion into `driver` module.
I think `janusgraph-release` profile should be the same for all modules because it is connected to populating new JanusGraph releases.

Issue #1525, opened 2019-04-08
hasId should be optimized to perform in O(1)


porunov:
Currently `hasId` step performs in O(n) complexity but should be O(1).
Currently `is` step complexity is O(1) but `hasId` step is O(n) which is strange.
I.e. with the big amount of outgoing edges the next code will take a lot of time:
```
Vertex outVertex = ...;
Vertex inVertex = ...;
graph.traversal().V(outVertex.id()).outE("my_label").where(__.otherV().hasId(inVertex.id())).hasNext();
```
But the same query with `is` will perform almost immediately: 
```
Vertex outVertex = ...;
Vertex inVertex = ...;
graph.traversal().V(outVertex.id()).outE("my_label").where(__.otherV().is(inVertex)).hasNext();
```

rngcntr:
I'll have a closer look at this as it's related to #1937 

At this point, I don't think this is a question of complexity. From what I can tell, both queries are O(n), but one of them uses the information encoded in the edges while the other one fetches the adjacent vertex of every edge. This should be easily avoidable.

porunov:
Fixed by #1960

PR #1960, opened 2020-02-12
Optimize hasId() in the same way as is()

rngcntr:
Solves #1525

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
@@ -14,7 +14,7 @@
 
 package org.janusgraph.graphdb.tinkerpop.optimize;
 
-import org.janusgraph.graphdb.types.system.ImplicitKey;
+import java.util.List;
 import org.apache.tinkerpop.gremlin.process.traversal.Compare;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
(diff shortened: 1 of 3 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
```diff
@@ -32,73 +32,191 @@
...
 
+                switch (type) {
+                case IS:
+                    replaceIsStep(traversal, originalStep, subSteps);
+                    break;
+                case HASID:
+                    replaceHasIdStep(traversal, originalStep, subSteps);
+                    break;
+                default:
+                    break;
```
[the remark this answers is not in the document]
rngcntr: It's not a big deal but I'd like to keep it in there just to avoid unnecessary compiler warnings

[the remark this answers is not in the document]
porunov: You are right. I didn't think about it. It makes sense then.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
```diff
@@ -32,73 +32,179 @@
...
-                }
-                P predicate = ((IsStep) steps.get(1)).getPredicate();
-                //Check that we have a valid direction and a valid vertex filter predicate
-                if (direction != null && predicate.getBiPredicate() == Compare.eq && predicate.getValue() instanceof Vertex) {
-                    Vertex vertex = (Vertex) predicate.getValue();
-
-                    //Now, check that this step is preceded by VertexStep that returns edges
-                    Step<?, ?> currentStep = originalStep.getPreviousStep();
-                    while (currentStep != EmptyStep.instance()) {
-                        if (!(currentStep instanceof HasStep) && !(currentStep instanceof IdentityStep)) {
```
rngcntr: I removed the repeated jumps because issues arose when using this strategy on queries that utilize `SideEffectStep`s.
The re-ordering of steps caused the logical order of these steps to break.

porunov: As I understand you found a bug which currently exists. Would you be able to provide a separate PR which removes this jumps and adds a test where we can see that bug? Or is the appears only in optimized queries with .hasId(id)?

porunov: If the bug can be reproduced only in this PR, would you be able to add a test to this PR which shows the issue you described?

rngcntr: This is a bug I encountered when applying the optimization techniques from this class in PR #1961 
As the basic idea behind both approaches is essentially the same, it should be possible to find a matching test case. I'll try to find one once I get to work on this the next time.

rngcntr: Nevermind, I introduced the bug myself when I was trying to catch even more optimizable sequences of steps. Thanks for your review!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
```diff
@@ -32,73 +32,162 @@
...
+
+    private void replaceStep(Traversal.Admin<?, ?> traversal, OptimizableQueryType type,
+                             TraversalFilterStep originalStep, List<Step> steps) {
+        // Get the direction in which we filter on the adjacent vertex (or null if not a valid
+        // adjacency filter)
+        Direction direction = parseDirection(steps);
+        P predicate = parsePredicate(type, steps);
+        // Check that we have a valid direction and a valid vertex filter predicate
+        if (direction != null && isValidPredicate(type, predicate)) {
+            if (isPreviousStepValid(originalStep, direction)) {
```
FlorianHockmann: Codacy Issue found: [These nested if statements could be combined](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=4989535)

FlorianHockmann:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=4989535&bid=16472911)

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexFilterOptimizerStrategy.java
```diff
@@ -32,73 +32,161 @@
...
+        Step<?, ?> previousStep = originalStep.getPreviousStep();
+        while (previousStep != EmptyStep.instance()) {
+            if (!(previousStep instanceof HasStep) && !(previousStep instanceof IdentityStep)) {
+                break;
+            }
+            previousStep = previousStep.getPreviousStep();
+        }
+
+        if (previousStep instanceof VertexStep) {
+            try (VertexStep<?> vertexStep = (VertexStep<?>) previousStep) {
```
porunov: I didn't find `vertexStep.close()` in the previous version. I don't know if we should close it here after `try with resources` but could you explain why you are closing `vertexStep` here? Is it connected to resources leak or other possible problems?

rngcntr: When not closing `vertexStep`, Java outputs a warning about a possible resource leak.

porunov: @rngcntr Can you confirm a possible resource leak here?
I am just not sure if there is a resource leak. What about other possible `VertexStep` which are not connected to this optimization? Do we have a resource leak there?
What about all other steps which implement `AutoCloseable` (most of the steps are either implementing it through `TraversalParent` or directly)? Do we have resource leak there or is this resource leak connected only to this optimization?
If the resource leak is connected to all steps, then we should close them somewhere else (no in particular optimization step). If the resource leak is directly connected to this optimization, then you are doing it right and we should close those steps here.
I am not very familiar with that code base, so I don't currently see resource leak here.
What I am thinking is that maybe we should close `originalStep` because we replace it with a new step but I am not sure about it yet.
Also, I don't want to introduce a bug here in case this vertex step is supossed to be closed at some other stage but we are closing it erlier, in such case we could introduce a new bug. I don't tell that this code introduces a new bug I just tell that I don't see this code to fix any existing bug (at least, I can't verify it yet).
I think maybe it is better to not mix the possible resource leak (which isn't connected to this PR) and `hasId` optimization.
I think a separate PR to fix a possible resource leak would be a better choice (if there is a resource leak).
If you can confirm resource leak, then I am OK with making this change in this PR. If you are able to add a test which verifies this resource leak, it would be great.

rngcntr: I can neither confirm the resource leak nor find any find any bugs (covered by the tests) that arise from closing the step here.
Internally, `close()` causes the underlying `FlatMapStep`'s Iterator to be closed _if it it implements_ `AutoCloseable`. This Iterator is initialized with a singleton instance of the class `EmptyIterator` which is replaced with an actual Iterator of the Elements emitted by the call of `flatMap()`.

Because `EmptyIterator` does not implement `AutoCloseable`, the call of `close()` has no other effect than satisfying the compiler ;)

I quickly discussed this with @farodin91 yesterday and we came to the conclusion that even if there's no point in calling `close()` in this particular case, it aids consistency of the source code in general.

Don't get me wrong here, I'm also fine with keeping this step unclosed if others prefer

porunov: What do you mean by `it aids consistency of the source code in general.`?
As for me, if we are calling `close` for some steps (without any logic behind this except removing the warning from compiler) and not calling `close` for other steps then it looks like inconsistent code.
Let's imagine the situation when `close` method actually making the step unreadable. In such case, we will experience a situation, that sometimes, some of `VertexStep`s are closed during the optimization phase, but some of `VertexStep`s are not closed during the same optimization phase. We will need to track it back to this point to find out that we are closing only some `VertexStep`s here for no reason actually (besides making the compiler happy).
I've described a really easy possible bug here, which I believe we will be able to find very quickly, but in the future `close` method may have a totally different logic, which will cause "something" and so, it will be hard to find out the bug which isn't easily reproducible.
As for me, adding such logic just to make a compiler happy is a bad idea.
I believe, this change isn't connected to this optimization directly. I would recommend opening a new issue and discussing this change there, because this may take some time before we have a consensus on calling `close` here (which we are not calling right now but your PR adds this logic for some reason. As for me, the reason isn't satisfying, but other members may have more thoughts on this).

rngcntr: I totally get your point concerning future changes. I reverted this change and I'm now keeping the handling of `close` calls the same as it was before this PR.

Issue #1962, opened 2020-02-17
Lucene query tokenizing for custom analyzer doesn't work properly


VladimirBogomolov:
If use CustomAnalyzer with filter that returns more than one token for one analyzing word then get wrong lucene query.
For reproduce that bug you shold configure CustomAnalyzer with, for example SynonymFilter, RussianLuceneMorphology or something like.
Example.
I use SynonymFilter and I have query "clear", after tokenizer has been applied we get next tokens: "clear", "change", "modify", "unclutter", "alter" (https://www.synonym.com/synonyms/clear). Now tokens are used for building query and I have 
`+fieldName:clear +field:change +fieldName:modify +fieldName:unclutter +fieldName:alter`,
but should be 
`fieldName:clear field:change fieldName:modify fieldName:unclutter fieldName:alter`

- Version:  0.4.1, master branch has that bug too
- Storage Backend: Berkeley
- Mixed Index Backend: Lucene
- Mailing list Thread URL:
- Steps to Reproduce:

porunov:
Fidex in #1963

PR #1963, opened 2020-02-17
Improve lucene tokenizer

VladimirBogomolov:
Resolve Issue #1962
-----

### For all changes:

### DIFF
--- a/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
+++ b/janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
@@ -14,6 +14,7 @@
 
 package org.janusgraph.diskstorage.lucene;
 
+import org.apache.lucene.analysis.tokenattributes.OffsetAttribute;
 import org.janusgraph.core.Cardinality;
 import org.janusgraph.core.attribute.Cmp;
 import org.janusgraph.core.attribute.Geo;
@@ -72,6 +73,7 @@ import java.util.AbstractMap.SimpleEntry;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -603,23 +605,27 @@ private static Query numericQuery(String key, Cmp relation, Number value) {
...
+        Map<Integer, List<String>> stemsByOffset = new HashMap<>();
         try (CachingTokenFilter stream = new CachingTokenFilter(analyzer.tokenStream(fieldName, value))) {
+            final OffsetAttribute offsetAtt = stream.getAttribute(OffsetAttribute.class);
             final TermToBytesRefAttribute termAtt = stream.getAttribute(TermToBytesRefAttribute.class);
             stream.reset();
             while (stream.incrementToken()) {
-                terms.add(termAtt.getBytesRef().utf8ToString());
+                int offset = offsetAtt.startOffset();
+                String stem = termAtt.getBytesRef().utf8ToString();
+                stemsByOffset.putIfAbsent(offset, new ArrayList<>());
```
porunov: (nitpick) 
So we are executing `new ArrayList<>()` even if it isn't absent. Generally it isn't a big deal because GC handles those objects easily but I am not sure if compilator is able to optimize that (I think it isn't).
You can either create a static function which returns a new array list (i.e. it is executed only if it is absent) or your can just return your list and create a new one if it is empty.
Something like:
```
List<String> stemList = stemsByOffset.get(offset);
if(stemList == null){
  stemList = new ArrayList<>();
  stemsByOffset.put(offset, stemList);
}
stemList.add(stem);
```

VladimirBogomolov: It makes sense, thanks.
Done.

VladimirBogomolov:
> @VladimirBogomolov LGTM thank you! But could you also squash you commits into a single commit?

done

VladimirBogomolov:
I think the test
HBaseOperationCountingTest> JanusGraphOperationCountingTest.testIdCounts
fails not because of my changes

Issue #1969, opened 2020-02-18
Improve Index cache lookup


VladimirBogomolov:
Sometimes a workaround should look for uncommitted vertices within a transaction. To do this, we use the index cache. The first condition, which has the corresponding composite index, is used as the key to search the cache.

`StandardJanusGraphTx.java`
``` java
private PredicateCondition<PropertyKey, JanusGraphElement> getEqualityCondition(Condition<JanusGraphElement> condition) {
...
        for (final Condition<JanusGraphElement> child : condition.getChildren()) {
                final PredicateCondition <PropertyKey, JanusGraphElement> p = getEqualityCondition (child);
                if (p! = null) return p;
        }
...
}
```
Thus, if the selected index is not selective (i.e., it returns too many elements), then a cache search is not optimal.
I think that we need to iterate over all indexed keys and choose a more suitable one to search in the cache.

PR #1970, opened 2020-02-18
Improve index cache lookup

VladimirBogomolov:
Resolve Issue #1963
-----

### For all changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -90,6 +90,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
@@ -1232,17 +1233,17 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
 
     public final QueryExecutor<GraphCentricQuery, JanusGraphElement, JointIndexQuery> elementProcessorImpl = new QueryExecutor<GraphCentricQuery, JanusGraphElement, JointIndexQuery>() {
 
-        private PredicateCondition<PropertyKey, JanusGraphElement> getEqualityCondition(Condition<JanusGraphElement> condition) {
+        private List<PredicateCondition<PropertyKey, JanusGraphElement>> getEqualityConditions(Condition<JanusGraphElement> condition) {
             if (condition instanceof PredicateCondition) {
                 final PredicateCondition<PropertyKey, JanusGraphElement> pc = (PredicateCondition) condition;
-                if (pc.getPredicate() == Cmp.EQUAL && TypeUtil.hasSimpleInternalVertexKeyIndex(pc.getKey())) return pc;
+                if (pc.getPredicate() == Cmp.EQUAL && TypeUtil.hasSimpleInternalVertexKeyIndex(pc.getKey())) return Collections.singletonList(pc);
             } else if (condition instanceof And) {
-                for (final Condition<JanusGraphElement> child : condition.getChildren()) {
-                    final PredicateCondition<PropertyKey, JanusGraphElement> p = getEqualityCondition(child);
-                    if (p != null) return p;
-                }
+                return StreamSupport.stream(condition.getChildren().spliterator(), false)
+                                    .map(this::getEqualityConditions)
+                                    .flatMap(Collection::stream)
+                                    .collect(Collectors.toList());
             }
-            return null;
+            return Collections.emptyList();
         }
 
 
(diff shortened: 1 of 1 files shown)


rngcntr:
Did you mean Issue #1969?

VladimirBogomolov:
> Did you mean Issue #1969?

Yes, I changed the description

porunov:
@VladimirBogomolov I didn't check this PR yet but is it possible to add a test to verify that the optimization works as expected?

VladimirBogomolov:
> @VladimirBogomolov I didn't check this PR yet but is it possible to add a test to verify that the optimization works as expected?

Test has been added.

VladimirBogomolov:
@porunov What about my PR ?

porunov:
@VladimirBogomolov thank you for the test. It definitely improves index lookup performance very much when we are contracting a query vice verse in your test case `.has("key2", "KeyValue1").has("key1", "SameKeyValue")` but I am concerned about best case scenario `.has("key1", "SameKeyValue").has("key2", "KeyValue1")` where the index lookup performance decreases about 2 times.
Sorry but I still didn't yet check the logic behind the code. Would you be able to describe disadvantages of this PR (for example, cases when the performance decreases)? On what parameters the performance decree depends in the second query? Did it happen by accident or does it depend on amount of indices which are available?

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1252,49 +1253,71 @@ public boolean isDeleted(VertexCentricQuery query, JanusGraphRelation result) {
...
-                        Preconditions.checkArgument(cond.getType() != Condition.Type.LITERAL || cond instanceof PredicateCondition);
-                        if (cond instanceof PredicateCondition) {
-                            keys.add(((PredicateCondition<PropertyKey, JanusGraphElement>) cond).getKey());
+                if (equalityConditions.isEmpty()) {
+                    vertices = withoutCache(query);
+                } else {
+                    int minSize = Integer.MAX_VALUE;
+                    PredicateCondition<PropertyKey, JanusGraphElement> bestCond = null;
+                    for (PredicateCondition<PropertyKey, JanusGraphElement> cond : equalityConditions) {
+                        int currentSize = Iterables.size(newVertexIndexEntries.get(cond.getValue(), cond.getKey()));
```
rngcntr, line 1266: Part of the decreased performance of QUERY#2 could be due to the call of `newVertexIndexEntries.get(...)` here.
Looking at the [current implementation](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/indexcache/ConcurrentIndexCache.java) this creates an `ArrayList` and populates it with the matching index entries. Although `Iterables` is clever enough to detect that the resulting `Iterable` is in fact a `Collection` and calls it's `size()` method directly, this looks to me like an overkill for counting the number of results.

Instead, it could be beneficial to extend the Interface `IndexCache` by a method `count(Object value, PropertyKey key)` which only returns the size of the result set directly.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1252,49 +1253,71 @@ public boolean isDeleted(VertexCentricQuery query, JanusGraphRelation result) {
...
                 return (Iterator) com.google.common.collect.Iterators.filter(vertices, query::matches);
             } else if ( (query.getResultType() == ElementCategory.EDGE || query.getResultType()==ElementCategory.PROPERTY)
                                         && !addedRelations.isEmpty()) {
                 return (Iterator) addedRelations.getView(relation -> query.getResultType().isInstance(relation) && !relation.isInvisible() && query.matches(relation)).iterator();
             } else return Collections.emptyIterator();
         }
 
+        private Iterator<JanusGraphVertex> withCache(PredicateCondition<PropertyKey, JanusGraphElement> standardIndexKey) {
+            Iterator<JanusGraphVertex> vertices;
+            vertices = com.google.common.collect.Iterators.transform(newVertexIndexEntries.get(standardIndexKey.getValue(), standardIndexKey.getKey()).iterator(), new com.google.common.base.Function<JanusGraphVertexProperty, JanusGraphVertex>() {
```
rngcntr, line 1287:
```java
return StreamSupport.stream(
    newVertexEntries.get(standardIndexKey.getValue(), standardIndexKey.getKey()).spliterator(), false)
    .map(p -> p.element())).iterator();
```

rngcntr:
@porunov The logic didn't change that much. Here's the difference:

The old query executor took the first `PredicateCondition` it could find in the query, loaded all matching elements from the cache and filtered them in memory to match all remaining conditions. If the first condition is not very selective, this is bad.
The new approach finds all `PredicateCondition`s recursively and queries the cache for the number of hits for each of those. It then selects the condition with the highest selectivity and does further in memory matching based on these results.

 - The reason for the improvement of QUERY#1 is obvious: We are now able to select the better condition.
 - The reason for the decrease in performance is also easy to see: Even if the first predicate is the most selective, we still need to check all of the remaining predicates to confirm this.

Issue #1916, opened 2019-12-30
Berkeley-Lucene: Error: Multiple entries with same key


VladimirBogomolov:
-master
-Berkeley
-Lucene

Sometimes I have an error 
```
java.lang.IllegalArgumentException: Multiple entries with same key: _list_id=[_list_id] and _list_id=[_list_id]
        at com.google.common.collect.ImmutableMap.conflictException(ImmutableMap.java:215) ~[guava-25.1-jre.jar:?]
        at com.google.common.collect.ImmutableMap.checkNoConflict(ImmutableMap.java:209) ~[guava-25.1-jre.jar:?]
        at com.google.common.collect.RegularImmutableMap.checkNoConflictInKeyBucket(RegularImmutableMap.java:147) ~[guava-25.1-jre.jar:?]
        at com.google.common.collect.RegularImmutableMap.fromEntryArray(RegularImmutableMap.java:110) ~[guava-25.1-jre.jar:?]
        at com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:393) ~[guava-25.1-jre.jar:?]
        at org.janusgraph.graphdb.types.indextype.IndexTypeWrapper.getField(IndexTypeWrapper.java:83) ~[janusgraph-core-0.5.0-SNAPSHOT.jar:?]
        at org.janusgraph.graphdb.types.indextype.MixedIndexTypeWrapper.getField(MixedIndexTypeWrapper.java:67) ~[janusgraph-core-0.5.0-SNAPSHOT.jar:?]
        at org.janusgraph.graphdb.database.management.JanusGraphIndexWrapper.getIndexStatus(JanusGraphIndexWrapper.java:85) ~[janusgraph-core-BIC-SNAPSHOT.jar:?]
        at org.janusgraph.graphdb.database.management.GraphIndexStatusWatcher.call(GraphIndexStatusWatcher.java:68) ~[janusgraph-core-0.5.0-SNAPSHOT.jar:?]
```
That error is not 100% reproducible. If you make retry after error, than you will be succeed.
I think that guava ImmutableMap could be changed on simple HashMap.
I can make PR.

PR #1973, opened 2020-02-18
Fix error with field key unique constraint violation

VladimirBogomolov:
Error occurs with berkeley backend

Signed-off-by: Vladimir Bogomolov <vladimir.bogomoloff@gmail.com>

-----
Resolve Issue #1916
-----

### For all changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/indextype/IndexTypeWrapper.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/indextype/IndexTypeWrapper.java
@@ -14,9 +14,12 @@
 
 package org.janusgraph.graphdb.types.indextype;
 
+import java.util.*;
+
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
+import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.janusgraph.core.PropertyKey;
 import org.janusgraph.core.schema.JanusGraphSchemaType;
 import org.janusgraph.graphdb.internal.ElementCategory;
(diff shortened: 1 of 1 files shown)


porunov:
@VladimirBogomolov could you please a test to cover this bugfix?

VladimirBogomolov:
> @VladimirBogomolov could you please a test to cover this bugfix?
Sorry, but this is very difficult. As I wrote in my issue, this error is not 100% reproducible. But I think that due to the fact that BerkeleyJEKeyValueStore::acquLock is not implemented, data race sometimes occurs, which leads to duplication of schema nodes for schema fields. That is why this error occurs.

porunov:
> Sorry, but this is very difficult. As I wrote in my issue, this error is not 100% reproducible. But I think that due to the fact that BerkeleyJEKeyValueStore::acquLock is not implemented, data race sometimes occurs, which leads to duplication of schema nodes for schema fields. That is why this error occurs.

I see. Yes, sometimes this is difficult to write a test case for race conditions. I would just like to understand a typical flow in which this bug happens. As I understand, sometimes we are inserting the same entry in the `ImmutableMap` and it causes conflicts.
This fix will fix the problem, but it would be good if we could understand the root of the problem. Do you know in which cases we might get this bug? As I understand, `getFieldKeys()` returns duplicate values for mixed indexes but it would be good if we could understand why it returns duplicate values and if we can prevent that.

VladimirBogomolov:
> > Sorry, but this is very difficult. As I wrote in my issue, this error is not 100% reproducible. But I think that due to the fact that BerkeleyJEKeyValueStore::acquLock is not implemented, data race sometimes occurs, which leads to duplication of schema nodes for schema fields. That is why this error occurs.
> 
> I see. Yes, sometimes this is difficult to write a test case for race conditions. I would just like to understand a typical flow in which this bug happens. As I understand, sometimes we are inserting the same entry in the `ImmutableMap` and it causes conflicts.
> This fix will fix the problem, but it would be good if we could understand the root of the problem. Do you know in which cases we might get this bug? As I understand, `getFieldKeys()` returns duplicate values for mixed indexes but it would be good if we could understand why it returns duplicate values and if we can prevent that.

I agree that this does not eliminate the root of the problem.
The case where this happens is when I add a lot of indexes and then wait for each index to be enabled after `JanusGraphManagement :: updateIndexStatus`.
Maybe I need to investigate the problem more. I close PR and issue to more details.

PR #1992, opened 2020-02-29
Emphasize that property key cannot be of primitive type [doc only]

### LINKED ISSUE #1983: [Doc only] Emphasize that property key cannot be of primitive type
Primitive types are not allowed to be used as property key:

https://github.com/JanusGraph/janusgraph/blob/d0e5eff1de4e0dcdc95b6292ebe64a3baa357646/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/StandardPropertyKeyMaker.java#L103-L106

It would be nice if we emphasize it in the docs: https://docs.janusgraph.org/basics/schema/#property-key-data-type

li-boxuan:
Closes #1983

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/docs/basics/schema.md
+++ b/docs/basics/schema.md
@@ -105,7 +105,8 @@ Use `dataType(Class)` to define the data type of a property key.
 JanusGraph will enforce that all values associated with the key have the
 configured data type and thereby ensures that data added to the graph is
 valid. For instance, one can define that the `name` key has a String
-data type.
+data type. Note that primitive types are not supported. Use the corresponding
+wrapper class, e.g. `Integer` instead of `int`.
 
 Define the data type as `Object.class` in order to allow any
 (serializable) value to be associated with a key. However, it is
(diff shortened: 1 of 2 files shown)


docs/basics/schema.md
```diff
@@ -105,7 +105,8 @@ Use `dataType(Class)` to define the data type of a property key.
 JanusGraph will enforce that all values associated with the key have the
 configured data type and thereby ensures that data added to the graph is
 valid. For instance, one can define that the `name` key has a String
-data type.
+data type. Note that primitive types are not supported. Use the corresponding
+object type, e.g. Integer instead of int.
```
FlorianHockmann: Please highlight _Integer_  and _int_ as code by using ``: `Integer` and `int`.

Also, are those really called _object types_? I just searched for a term for these and found either _wrapper classes_ or _reference types_.

li-boxuan: Edited. I agree wrapper class or reference type sounds more correct. I was using 'object type' simply because the code said so: https://github.com/JanusGraph/janusgraph/blob/d0e5eff1de4e0dcdc95b6292ebe64a3baa357646/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/StandardPropertyKeyMaker.java#L103-L106

FlorianHockmann: Oh, I didn't compare the two messages. Would be good to have this consistent between docs and the error message so it would be good if you could also update the error message.

li-boxuan: Sure, done

PR #1330, opened 2018-11-29
fixing issue #1125 premature limit application when using multiple indexes in graph query

dk-github:
This fix makes sure that we only apply limit to the query result stream when it is safe to do without a risk of losing valid results. 
Also includes a new unit test which illustrates the issue being fixed in a way easier to reproduce than originally reported on the issue #1125.

Issue #270, opened 2017-05-11
User-defined index prioritization


sharpau:
As a JanusGraph user, I would like to be able to tell JanusGraph to prioritize certain indices higher than others, because I know things about the distribution of the keys in those indices that makes one index particularly slow.

Full background:

We've run into an issue with JanusGraph (preexisting from Titan) with index prioritization (referring to GraphCentricQueryBuilder#constructQueryWithoutProfile). We have 2 indices: a mixed index in ElasticSearch, and an exact-match composite index on two properties. The mixed index is a pretty normal name index. However, the composite index is extremely dense - we basically switched to using this composite index because we kept ending up with supernodes in the graph that made certain vertices impractical to work with, so the composite index may contain millions of entries on only a single-digit number of keys.

Each of these indices work fine on their own, but when you have a query that both indices apply to it gets problematic. In short: we'd like to always hit the mixed index first, because it narrows down the number of results significantly more quickly than the composite index. But the index scoring in query construction automatically gives composite indices a weight of 2000 just for being composite indices. So this means we use the super dense index first and don't really narrow down the number of results any, and it's much slower (whereas if we remove the clauses that match the composite index, the mixed index alone performs fine).

So far the best idea we have to fix this is to only query the mixed index and do the composite index filtering in memory. This is just a) a hassle and b) very error-prone, because when all your queries have a limit, doing any filtering outside of the graph traversal means that you don't know what limit to give to the traversal, and traversals without limits can have problems with the Cassandra backend (i.e. exceeding max frame size).

In my ideal world, I would have some way of telling JanusGraph "hey, when you are considering these two indices, always go to this mixed index first!"

davidclement90:
I have the same kind of issue and I find two ways to deal with it thanks to PR 233.

1) If I want an exact match on my property, I configure my composite index to match equal operator and my mixed index to match textContains operator. Then I set ES mapping with an exact match.

2) If not, I create a fake property. The composite index still match equal operator. The mixed index is composed by the property and the fake property. I use copy_to on the ES mapping to copy the property to the fake property. So I query the fake property to use mixed index.

PR #2040, opened 2020-03-14
Refactor SubqueryIterator and the way we handle multiple index queries

### LINKED ISSUE #2032: Performance issue for multiple index queries when limit is given
`g.V().has("name", "bob").limit(1)` query is very fast, provided index is built for `name`.

However, `g.V().has("name", "bob").has("age", 20).limit(1)` query is slow, provided `name` and `age` have separate indices. Ideally this should not (often) happen because a small limit is given.

I've looked through several relevant PRs:

https://github.com/JanusGraph/janusgraph/pull/481 starts to stream the result from the index backend. It leverages scroll functionality provided by some external index backends to speed up retrieval (especially the first few results). To achieve so, [this](https://github.com/JanusGraph/janusgraph/pull/481/files#diff-022be93bbc23d60a3d1be7dc205a12c3R1280-R1281) divides queries into two parts: first part fed into [SubqueryIterator](https://github.com/JanusGraph/janusgraph/pull/481/files#diff-d3269581786b8b3f748ed9ec0b653afb) directly; the rest of queries go to `QueryUtil.processIntersectingRetrievals` first where actual results are fetched and collected, and then fed into `SubqueryIterator` as well. The SubqueryIterator then does an intersection of results of two parts and returns the iterator.

The above PR introduces a bug: it feeds `QueryUtil.processIntersectingRetrievals` with the limit. As pointed out by https://github.com/JanusGraph/janusgraph/issues/1125, since the total queries are divided into two parts now, we should not use the given limit to retrieve the rest of the queries and do intersection with the first query. Thus this PR https://github.com/JanusGraph/janusgraph/pull/1330 fixes the bug by passing through Infinity as the limit. However, this means when there are more than one index query, we always need to fetch all results (except the first query whose results are streamed). This means the streaming strategy, in this case, does harm to the overall performance, though it is still very useful when there is only one index query.

li-boxuan:
When LimitAdjustingIterator is disabled, limit is given, and
there is more than one index query,  we progressively execute
index queries with proper limit rather than Query.NO_LIMIT. The
optimization in average provides better performance, especially
when the limit is much smaller than total number of indexed elements.

Closes #2032 

-----

As I mentioned in https://github.com/JanusGraph/janusgraph/issues/2032#issuecomment-596083134, this PR tries to reinstate the original query before https://github.com/JanusGraph/janusgraph/pull/481. The difference is this PR tries to estimate the 'appropriate' subLimit in a smarter way and tries to pick up queries that are more likely to be the bottlenecks first. It doesn't mean it is necessarily better than the original logic though.

When only one query is given, or the given limit is infinity, or the LimitAdjustingIterator is enabled, the logic in https://github.com/JanusGraph/janusgraph/pull/481 and https://github.com/JanusGraph/janusgraph/pull/1330 is (mostly) retained. Note that when LimitAdjustingIterator is enabled, we must ensure the order, i.e. limit(i) results = the first i results of limit(j), if i < j. Therefore, we must use the current logic under this circumstance.

There are a few enhancements that can be done in the future, and not included in this PR. I will add more test cases, but before that, I would like to collect some initial feedback.

-----

### For all changes:

### For code changes:

### Note:


janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,80 +17,200 @@
...
+
+    private JointIndexQuery.Subquery streamedQuery;
 
     private boolean isTimerRunning;
 
-    public SubqueryIterator(JointIndexQuery.Subquery subQuery, IndexSerializer indexSerializer, BackendTransaction tx,
-            Cache<JointIndexQuery.Subquery, List<Object>> indexCache, int limit,
-            Function<Object, ? extends JanusGraphElement> function, List<Object> otherResults) {
-        this.subQuery = subQuery;
+    public SubqueryIterator(List<JointIndexQuery.Subquery> queries, IndexSerializer indexSerializer,
```
FlorianHockmann: Codacy Issue found: [The constructor 'SubqueryIterator(List, IndexSerializer, BackendTransaction, Cache, int, Function)' has an NPath complexity of 8333](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=5179944)

FlorianHockmann:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java  26
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=5179944&bid=16999687)

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
FlorianHockmann: Codacy Issue found: [Avoid really long methods.](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=5179944)

li-boxuan:
I wrote a simple program to compare the performance improvement. I used Elasticsearch and Cassandra. The test code is as below:

```java
@Test
public void testMultipleIndexQueryWithHeavyLoad() {
    final PropertyKey name = makeKey("name", String.class);
    final PropertyKey weight = makeKey("weight", Double.class);
    final PropertyKey text = makeKey("text", String.class);
    final PropertyKey flag = makeKey("flag", Boolean.class);

    final JanusGraphIndex composite1 = mgmt.buildIndex("composite1", Vertex.class).addKey(name).buildCompositeIndex();
    final JanusGraphIndex composite2 = mgmt.buildIndex("composite2", Vertex.class).addKey(weight).buildCompositeIndex();
    final JanusGraphIndex mixed1 = mgmt.buildIndex("mixed1", Vertex.class).addKey(text, getTextMapping()).buildMixedIndex(INDEX);
    final JanusGraphIndex mixed2 = mgmt.buildIndex("mixed2", Vertex.class).addKey(flag).buildMixedIndex(INDEX);
    mixed1.name();
    mixed2.name();
    composite1.name();
    composite2.name();
    finishSchema();

    final int numV = 100000;
    final String[] strings = {"houseboat", "humanoid", "differential", "extraordinary"};
    final int modulo = 5;

    for (int i = 0; i < numV; i++) {
        final JanusGraphVertex v = tx.addVertex();
        v.property("name", strings[i % 4]);
        v.property("text", strings[i % 3]);
        v.property("weight", (i % modulo) + 0.5);
        v.property("flag", Boolean.valueOf(i % 2 == 0));
    }
    tx.commit();

    long startTimeMs;
    long measuredTimeMs;

    for (int noOfSubqueries : Arrays.asList(1, 2, 3, 4)) {
        System.out.println("############################");
        System.out.println("Number of index subqueries is: " + noOfSubqueries);
        for (boolean adjustLimit : Arrays.asList(true, false)) {
            clopen(option(ADJUST_LIMIT), adjustLimit);
            System.out.println("ADJUST_LIMIT is: " + adjustLimit);
            for (int limit : Arrays.asList(1, 10, 100, 1000, Query.NO_LIMIT)) {
                startTimeMs = System.currentTimeMillis();
                GraphTraversal<Vertex, Vertex> query = g.V().has("name", "houseboat");
                if (noOfSubqueries > 1) query = query.has("text", Text.textContains("houseboat"));
                if (noOfSubqueries > 2) query = query.has("weight", 0.5);
                if (noOfSubqueries > 3) query = query.has("flag", true);
                Long count = query.limit(limit).count().next();
                measuredTimeMs = System.currentTimeMillis() - startTimeMs;
                System.out.println(limit + " limit search time: " + measuredTimeMs + ", count = " + count);
                g.tx().rollback();
            }
        }
    }
}
```

And here are the statistics for running the program above:

Throughput improvement with LimitAdjustingIterator enabled
| Limit    | 1 query | 2 queries | 3 queries | 4 queries |
|----------|---------|-----------|-----------|-----------|
| 1        | -21.70% | 15.17%    | -2.94%    | -2.07%    |
| 10       | -37.50% | 10.54%    | -3.60%    | 7.95%     |
| 100      | -27.78% | 3.85%     | -0.02%    | 6.98%     |
| 1000     | -7.89%  | 8.48%     | 4.07%     | -0.64%    |
| NO_LIMIT | 24.06%  | 20.57%    | 5.20%     | -0.22%    |

image

Basically, the logic is almost the same as the current logic when LimitAdjustingIterator is enabled, thus the performances are almost the same.

Throughput improvement with LimitAdjustingIterator disabled
| Limit    | 1 query | 2 queries | 3 queries | 4 queries |
|----------|---------|-----------|-----------|-----------|
| 1        | 14.29%  | 668.12%   | -2.97%    | 2217.79%  |
| 10       | 14.29%  | 437.00%   | -2.87%    | 127.17%   |
| 100      | 14.29%  | -8.90%    | -23.81%   | 0.90%     |
| 1000     | 20.00%  | -28.85%   | -0.56%    | -1.40%    |
| NO_LIMIT | 9.60%   | 14.11%    | 13.40%    | -2.66%    |

image

The improvement can be huge when the limit is small.

I only ran the program on my laptop once, so the statistics above are just for reference only. I would be happy to run more tests if any other specific use case is suggested by reviewers.

FlorianHockmann:


```diff

Complexity increasing per file
==============================
- janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=5179944&bid=16999687)

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+    /**
+     * Compute score for each query. Lower score means the query is more selective and more likely to
+     * be the bottleneck. Exhausted queries always have 0 score. Unless more factors are taken into consideration,
+     * at the moment it does not make sense to compare queries if we only have two.
+     */
+    private double[] computeScores(final int resultsExhaustedCount, boolean[] resultsExhausted,
+                                   int[] offsets, final int numberOfQueries, final int numberOfResults,
+                                   final int limit, Map<Object, List<Integer>> subResultToQueryMap) {
+        double[] scores = new double[numberOfQueries];
+        // TODO: take more factors into consideration, e.g. query latency
```
li-boxuan: If we can find a way to let users define score/priority for indices, we can solve https://github.com/JanusGraph/janusgraph/issues/270 here

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1321,24 +1320,13 @@ else if (query.getResultType() == ElementCategory.VERTEX) {
...
-                        }
-                    });
-                }
-                // Constructs an iterator which lazily streams results from 1st index, and filters by looking up in the intersection of results from all other indices (if any)
-                // NOTE NO_LIMIT is passed to processIntersectingRetrievals to prevent incomplete intersections, which could lead to missed results
-                iterator = new SubqueryIterator(indexQuery.getQuery(0), indexSerializer, txHandle, indexCache, indexQuery.getLimit(), getConversionFunction(query.getResultType()),
-                        retrievals.isEmpty() ? null: QueryUtil.processIntersectingRetrievals(retrievals, Query.NO_LIMIT));
+                // Constructs an iterator which lazily streams results if only one query is given, otherwise filters results
+                // by doing an intersection from all subQueries. When there is more than one query and adjustQuery is
+                // enabled, we must use NO_LIMIT to ensure result orders remain unchanged among executions.
```
li-boxuan: It seems that when backendQueryHolder.isSorted is true, we also MUST ensure the order of overall results conforms to the order the first subquery results.  Appreciate it if anyone can confirm.

rngcntr: https://github.com/JanusGraph/janusgraph/blob/b2a6e8125b83038873db19ed62b43a0f6190d812/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/BackendQueryHolder.java#L27-L28

This looks to me like `isSorted` causes the responsible logic to skip sorting the results as it was originally requested by the user, because we somehow know that the backend will produce the correct order on it's own. This means we are not allowed to alter the ordering of elements at any later point in time.

Please notice that although I strongly support your consideration, I can not definitively confirm it.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+            QueryProfiler profiler = subQuery.getProfiler();
+            QueryProfiler.startProfile(profiler, subQuery);
+            // TODO: leverage the scrolling capability of indexing backends ratherthan throw away old results
+            indexSerializer.query(subQuery, tx).skip(offsets[idx]).forEachOrdered(result -> {
+                offsets[idx]++;
+                subResultToQueryMap.computeIfAbsent(result, k -> new ArrayList<>()).add(idx);
+            });
             profiler.stopTimer();
+            profiler.setResultSize(offsets[idx]);
+            // TODO: should we put it into cache?
```
rngcntr, line 206: I don't really have an opinion on that but I think it's at least worth a discussion.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+        // TODO: take more factors into consideration, e.g. query latency
+        if (resultsExhaustedCount < numberOfQueries && numberOfResults < limit && numberOfQueries > 2) {
+            // A query whose results have many intersections with other queries is less likely to be selective.
+            // A query with few intersections with other queries is more valuable in the sense that it contains
+            // less redundant information, and thus is more useful.
+            // In other words, if multiple index queries have huge overlaps with each other except one query,
+            // then that query is more likely to be the bottleneck.
+            for (List<Integer> queryNoList : subResultToQueryMap.values()) {
+                for (int idx : queryNoList) {
+                    if (!resultsExhausted[idx]) scores[idx] += Math.log(queryNoList.size());
```
rngcntr, line 243: I like this approach! Did you come up with it yourself or is there any scientific research that suggests this as a measure of selectivity?

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+                } else {
+                    assert otherResults != null;
+                    otherResults.retainAll(subResults);
+                }
+            } catch (Exception e) {
+                throw new JanusGraphException("Could not execute index query", e.getCause());
+            }
+        }
+
+        Set<Object> others = otherResults;
```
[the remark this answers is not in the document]
li-boxuan: Yeah, because the variable used in lambda function must be final or effectively final

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+        // combine the results of lazily streamed first query and the results of rest queries
+        elementIterator = stream.filter(e -> others == null || others.contains(e))
+            .map(conversionFunction).map(r -> (JanusGraphElement) r).iterator();
+    }
+
+    private void intersectAll(List<JointIndexQuery.Subquery> queries, IndexSerializer indexSerializer,
+                              BackendTransaction tx, final int limit,
+                              Function<Object, ? extends JanusGraphElement> conversionFunction) {
+        final int multiplier = Math.min(128, (int) Math.pow(2, Math.pow(2, queries.size() - 1)));
```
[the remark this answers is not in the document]
li-boxuan: Thanks for reminding me of the overflow problem! Regarding this heuristic - this was created by the original author, I was just moving this line around and changed the constant a bit.

Previously it was:

```java
final int multiplier = Math.min(16, (int) Math.pow(2, retrievals.size() - 1));
```

Interestingly, this logic has been in the codebase for years. I guess in practice people don't really have 6 or more index queries in one single query.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+        int baseSubLimit = Math.min(limit * multiplier, Query.NO_LIMIT);
```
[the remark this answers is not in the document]
li-boxuan: Thanks. It was not only redundant, but also wrong; I didn't take overflow problem into consideration.  The author's original code (below) seems correct though:

```java
        int subLimit = Integer.MAX_VALUE;
        if (Integer.MAX_VALUE / multiplier >= limit) subLimit = limit * multiplier;
```

I was just trying to make it more succinct, without noticing I actually introduced a new bug

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
li-boxuan: Yeah, there is one point I am unsure of, but I didn't have time to look into yet: if I put the first 5000 results of index query X into the cache, then when another index query X with limit 1000 comes, will that be effectively loaded from the cache? In other words, does the cache do exact match comparison or subset comparison based on the limit?

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
li-boxuan: I came up with this by myself based on intuition. My thoughts might be wrong though.

My basic assumption is that no index is perfect; there are indexes that more or less store some sort of 'redundant' information. In the extreme case, some indexes might return exactly the same results.

li-boxuan: In fact, now I doubt if we should do such selection at all since we can possibly execute all backend queries in parallel (I don't understand why the original author did not do this in the first place).

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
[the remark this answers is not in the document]
rngcntr: Have you tried other values or just the old one and this one? I'm curious about how much of a difference this makes in terms of performance gain.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
rngcntr: I don't know anything about the cache implementation either, but maybe there's someone else here who does ;)

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
rngcntr: Parallelization sounds like a good idea here. Would you like to address it in this PR or rather leave it as a TODO?

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -17,81 +17,236 @@
...
+    }
+
+    private void intersectResults(Map<Object, List<Integer>> subResultToQueryMap,
+                                  List<Object> results, final int limit, final int numberOfQueries) {
+        for (Iterator<Map.Entry<Object, List<Integer>>> it = subResultToQueryMap.entrySet().iterator();
+             it.hasNext(); ) {
+            Map.Entry<Object, List<Integer>> entry = it.next();
+            if (entry.getValue().size() == numberOfQueries) {
+                // this particular result satisfies every index query
+                results.add(entry.getKey());
```
rngcntr, line 218: How do you ensure an entry gets only added to the result list once here? `intersectResults` is called in every iteration of the outer loop in `intersectAll`. But in the (N+1)st iteration, `subResultToQueryMap` still contains all the valid results that were found in the N-th iteration or am I missing something?

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
li-boxuan: Not sure, I would like to hear more voice on this. A possible reason for not using parallelization is due to the additional overhead, especially when the overall read load is very heavy, even though latency for a single read request can be reduced drastically by parallelization.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
li-boxuan: On [line 220](https://github.com/JanusGraph/janusgraph/pull/2040/files/7de65fb3da9fec3b9c05ab9c18c6b91dd55d0204#diff-d3269581786b8b3f748ed9ec0b653afbR220) I removed it from `subResultToQueryMap`

rngcntr: I should have looked just two lines further... :sweat_smile: 
Resolved!

PR #2055, opened 2020-03-22
GraphCentricQueryBuilder: Optimize index selection algorithm

### LINKED ISSUE #2048: Optimize index select algorithm to reduce number of index queries
Sometimes an additional composite index query is picked up even if a mixed index query itself is sufficient. See the following example:

```java
graph = JanusGraphFactory.open('conf/janusgraph-cql-es.properties')

mgmt = graph.openManagement()

name = mgmt.makePropertyKey('name').dataType(String.class).make()
nickname = mgmt.makePropertyKey('nickname').dataType(String.class).make()


mgmt.buildIndex('composite', Vertex.class).addKey(name).buildCompositeIndex()
mgmt.buildIndex('mixed',Vertex.class).addKey(name,Mapping.STRING.asParameter()).addKey(nickname,Mapping.STRING.asParameter()).buildMixedIndex('search')

mgmt.commit()


g = graph.traversal()
g.V().has('name', 'Robert').has('nickname', 'Bob').profile()
```

The result shows that it creates a JointIndexQuery consisting of two subqueries, one using Composite Index and the other one using Mixed Index:

```
==>Traversal Metrics
JanusGraphStep([],[name.eq(Robert), nickname.eq...                                           135.230   100.00
    \_condition=(name = Robert AND nickname = Bob)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=[(name = Robert AND nickname = Bob)]:mixed
    \_index=mixed
    \_index_impl=search
  optimization                                                                                 0.023
  optimization                                                                                35.363
  backend-query                                                        0                      89.864
    \_query=mixed:[(name = Robert AND nickname = Bob)]:mixed
  backend-query                                                        0                       1.583
    \_query=composite:multiKSQ[1]@2147483647
                                            >TOTAL                     -           -         135.230        -
```

Given that in this case, the mixed index already covers all conditions, I believe there is no need to fetch the composite index. Plus, a 'multiple indices' query is much more expensive than a 'single index' query.

This is because Composite Index always has a very high score in GraphCentricQueryBuilder. If we have two composite indices, one for `name` and one for `nickname` (meaning that composite indices can cover all conditions), then two composite indices might be better than a single mixed index. However, in the case that one single mixed index can cover all conditions, it does not make sense to involve a composite index.

In general, if an index A is necessary (in the sense that some condition can only be covered by it), and it is a superset of some other index B (regardless of whether it's composite or mixed), then B can be removed from the JointIndexQuery. This strategy can solve the problem mentioned beforehand but can fail to deal with situations that are more complicated.

To make it clearer, let's assume the conditions given by the user are condition 1, condition 2 &  condition 3. Consider three index candidates: 

- composite index A that covers condition 1
- mixed index B that covers condition 1 & 2
- mixed index C that covers condition 2 & 3

According to the current scoring scheme, composite index A will firstly be picked up (composite index is preferred). The next index query that gets selected can be B or C. C is better in the sense that it covers two additional conditions, while B only covers one, but let's just assume B is better in some other way (e.g. support sorting). Then B is selected after A. Finally we still have to pick up index C because we yet to cover condition 3.

So finally, all three indices are selected. However, we can see that the real optimal index combination shall be composite index A + mixed index C. Although B shall be removed, it is not a subset of any other index. Now we can see this problem is indeed the famous [Set Cover Problem](https://en.wikipedia.org/wiki/Set_cover_problem). To be more precise, it is the weighted set cover problem.

li-boxuan:
This intends to optimize index selection algorithm by introducing
a brute force enumeration way to select appropriate indexes in
GraphCentricQueryBuilder.

Closes #2048 

-----

I will add test cases, but any early feedback would be very helpful.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
@@ -18,11 +18,12 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
+import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.janusgraph.core.*;
 import org.janusgraph.core.attribute.Cmp;
 import org.janusgraph.core.attribute.Contain;
-import org.janusgraph.core.schema.SchemaStatus;
 import org.janusgraph.core.schema.JanusGraphSchemaType;
+import org.janusgraph.core.schema.SchemaStatus;
 import org.janusgraph.graphdb.database.IndexSerializer;
 import org.janusgraph.graphdb.internal.ElementCategory;
 import org.janusgraph.graphdb.internal.InternalRelationType;
@@ -34,6 +35,8 @@ import org.janusgraph.graphdb.query.profile.QueryProfiler;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.graphdb.types.*;
 import org.janusgraph.graphdb.types.system.ImplicitKey;
+import org.janusgraph.graphdb.util.IndexCandidate;
+import org.janusgraph.graphdb.util.IndexCandidateGroup;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -49,7 +52,6 @@ import java.util.stream.StreamSupport;
 public class GraphCentricQueryBuilder implements JanusGraphQuery<GraphCentricQueryBuilder> {
 
     private static final Logger log = LoggerFactory.getLogger(GraphCentricQueryBuilder.class);
-
     /**
      * Transaction in which this query is executed.
      */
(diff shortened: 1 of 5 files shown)


FlorianHockmann:


```diff

Complexity increasing per file
==============================
- janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexCandidate.java  1
- janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexCandidateGroup.java  6
         

Complexity decreasing per file
==============================
+ janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java  -25
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=5223020&bid=17124290)

FlorianHockmann:


```diff

Complexity increasing per file
==============================
- janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexCandidate.java  1
- janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexCandidateGroup.java  6
         

Complexity decreasing per file
==============================
+ janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java  -25
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/JanusGraph/janusgraph/pullRequest?prid=5223020&bid=17124290)

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexCandidate.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public Object getSubCondition() {
+        return subCondition;
+    }
+
+    public double getScore() {
+        return score;
+    }
+
+    public void setScore(double score) {
```
rngcntr, line 57: Isn't it a cleaner separation of logical components if the score calculation is done inside the `IndexCandidate` class?
Instead of setting the code from anywhere outside, it can be calculated by the candidate itself.

This would also enable us to move all `SCORE` constants to the `IndexCandidate` class.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -490,5 +409,189 @@ private static boolean coversAll(final MixedIndexType index, Condition<JanusGrap
...
+        return jointQuery;
+    }
+
+    /**
+     * Validate given index candidate and enrich it
+     * @param index
+     * @param conditions
+     * @return
+     */
+    private IndexCandidate enrichIndex(final IndexType index, final MultiCondition<JanusGraphElement> conditions) {
```
rngcntr, line 538: It looks like the only purpose of this method is to construct an instance of `IndexCandidate` from an `IndexType` and a `MultiCondition`, am I right?
If so, it would make sense to use this as a constructor of `IndexCandidate`.

li-boxuan:
@rngcntr Thanks for your valuable reviews! Due to some pending compliance issue, I was suggested to put open source contribution on hold for a while, thus I cannot update my commit as of now.

Apart from stylish changes, do you have any more comment on the functional aspect? I’d like to hear more about how people like this approach in general.

rngcntr:
I admit I don't have a sense for how much difference there is between the approximation and the exact solution. Especially because even the exact solution is heavily bases on estimations of scoring features itself.
Therefore, I felt unable to review your additions from a functional standpoint.

Do you have a rough schedule on when you are able to contribute again? I appreciate your recent ideas and the effort you put into implementing them.

li-boxuan:
> Do you have a rough schedule on when you are able to contribute again

It's difficult to say during this pandemic period, but in case anyone would like to work on top of my commit, please feel free to take anything useful. That being said, I am hoping to be back not too long after.

li-boxuan:
Closing this PR as there is a better one (https://github.com/JanusGraph/janusgraph/pull/2099) on top of this

PR #2067, opened 2020-04-01
Fix gremlin-server should run inmemory by default

porunov:

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -65,6 +65,13 @@ The default version of the distribution package does no longer contain the `janu
 This includes a packaged version of cassandra and elasticsearch. If you want to have `janusgraph.sh`, 
 you have to download distribution with the suffix `-full`.
 
+##### Gremlin Server distributed with the release uses inmemory storage backend and no search backend by default
+
+Gremlin Server is by default configured for the inmemory storage backend and no search backend when started with 
+`bin/gremlin-server.sh`.  
+You can provide configuration for another storage backend and/or search backend by providing a path to the appropriate 
+configuration as a second parameter (`./bin/gremlin-server.sh ./conf/gremlin-server/[...].yaml`).
+
 ### Version 0.5.0 (Release Date: March 10, 2020)
 
 ```xml tab='Maven'
--- a/janusgraph-dist/src/assembly/static/conf/gremlin-server/gremlin-server.yaml
+++ b/janusgraph-dist/src/assembly/static/conf/gremlin-server/gremlin-server.yaml
@@ -17,7 +17,7 @@ port: 8182
 scriptEvaluationTimeout: 30000
 channelizer: org.apache.tinkerpop.gremlin.server.channel.WebSocketChannelizer
 graphs: {
-  graph: conf/gremlin-server/janusgraph-cql-es-server.properties
+  graph: conf/janusgraph-inmemory.properties
 }
 scriptEngines: {
   gremlin-groovy: {
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -115,11 +115,13 @@ $ ./bin/gremlin-server.sh start
 ```
 or
 ```bash
-$ ./bin/gremlin-server.sh /conf/gremlin-server/[...].yaml
+$ ./bin/gremlin-server.sh ./conf/gremlin-server/[...].yaml
 ```
 
 !!! info
-    By default, this will instruct JanusGraph to use it's own inmemory backend instead of a dedicated database server.
+    The default configuration (`gremlin-server.yaml`) uses it's own inmemory backend instead of a dedicated database server.
+    No search backend is used by default, so mixed indices aren't supported as search backend isn't specified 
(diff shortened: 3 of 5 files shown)


janusgraph-dist/src/assembly/static/conf/gremlin-server/gremlin-server.yaml
```diff
@@ -17,7 +17,7 @@ port: 8182
 scriptEvaluationTimeout: 30000
 channelizer: org.apache.tinkerpop.gremlin.server.channel.WebSocketChannelizer
 graphs: {
-  graph: conf/gremlin-server/janusgraph-cql-es-server.properties
```
farodin91, line 20: I think we should think of removing all different gremlin-server.yaml and just have a single the user just have to decide between `janusgraph-*.properties` files. This would be a major breaking change.

Besides, this is also a breaking change and should be mentioned in the changelog.

janusgraph-dist/src/assembly/static/conf/gremlin-server/gremlin-server.yaml
```diff
@@ -17,7 +17,7 @@ port: 8182
...
+  graph: conf/gremlin-server/janusgraph-inmemory.properties
```
farodin91, line 20: Why not use conf/janusgraph-inmemory.properties?

porunov: Changed to `conf/janusgraph-inmemory.properties`

janusgraph-dist/src/assembly/static/conf/gremlin-server/gremlin-server.yaml
porunov: Agree. We should think about removing them but I think we should not spend much time on the current release and think about adding those changes in future releases.
I have added a not to the changelog

farodin91: If you change this behaviour here, you have to update janusgraph.sh. The janusgraph.sh uses the gremlin-server.yaml and expects cql-es is started.

porunov:
@farodin91 what do you mean by that? Do you have problems with running `janusgraph.sh`? I have tested it on Fedora and MacOS. Works as expected.

farodin91:
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-dist/src/assembly/static/bin/janusgraph.sh#L174
The janusgraph.sh would run through but you would connect to a inmemory database.

porunov:
interesting.. I didn't research that. Did you test it? I am executing it right now:
```
gremlin> g = graph.traversal()
==>graphtraversalsource[standardjanusgraph[cql:[127.0.0.1]], standard]
gremlin> g.E().has('place', geoWithin(Geoshape.circle(37.97, 23.72, 50)))
==>e[368-39c-9hx-j1c][4224-battled->24672]
==>e[2s0-39c-9hx-394][4224-battled->4216]
```
Looks like both cql and es is correctly used.
Notice, my `gremlin-server.yaml` contains `graph: conf/gremlin-server/janusgraph-inmemory.properties` as stated in this PR.

farodin91:
If you have to connect remotely to the started gremlin-server.yaml, not just start a local janusgraph in your console.

janusgraph-dist/src/assembly/static/conf/gremlin-server/gremlin-server.yaml
FlorianHockmann: I think the best solution for this PR would be to let `janusgraph.sh` start `gremlin-server.sh` explicitly with a `cql-es` YAML. Since cql-es was the default until now, there is no YAML for that. So, we should create another one and then use that in `janusgraph.sh`.
That way, users of `janusgraph.sh` wouldn't notice a difference and `gremlin-server.sh` can work without CQL and ES in the truncated distribution.

docs/changelog.md
```diff
@@ -65,6 +65,11 @@ The default version of the distribution package does no longer contain the `janu
 This includes a packaged version of cassandra and elasticsearch. If you want to have `janusgraph.sh`, 
 you have to download distribution with the suffix `-full`.
 
+##### Gremlin server distributed with the release uses inmemory storage backend and no search backend
+
+Gremlin server distributed with the release uses inmemory storage backend and no search backend that is why
```
FlorianHockmann, line 70: I think this section is a bit ambiguous as users could understand this as: _Gremlin Server now only works with inmemory_ whereas what you actually mean is that Gremlin Server is by default configured for the inmemory storage backend when started with `bin/gremlin-server.sh`.

Another nitpick here: It should always be _Gremlin **S**erver_.

porunov:
@FlorianHockmann @farodin91 thank you for suggestions!
I have applied your requested changes. I have tested `gremlin-server.sh` - by default uses inmemory storage backend. Tested `janusgraph.sh` localy, uses cql and es. Tested `janusgraph.sh` remotely, uses cql and es. Loading `GraphOfTheGodsFactory` works as expected, basic queries and queries against mixed index works also as expected.

porunov:
Updated release jars

PR #2080, opened 2020-04-15
Remove redundant data structure in CQL adapter

rngcntr:
The CQL storage adapter currently parses results retrieved from the backend into an array from which they are further evaluated by some method outside the storage adapter.

https://github.com/JanusGraph/janusgraph/blob/db8bcb1c43c1f500d71a71d7f0666c6f9a03917c/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java#L318

This structure causes a huge overhead, because all data is now stored both in the ResultSet and in the ArrayList. Instead, we can use a custom Iterator which wraps the `ResultSet`'s iterator and returns the results one-by-one when needed.
This improvement has a massive impact on the performance of all range reads on CQL backends:

edgeCount

For both implementations, counting the neighbors of a vertex serves as a query resulting in a large range read. For large instances (>900k adjacent vertices) the gain in performance is very clear to see: While the current implementation fails to return results within the default limit of 30s, the improved solution is stable even for more than twice as large instances.

But taking a closer look, even smaller instances show an increase in performance of about 25%

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
@@ -46,6 +46,7 @@ import static org.janusgraph.diskstorage.cql.CQLConfigOptions.COMPACTION_OPTIONS
 import static org.janusgraph.diskstorage.cql.CQLConfigOptions.COMPACTION_STRATEGY;
 import static org.janusgraph.diskstorage.cql.CQLTransaction.getTransaction;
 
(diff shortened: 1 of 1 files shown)


farodin91:
Could you rebase on 0.5? 0.2-0.4 are no longer supported.

porunov:
@rngcntr Thank you for the contribution! I have changed target branch to `v0.5`. Would you be able to resolve conflicts?

rngcntr:
I thought we would maybe merge this into older versions anyways but I'll do a rebase.

PR #2088, opened 2020-04-17
Fix global config overriding

mad:
When evaluate discrepancies options we need check for existense, otherwise we catch exception when local config try to pass new one managed option with disallow empty predicate

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilder.java
@@ -248,7 +248,12 @@ public class ReadConfigurationBuilder {
             Object localValue = entry.getValue();
 
             // Get the storage backend's setting and compare with localValue
-            Object storeValue = globalWrite.get(configOption, pathId.umbrellaElements);
+            Object storeValue;
+            if (globalWrite.has(configOption, pathId.umbrellaElements)) {
+                storeValue = globalWrite.get(configOption, pathId.umbrellaElements);
+            } else {
+                storeValue = configOption.getDefaultValue();
+            }
 
             // Check if the value is to be overwritten
             if (overwrite.has(configOption, pathId.umbrellaElements))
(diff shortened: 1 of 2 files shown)


mad:
Existing test modified for test this fix

porunov:
@mad I just trying to figure out what is the root of this bug.
I see that the exception is thrown here: https://github.com/JanusGraph/janusgraph/blob/0668d7f6dd3ff428e9711e2b2b4d10dc85bc3e42/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ConfigOption.java#L229

I am just trying to figure out the root problem here. I assume, that if `defaultValue` is `null` and `verificationFct.apply((O) input)` is `false`, than we should not use that config option because it goes against the current logic (which maybe contains a bug).

Currently I am checking that we are getting `false` from `verificationFct.apply((O) input)` due to `disallowEmpty` checker:
https://github.com/JanusGraph/janusgraph/blob/0668d7f6dd3ff428e9711e2b2b4d10dc85bc3e42/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ConfigOption.java#L261

Thus, if we force to use `storeValue = configOption.getDefaultValue();` it means that we just ignore that check. Wouldn't it break some logic?
It would be really helpful to see the configuration which you are trying to use and you are getting that error.

mad:
@porunov This `storeValue = configOption.getDefaultValue();` need only for determine conflict options, not for real usages

Example of code for clarification

```
graph = JanusGraphFactory.build().
set("storage.backend","berkeleyje").
set("storage.directory","/tmp/graph").
open()
graph.close()

graph = JanusGraphFactory.build().
set("storage.backend","berkeleyje").
set("storage.directory","/tmp/graph").
set("attributes.custom.attribute10.attribute-class", "SomeClass").
set("attributes.custom.attribute10.serializer-class", "SomeClass").
open()
```

will cause of
```
Need to set configuration value: root.attributes.custom.attribute-class
Type ':help' or ':h' for help.
Display stack trace? [yN]y
java.lang.IllegalStateException: Need to set configuration value: root.attributes.custom.attribute-class
```

but should log warn by default https://github.com/JanusGraph/janusgraph/blob/051c56053c398a7b032838908afb4e7bee7c6c4f/janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilder.java#L264-L272

PR #2099, opened 2020-04-27
GraphCentricQueryBuilder: Optimize index selection algorithm [continuation]

rngcntr:
As @li-boxuan [said](https://github.com/JanusGraph/janusgraph/pull/2055#issuecomment-615805172), his schedule does currently not allow any contribution to JanusGraph.
Thus, this PR is intended to continue @li-boxuan's work on #2055 to solve #2048.

-----

### For all changes:

### For code changes:

### Note:


janusgraph-core/src/main/java/org/janusgraph/util/datastructures/SetUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author Florian Grieskamp (Florian.Grieskamp@gdata.de)
+ */
+public class SetUtil {
+    
+    public static <T> Set<Set<T>> powerSet(final Set<T> originalSet) {
```
li-boxuan, line 25: Can we maybe follow guava's powerSet contract? They look reasonable to me and nice to have IMHO.
```
IllegalArgumentException: If set has more than 30 unique elements, because this causes the power set size to exceed the int range.
NullPointerException: If set is or contains null.
```

janusgraph-core/src/main/java/org/janusgraph/util/datastructures/SetUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+        if (originalSet.isEmpty()) {
+            return resultSets;
+        }
+
+        Set<T> remaining = new HashSet<T>(originalSet);
+        T first = remaining.iterator().next();
+        remaining.remove(first);
+
+        for (Set<T> intermediateSet : powerSet(remaining)) {
```
li-boxuan, line 39: Did you explore Guava's PowerSet implementation? Their implementation seems to be more efficient: https://www.baeldung.com/java-power-set-of-a-set#optimization, and more importantly, it only uses O(n) space by utilizing the Iterator interface.

li-boxuan:
Just did a simple test using the same settings as in https://github.com/JanusGraph/janusgraph/issues/2048#issue-583853695, now no additional composite index query is picked up (better than old behaviour):
```gremlin
gremlin> g.V().has('name', 'Robert').has('nickname', 'Bob').profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(Robert), nickname.eq...                                           159.248   100.00
    \_condition=(name = Robert AND nickname = Bob)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=[(name = Robert AND nickname = Bob)](4000):mixed
    \_index=mixed
    \_index_impl=search
  optimization                                                                                 0.031
  optimization                                                                                56.180
  backend-query                                                        0                      96.603
    \_query=mixed:[(name = Robert AND nickname = Bob)](4000):mixed
    \_limit=4000
                                            >TOTAL                     -           -         159.248        -
```

When one composite index is sufficient, mixed index will not be selected (the same as previous behaviour): 

```gremlin
gremlin> g.V().has('name', 'Robert').profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(Robert)])                                                           5.770   100.00
    \_condition=(name = Robert)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@2000
    \_index=composite
  optimization                                                                                 0.023
  optimization                                                                                 3.494
  backend-query                                                        0                       2.016
    \_query=composite:multiKSQ[1]@2000
    \_limit=2000
                                            >TOTAL                     -           -           5.770        -
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -265,6 +265,13 @@
                     "performance improvement if there is a non-trivial latency to the backend.",
             ConfigOption.Type.MASKABLE, false);
 
+    public static final ConfigOption<Integer> INDEX_SELECT_BRUTE_FORCE_THRESHOLD = new ConfigOption<>(QUERY_NS, "index-select-threshold",
+            "Threshold of deciding whether to use brute force enumeration algorithm or fast approximation algorithm " +
+                    "for selecting suitable indexes. Selecting optimal indexes for a query is a NP-complete set cover problem. " +
+                    "When number of suitable index candidates is no larger than threshold, JanusGraph uses brute force search " +
+                    "with exponential time complexity to ensure the best combination of indexes is selected.",
+            ConfigOption.Type.MASKABLE, 10);
+
```
porunov, line 274: I didn't check the code yet, but have general questions about this optimization.
1) Do I understand it correctly that if we use `index-select-threshold = 0` it means that current logic (which is currently in `master` branch) will be used? I.e., as I understand, if we use `0` it means that the behaviour will be the same as in current `master` branch and complexity of the current algorithm to select indexes will not be increased.

2) If `1)` is `yes` than I see that this value is set by default to `10`. It means that it is a breaking change. If so, we should describe this breaking change here https://github.com/JanusGraph/janusgraph/blob/master/docs/changelog.md#upgrade-instructions

3) Is it suggested to use `10` here or should user somehow decide what value to use here? Just interested, if there is 10 or less candidates, is it faster to use `brute force` algorithm or is it just sufficiently fast so that it is OK to sacrifice candidate selection performance if favour to choosing a good candidate?

li-boxuan: Yes, if set to 0, then the existing logic in current master branch will be used.
10 is an arbitrary number picked by me originally. My primary purpose was to pick up a threshold that is "just sufficiently fast so that it is OK to sacrifice candidate selection performance if favour to choosing a good candidate". We could make it even smaller just to be safe it will make as least impact as possible.

In practice, I think the chance is small that people have many index candidates (e.g. >= 5) for one index query.

janusgraph-core/src/main/java/org/janusgraph/util/datastructures/SetUtil.java
rngcntr: I refactored the code, it should not exceed the O(n) space bound now

rngcntr:
I will squash the commits as soon as review is finished

rngcntr:
Speaking of squashing the commits, is there a policy on how to handle commits by multiple authors? If I squash them into one commit, git will see this as "my work" although it's mostly @li-boxuan's idea.

porunov:
@rngcntr The only policy is that commits should be signed. If there are several commits from different authors, it is completely fine. 
We are asking to use 1 commit per PR not because of some policy but because of cleaner commits history.
If your PR requires several commits, that's totally fine but most of the time all PRs should be a single commit which represents a contribution. The logic behind this is the next:
1) It is much easier for reviewers to review a single commit instead of multiple commits. For example, some malicious code could be added between commits and hidden by adding and removing it in different commits, thus we are asking to squash commits to not spending time checking each commit separately.
2) Cleaner history. Easier to understand the contribution when searching for specific commits.

In this PR it is totally fine to have several commits from different authors as they are related to a single contribution. In this scenario it isn't a problem to check several commits for such a big contribution

rngcntr:
> 1. As per [this commit](https://github.com/JanusGraph/janusgraph/pull/2099#discussion_r419153145), I am suggesting to add a section into [upgrade instructions](https://github.com/JanusGraph/janusgraph/blob/master/docs/changelog.md#upgrade-instructions) which describes new index selection behaviour and what should user do to preserve old behaviour.

Check

li-boxuan:
I did an experiment on my laptop (Macbook pro 15' 2015), using Berkeley DB and Elasticsearch, and here is the comparison result:

image

| index candidates size | selection time of brute force | selection time of greedy algo | total time of brute force  | total time of greedy algo |
|-----------------------|-------------------------------|-------------------------------|----------------------------|---------------------------|
| 2                     | 3                             | 3                             | 812                        | 1036                      |
| 3                     | 4                             | 5                             | 831                        | 1003                      |
| 4                     | 6                             | 5                             | 836                        | 1107                      |
| 5                     | 6                             | 5                             | 870                        | 1096                      |
| 6                     | 10                            | 7                             | 910                        | 979                       |
| 7                     | 11                            | 9                             | 875                        | 1023                      |
| 8                     | 15                            | 12                            | 1000                       | 1142                      |
| 9                     | 17                            | 10                            | 940                        | 1105                      |
| 10                    | 24                            | 13                            | 967                        | 1098                      |
| 11                    | 19                            | 23                            | 880                        | 1124                      |
| 12                    | 20                            | 12                            | 945                        | 1163                      |
| 13                    | 32                            | 12                            | 1024                       | 1148                      |
| 14                    | 69                            | 17                            | 1013                       | 1256                      |
| 15                    | 98                            | 14                            | 967                        | 1025                      |
| 16                    | 143                           | 12                            | 1192                       | 1121                      |
| 17                    | 260                           | 8                             | 1122                       | 901                       |
| 18                    | 516                           | 13                            | 1390                       | 1005                      |
| 19                    | 695                           | 19                            | 1409                       | 1103                      |
| 20                    | 1474                          | 19                            | 2269                       | 1370                      |
| 21                    | 2540                          | 14                            | 3350                       | 939                       |

Code snippet that I used to generate the above results (parameters: numOfAdditionalProps and threshold).
```java
@Test
public void testMultipleIndexQueryWithHeavyLoad() {
    final int numOfAdditionalProps = 1;
    final int threshold = Integer.MAX_VALUE;
//    final int threshold = 0;

    final PropertyKey name = makeKey("name", String.class);
    final JanusGraphIndex compositeNameIndex = mgmt.buildIndex("composite1", Vertex.class).addKey(name).buildCompositeIndex();
    compositeNameIndex.name();

    for (int i = 0; i < numOfAdditionalProps; i++) {
        final PropertyKey prop = makeKey("prop" + i, String.class);
        final JanusGraphIndex mixedIndex = mgmt.buildIndex("mixed" + i, Vertex.class)
            .addKey(name, Mapping.STRING.asParameter())
            .addKey(prop, Mapping.STRING.asParameter()).buildMixedIndex(INDEX);
        mixedIndex.name();
    }

    finishSchema();

    final int numV = 20000;
    List<String> strings = new ArrayList<>();
        strings.add("value" + i);
    }
    final int modulo = 5;

    for (int i = 0; i < numV; i++) {
        final JanusGraphVertex v = tx.addVertex();
        v.property("name", strings.get(0));
        for (int j = 0; j < numOfAdditionalProps; j++) {
            v.property("prop" + j, strings.get(i % 100));
        }
    }
    tx.commit();

    long startTimeMs;
    long measuredTimeMs;

    clopen(option(INDEX_SELECT_BRUTE_FORCE_THRESHOLD), threshold);
    startTimeMs = System.currentTimeMillis();
    long count = g.V().has("name", strings.get(0))
        .has("prop0", strings.get(0)).count().next();
    measuredTimeMs = System.currentTimeMillis() - startTimeMs;
    System.out.println("total search time (ms): " + measuredTimeMs + ", count = " + count);
    g.tx().rollback();
}
```

rngcntr:
Thanks for the numbers @li-boxuan!
I think the total time measurements are not as relevant here as the selection time. Are these average numbers from a set of runs with the same amount of index candidates?
Even though the selection times for the greedy algorithm are fluctuating, the measurements support the claim that a threshold of 10 is reasonable.

li-boxuan:
> I think the total time measurements are not as relevant here as the selection time.

Yeah, I just wanted to give an intuitive sense of how much time index selection takes up to. 

> Are these average numbers from a repeated set of runs?

Nope, only one run. I did multiple runs at the very beginning and did not observe much fluctuation, so I ended up doing one run only. Since there are no obvious outliers, I tend to believe these numbers are representative enough.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -497,5 +407,122 @@ private static boolean coversAll(final MixedIndexType index, Condition<JanusGrap
         return match != null && indexInfo.supports(index, match, atom.getPredicate());
     }
 
+    private JointIndexQuery selectIndices(final Set<IndexType> indexCandidates, final MultiCondition<JanusGraphElement> conditions,
```
porunov, line 410: I would suggest to move `selectIndices` method into a separate class. Something like `IndexSelectionService`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -497,5 +407,122 @@ private static boolean coversAll(final MixedIndexType index, Condition<JanusGrap
...
+                                          final Set<Condition> coveredClauses, final MutableBoolean isSorted) {
+        /* TODO: smarter optimization:
+        - use in-memory histograms to estimate selectivity of PredicateConditions and filter out low-selectivity ones
+                if they would result in an individual index call (better to filter afterwards in memory)
+        */
+        if (indexCandidates.size() <= tx.getGraph().getConfiguration().getIndexSelectThreshold()) {
+            return selectIndicesByBruteForce(indexCandidates, conditions, coveredClauses, isSorted);
+        } else {
+            return selectIndicesByApprox(indexCandidates, conditions, coveredClauses, isSorted);
+        }
```
porunov, line 420: I would suggest to move `selectIndicesByBruteForce` and `selectIndicesByApprox` to separate classes. As for me they are strategies. So I would move them to `BruteForceIndexSelectionStrategy` and `ApproximationIndexSelectionStrategy`.

rngcntr:
@porunov These are valid points and nice suggestions. I will try to start refactoring either at the end of this week or early next week.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/IndexCandidate.java
```diff
@@ -0,0 +1,170 @@
...
+        double score = 0.0;
+        if (index.isCompositeIndex()) {
+            if (((CompositeIndexType)index).getCardinality() == Cardinality.SINGLE) {
+                score = CARDINALITY_SINGE_SCORE;
+            } else {
+                score = CARDINALITY_OTHER_SCORE;
+            }
+        }
+        return score;
+    }
```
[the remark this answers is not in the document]
rngcntr: To me it looks like these constants were introduced with thought in mind that there _should_ be a distinction in scores between cardinalities. It just looks like the original author prepared this distinction but did evaluate the performance impact of different scores right away.
Therefore in my opinion, we should keep this code as an entry point for further optimizations as it already hints to a tweak which can increase performance.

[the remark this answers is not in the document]
porunov: Good points! Agree!

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
rngcntr: I'm currently working on this approach but it looks like this requires moving more code than expected from the `GraphCentricQueryBuilder` to the strategy classes. I'll push a draft later today so you can at least have a look at what I mean.

rngcntr:
Note: I tried to follow a hardcore approach to keep the index selection strategies completely independent from the `GraphCentricQueryBuilder`. I don't know if this is actually desirable, but I thought it would be cleaner to move all index related stuff to a separate class. This results in many static methods moving from `GraphCentricQueryBuilder` to `IndexSelectionStrategy`. As of now, not entirely happy with the results, because having all of these methods in an interface does not look as clean as I thought it would :sweat_smile:

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionStrategy.java
```diff
@@ -0,0 +1,231 @@
...
+import org.javatuples.Pair;
+
+import javax.annotation.Nullable;
+import java.util.*;
+import java.util.stream.StreamSupport;
+
+/**
+ * @author Florian Grieskamp (Florian.Grieskamp@gdata.de)
+ */
+public interface IndexSelectionStrategy {
```
porunov, line 41: Most of methods of this interface are related to implementation, not abstraction. Thus they should either be private or package protected or protected as they are not intended to be used publicly.
Thus, I suggest create an abstract class `public abstract class `AbstractIndexSelectionStrategy` implements IndexSelectionStrategy` and move implementations there (make everything non-static. Methods which should be static can be left here). After that make `BruteForceIndexSelectionStrategy` and `ApproximateIndexSelectionStrategy` to be extended from `AbstractIndexSelectionStrategy`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/DefaultIndexSelector.java
```diff
@@ -0,0 +1,61 @@
...
+import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
+import org.janusgraph.graphdb.types.IndexType;
+import org.javatuples.Pair;
+
+import java.util.Set;
+
+/**
+ * @author Florian Grieskamp (Florian.Grieskamp@gdata.de)
+ */
+public class DefaultIndexSelector implements IndexSelectionStrategy {
```
porunov, line 32: This looks more like `Service` because it contains a business logic to interact with different strategies. I would suggest to rename it to `IndexSelectorService` and do not implement `IndexSelectionStrategy`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/DefaultIndexSelector.java
```diff
@@ -0,0 +1,61 @@
...
+    private StandardJanusGraphTx tx;
+
+    private IndexSelectionStrategy bruteForce;
+    private IndexSelectionStrategy approximation;
+
+    public DefaultIndexSelector(StandardJanusGraphTx tx) {
+        this.tx = tx;
+
+        this.bruteForce = new BruteForceIndexSelector();
+        this.approximation = new ApproximateIndexSelector();
```
porunov, line 46: Could you please make it as direct constructor dependency to follow DI principle? It will simplify this class testing in future

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/DefaultIndexSelector.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public Pair<JointIndexQuery, Boolean> selectIndices(final Set<IndexType> indexCandidates, final MultiCondition<JanusGraphElement> conditions,
+                                                        final Set<Condition> coveredClauses, OrderList orders, IndexSerializer serializer) {
+        /* TODO: smarter optimization:
+        - use in-memory histograms to estimate selectivity of PredicateConditions and filter out low-selectivity ones
+                if they would result in an individual index call (better to filter afterwards in memory)
+        */
+        int bruteForceThreshold = tx.getGraph().getConfiguration().getIndexSelectThreshold();
```
porunov, line 56: Can `bruteForceThreshold` be changed during different calls of this method? If no, could you provide `bruteForceThreshold` as a dependency of this class?

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ApproximateIndexSelector.java
```diff
@@ -0,0 +1,76 @@
...
+import org.janusgraph.graphdb.types.IndexType;
+import org.janusgraph.graphdb.types.MixedIndexType;
+import org.javatuples.Pair;
+
+import java.util.Set;
+
+/**
+ * @author Florian Grieskamp (Florian.Grieskamp@gdata.de)
+ */
+public class ApproximateIndexSelector implements IndexSelectionStrategy {
```
porunov, line 32: Do not remove `Strategy` of concrete implementations. I.e. call it `ApproximateIndexSelectionStrategy`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/BruteForceIndexSelector.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Florian Grieskamp (Florian.Grieskamp@gdata.de)
+ */
+public class BruteForceIndexSelector implements IndexSelectionStrategy {
```
porunov, line 36: Do not remove `Strategy` of concrete implementations. I.e. call it `BruteForceIndexSelectionStrategy`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionStrategy.java
```diff
@@ -0,0 +1,231 @@
...
+    Pair<JointIndexQuery, Boolean> selectIndices(final Set<IndexType> indexCandidates, final MultiCondition<JanusGraphElement> conditions,
```
porunov, line 42: Instead of using `Pair` as a return type, I would create some POJO which contains `JointIndexQuery jointQuery` and `Boolean sorted` because 1) It isn't obvious what does `Boolean` means here. 2) If we decide to extend or shrink our return type, we won't be able to use Pair anymore.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/DefaultIndexSelector.java
rngcntr: Good point

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/DefaultIndexSelector.java
rngcntr: Why not see this as a Strategy, too? Sure it combines and calls other strategies based on the size of the problem. But from my point of view it serves the same purpose as other implementations of `IndexSelectionStrategy` and acts as a drop-in replacement for other strategies.

porunov: Good point. I guess it can be seen as a strategy also but from the other hand it just contains logic to choose another strategy based on some logic. I guess if we add another strategy it becomes a dependency of this class also because the logic behind this class is basically to select a strategy.
I see it more as a service but I don't mind leaving it as a strategy. Just give it some more clear name. Maybe something like 'ThresholdBasedIndexSelectionStrategy'.

rngcntr: LGTM
I'll see if I can find some spare time today or tomorrow to implement your requested changes

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -81,13 +81,21 @@
...
     public GraphCentricQueryBuilder(StandardJanusGraphTx tx, IndexSerializer serializer) {
         Preconditions.checkNotNull(tx);
         Preconditions.checkNotNull(serializer);
         useSmartLimit = tx.getGraph().getConfiguration().adjustQueryLimit();
         this.tx = tx;
         this.serializer = serializer;
+        this.indexSelector =
+            new ThresholdBasedIndexSelectionStrategy(tx.getGraph().getConfiguration().getIndexSelectThreshold(),
+                new BruteForceIndexSelectionStrategy(),
+                new ApproximateIndexSelectionStrategy());
```
porunov, line 98: I think we can add `indexSelector` as a dependency to this constructor instead of initializing it here to follow DI

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -31,12 +33,11 @@
 import org.janusgraph.graphdb.query.profile.QueryProfiler;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.graphdb.types.*;
-import org.janusgraph.graphdb.types.system.ImplicitKey;
+import org.javatuples.Pair;
```
porunov, line 36: Unused import

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,217 @@
...
+        if (indexCandidate.getIndex().isCompositeIndex()) {
+            jointQuery.add((CompositeIndexType) indexCandidate.getIndex(), serializer.getQuery(
+                (CompositeIndexType) indexCandidate.getIndex(), (List<Object[]>) indexCandidate.getSubCondition()));
+        } else {
+            jointQuery.add((MixedIndexType) indexCandidate.getIndex(), serializer.getQuery(
+                (MixedIndexType) indexCandidate.getIndex(), (Condition) indexCandidate.getSubCondition(), orders));
+        }
+    }
+
+    protected List<Object[]> indexCover(final CompositeIndexType index, Condition<JanusGraphElement> condition,
```
porunov, line 101: `protected` -> `private`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,217 @@
...
+        final List<Object[]> indexCovers = new ArrayList<>(4);
+
+        constructIndexCover(indexValues,0,fields,condition,indexCovers,coveredClauses);
+        if (!indexCovers.isEmpty()) {
+            covered.addAll(coveredClauses);
+            return indexCovers;
+        } else return null;
+    }
+
+    protected void constructIndexCover(Object[] indexValues, int position, IndexField[] fields,
```
porunov, line 120: `protected` -> `private`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,217 @@
...
+                for (final Object value : equalCon.getValue()) {
+                    final Object[] newValues = Arrays.copyOf(indexValues,fields.length);
+                    newValues[position]=value;
+                    constructIndexCover(newValues,position+1,fields,condition,indexCovers,coveredClauses);
+                }
+            }
+        }
+    }
+
+    protected Condition<JanusGraphElement> indexCover(final MixedIndexType index,
```
porunov, line 140: `protected` -> `private`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,217 @@
...
+        for (final Condition<JanusGraphElement> subClause : condition.getChildren()) {
+            if (coversAll(index,subClause,indexInfo)) {
+                subCondition.add(subClause);
+                covered.add(subClause);
+            }
+        }
+        return subCondition.isEmpty()?null:subCondition;
+    }
+
+    protected boolean coversAll(final MixedIndexType index, Condition<JanusGraphElement> condition,
```
porunov, line 175: `protected` -> `private`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,217 @@
...
+        final PropertyKey key = (PropertyKey) atom.getKey();
+        final ParameterIndexField[] fields = index.getFieldKeys();
+        final ParameterIndexField match = Arrays.stream(fields)
+            .filter(field -> field.getStatus() == SchemaStatus.ENABLED)
+            .filter(field -> field.getFieldKey().equals(key))
+            .findAny().orElse(null);
+        return match != null && indexInfo.supports(index, match, atom.getPredicate());
+    }
+
+    protected Map.Entry<Condition,Collection<Object>> getEqualityConditionValues(
```
porunov, line 199: `protected` -> `private`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,217 @@
...
+import org.janusgraph.graphdb.query.QueryUtil;
+import org.janusgraph.graphdb.query.condition.*;
+import org.janusgraph.graphdb.query.graph.JointIndexQuery;
+import org.janusgraph.graphdb.types.*;
+import org.janusgraph.graphdb.types.system.ImplicitKey;
+
+import java.util.*;
+import java.util.stream.StreamSupport;
+
+public abstract class AbstractIndexSelectionStrategy {
```
porunov, line 37: This class looks like self sufficient class. Self sufficient classes should not be `abstract`. Now it looks more like Utility class. 
To make it logically abstract I suggest to add `implements IndexSelectionStrategy` but don't implement it in this class directly. Instead, children of this class will need to implement methods from `IndexSelectionStrategy`.
Thus, you don't need to add `implements IndexSelectionStrategy` to any children of this class.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ThresholdBasedIndexSelectionStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @author Florian Grieskamp (Florian.Grieskamp@gdata.de)
+ */
+public class ThresholdBasedIndexSelectionStrategy
+    extends AbstractIndexSelectionStrategy implements IndexSelectionStrategy {
+
+    private int threshold;
+
+    private IndexSelectionStrategy usedIfLessOrEqualThreshold;
+    private IndexSelectionStrategy usedIfGreaterThanThreshold;
```
porunov, line 37: Mark all these fields as `final`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ApproximateIndexSelectionStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+                if (coveredClauses.isEmpty()) {
+                    isSorted = candidateSupportsSort;
+                }
+                coveredClauses.addAll(bestCandidate.getSubCover());
+                addToJointQuery(bestCandidate, jointQuery, serializer, orders);
+            } else {
+                break;
+            }
+        }
+        return new SelectedIndexQuery(jointQuery, Boolean.valueOf(isSorted));
```
porunov, line 81: Redundant boxing. I.e. use `isSorted` directly instead of `Boolean.valueOf(isSorted)`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/BruteForceIndexSelectionStrategy.java
```diff
@@ -0,0 +1,90 @@
...
+import org.janusgraph.core.JanusGraphElement;
+import org.janusgraph.graphdb.database.IndexSerializer;
+import org.janusgraph.graphdb.internal.OrderList;
+import org.janusgraph.graphdb.query.condition.Condition;
+import org.janusgraph.graphdb.query.condition.MultiCondition;
+import org.janusgraph.graphdb.query.graph.JointIndexQuery;
+import org.janusgraph.graphdb.types.IndexType;
+import org.janusgraph.graphdb.types.MixedIndexType;
+import org.janusgraph.util.datastructures.PowerSet;
+import org.javatuples.Pair;
```
porunov, line 30: Unused import

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ApproximateIndexSelectionStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Set;
+import org.janusgraph.core.JanusGraphElement;
+import org.janusgraph.graphdb.database.IndexSerializer;
+import org.janusgraph.graphdb.internal.OrderList;
+import org.janusgraph.graphdb.query.condition.Condition;
+import org.janusgraph.graphdb.query.condition.MultiCondition;
+import org.janusgraph.graphdb.query.graph.JointIndexQuery;
+import org.janusgraph.graphdb.types.IndexType;
+import org.janusgraph.graphdb.types.MixedIndexType;
+import org.javatuples.Pair;
```
porunov, line 26: Unused import

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ThresholdBasedIndexSelectionStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+import org.janusgraph.core.JanusGraphElement;
+import org.janusgraph.graphdb.database.IndexSerializer;
+import org.janusgraph.graphdb.internal.OrderList;
+import org.janusgraph.graphdb.query.condition.Condition;
+import org.janusgraph.graphdb.query.condition.MultiCondition;
+import org.janusgraph.graphdb.query.graph.JointIndexQuery;
+import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
+import org.janusgraph.graphdb.types.IndexType;
+import org.javatuples.Pair;
+
```
porunov, line 27: Several unused imports

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+                                     final Set<Condition> coveredClauses, OrderList orders,
+                                     IndexSerializer serializer);
+
+    static boolean indexCoversOrder(MixedIndexType index, OrderList orders) {
+        for (int i = 0; i < orders.size(); i++) {
+            if (!index.indexesKey(orders.getKey(i)))
+                return false;
+        }
+        return true;
+    }
```
porunov, line 42: This is an utility method, so I would recommend to move it to another class. I.e. something like `IndexSelectionUtil`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
rngcntr: Are you ok with configuring it in `StandardJanusGraph`?

porunov: Also, I guess `indexSelectThreshold` doesn't change from query to query, so we don't need to create a new implementation of a strategy each time. We should create indexSelectionStrategy only once when we establish a connection.

rngcntr: I assume this is a "yes" :smile:

porunov: Ah, I didn't see you past comment) Yes, configuring it in `StandardJanusGraph` looks good!

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
```diff
@@ -0,0 +1,125 @@
...
+            if (coveredClauses.contains(c)) {
+                subScore = subScore * ALREADY_MATCHED_ADJUSTOR;
+            }
+            score += subScore + getIndexTypeScore(index);
+        }
+
+        if (supportsSort) {
+            score += ORDER_MATCH;
+        }
+    }
```
porunov, line 86: Move it to `ApproximateIndexSelectionStrategy`. You can just add necessary parameters and return a calculated score. I.e. `private double calculateScore(Set<Condition> subCover, IndexType index, final Set<Condition> coveredClauses, boolean supportsSort)`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
```diff
@@ -0,0 +1,125 @@
...
+    public Object getSubCondition() {
+        return subCondition;
+    }
+
+    public void calculateScoreBruteForce() {
+        score = 0.0;
+        for (final Condition c : subCover) {
+            score += getConditionBasicScore(c) + getIndexTypeScore(index);
+        }
+    }
```
porunov, line 70: Move it to BruteForceIndexSelectionStrategy. You can just add necessary parameters and return a calculated score. I.e. `private double calculateScore(Set<Condition> subCover, IndexType index)`

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
porunov: Also, you can just add a method `setScore` into `IndexCandidate` and update score after the calculation

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
```diff
@@ -0,0 +1,125 @@
...
+        double score = 0.0;
+        if (index.isCompositeIndex()) {
+            if (((CompositeIndexType)index).getCardinality() == Cardinality.SINGLE) {
+                score = CARDINALITY_SINGE_SCORE;
+            } else {
+                score = CARDINALITY_OTHER_SCORE;
+            }
+        }
+        return score;
+    }
```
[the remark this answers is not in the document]
rngcntr: What's the point in moving everything from the `IndexCandidate` to `AbstractIndexSelectionStrategy`? Why would we want to have methods that are only used for `IndexCandidate`s in another class?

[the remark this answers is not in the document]
porunov: To follow [SOLID](https://en.wikipedia.org/wiki/SOLID) principles. Currently this class violates [Single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).
1) The class is responsible for 2 separate algorithms. Brute force and approximation.
2) The class is responsible for both business logic (calculation logic) and data (POJO).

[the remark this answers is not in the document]
porunov: Notice, those methods are not used by `IndexCandidate` as you are telling.
private methods are used by brute force and approximation calculation logic (public methods). That calculation logic is used only in concrete strategies. Brute force calculation logic is used only in brute force strategy. Approximation calculation logic is used only in approximation strategy. Whereas `getConditionBasicScore` and `getIndexTypeScore` is general logic which is used by both score calculation logic.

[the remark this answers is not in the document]
rngcntr: Thanks for the explanation! Nice and comprehensive :smile:

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
```diff
@@ -0,0 +1,125 @@
...
+     * @return true if the IndexCandidate is fully covered by the given set. false if not.
+     */
+    public boolean isCoveredBy(final Set<Condition> coveringClauses) {
+        for (final Condition c : subCover) {
+            if (!coveringClauses.contains(c)) {
+                return false;
+            }
+        }
+        return true;
+    }
```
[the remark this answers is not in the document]
rngcntr: I think we can even replace this method by just calling `coveringClauses.containsAll(subCover)`.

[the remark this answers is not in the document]
porunov: Totally agree

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.stream.StreamSupport;
+
+public abstract class AbstractIndexSelectionStrategy implements IndexSelectionStrategy {
+    private static final double EQUAL_CONDITION_SCORE = 4;
+    private static final double OTHER_CONDITION_SCORE = 1;
+    private static final double CARDINALITY_SINGE_SCORE = 1000;
+    private static final double CARDINALITY_OTHER_SCORE = 1000;
+
+    protected static final double ORDER_MATCH = 1;
+    protected static final double ALREADY_MATCHED_ADJUSTOR = -1.1;
```
porunov, line 41: Missed on this. It is used only by `ApproximateIndexSelectionStrategy`, so I think it should be moved there

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ApproximateIndexSelectionStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+                coveredClauses.addAll(bestCandidate.getSubCover());
+                addToJointQuery(bestCandidate, jointQuery, serializer, orders);
+            } else {
+                break;
+            }
+        }
+        return new SelectedIndexQuery(jointQuery, isSorted);
+    }
+
+    private double calculateIndexCandidateScore(final IndexCandidate indexCandidate,
```
porunov, line 83: It would be better to pass only necessary parameters to the method. Even so, we will need to pass 2 parameters (subCover and index), it is better to pass them instead of indexCandidate because `indexCandidate` has more details (which are not needed by this method, i.e. subCondition)

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/BruteForceIndexSelectionStrategy.java
```diff
@@ -0,0 +1,97 @@
...
+                                            (MixedIndexType) bestIndexes.get(0).getIndex(), orders);
+            for (IndexCandidate c : bestIndexes) {
+                addToJointQuery(c, jointQuery, serializer, orders);
+            }
+        }
+
+        return new SelectedIndexQuery(jointQuery, isSorted);
+    }
+
+    private double calculateIndexCandidateScore(final IndexCandidate ic) {
```
porunov, line 90: Same here. I would pass 2 separate arguments instead of a single but with redundant information

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
```diff
@@ -0,0 +1,55 @@
...
+    private double score;
+
+    public IndexCandidate(final IndexType index,
+                          final Set<Condition> subCover,
+                          final Object subCondition) {
+        this.index = index;
+        this.subCover = subCover;
+        this.subCondition = subCondition;
+        // initialize with the worst possible score
+        this.score = Double.NEGATIVE_INFINITY;
```
porunov, line 39: I would put it back to inline initialization because:
1) The field isn't final, it means that we may add a bug if we add another constructor later (i.e. forget to initialize it)
2) I'm not sure about non final fields in sense of initialization. If final field is initialized only once, what about non final field? Will it be initialized with `0.0` at class fields initialization phase and them `Double.NEGATIVE_INFINITY` at constructor initialization phase? Again, not sure about it but if we use inline initialization we guarantee that it will be `Double.NEGATIVE_INFINITY` in both class fields initialization phase and constructor initialization phase.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ApproximateIndexSelectionStrategy.java
rngcntr: I think it's convenient to pass the entire `IndexCandidate` object to the score calculation. Especially as we trimmed down the class `IndexCandidate` to contain only relevant information and no logic, in my opinion every attribute of of an `IndexCandidate` qualifies as a valid foundation to be considered for the calculation of a score.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexCandidate.java
rngcntr: Yes, you're right. Good point.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
rngcntr: I briefly thought about that, but I decided (for no reason) to have all constants in one place. I'm also fine with moving them to the strategy that uses them, if it is only a single one.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/ApproximateIndexSelectionStrategy.java
porunov: Good point. Agree

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/AbstractIndexSelectionStrategy.java
porunov: I think it is better to move them into a concrete strategy while they are used only by that strategy

rngcntr:
Some builds are failing, but I don't have time to have a closer look at it right now. I will probably look into it on tuesday.

porunov:
@rngcntr all tests passed. It was a temporary problem with docker containers

PR #2103, opened 2020-04-30
Fix ElasticSearch INDEX_STORE_NAMES_CACHE should not be shared between instances

### LINKED ISSUE #2102: ElasticSearch INDEX_STORE_NAMES_CACHE is shared between instances
`ElasitcSearchIndex` should not share same `INDEX_STORE_NAMES_CACHE`.

- Version: 0.5.0, 0.5.1
- Mixed Index Backend: ElasticSearch
- Mailing list Thread URL: https://groups.google.com/forum/#!topic/janusgraph-users/YxR1HY6AbS8
- Steps to Reproduce: Create 2 instances of `ElasticSearchIndex`

porunov:
- Move INDEX_STORE_NAMES_CACHE from static to instance specific scope.
- Add possibility to disable INDEX_STORE_NAMES_CACHE cache

Fixes #2102

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -72,6 +72,7 @@ import java.io.UncheckedIOException;
 import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -243,6 +244,12 @@
         new ConfigOption<>(ELASTICSEARCH_NS, "retry_on_conflict",
             "Specify how many times should the operation be retried when a conflict occurs.", ConfigOption.Type.MASKABLE, 0);
 
+    public static final ConfigOption<Boolean> ENABLE_INDEX_STORE_NAMES_CACHE =
```
mad, line 247: I think we don't need addition option for that

Simply we may use cache with 10k LRU enties

porunov: Did you check performance of 10k LRU cache and concurrentHashMap? I checked that it is faster to get a name from map than making contact for 50k and less entities. What if user has more indexes (unlikely but possible)? In such scenario the will be forced to use cache that is slower than string concatenation.

porunov: Also, LRU cache may evict some names which we would like to avoid because it is a permanent cache

mad: Yes you right concurrenthashmap mush faster than cache. But is so rare case when user would be change this option by memory usages cause, i think

porunov: You are right. Most likely nobody will use this option and most likely it will be `true` for all usecases but I think, it would be good to have this option in case the user has very unusual usecase. 

Also, this optimization was introduced here #1695 . This optimization is really tiny, so, regular user won't see any difference but still it speedups index store name retrieval for about 3-4 times faster for regular usages and reduces memory consumptions for unnecessary string concatenations.

PR #2105, opened 2020-05-02
Optimize KVCMutation construct for deletions in CacheTransaction

porunov:
Instead of copying elements to a list and then again to a new list, this PR provides a function to copy all deletions in a single pass.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
@@ -18,6 +18,7 @@ import com.google.common.base.Preconditions;
 
 import java.util.*;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Container for collection mutations against a data store.
@@ -33,12 +34,23 @@ public abstract class Mutation<E,K> {
     private List<K> deletions;
 
     public Mutation(List<E> additions, List<K> deletions) {
-        Preconditions.checkNotNull(additions);
-        Preconditions.checkNotNull(deletions);
-        if (additions.isEmpty()) this.additions=null;
-        else this.additions = new ArrayList<>(additions);
-        if (deletions.isEmpty()) this.deletions=null;
-        else this.deletions = new ArrayList<>(deletions);
+        assert additions!=null;
+        assert deletions!=null;
+        this.additions = additions.isEmpty()? null : new ArrayList<>(additions);
+        this.deletions = deletions.isEmpty()? null : new ArrayList<>(deletions);
+    }
+
+    public Mutation(Supplier<List<E>> additionsCopySupplier, Supplier<List<K>> deletionsCopySupplier) {
+        additions = additionsCopySupplier.get();
+        deletions = deletionsCopySupplier.get();
+        assert additions!=null;
+        assert deletions!=null;
+        if (additions.isEmpty()){
+            additions=null;
+        }
+        if (deletions.isEmpty()){
+            deletions=null;
+        }
     }
 
     public Mutation() {
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/CacheTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/CacheTransaction.java
@@ -15,7 +15,6 @@
 package org.janusgraph.diskstorage.keycolumnvalue.cache;
 
 import com.google.common.base.Preconditions;
-
 import org.janusgraph.diskstorage.*;
 import org.janusgraph.diskstorage.keycolumnvalue.KCVMutation;
 import org.janusgraph.diskstorage.keycolumnvalue.KeyColumnValueStore;
@@ -29,7 +28,6 @@ import org.janusgraph.graphdb.database.serialize.DataOutput;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.Callable;
-import java.util.stream.Collectors;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
@@ -105,10 +103,19 @@ public class CacheTransaction implements StoreTransaction, LoggableTransaction {
 
     private KCVMutation convert(KCVEntryMutation mutation) {
         assert !mutation.isEmpty();
-        if (!mutation.hasDeletions())
-            return new KCVMutation(mutation.getAdditions(), KeyColumnValueStore.NO_DELETIONS);
-        else
-            return new KCVMutation(mutation.getAdditions(), mutation.getDeletions().stream().map(KCVEntryMutation.ENTRY2COLUMN_FCT).collect(Collectors.toList()));
+        if (mutation.hasDeletions()) {
+            return new KCVMutation(
+                () -> new ArrayList<>(mutation.getAdditions()),
+                () -> {
+                    List<Entry> deletions = mutation.getDeletions();
+                    ArrayList<StaticBuffer> convertedDeletions = new ArrayList<>(deletions.size());
+                    for(Entry entry : deletions){
+                        convertedDeletions.add(KCVEntryMutation.ENTRY2COLUMN_FCT.apply(entry));
+                    }
+                    return convertedDeletions;
+                });
+        }
+        return new KCVMutation(mutation.getAdditions(), KeyColumnValueStore.NO_DELETIONS);
     }
 
     private void flushInternal() throws BackendException {
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KCVMutation.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KCVMutation.java
@@ -21,6 +21,7 @@ import org.janusgraph.diskstorage.keycolumnvalue.cache.KCVEntryMutation;
 
 import java.util.List;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * {@link Mutation} type for {@link KeyColumnValueStore}.
@@ -32,6 +33,10 @@ public class KCVMutation extends Mutation<Entry,StaticBuffer> {
 
     public KCVMutation(List<Entry> additions, List<StaticBuffer> deletions) {
         super(additions, deletions);
+    }
+
+    public KCVMutation(Supplier<List<Entry>> additionsCopySupplier, Supplier<List<StaticBuffer>> deletionsCopySupplier) {
+        super(additionsCopySupplier, deletionsCopySupplier);
     }
 
     @Override


janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
```diff
@@ -35,10 +36,27 @@
...
+            this.additions = new ArrayList<>(additions);
+        }
+        if (deletions.isEmpty()){
+            this.deletions=null;
+        } else {
+            this.deletions = new ArrayList<>(deletions);
+        }
+    }
+
+    public Mutation(CopyArrayListFunction<E> additionsCopyFunction, CopyArrayListFunction<K> deletionsCopyFunction) {
```
rngcntr, line 51: Why not use a `Supplier<List<E>>` instead of `CopyArrayListFunction<E>`?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
```diff
@@ -35,10 +36,27 @@
     public Mutation(List<E> additions, List<K> deletions) {
```
rngcntr, line 36: nitpick: You can redirect calls of this constructor to the new constructor:
```java
public Mutation(List<E> additions, List<K> deletions) {
    this(() -> new ArrayList<E>(additions), () -> new ArrayList<E>(deletions));
}
```

Saves lines and reduces potential for future mistakes ;)

porunov: I am not sure I get it here. How where do you get converted deletions here?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
porunov: This is very good suggestion. Changed to Supplier

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
porunov: Ah, got it. I didn't check the code

porunov: I'm in doubt of it because we used to check `Preconditions.checkNotNull(additions)` before starting working with the instance. If we change to above code, we may get NPE in ArrayList construction logic. If we move `checkNotNull` it means we add duplicate `checkNotNull` call.
From another hand `checkNotNull` throws NPE if the instance is `null`. Which is a little bit strange to me. I'm not sure why we really need this check here at all. If an instance is null -> we will get the same `NPE` because of `additions.isEmpty()` call. From another hand, we won't get `NPE` here because we never pass null here. So, I guess, this check was added here just for readability to help other developers that they should not pass null here.
If so, I guess we can remove `checkNotNull` from this place at all and use either `assert additions!=null;` or just add javadoc to these constructors and mention there that passed parameters should not be null.
What do you think about it?

rngcntr: I think Assertions are a clean way of handling this. Also, I don't mind having duplicate assertions as much as having duplicate business logic

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/CacheTransaction.java
```diff
@@ -105,10 +103,19 @@ public String toString() {
...
+        if (mutation.hasDeletions()) {
+            return new KCVMutation(
+                () -> new ArrayList<>(mutation.getAdditions()),
+                () -> {
+                    List<Entry> deletions = mutation.getDeletions();
+                    ArrayList<StaticBuffer> convertedDeletions = new ArrayList<>(deletions.size());
+                    for(Entry entry : deletions){
+                        convertedDeletions.add(KCVEntryMutation.ENTRY2COLUMN_FCT.apply(entry));
+                    }
+                    return convertedDeletions;
```
rngcntr, line 115: Wouldn't it be more readable if we do:
```
return mutations.getDeletions()
    .stream()
    .map(KCVEntryMutation.ENTRY2COLUMN_FCT::apply)
    .collect(Collectors.toList());
```

rngcntr: Oh this way we can even omit the code block and the return statement

porunov: As noted and benchmarked by @mad , vanilla java is faster than stream api (https://github.com/JanusGraph/janusgraph/pull/2098#discussion_r415074443) and thus it is recommended to use vanilla java in performance critical sections (like mutations).
Personally, I didn't compare ArrayList transformation performance for vanilla java and stream API. Also, I'm not sure if StreamAPI is using initial ArrayList length to construct an array of needed size without resizing an array during additions. 
I think that StreamAPI is a cleaner version here, but I didn't check benchmarks. If you can confirm that Stream API uses `new ArrayList<>(initialArrayListSize)` instead of `new ArrayList<>()` for `Collectors.toList()` for stream from ArrayList and you checked performance and it is the same as vanilla java, I am more than happy to apply this suggestion. Otherwise, I think it would be better to use vanilla java code in this section.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
porunov: I will change `checkNotNull` checks here to:
```
assert additions!=null;
assert deletions!=null;
```
As they are not enabled by default and thus will be used for testing only. And they will clearly show developers that they should not pass `null` here.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/CacheTransaction.java
rngcntr: As far as I know, one can initialize a custom list to be used by `collect()`. But I have no benchmarks. Will do one if I have some time later today.

porunov: @rngcntr indeed we could do `.collect(Collectors.toCollection(() -> new ArrayList<>(data.size())))`.
It would be good to check performance for this scenario. If it has the same performance, then we could use Steam API here.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
porunov: Hm. Not sure now. The code looks strange if we are using this logic:
```
public Mutation(List<E> additions, List<K> deletions) {
    this(() -> {
        assert additions!=null;
        return new ArrayList<E>(additions);
    }, () -> {
        assert deletions!=null;
        return new ArrayList<K>(deletions);
    });
}
```
I think the below code is cleaner and also doesn't create an unnecessary suppliers:
```
public Mutation(List<E> additions, List<K> deletions) {
    assert additions!=null;
    assert deletions!=null;
    this.additions = additions.isEmpty()? null : new ArrayList<>(additions);
    this.deletions = deletions.isEmpty()? null : new ArrayList<>(deletions);
}
```
Overall, I see that the whole `Mutation` and usages of `Mutation` should be refracted but I guess in other PRs.

porunov: What do you think about the updated implementation?

rngcntr: I'd prefer to have the logic only once, either in the old constructor or in the new one. Maybe you could replace the new constructor to only do
```java
this(additionsCopySupplier.get(), deletionsCopySupplier.get());
```
But this solution would not check if the suppliers themselves are `null`. On the other hand, the current solution doesn't do this either...
This is just a nitpick. I'm also ok with the current implementation but it would be nicer to keep it as clean as possible.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/CacheTransaction.java
rngcntr: Using `ArrayList` with the stream API brings a slight improvement, but nothing beats vanilla:
```java
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@BenchmarkMode(Mode.AverageTime)
@OperationsPerInvocation(StreamVsVanilla.N)
@Warmup(iterations = 3, time = 2, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 3, time = 2, timeUnit = TimeUnit.SECONDS)
public class StreamVsVanilla {
    public static final int N = 100_000;

    static List<String> sourceList = new ArrayList<>(N);
    static Function<String, String> identity = Function.identity();

    static {
        for (int i = 0; i < N; i++) {
            sourceList.add("hello");
        }
    }

    @Benchmark
    public List<String> vanilla() {
        List<String> adds = new ArrayList<>(N);
        for (final String s : sourceList) {
            adds.add(identity.apply(s));
        }
        return adds;
    }

    @Benchmark
    public List<String> stream() {
        List<String> adds = sourceList.stream()
            .map(identity)
            .collect(Collectors.toCollection(() -> new ArrayList<>(sourceList.size())));
        return adds;
    }

    @Benchmark
    public List<String> naiveStream() {
        List<String> adds = sourceList.stream()
            .map(identity)
            .collect(Collectors.toList());
        return adds;
    }
}
```
Results:
```
Benchmark                    Mode  Cnt   Score   Error  Units
StreamVsVanilla.naiveStream  avgt   15  11.339 ± 0.461  ns/op
StreamVsVanilla.stream       avgt   15  10.594 ± 0.548  ns/op
StreamVsVanilla.vanilla      avgt   15   6.362 ± 0.088  ns/op
```

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
porunov: The problem with this solution:
`this(additionsCopySupplier.get(), deletionsCopySupplier.get());` is that it will cause a copy of supplier's returned `ArrayList` to a new `ArrayList` which makes suppliers redundant and removes this optimization.
I think it would be better to leave current optimization as is and make re-factoring of `Mutation` and usages of `Mutation` in other PRs

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/CacheTransaction.java
porunov: @rngcntr Thank you for these benchmarks. If so, we need to use vanilla code in performance critical sections. Also, I have created an issue here #2110 to move code to Vanilla Java in other performance critical sections.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Mutation.java
rngcntr: I see

porunov:
Merging with lazy consensus

PR #2148, opened 2020-06-25
Fix #2147: graph is created again during the drop process

### LINKED ISSUE #2147: Graph is recreated when dropping with ConfiguredGraphFactory
## Description
- Version: Tested with 0.4.0, the code is still present in the current `master` version.
- Storage Backend: Cassandra and Scylla
- Mixed Index Backend: No

### Steps to Reproduce
- Logging to a gremlin shell.
- Create a graph with `ConfiguredGraphFacory.create("drop_test")`
- Drop the graph with `ConfiguredGraphFacory.drop("drop_test")`

#### Expected
The graph get's deleted from:
- `ConfigurationManagementGraph`
- The underlay storage backend

#### Actual
The graph is deleted from the `ConfigurationManagementGraph`, but it's _truncated_ in the storage backend. For the case of Cassandra/Scylla the keyspaces are dropped and recreated before the drop method return.

This is an extract of the attached log showing the DROPPED event being followed by a CREATED:
```
2020-06-25 19:20:19,478 [JanusGraph Cluster-nio-worker-0] DEBUG com.datastax.driver.core.Cluster  - Received event EVENT DROPPED KEYSPACE drop_test, scheduling delivery
....
2020-06-25 19:20:21,802 [JanusGraph Cluster-nio-worker-0] DEBUG com.datastax.driver.core.Cluster  - Received event EVENT CREATED KEYSPACE drop_test, scheduling delivery
```

Here is the complete log:
[gremlin-server.log](https://github.com/JanusGraph/janusgraph/files/4833137/gremlin-server.log)

BrunoBerisso:
This PR address issue #2147 by refactoring the `drop()` method in the `ConfiguredGraphFactory` class.

The bug was in the call to `removeGraphFromCache(String graphName)` which internally tries to open the graph. Because of the order of the calls, the graph was already dropped by a call to `JanusGraphFactory.drop(graph)` which then result in the storage to be recreated.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/.github/workflows/ci-backend-cql.yml
+++ b/.github/workflows/ci-backend-cql.yml
@@ -71,6 +71,10 @@ jobs:
           - module: cql
             args: "-Pcassandra3-murmur -Dtest=\"**/hadoop/*\""
           - module: cql
+            args: "-Pcassandra3-byteordered -Dtest=\"**/core/cql/*\""
+          - module: cql
+            args: "-Pcassandra3-murmur -Dtest=\"**/core/cql/*\""
+          - module: cql
             args: "-Pcassandra3-murmur-ssl -Dtest=\"**/diskstorage/cql/CQLStoreTest.java\""
           - module: cql
             args: "-Pcassandra3-murmur-client-auth -Dtest=\"**/diskstorage/cql/CQLStoreTest.java\""
--- a/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
@@ -34,6 +34,8 @@ import java.util.Map;
 import java.util.List;
 import java.util.Set;
 import java.util.Objects;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -153,13 +155,59 @@ public class ConfiguredGraphFactory {
      *
      * <p><b>WARNING: This is an irreversible operation that will delete all graph and index data.</b></p>
      * @param graphName String graphName. Corresponding graph can be open or closed.
-     * @throws BackendException If an error occurs during deletion
-     * @throws ConfigurationManagementGraphNotEnabledException If ConfigurationManagementGraph not
      */
-    public static void drop(String graphName) throws Exception {
-        final StandardJanusGraph graph = (StandardJanusGraph) ConfiguredGraphFactory.close(graphName);
-        JanusGraphFactory.drop(graph);
-        removeConfiguration(graphName);
+    public static void drop(String graphName) {
+        final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
+        Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
+
+        final StandardJanusGraph graph = (StandardJanusGraph) jgm.getGraph(graphName);
+        // Evict the graph from the cache in all nodes in the cluster
+        DropGraphOnEvictionTrigger dropTrigger = new DropGraphOnEvictionTrigger(graphName, graph, getConfigGraphManagementInstance());
+        removeGraphFromCache(graph, dropTrigger);
+        // Block the current thread until the drop operation finish
+        dropTrigger.waitUntilDropFinish();
+    }
+
+    /**
+     * This trigger will fire when the give graph was evicted from all the nodes in the cluster.
+     */
+    private static class DropGraphOnEvictionTrigger implements Callable<Boolean> {
+        private static final Logger log = LoggerFactory.getLogger(DropGraphOnEvictionTrigger.class);
+        private final String graphName;
+        private final JanusGraph graph;
+        private final ConfigurationManagementGraph configManagementGraph;
+        private final CountDownLatch latch;
+
+        private DropGraphOnEvictionTrigger(String graphName, JanusGraph graph, ConfigurationManagementGraph configManagementGraph) {
+            this.graphName = graphName;
+            this.graph = graph;
+            this.configManagementGraph = configManagementGraph;
+            this.latch = new CountDownLatch(1);
+        }
+
+        @Override
+        public Boolean call() throws BackendException {
+            try {
+                log.info("Graph {} has been removed from the graph cache on every JanusGraph node in the cluster.", graphName);
+                log.warn("Attempting to drop the graph {}.", graphName);
+                configManagementGraph.removeConfiguration(graphName);
+                JanusGraphFactory.drop(graph);
+                log.warn("Graph {} has been dropped.", graphName);
+                return true;
+            } finally {
+                latch.countDown();
+            }
+        }
+
+        public void waitUntilDropFinish() {
+            log.debug("Waiting for graph {} to be evicted from the cache in all JanusGraph nodes...", graphName);
+            try {
+                latch.await();
+            } catch (InterruptedException e)  {
+                Thread.currentThread().interrupt();
+                log.error("Thread interrupted while waiting for graph to be dropped.", e);
+            }
+        }
     }
 
     private static ConfigurationManagementGraph getConfigGraphManagementInstance() {
@@ -233,7 +281,7 @@ public class ConfiguredGraphFactory {
 
         try {
             final JanusGraph graph = open(graphName);
-            removeGraphFromCache(graph);
+            removeGraphFromCache(graph, null);
         } catch (Exception e) {
             // cannot open graph, do nothing
             log.error(String.format("Failed to open graph %s with the following error:\n %s.\n" +
@@ -241,14 +289,14 @@ public class ConfiguredGraphFactory {
         }
     }
 
-    private static void removeGraphFromCache(final JanusGraph graph) {
+    private static void removeGraphFromCache(final JanusGraph graph, Callable<Boolean> evictionTrigger) {
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
         Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
         final String graphName = ((StandardJanusGraph) graph).getGraphName();
         jgm.removeGraph(graphName);
         jgm.removeTraversalSource(toTraversalSourceName(graphName));
         final ManagementSystem mgmt = (ManagementSystem) graph.openManagement();
-        mgmt.evictGraphFromCache();
+        mgmt.evictGraphFromCache(evictionTrigger);
         mgmt.commit();
     }
 
--- a/janusgraph-test/pom.xml
+++ b/janusgraph-test/pom.xml
@@ -30,6 +30,13 @@
     </repositories>
 
     <dependencies>
+        <!-- Enforce a classpath ordering constraint to ensure log4j is used for logging. -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <!-- End logging backends. -->
         <dependency>
             <groupId>org.janusgraph</groupId>
             <artifactId>janusgraph-core</artifactId>
--- a/janusgraph-backend-testutils/pom.xml
+++ b/janusgraph-backend-testutils/pom.xml
@@ -25,6 +25,10 @@
             <artifactId>gremlin-test</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.tinkerpop</groupId>
+            <artifactId>gremlin-server</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.junit.platform</groupId>
             <artifactId>junit-platform-launcher</artifactId>
         </dependency>
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/core/AbstractConfiguredGraphFactoryTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/core/AbstractConfiguredGraphFactoryTest.java
@@ -14,40 +14,65 @@
 
 package org.janusgraph.core;
 
+import org.apache.commons.configuration.MapConfiguration;
 import org.apache.tinkerpop.gremlin.groovy.engine.GremlinExecutor;
 import org.apache.tinkerpop.gremlin.jsr223.GremlinScriptEngineManager;
-import org.janusgraph.graphdb.configuration.builder.GraphDatabaseConfigurationBuilder;
-import org.janusgraph.graphdb.management.JanusGraphManager;
-import org.janusgraph.graphdb.management.ConfigurationManagementGraph;
-import org.janusgraph.graphdb.management.utils.ConfigurationManagementGraphNotEnabledException;
-import org.janusgraph.diskstorage.configuration.backend.CommonsConfiguration;
-import org.janusgraph.graphdb.database.StandardJanusGraph;
-import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.GRAPH_NAME;
-import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_BACKEND;
-
 import org.apache.tinkerpop.gremlin.server.Settings;
-import org.apache.commons.configuration.MapConfiguration;
-
-import java.util.Map;
-import java.util.HashMap;
-
-import org.junit.jupiter.api.Test;
+import org.janusgraph.diskstorage.configuration.backend.CommonsConfiguration;
+import org.janusgraph.graphdb.configuration.builder.GraphDatabaseConfigurationBuilder;
+import org.janusgraph.graphdb.database.StandardJanusGraph;
+import org.janusgraph.graphdb.management.ConfigurationManagementGraph;
+import org.janusgraph.graphdb.management.JanusGraphManager;
+import org.janusgraph.graphdb.management.utils.ConfigurationManagementGraphNotEnabledException;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
 import javax.script.Bindings;
 import javax.script.SimpleBindings;
+import java.util.Map;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.GRAPH_NAME;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_BACKEND;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 
-public class ConfiguredGraphFactoryTest {
-    private static final JanusGraphManager gm;
-    static {
+public abstract class AbstractConfiguredGraphFactoryTest {
+
+    private static JanusGraphManager gm;
+
+    /**
+     * Getter for the settings to use to instantiate the graph supporting the
+     * ConfigurationManagementGraph instance.
+     * @return a MapConfiguration instance
+     */
+    protected abstract MapConfiguration getManagementConfig();
+
+    /**
+     * Getter for the settings to use to create the template configurations
+     * (createTemplateConfiguration)
+     * @return a MapConfiguration instance
+     */
+    protected abstract MapConfiguration getTemplateConfig();
+
+    /**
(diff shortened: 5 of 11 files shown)


janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
```diff
@@ -222,6 +223,25 @@ public void removeConfigurationShouldRemoveGraphFromCache() throws Exception {
         }
     }
 
+    @Test
+    public void dropGraphShouldRemoveGraphFromCache() throws Exception {
```
farodin91, line 227: I just added test in it pass without changes which i expected not to happen.

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
```diff
@@ -155,10 +155,16 @@ public static JanusGraph close(String graphName) throws Exception {
...
-        final StandardJanusGraph graph = (StandardJanusGraph) ConfiguredGraphFactory.close(graphName);
+    public static void drop(String graphName) throws BackendException, ConfigurationManagementGraphNotEnabledException, Exception {
+        final ConfigurationManagementGraph configManagementGraph = getConfigGraphManagementInstance();
+        configManagementGraph.removeConfiguration(graphName);
+
+        final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
+        Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
+
+        final StandardJanusGraph graph = (StandardJanusGraph) jgm.getGraph(graphName);
+        removeGraphFromCache(graph);
```
farodin91, line 166: > The bug was in the call to removeGraphFromCache(String graphName) which internally tries to open the graph. Because of the order of the calls, the graph was already dropped by a call to JanusGraphFactory.drop(graph) which then result in the storage to be recreated.

This sentence confuses me a bit, due the fact you didn't change the functionality of it.

Can you clarify it for me?

BrunoBerisso: There are two versions of `removeGraphFromCache()` one expects a [String graphName](https://github.com/JanusGraph/janusgraph/blob/89c18d053b81a45145a834e81e53f847abfa30d4/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java#L231) and the other expects a [JausGraph graph](https://github.com/JanusGraph/janusgraph/blob/89c18d053b81a45145a834e81e53f847abfa30d4/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java#L243).

The previous version of this method used the version that expects a _String graphName_ which [open the graph](https://github.com/JanusGraph/janusgraph/blob/89c18d053b81a45145a834e81e53f847abfa30d4/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java#L234) and call the other version that expect a `JanusGraph` instance.

The problem with that approach is that the `open()` call will create the graph if it doesn't exist, with the previous implementation the graph was dropped before calling `removeGraphFromCache()` which caused the graph to be created again.

What I did was use the version of `removeGraphFromCache()` that expect a `JanusGraph` instance which doesn't need to open the graph.

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
BrunoBerisso: Yeah, I add a test for `drop()` because I didn't found one and wanted to check that my changes don't introduce any regressions.

I'm not sure how I can test this particular case using the `inmemory` backend, I still can reproduce the bug but there are no measurable effects. We probably could use BerkeleyDB to check this case but that involves adding a dependency to _janusgraph-test_ target.

farodin91: Valid point

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
farodin91: Thanks for explanation

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
farodin91: It would be okay to add BerkeleyDB to this project. I think a CQL we better so we could also test distributed database in future.

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
```diff
@@ -26,16 +27,35 @@
...
+
+    @Container
+    public static final CassandraContainer cqlContainer;
     private static final JanusGraphManager gm;
+
     static {
+        cqlContainer = new CassandraContainer<>(CASSANDRA_IMAGE + ":" + CASSANDRA_VERSION)
+            .withExposedPorts(9042)
+            .withStartupAttempts(1)
+            .withStartupTimeout(Duration.ofMinutes(10));
```
BrunoBerisso, line 58: This long delay was needed because the Cassandra docker image takes around 5 min to get ready (measured in my machine) when running the `verify` command so I choose to set a safe default. I believe it's because the tests are running in parallel.

janusgraph-test/pom.xml
```diff
@@ -30,6 +30,13 @@
     </repositories>
 
     <dependencies>
+        <!-- Enforce a classpath ordering constraint to ensure log4j is used for logging. -->
```
BrunoBerisso, line 33: I hit this error from Travis
```
The job exceeded the maximum log length, and has been terminated.
```
and found out that the log4j settings were not being used because of a conflict on the available SLF4J available at run time. This change enforces the use of log4j which make the settings in `log4j.properties` being picked up.

The logs are much quieter now, the full DEBUG output is being appended to a file. Let me know if this is the right approach.

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
farodin91: I think it would be better use the JanusGraphCassandraContainer.

janusgraph-test/pom.xml
farodin91: Yeah this is a known issue.

BrunoBerisso:
I did try to use JanusGraphCassandraContainer but I  had issues with the tests that use "inmemory" storage.As far as I understand JanusGraphCassandraContainer replace all values on StandarStorageManager for some reason here https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java#L178If I comment the call to setWrapperStorageManager it works. I believe I'm missing something 🤔
-------- Original message --------From: Jan Jansen <notifications@github.com> Date: 7/5/20  13:10  (GMT+00:00) To: JanusGraph/janusgraph <janusgraph@noreply.github.com> Cc: Bruno Berisso <bruno.sebastian.berisso@gmail.com>, Author <author@noreply.github.com> Subject: Re: [JanusGraph/janusgraph] Fix #2147: graph is created again during the drop process (#2148) 
@farodin91 commented on this pull request.



In janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java:
>      static {
+        cqlContainer = new CassandraContainer<>(CASSANDRA_IMAGE + ":" + CASSANDRA_VERSION)
+            .withExposedPorts(9042)
+            .withStartupAttempts(1)
+            .withStartupTimeout(Duration.ofMinutes(10));

I think it would be better use the JanusGraphCassandraContainer.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/JanusGraph/janusgraph/pull/2148#discussion_r449870365",
"url": "https://github.com/JanusGraph/janusgraph/pull/2148#discussion_r449870365",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
BrunoBerisso: I did try to use JanusGraphCassandraContainer but I had issues with the tests that use "inmemory" storage. As far as I understand JanusGraphCassandraContainer replace all values on StandarStorageManager for some reason [here](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java#L178). If I comment the call to setWrapperStorageManager all tests cases works. I might be missing something 🤔

farodin91: Your assumption is correct.

BrunoBerisso:
Thanks for the time you put on this review @farodin91 
Please don't merge this yet, I'm investigating an issue that could be related to this change. I'll keep this PR updated.

farodin91:
@BrunoBerisso Any progress?

BrunoBerisso:
Hi Jan, sorry for the delay on this one. I got caught by a stream of work
and couldn't push the latest changes.

The problem I found with the approach is that when running in a cluster we
should wait for all the nodes in the cluster to ACK the eviction of the
graph from their internal cache before dropping. I have the code in place
to handle it correctly but I'm not happy with it and I need to sit down for
an hour and finish it.

I didn't forget about this, just need that extra hour.

farodin91 (approved):
@BrunoBerisso Thank you for coming back. LGTM! Cool idea of writing your own trigger.

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
```diff
@@ -152,13 +153,62 @@ public static JanusGraph close(String graphName) throws Exception {
...
+                dropFinished = true;
+                notifyAll();
+            }
+        }
+
+        public synchronized void waitUntilDropFinish() {
+            log.debug("Waiting for graph {} to be evicted from the cache in all JanusGraph nodes...", graphName);
+            while (!dropFinished) {
+                try {
+                    wait();
```
li-boxuan, line 205: The blocking mechanism you implemented looks correct to me, but I recommend using CountDownLatch which is simpler and less error-prone.

li-boxuan (commented):
I tried running locally to understand your code better, but I could not run it on my Macbook (I can run other CQL tests though). That being said, I left some stylish comments.

Another thing is, I think this PR should be retargeted at v0.5 branch. @farodin91 Am I correct?

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
BrunoBerisso: This was the original intention before adding `CassandraContainer` to the test. Now the actual fix is checked with the [assertNull(graph_keyspace);]( https://github.com/JanusGraph/janusgraph/pull/2148/files#diff-6d6dec4c758c73971b028a06681ea37f2c7e63e828229bdd1e4d7c896e1fe409R266) in this same method.

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
BrunoBerisso: Thanks for the recommendation, I wasn't familiar with the `CountDownLatch`. Now that I'm learning about it I'm not seeing how this would be simpler for this case, given that we are waiting for only one operation to complete.

Could you please elaborate a little on it? Or just point me to the right place so I can dig some more.

Thanks

li-boxuan: I guess when I say 'simpler' I mean It's simpler for programmers & reviewers to reason about its correctness. For example, I had to check carefully you had "synchronized" keywords in-place and you used a while loop to wait. In terms of lines of code, there might not be much of a difference.

I recommend reading _Item 81: Prefer concurrency utilities to wait and notify_ from _Effective Java, Third Edition_. [Here](http://reader.epubee.com/books/mobile/e0/e0931e78a3cb5e4e6a04116a22bcbded/text00005.html#filepos0001169971) is a resource online. As it suggests,

> Given the difficulty of using wait and notify correctly, you should use the higher-level concurrency utilities instead.

There are some discussions around this on StackOverflow as well, like [this](https://stackoverflow.com/questions/2536692/a-simple-scenario-using-wait-and-notify-in-java) and [this](https://stackoverflow.com/questions/5999100/is-there-a-block-until-condition-becomes-true-function-in-java).

Regarding the usage of CountDownLatch, I am sure you could find tons of examples online. There are also a few usages of it in the JanusGraph codebase.

BrunoBerisso: Thanks for the detailed answer, I appreciate your time investment in it and in this review.

I will update this PR during the holidays to include your suggestions.

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
BrunoBerisso: I push this change. `ConfiguredGraphFactoryTest` is now an abstract class extended by `CQLConfiguredGraphFactoryTest` and `InmemoryConfiguredGraphFactoryTest`.

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
BrunoBerisso: I update this to use `CountDownLatch` instead of _wait/notifyAll_.

I also removed the _while (!dropFinished)_ check because it was causing the Gremlin script engine to lock forever if the script times out before the eviction completes.

janusgraph-test/src/test/java/org/janusgraph/core/CQLConfiguredGraphFactoryTest.java
```diff
@@ -0,0 +1,89 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+@Testcontainers
+public class CQLConfiguredGraphFactoryTest extends ConfiguredGraphFactoryTest {
+
+    private static final String CASSANDRA_VERSION = "3.11.6";
+    private static final String CASSANDRA_IMAGE = "cassandra";
+
+    @Container
+    public static final CassandraContainer cqlContainer;
```
li-boxuan, line 43: Can you use JanusGraphCassandraContainer if possible?

BrunoBerisso: I try to use it when I first added the tests but it seems to not allow multiple storage backends. I added some details about this before [in this conversation](https://github.com/JanusGraph/janusgraph/pull/2148#discussion_r450122533)

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
```diff
@@ -14,37 +14,65 @@
...
     }
 
+    /**
+     * Getter for the settings to use to instantiate the graph supporting the
+     * ConfigurationManagementGraph instance.
+     * @return a MapConfiguration instance
+     */
+    protected static MapConfiguration getManagementConfig() {
+        final Map<String, Object> map = new HashMap<>();
+        map.put(STORAGE_BACKEND.toStringWithoutRoot(), "inmemory");
```
li-boxuan, line 59: Is it possible not to use in-memory backend in CQL test? Not familiar with this, so not sure if this is the reason you are unable to use JanusGraphCassandraContainer.

BrunoBerisso: This is a good idea, thanks.

I try it locally but `JanusGraphCassandraContainer` seems unreliable for me, I'm having a consistent `java.lang.IllegalStateException: Could not send request, session is closed` in the first call to `ConfiguredGraphFactory.removeConfiguration(graphName);`. I don't have this behavior with the Cassandra container.

I think this could be an issue with my local machine (I also had to restart docker sometimes for the container to start running) so I pushed this change to see if it works on the CI pipeline.

janusgraph-test/src/test/java/org/janusgraph/core/CQLConfiguredGraphFactoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.GRAPH_NAME;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_BACKEND;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_HOSTS;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_PORT;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+@Testcontainers
+public class CQLConfiguredGraphFactoryTest extends ConfiguredGraphFactoryTest {
```
li-boxuan, line 36: Can you move CQLConfiguredGraphFactoryTest to janusgraph-cql module,
ConfiguredGraphFactoryTest to janusgraph-backend-testutils module (you might need to introduce additional dependency),
and InmemoryConfiguredGraphFactoryTest to janusgraph-inmemory module?

This is a common practice for tests involving different backends. We have workflows like ci-backend-cql.yml which picks up tests only under certain folders.

li-boxuan (commented):
Thank you @BrunoBerisso, I think it's in a good shape now! Can you rebase your PR? We recently activated codacy and codecov checks.

janusgraph-test/src/test/java/org/janusgraph/core/CQLConfiguredGraphFactoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Container
+    public static final JanusGraphCassandraContainer cqlContainer = new JanusGraphCassandraContainer();
+
+    static {
+        cqlContainer.start();
```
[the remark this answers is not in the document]
BrunoBerisso: I also thought it was not needed, but without it, the container won't start when running locally.

li-boxuan:
If you check out https://github.com/JanusGraph/janusgraph/blob/master/.github/workflows/ci-backend-cql.yml#L63-L76 you can see we only include tests under specific folders. The simplest way is to move CQLConfiguredGraphFactoryTest.java to graphdb.cql package. Alternatively, you can add include new folder in ci-backend-cql.yml. Otherwise CQLConfiguredGraphFactoryTest is not executed in CI.

BrunoBerisso:
@li-boxuan The new tests for janusgraph-cql are failing with the same error I commented before https://github.com/JanusGraph/janusgraph/pull/2148#discussion_r551276910

The problem seems to be related to `JanusGraphCassandraContainer` patching `StandardStoreManager` [here](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java#L178). If I comment that out all tests pass, I'm not sure why that's needed.

Do you have any suggestions?
Thanks

farodin91:
> @li-boxuan The new tests for janusgraph-cql are failing with the same error I commented before https://github.com/JanusGraph/janusgraph/pull/2148#discussion_r551276910
> 
> 
> 
> The problem seems to be related to `JanusGraphCassandraContainer` patching `StandardStoreManager` [here](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java#L178). If I comment that out all tests pass, I'm not sure why that's needed.
> 
> 
> 
> Do you have any suggestions?
> 
> Thanks

The code is used to speed up the tests. I didn't check if it is need anymore.

.github/workflows/ci-backend-cql.yml
```diff
@@ -70,6 +70,8 @@ jobs:
             args: "-Pcassandra3-murmur -Dtest=\"**/graphdb/cql/*\""
           - module: cql
             args: "-Pcassandra3-murmur -Dtest=\"**/hadoop/*\""
+          - module: cql
+            args: "-Pcassandra3-murmur -Dtest=\"**/core/cql/*\""
```
li-boxuan, line 74: Can you add
```
          - module: cql
            args: "-Pcassandra3-byteordered -Dtest=\"**/core/cql/*\""
```
so that we test against byteordered partitioner as well?

li-boxuan (approved):
Looks good to me, thank you! Now I can also run the test locally! Just left a comment on the workflow definition.

@farodin91 Do you want to review again, especially given the changes on JanusGraphCassandraContainer?

BrunoBerisso:
Thanks @li-boxuan and @farodin91 for your effort in reviewing this PR and keeping the project moving forward. I hope I can continue contributing to JanusGraph in the future.

farodin91:
I will try to review the changes tomorrow.

farodin91:
Cql doesn't take longer than expected.

PR #2165, opened 2020-07-12
Optimize LevenshteinDistance used in Fuzzy predicates

### LINKED ISSUE #2164: Optimize LevenshteinDistance used in Fuzzy predicates
For Fuzzy predicates, LevenshteinDistance is used to calculate the edit distance between strings:
https://github.com/JanusGraph/janusgraph/blob/68d7d703b430044bdb36f7c7fad934359fd017a8/janusgraph-driver/src/main/java/org/janusgraph/core/attribute/Text.java#L256-L267

which is of time complexity O(nm), where m and n represent string length.

The same LevenshteinDistance implementation, in fact, provides a short-cut option:
https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/similarity/LevenshteinDistance.java#L175-L182
which has time complexity O(km) where k is the distance threshold and m is the length of the longer string. In our case, we have k = 2. This can greatly reduce the latency especially when both strings are long.

li-boxuan:
This reduces time complexity of fuzzy match from O(n^2) to O(n), where n
is the length of the string. This also adds tests which use fuzzy
predicates without index.

Closes #2164

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-driver/src/main/java/org/janusgraph/core/attribute/Text.java
+++ b/janusgraph-driver/src/main/java/org/janusgraph/core/attribute/Text.java
@@ -243,6 +243,9 @@ public enum Text implements JanusGraphPredicate {
 
     };
 
+    private static final LevenshteinDistance ONE_LEVENSHTEIN_DISTANCE = new LevenshteinDistance(1);
+    private static final LevenshteinDistance TWO_LEVENSHTEIN_DISTANCE = new LevenshteinDistance(2);
+
     /**
      * Whether {@code term} is at X Levenshtein of a {@code value} 
      * with X=:
@@ -254,16 +257,15 @@ public enum Text implements JanusGraphPredicate {
      * @return true if {@code term} is similar to {@code value} 
      */
     private static boolean isFuzzy(String term, String value){
-        int distance;
         term = term.trim();
         if (term.length() < 3) {
-            distance = 0;
+            return term.equals(value);
         } else if (term.length() < 6) {
-            distance = 1;
-        } else {
-            distance = 2;
+            int levenshteinDistance = ONE_LEVENSHTEIN_DISTANCE.apply(value, term);
+            return levenshteinDistance <= 1 && levenshteinDistance >= 0;
         }
-        return LevenshteinDistance.getDefaultInstance().apply(value, term)<=distance;
+        int levenshteinDist = TWO_LEVENSHTEIN_DISTANCE.apply(value, term);
+        return levenshteinDist <= 2 && levenshteinDist >= 0;
     }
 
     private static final Logger log = LoggerFactory.getLogger(Text.class);
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/query/QueryTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/query/QueryTest.java
@@ -18,6 +18,7 @@ import com.google.common.collect.Iterators;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.janusgraph.core.*;
 import org.janusgraph.core.attribute.Contain;
+import org.janusgraph.core.attribute.Text;
 import org.janusgraph.core.schema.JanusGraphManagement;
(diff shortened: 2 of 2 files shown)


janusgraph-driver/src/main/java/org/janusgraph/core/attribute/Text.java
```diff
@@ -263,7 +263,9 @@ private static boolean isFuzzy(String term, String value){
         } else {
             distance = 2;
         }
-        return LevenshteinDistance.getDefaultInstance().apply(value, term)<=distance;
+        // LevenshteinDistance returns -1 if distance is larger than threshold
+        int levenshteinDist = new LevenshteinDistance(distance).apply(value, term);
+        return levenshteinDist <= distance && levenshteinDist >= 0;
```
porunov, line 268: Please, don't create a new instance of `LevenshteinDistance` for each search.
We have 3 possible `distance` values (`0`, `1` and `2`). So, just create 3 constants of `LevenshteinDistance` and reuse them.
I.e.
```
private static final LevenshteinDistance ZERO_LEVENSHTEIN_DISTANCE = new LevenshteinDistance(0);
private static final LevenshteinDistance ONE_LEVENSHTEIN_DISTANCE = new LevenshteinDistance(1);
private static final LevenshteinDistance TWO_LEVENSHTEIN_DISTANCE = new LevenshteinDistance(2);
```

After that we can rewrite the method to be as following:
```
private static boolean isFuzzy(String term, String value){
    term = term.trim();
    if (term.length() < 3) {
        return ZERO_LEVENSHTEIN_DISTANCE.apply(value, term) == 0;
    } else if (term.length() < 6) {
        int levenshteinDist = ONE_LEVENSHTEIN_DISTANCE.apply(value, term);
        return levenshteinDist <= 1 && levenshteinDist >= 0;
    }
    int levenshteinDist = TWO_LEVENSHTEIN_DISTANCE.apply(value, term);
    return levenshteinDist <= 2 && levenshteinDist >= 0;
}
```

li-boxuan:
Here are some performance test results (I just ran it on my laptop):

Code snippet:
```groovy
graph = JanusGraphFactory.open("inmemory")
g = graph.traversal()
for (i = 0; i < 100000; i++) {
    g.addV().property("prop", value1).iterate();
}
g.V().has("prop", textFuzzy(value2)).count().profile()
```

**1.  Identical string (in theory, there should be no change in performance)**
value1 = value2 = "identical string value"

the new version takes 1120, 1053, 1056 ms for three runs. Avg: 1076ms
0.5.2 version takes 997, 1051, 1047 ms for three runs. Avg: 1031ms

**2. Totally different string (in theory, we should see a clear performance boost)**
value1 = "this is some random long string value that I made up just now"
value2 = "random query string that is not gonna be matched within the distance limit"

the new version takes 1212, 1036, 1083 ms for three runs. Avg:  1110ms
0.5.2 version takes 1478, 1491, 1373 ms for three runs. Avg: 1447ms

The performance difference will be more obvious if you use an even longer string.

janusgraph-driver/src/main/java/org/janusgraph/core/attribute/Text.java
```diff
@@ -254,16 +258,15 @@ public String toString() {
      * @return true if {@code term} is similar to {@code value} 
      */
     private static boolean isFuzzy(String term, String value){
-        int distance;
         term = term.trim();
         if (term.length() < 3) {
-            distance = 0;
+            return ZERO_LEVENSHTEIN_DISTANCE.apply(value, term) == 0;
```
porunov, line 263: Just thinking. Won't it be a little bit more efficient if we would use `return value.equals(term);` here?
As I understand `LevenshteinDistance` is 0 only for equal strings, if I am not missing anything.

li-boxuan: Ouch, why didn't I think about this in the first place... I think you are right. I'll update it later today.

porunov:

Issue #2176, opened 2020-07-26
Deleting vertex followed by adding property in nested transactions leads to wrong index entry


li-boxuan:
Look at this test case:
https://github.com/JanusGraph/janusgraph/blob/33967fdb9144be2f2013e96c684eed3a8ce86251/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java#L1695-L1698

which calls this method (initialValue = null, updatedValue != null):

https://github.com/JanusGraph/janusgraph/blob/33967fdb9144be2f2013e96c684eed3a8ce86251/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java#L1786-L1817

It creates two nested transactions. The first transaction removes the vertex and then commits, while the second transaction adds a property value and then commits. Looking at line 1816:

https://github.com/JanusGraph/janusgraph/blob/33967fdb9144be2f2013e96c684eed3a8ce86251/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java#L1816

This assertion does not check the secondary store, because the mixed index is built upon full-text search, which does not support equality check. In other words, this traversal simply checks storage backend.

However, if we make the index using string search (which means it supports equality check), as below (https://github.com/li-boxuan/janusgraph/commit/d9d79259295e):

```java
mgmt.buildIndex("mixedIndex", Vertex.class).addKey(prop, Mapping.STRING.asParameter()).buildMixedIndex(INDEX);
```

Then this test case will start to fail:

https://travis-ci.org/github/li-boxuan/janusgraph/jobs/711858999#L1397

`CQLElasticsearchTest>JanusGraphIndexTest.testDeleteVertexThenAddProperty:1708->JanusGraphIndexTest.testNestedWrites:1816 expected: <true> but was: <false>`

So this test case itself, if fixed as mentioned above, demonstrates the bug: the second transaction commits into secondary index store, even if this vertex does not exist anymore.

li-boxuan:
After reading relevant code in StandardJanusGraph and CQLStoreManager, now I can try to conclude how it works (using Cassandra and Elasticsearch for example):

1) the first transaction successfully deletes the vertex
[...]

PR #2182, opened 2020-08-09
Fix testNestedWrites testcase in JanusGraphIndexTest

li-boxuan:
The test case builds a mixed index using TEXT mapping, but uses
equality condition in query (which does not use this mixed
index). This makes one assertion does not work as expected, and
as a consequence, passes the test by chance. This commit fixes the
mixed index usage and the assertion to uncover the real problem.

Related to #2176

**Note that this does not fix the problem stated in #2176, but fix the
flaky test itself to uncover the problem.**

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -62,6 +62,7 @@ import org.janusgraph.graphdb.log.StandardTransactionLogProcessor;
 import org.janusgraph.graphdb.query.index.ApproximateIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.index.BruteForceIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.index.ThresholdBasedIndexSelectionStrategy;
+import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.graphdb.types.ParameterType;
 import org.janusgraph.graphdb.types.StandardEdgeLabelMaker;
 import org.janusgraph.testutil.TestGraphConfigs;
@@ -1825,7 +1826,7 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
 
         // Write schema and one vertex
         final PropertyKey prop = makeKey(propName, String.class);
-        createExternalVertexIndex(prop, INDEX);
+        mgmt.buildIndex("mixed", Vertex.class).addKey(prop, Mapping.STRING.asParameter()).buildMixedIndex(INDEX);
         finishSchema();
 
         final JanusGraphVertex v = graph.addVertex();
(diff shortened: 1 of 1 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1809,14 +1810,39 @@ private void testNestedWrites(String initialValue, String updatedValue, TestInfo
...
+     * @param vertices An iterable of vertices
+     * @return boolean indicating if given vertices do not exist
+     */
+    private boolean verticesRemoved(Iterable<JanusGraphVertex> vertices) {
+        if (Iterables.isEmpty(vertices)) {
+            return true;
+        }
+        StandardJanusGraphTx queryTx = (StandardJanusGraphTx) graph.newTransaction();
+        for (JanusGraphVertex v : vertices) {
+            if (!graph.edgeQuery(v.longId(), graph.vertexExistenceQuery, queryTx.getTxHandle()).isEmpty()) {
```
rngcntr, line 1837: What if there are valid non-ghost vertices within the `Iterable` that just have no neighbors?
I think they are considered "removed" here, right? Should they?

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1785,7 +1786,7 @@ private void testNestedWrites(String initialValue, String updatedValue, TestInfo
 
         // Write schema and one vertex
         final PropertyKey prop = makeKey(propName, String.class);
-        createExternalVertexIndex(prop, INDEX);
+        mgmt.buildIndex("mixed", Vertex.class).addKey(prop, Mapping.STRING.asParameter()).buildMixedIndex(INDEX);
```
rngcntr, line 1789: What's the logical difference between this line and the original one?

li-boxuan: The original one builds a mixed index backed by TEXT, which supports full-text search but not equality check. Therefore, original line 1816
```java
assertEmpty(graph.query().has(propName, updatedValue).vertices());
```
does a full scan without using mixed index.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
li-boxuan: I guess you are misguided by the method name, "edgeQuery". It does not check neighbors, the method name is "edgeQuery" because almost everything is an edge in JanusGraph to my understanding. The method queries "edgeStore", which stores all vertices, properties, and edges.

rngcntr: Thanks for pointing that out! You are completely right, I misunderstood `edgeQuery`.

rngcntr (approved):

PR #1961, opened 2020-02-14
Apply optimizations to queries to make use of adjacent ids

rngcntr:
This feature was originally part of #1937 which was splitted to decrease per-PR-complexity.

This PR introduces two new OptimizerStrategies, which further improve the existing optimizations based on adjacent vertex ids. In the current state, these queries will be optimized to check the adjacency on the edge before iterating to the vertex itself:

- `g.V(left).outE('label').where(_.inV().is(right))`
- `g.V(left).outE('label').filter(_.inV().is(right))`

The optimization of these queries is within the scope of #1960:
- `g.V(left).outE('label').where(__.inV().hasId(right))`
- `g.V(left).outE('label').filter(__.inV().hasId(right))`

With the new optimizer, these queries will also profit of optimization:
- `g.V(left).outE('label').inV().is(right)`
- `g.V(left).outE('label').inV().hasId(right.id())`
- `g.V(left).out('label').is(right)`
- `g.V(left).out('label').hasId(right.id())`
- `g.V(left).out('label').filter(is(right))`
- `g.V(left).out('label').where(is(right))`
- `g.V(left).out('label').filter(hasId(right.id()))`
- `g.V(left).out('label').where(hasId(right.id()))`

All of this also applies to all other valid variations of `in()/out()/both()`.

### For all changes:
- [ ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

PR #2189, opened 2020-08-24
Optimize adjacency checks with unique index

rngcntr:
This PR aims to make the changes from #1961 more convenient to use. Since then, JanusGraph users are able to make use of adjacent IDs when traversing graphs like this:
```
alice = g.V().has('user', 'name', 'alice').next()
bob = g.V().has('user', 'name', 'bob').next()
g.V(alice).out('knows').as('x').in('knows').is(bob).select('x')
```
This query will efficiently return the people known by both Alice and Bob, even if the person in between is known by very many people. With this PR, we don't need to fetch the vertices of Alice and Bob in advance anymore:
```
g.V().has('user', 'name', 'alice').out('knows').as('x').in('knows').has('user', 'name', 'bob').select('x')
```
Again, we do not want to traverse all neighbor vertices of `x` and check their `name` property for the value `bob`, if we already know that there will be only a single vertex matching this property. Instead, we can determine the matching vertex and compare it directly as done before in #1961. To do so, we need to ensure that there _actually is_ only one matching vertex. We therefore check for each `has` step, whether it's combination of properties can be answered by a unique index. If so, this step qualifies for optimization.

So far, so good. But there are two little catches which come down to the transaction level:
1. At the time the query is optimized, a transaction is necessary to retrieve the uniqueness information about the indexes. So it is theoretically possible that this information changes between the query's optimization phase and it's actual execution. This can have two outcomes:
   - The index or it's uniqueness constraint are _added_ after the optimization phase: This is not really a problem because the query will just run without the optimization.
   - The index or it's uniqueness constraint are _dropped_ after the optimization phase: This can have quite an impact on performance. On sparsely connected large graphs, it is obviously better to traverse all neighbors of `x` instead of searching the _whole graph_ for `Bob`.
2. Please see my comment in the code. I don't know for sure if this is a problem so I kindly ask you to share your knowledge here.

Because of these two points, please see this PR as a PoC and as an opportunity to request comments on the general idea.

-----

### For all changes:

### For code changes:

### Note:


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/LazyHasStep.java
```diff
@@ -0,0 +1,54 @@
...
+        this.filterTraversal = filterTraversal.asAdmin();
+        integrateChild(this.filterTraversal);
+
+        HasContainer hc = new LazyHasContainer(ADJACENT_ID.name(), this::initializeFilter);
+        this.addHasContainer(hc);
+    }
+
+    private P<?> initializeFilter() {
+        if (filterTraversal.hasNext()) {
+            return P.eq(filterTraversal.next());
```
rngcntr, line 46: At this point, I'm not entirely sure if the `filterTraversal` uses the same transaction as the surrounding `traversal`. Does anyone know off the top of their heads, how such an access is handled by JanusGraph? And if not, can you hint me, where I have to look for it? So far, I was not able to find out, whether the same transaction is reused.

li-boxuan:
I like the idea! In general, I feel it's a good optimization. That being said, I wonder if it would lower the performance when (still, take g.V().has('user', 'name', 'alice').out('knows').as('x').in('knows').has('user', 'name', 'bob').select('x') as example):
1) x has no 'knows' edge (if there is vertex-centric index), or no edge at all.
2) x has very few (say, 1 or 2) incident edges.

I would guess this strategy makes more sense when 'x' has more than just a few neighbors. Would you consider making this strategy as optional and configurable (if that's possible), so that users can choose to turn this off (especially for advanced users who are able to write optimal queries)?

rngcntr:
Thanks for your comment! In both cases, I don't think we will see a (relevant) impact in performance for the following reasons:

1. If `x` has no (qualifying) edge, no traverser would be entering the following step, which would cause the unique index to be queried.
2. If `x` has 1 edge, one would normally load it's neighbor using the ID. What we now do instead is to load the neighbor using an index call. So the amount of loaded vertices stays the same but one additional index call is executed.

I will of course do some further testing and benchmarking before releasing this into the wild ;)

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/LazyHasStep.java
rngcntr: While verifying that both traversals use the same transaction, I realized that not only those two transactions are equal. The same transaction is actually used in the optimization phase. With this knowledge, there is no need to wait until runtime before evaluating the `filterTraversal`. The matching ID can already be retrieved at optimization time and embedded into the query.

That makes the optimization *much* easier and also drops the need for custom classes like `LazyHasStep` and `LazyHasContainer`.

rngcntr:
So now that the uncertanity about transactions has been eliminated, I consider this PR ready for review. In terms of performance, I promised to compare the query execution time with and without the optimization. The plotted runtimes are measured on a setup with Cassandra 3.11 and Elasticsearch 6.6.0 with JanusGraph 0.5.2 as a reference. All data points are median values of 100 consecutive runs on the same graph.
performance

li-boxuan:
> To be honest, I can not tell why queries execute slower on extremely small graphs.

Could it be due to database level cache, depends on how you run these tests? It would be awesome if you could share your benchmark code.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+            if (condition instanceof PredicateCondition) {
+                final RelationType type = ((PredicateCondition<RelationType,JanusGraphElement>) condition).getKey();
+                Preconditions.checkArgument(type != null && type.isPropertyKey());
+                Iterables.addAll(availableIndexes, ((InternalRelationType) type).getKeyIndexes());
+            }
+            return true;
+        });
+
+        // select only valid unique indexes
+        return availableIndexes.stream()
```
li-boxuan, line 102: Maybe you can put the selection logic into a new class that implements IndexSelectionStrategy interface

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+     * Determines whether this HasStep can be answered by a unique index and thus, returns either 0 or 1 match
+     */
+    @Override
+    protected boolean isValidStep(HasStep<?> step) {
+        StandardJanusGraphTx tx;
+
+        try {
+            tx = (StandardJanusGraphTx) JanusGraphTraversalUtil.getTx(step.getTraversal());
+        } catch (IllegalArgumentException iaex) {
+            // The traversal needs to be bound to a graph in order to get information about the indexes
```
li-boxuan, line 77: When would IllegalArgumentException happen? I don't quite understand the logic here.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexOptimizerStrategy.java
```diff
@@ -56,21 +51,18 @@
     }
 
     protected void optimizeStep(T step) {
-        P<?> predicate = parsePredicate(step);
-        if (isValidPredicate(predicate)) {
-            OptimizablePosition pos = getOptimizablePosition(step);
-            replaceSequence(step, pos, predicate);
+        OptimizablePosition pos = getOptimizablePosition(step);
+        if (pos != OptimizablePosition.NONE && isValidStep(step)) {
+            replaceSequence(step, pos);
```
li-boxuan, line 56: Can you profile these optimizations if possible? They could take some time (especially the index selection part). Profiling also makes it clear to users whether this optimization is applied or not.

li-boxuan (commented):
I haven't gone through everything in detail yet, but in general, this looks nice.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final AdjacentVertexHasUniquePropertyOptimizerStrategy INSTANCE =
+        new AdjacentVertexHasUniquePropertyOptimizerStrategy();
+
+    private AdjacentVertexHasUniquePropertyOptimizerStrategy() {}
+
+    public static AdjacentVertexHasUniquePropertyOptimizerStrategy instance() { return INSTANCE; }
+
+    @Override
+    public void apply(final Traversal.Admin<?, ?> traversal) {
+        TraversalHelper.getStepsOfClass(HasStep.class, traversal)
```
li-boxuan, line 63: I am not sure here, but maybe you want to skip this optimization if `TraversalHelper.onGraphComputer(traversal)`. TBH I didn't even know we could create nested traversal, thus I don't know what impact it will have on OLAP traversal.

li-boxuan (commented):
I think I understand the majority if not all logic now. Adding more comments below.

rngcntr:
> > To be honest, I can not tell why queries execute slower on extremely small graphs.
> 
> Could it be due to database level cache (depending on how you run these tests)? It would be awesome if you could share your benchmark code.
> 
> I am also interested if you can make your x-axis value (g.V(start).out().count()) start at 0, and also include small numbers 1, 2, 3, ... 10?

The benchmark pseudo code is:
```init_graph() // with labels and indexes
create_central_vertex()
for i in {10, 20, ..., 1000}:
    add_neighbors(10) // add 10 neighbors to central vertex
    measure_query(A, 100) // run query A 100 times
    measure_query(B) // run query B 100 times
```

To verify the results are not impacted by the order of A and B, I ran the test a second time with B first and then A. All measurements are merged together and the median is shown in the plot. Both results looked similar anyways, so the run times are probably not influenced by the execution order very much.

I will re-run the tests later today as you suggested, ranging from 0 to 100 and disabling `db-cache` for JanusGraph just to make sure it has no impact.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: It works perfectly fine without the `try...catch` block except for a single TinkerPop test class which somehow creates traversals which are not bound to a graph. I didn't try to find out how or why these traversals are created that way. But as the `getTx(...)` method explicitly checks if the traversal is bound to a graph, it looks to me like a valid case to have an unbound traversal.
My first plan was to avoid exception handling but then I would have to duplicate most of the code from `getTx(...)`. That's why I went with `try...catch`.

rngcntr: On second thought: I can try to extract only the line `traversal.getGraph().isPresent()` and see if that's already sufficient. Just saw that this is how other optimizers handle such cases.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexOptimizerStrategy.java
rngcntr: By profiling, do you mean to investigate the output of `query.profile()`? Or something else that I miss?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasIdOptimizerStrategy.java
```diff
@@ -73,4 +77,10 @@ protected boolean isValidPredicate(P<?> predicate) {
         Object predicateValue = predicate.getValue();
         return predicateValue instanceof Vertex || predicateValue instanceof Long;
     }
+
+    @Override
+    protected FilterStep<Edge> makeFilterByAdjacentIdStep(Traversal.Admin<?, ?> traversal, HasStep<?> originalStep) {
+        HasContainer hc = new HasContainer(ADJACENT_ID.name(), P.eq(parsePredicate(originalStep).getValue()));
+        return new HasStep<Edge>(traversal, hc);
```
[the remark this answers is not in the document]
rngcntr: Labels are not needed in this case because the ID is unique even across labels. What we do need is the edge label, but this is already contained in the preceeding `in()`/`out()` step.
Your comment howewer pointed me to the [edge layout](https://docs.janusgraph.org/advanced-topics/data-model/#individual-edge-layout) docs page where I realized we should check if there is a sort key defined for the given edge label before optimizing. Because as one can see, only when the sort key is empty, the sort order will be defined by the IDs.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+                new GraphStep<>(traversal, Vertex.class, true)
+            );
+
+        filterTraversal.addStep(originalStep);
+        filterTraversal.addStep(new IdStep<>(traversal));
+        return filterTraversal;
+    }
+
+    private long evaluateFilter(Traversal.Admin<?,Long> filterTraversal) {
+        // negative IDs are not valid in JanusGraph. Therefore, if the filterTraversal has no result, the newly
```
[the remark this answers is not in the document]
rngcntr: Agreed. Although it would be an `outE()`, `inE()`,... step instead of `out()`, `in()`, ... which is not as bad performance wise. But still bad :D
So I think the cleanest way to so is to create a new strategy which eliminates traversals with unsatisfiable conditions within them completely. Sounds nice to me :grin:

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private Traversal.Admin<?,Long> generateFilter(Traversal.Admin<?,?> traversal, HasStep<?> originalStep) {
+        Traversal.Admin filterTraversal = new DefaultGraphTraversal<>(traversal.getGraph().get())
+            .addStep(
+                new GraphStep<>(traversal, Vertex.class, true)
+            );
+
+        filterTraversal.addStep(originalStep);
+        filterTraversal.addStep(new IdStep<>(traversal));
```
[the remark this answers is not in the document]
rngcntr: Good catch!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+        if (filterTraversal.hasNext()) {
+            determinedId = (long) filterTraversal.next();
+        }
+        return determinedId;
+    }
+
+    @Override
+    protected FilterStep<Edge> makeFilterByAdjacentIdStep(Traversal.Admin<?, ?> traversal, HasStep<?> originalStep) {
+        Traversal.Admin<?,Long> filterTraversal = generateFilter(traversal, originalStep);
+        HasContainer hc = new HasContainer(ADJACENT_ID.name(), P.eq(evaluateFilter(filterTraversal)));
```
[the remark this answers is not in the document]
rngcntr: Yes, that's what it does. Same question as above: What exactly to you mean by profiling?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: You're right. I never did any experiments with OLAP traversals so it's probably wise to not tinker (pun intended) with them before gaining a deeper knowledge.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexOptimizerStrategy.java
```diff
@@ -90,37 +82,36 @@ private OptimizablePosition getOptimizablePosition(T originalStep) {
...
 
         Step<Edge,Vertex> e2vStep;
         e2vStep = (Step<Edge, Vertex>) originalStep.getPreviousStep();
 
         // create new has("~adjacent", id_value) step before e2v step
-        HasStep<Edge> hasAdjacentIdStep = makeHasAdjacentIdStep(traversal, predicate);
+        FilterStep<Edge> filterByAdjacentIdStep = makeFilterByAdjacentIdStep(traversal, originalStep);
 
         // remove original selection step
         traversal.removeStep(originalStep);
```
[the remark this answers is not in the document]
rngcntr: First of all, I don't know where I would get the exact step position from, in order to use `removeStep(final int index)`. My only idea would be to iterate all steps myself. 
And in addition, I'm not sure why it would be better to delete the old step beforehand. Notice that the new step is not added to the query at this point so it makes no difference in comparing all steps of the traversal.

rngcntr:
> I am also interested if you can make your x-axis value (g.V(start).out().count()) start at 0, and also include small numbers 1, 2, 3, ... 10?

It is again time for some benchmarks! First, I disabled the cache to see how much of a difference it makes. Additionally, I moved the range of neighbors from [1:1000] to [0:100]. Here are the results:
full
As one would expect, caching has a huge impact on the unoptimized version, because it loads all vertices on every access. For the optimized query, we see that — within the limits of accuracy — caching does not impact the query execution time.

So let's have a closer look at vertices with few neighbors as @li-boxuan suggested:
zoomed
Indeed, there is a noticable difference in favor of the unoptimized query for sparse neighborhoods. While most prominent at zero neighbors, this difference vanishes as soon as we hit at least three or four neighbors. Again, this is within the limits of accuracy of the measurement.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: Confirmed that it works like I described in the comment above.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexOptimizerStrategy.java
li-boxuan: Oh I mean you should make the query.profile() result contain the optimization latency here.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
li-boxuan: If you run query.profile(), the latency metrics would be inaccurate because it does not take this query latency into account. By profiling I mean you need to have a timer here, which is a part of the query profile.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexOptimizerStrategy.java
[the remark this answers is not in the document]
li-boxuan: I agree at the moment, putting this step beforehand makes no difference. What I meant was:

We know that `removeStep(originalStep)` essentially traverses all steps and compares the steps with originalStep using `equals()` method. We have to ensure originalStep is not modified after we extract it from the traversal, otherwise `equals()` will always return false.

In the last operation, `FilterStep<Edge> filterByAdjacentIdStep = makeFilterByAdjacentIdStep(traversal, originalStep);`, the `originalStep` is passed in as a parameter. Within that function, `originalStep` is added to a new ad-hoc traversal. This means, `originalStep` is in fact shared by the original traversal and the nested ad-hoc traversal. As discussed above, we have to make sure this `originalStep` is not mutated in the nested ad-hoc traversal, otherwise it cannot be correctly removed from the original traversal afterward. This sounds a bit risky to me because I don't see a way to enforce this "contract".

Using an analogy below, this is a competition between
```java
List<Object> list = new ArrayList<>(Arrays.asList(obj1, obj2));
runFunc(obj1);    // a function call that arguably might mutate obj1
list.remove(obj1);
```

vs


```java
List<Object> list = new ArrayList<>(Arrays.asList(obj1, obj2));
list.remove(obj1);
runFunc(obj1);
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasIdOptimizerStrategy.java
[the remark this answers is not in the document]
li-boxuan: Oh, I meant by label() step in gremlin query, not vertex label or edge label. 😉

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: Now I understand you and what you say makes perfect sense. Thanks for clarification!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: Alright, I'll try to find out how to add a timer to the query profile. I thought that would happen automatically for optimization strategies because `optimization` appears quite often in the query profiles that I see.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasIdOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: Can you give an example of a query where you think a `label()` step would be relevant? I can't think of one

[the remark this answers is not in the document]
rngcntr: Regarding the sort key, applying the optimization is still advantageouos even with sort keys defined. This may cause a linear search across the edges for the one with the correct ID but this if of course still faster than a linear search across the neighbors with the correct property value.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: Adding custom profiling outside the actual query execution is not intended. The metrics, in which the profile is stored, are created within the first `next()` or `hasNext()` call of a `ProfileStep`. Therefore, there is no profiler available at optimization time.
The only "solution" that comes to my mind is a pretty ugly one and in fact a complete re-write of the current state:
Instead of doing the ID resolution at optimization time, we would have to do it at runtime. To do so, we would need to implement a custom step which holds both the unoptimized query and a skeleton of the optimized query as it's children. Then at runtime, this step could evaluate the index uniqueness and dynamically decide which path to use. It would then also have to inject the resolved ID into the optimized query somehow. In my eyes, this is a mess...

If you have any other idea, please let me know.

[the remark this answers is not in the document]
li-boxuan: To do profiling in the optimization step, you can actually create a new `TP3ProfileWrapper` and later in runtime, add the metrics of this "optimization profiler" to the query profiler. I did something similar in another issue I have been working on which I would probably submit a PR in 1-2 weeks.

Actually I am thinking that this ad-hoc traversal shall indeed be done only in runtime. Let's say we have
```groovy
g.V().has("key1", "value1").out(label).has("key2", "value2")
```
If `g.V().has("key1", "value1")` does not exist, the query will finish. It could also be the case that this `g.V().has("key1", "value1")` vertex does not have any `out(label)` edge at all. Under such "early-stopping" circumstances, the ad-hoc traversal done for `has("key2", "value2")` in the optimization stage becomes meaningless and brings only overhead. If this ad-hoc traversal is lazily computed, we don't have such a problem.

That being said, I do also see the potential performance improvement if we have `has("key2", "value2")` in the optimization stage, since it can eliminate the need of the whole traversal when `has("key2", "value2")` returns null. If we don't put this logic in this strategy, then it does not make much sense to have the ad-hoc traversal in optimization stage.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasIdOptimizerStrategy.java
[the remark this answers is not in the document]
li-boxuan: whoops sorry, I meant by as(stepLabel) step, e.g. `g.V(vertex).out().has("key", "value").as("v").addE("connects").from("v").next()`

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: As mentioned [earlier](https://github.com/JanusGraph/janusgraph/pull/2189#discussion_r479911325), I would prefer not to have this logic in the same strategy but instead have a "cleanup" strategy which runs at the end of the optimization phase and prunes the query by eliminating all sub queries that contain unsatisfiable condition.
Other optimization strategies (like this one) could then mark conditions as unsatisfiable by adding/replacing with a `hasId(-1)` step for example. This way, we can ensure we don't have to implement a similar logic in every strategy that makes use of it.

Also thanks for pointing me towards `TP3ProfileWrapper`. If you don't mind, I would like to wait until I see how you are handling this in your PR.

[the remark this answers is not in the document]
li-boxuan: > I would prefer not to have this logic in the same strategy but instead have a "cleanup" strategy...

That's fair.

> I would like to wait until I see how you are handling this in your PR

Sure 😉

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasIdOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: Finally coming back to this. Very interesting thought, but I think as the only thing that we actually replace are `has` steps, step labels should not interfere. The [TinkerPop documentation](https://tinkerpop.apache.org/docs/current/reference/#has-step) shows various usages of the `has` step. Most of them use only literal values which cannot be derived from step labels. The only exceptions are:
- `has(key, predicate)` which is ok because predicates also only take literal values.
- `has(key, traversal)` which is also ok because nested traversals don't have access to the surrounding traversal's step labels in the original query either.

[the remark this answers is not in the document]
li-boxuan: I am not sure if I got you, but let me explain what I am worried about:

Let's say we have a normal query:
```groovy
g.V(v1).as("v1").out().has("prop", "value2").as("v2").select("v1", "v2").by("prop")
```

I suspect that with your change, it would be translated to something equivalent to  (note the missing 'as("v2")' step):
```groovy
g.V(v1).as("v1").out().hasId(v2).select("v1", "v2").by("prop")
```
which leads to the wrong result.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
li-boxuan: nice pun!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasIdOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: Yes, you were right. What I thought about was the impact of labels for the `has` step not _on_ the `has` step. I wrote a test to verify this and indeed had to fix two lines. 
Glad you found that because two existing strategies were also affected.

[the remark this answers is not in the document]
li-boxuan: Do you mean the same problem exists in two strategies released in v0.5? If that's the case, I suggest you open up a separate PR to fix them, in case this PR cannot be merged in the next release.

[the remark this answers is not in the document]
rngcntr: Sounds like the right thing to do

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
rngcntr: It's been a while and after this PR was requested to be merged soon in yesterday's JanusGraph meetup, I came back to look at what's still missing here in order to merge the PR. Throughout the day I tried different ways to integrate the collection of metrics into this optimization strategy. But in contrast to what you did in your PR, the step within the traversal that is created by my optimization strategy, is later merged with previous `JanusGraphVertexStep`s and will therefore not receive it's own query metrics by the query executor.
The problem is that the `has` step created by my strategy is only intermediate and will never remain once the query is completely optimized by the `JanusGraphLocalQueryOptimizerStrategy`. Therefore at the time the initial `Metrics` Object is created, there is no step remaining that can possibly be responsible for accounting the time spent in the optimization phase.

One possible solution would be to create a custom `IdentityStep` within the query that does nothing but hold the metrics from the optimization phase. But this step could possibly interfere with other optimization strategies and get removed by them or make it harder for them to be applied.

Concluding, I don't see a really good solution for this problem. IMHO the cleanest way would be to somehow measure the time for each optimization independently from the query execution, bundle them and somehow integrate them in the profile result afterwards. But this would require a lot of work to be done and design decisions to be made.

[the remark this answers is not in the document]
rngcntr: Quick addition: As some optimizations are already applied by TinkerPop and therefore outside of JanusGraph, it would obviously preferred to have this kind of functionality as part of TinkerPop, not JanusGraph.

li-boxuan:
> Indeed, there is a noticeable difference in favor of the unoptimized query for sparse neighborhoods. While most prominent at zero neighbors, this difference vanishes as soon as we hit at least three or four neighbors

My understanding is, it can be very common for a node to have only 1-2 neighbors in a sparse graph. Although this can be a nice optimization in some scenarios, I'd suggest making it possible for advanced developers to disable it. For example, in our use cases, gremlin queries are written by developers only, and we generally hold the assumption that developers are able to write the most optimized query.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
li-boxuan: I yet to find some time to review this, but 
> this PR was requested to be merged soon in yesterday's JanusGraph meetup

this grasps my attention 😄.  I didn't attend the meetup but I am curious about the request: is there a planned date for the 0.6.0 release? If not, why would users (I assume this was a request by some user) think merging this PR soon will help them? Is it because they are using the master version of JanusGraph?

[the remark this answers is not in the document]
rngcntr: I don't think so. But from a user perspective if they neither know when 0.6.0 will be released nor when this feature will be merged, chances of getting it with v0.6.0 are better the earlier this change is in.
If you would like to view the recording of the meetup (where I gave a talk, too :tada:), I think @twilmes said he plans to upload it somewhere it's accessible for those who were not able to make it to the meetup.

rngcntr:
@li-boxuan Making the backend access for optimization strategies configurable sounds like a good idea. I created a global flag to disable it if needed.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+        Set<Class<? extends ProviderOptimizationStrategy>> postStrategies = new HashSet<>();
+        postStrategies.add(JanusGraphStepStrategy.class);
+        postStrategies.add(JanusGraphLocalQueryOptimizerStrategy.class);
+        return postStrategies;
+    }
+
+    @Override
+    public void apply(final Traversal.Admin<?, ?> traversal) {
+        if (TraversalHelper.onGraphComputer(traversal)
+            || traversal.getStartStep() instanceof TraversalVertexProgramStep // not caught by TraversalHelper
```
li-boxuan, line 72: This looks interesting. How did you find this step? Did you know why it is not caught by TraversalHelper.onGraphComputer?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+            || !traversal.getGraph().isPresent()) {
+            return;
+        }
+
+        Graph graph = traversal.getGraph().get();
+        StandardJanusGraph janusGraph = graph instanceof StandardJanusGraph
+            ? (StandardJanusGraph) graph
+            : ((StandardJanusGraphTx) graph).getGraph();
```
li-boxuan, line 80: Curious: when is it a StandardJanusGraph instance and when is it a StandardJanusGraphTx instance?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+
+        if (!janusGraph.getConfiguration().optimizerBackendAccess()) {
+            return;
```
li-boxuan, line 83: Nice. Can you add a test case for this?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: This odd case occured to me while I was trying to fix failing tests that were atually failing because of the `StandardJanusGraph` vs `StandardJanusGraphTx` problem. While debugging I found out that `TraversalHelper.onGraphComputer()` does return false for traversals that start with `TraversalVertexProgramStep`s, which are obviously graph computer traversals. So I had a look into `TraversalHelper` (which is provided by TinkerPop, not JanusGraph) and what I found is that it only considers traversals as graph computer traversals if their _parent_ traversal begins with such a step. For the top layer traversal itself, it returns `false`.
What I also found out is that in current versions of TinkerPop, the implementation of this method has changed. However, I did not investigate whether or not this case is included in newer versions of TinkerPop.

When debugging, I saw that this change reduced the number of failing tests from 11 down to 2, so there must be some actual difference. However, all the 11 failing tests were fixed by introducing the `StandardJanusGraph`/`StandardJanusGraphTx` distinction, so this check isn't actually required. I still kept it in just to be sure, because it seemingly had a positive impact on the test runs earlier.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: I found out that after introducing the config option, some tests failed because the graph was not a `StandardJanusGraph` as I expected. Instead, the traversal's graph was a `StandardJanusGraphTx`. I did not take a deeper look into the reason for that tbh.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: Done as a separate commit. I will squash the commits once the PR is going to be merged.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
li-boxuan: Can you point out some test cases where the graph is a StandardJanusGraphTx?

rngcntr: `InMemoryJanusGraphComputerTest#testMessagePassingOut` for example. But there are some others in the same test class.
You can just comment out the distinction between `StandardJanusGraph` and `StandardJanusGraphTx` and you will end up with `ClassCastException`s.

li-boxuan: I can see that comes from FulgoraGraphComputer which I am unfamiliar with. Anyway, there are only two implementations of graph interface in JanusGraph codebase, namely StandardJanusGraph and StandardJanusGraphTx, so I think this is good enough.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+
+        filterTraversal.addStep(originalStep);
+        filterTraversal.addStep(new IdStep<>(filterTraversal));
+        return filterTraversal;
+    }
+
+    private long evaluateFilter(Traversal.Admin<?,Long> filterTraversal) {
+        // negative IDs are not valid in JanusGraph. Therefore, if the filterTraversal has no result, the newly
+        // inserted HasStep should not match anything
+        long determinedId = -1;
```
li-boxuan, line 132: This does not work due to https://github.com/JanusGraph/janusgraph/blob/4f44a1db63513f01cd118b6ccbe24b5cdb36878d/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BaseVertexCentricQueryBuilder.java#L107-L111

You can see this problem by adding a test case where the adjacent vertex does not exist, e.g. `assertNumStep(1, 0, g.V(sv[0]).out().has("uniqueId", 10000), HasStep.class);`

li-boxuan (commented):
I just again did a pass on this PR and left some comments, but I am still not confident to say I understand every detail. That being said, I think it's in general in a good shape 👍 

Since there are so many subtleties in this PR (in fact, I find every bug in traversal strategies to be very difficult to spot), can you check https://codecov.io/gh/JanusGraph/janusgraph/commit/5f26b1ed7a450b1aff1c2146605c5ba38321dffc and see if you are able to cover more if not all of your changes in tests? If you can squash your commits, I think it would be easier to check on codecov since seems its coverage report is generated on a per-commit basis.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    @Override
+    protected boolean isValidStep(HasStep<?> step) {
+        StandardJanusGraphTx tx = (StandardJanusGraphTx) JanusGraphTraversalUtil.getTx(step.getTraversal());
+
+        List<String> givenKeys = step.getHasContainers().stream()
+            .map(HasContainer::getKey).collect(Collectors.toList());
+
+        List<PredicateCondition<String, JanusGraphElement>> constraints = step.getHasContainers().stream()
+            .filter(hc -> hc.getBiPredicate() == Compare.eq)
```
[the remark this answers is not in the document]
rngcntr: Wouldn't it be better to avoid nesting the necessary logic in here and instead have a strategy that replaces P.within(singleton) by P.eq?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+public class AdjacentVertexHasUniquePropertyOptimizerStrategy
+    extends AdjacentVertexOptimizerStrategy<HasStep<?>> {
+
+    private static final AdjacentVertexHasUniquePropertyOptimizerStrategy INSTANCE =
+        new AdjacentVertexHasUniquePropertyOptimizerStrategy();
+
+    private AdjacentVertexHasUniquePropertyOptimizerStrategy() {}
+
+    public static AdjacentVertexHasUniquePropertyOptimizerStrategy instance() { return INSTANCE; }
```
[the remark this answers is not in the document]
rngcntr: I actually prefer line breaks, too. I went and applied them to every strategy now.

rngcntr:
> Do you want to ask or report to the TinkerPop community?

There are no reply or resolve conversation options below your comment so I continue this in a separate post.
I'm pretty confident this is by design. Graph computer traversals are wrapped inside `TraversalVertexProgramStep`s while the traversal that contains this step isn't actually executed on the graph computer.
Since my strategy is not applied recursively, nested graph computer traversals are not optimized either way, so I removed this line again.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: Crucial! I think I found a better solution now, since I discovered Gremlin supports a `none()` step, which is exactly what I tried to emulate here with `has('~adjacent', -1)` but with style :grin: 

I also included your test case (which should be `assertNumStep(0, 0, ...`, right?) and will now wait for the build to have look at the code coverage and see where I need additional tests. Thanks for taking the time once again to review my PR!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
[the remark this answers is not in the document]
li-boxuan: well said, now it surprises me that this optimization hasn't been done yet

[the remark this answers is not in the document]
rngcntr: There are lots of small things that _could_ be done, but the sheer amount of these little tweaks makes it hard to cover all of them

janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategyTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static TraversalStrategy optimizer = AdjacentVertexHasUniquePropertyOptimizerStrategy.instance();
+
+    @Test
+    public void testWithBackendAccess() {
+        makeSampleGraph();
+
+        // AdjacentVertexHasUniquePropertyOptimizer
+        assertNumStep(1, 0, g.V(sv[0]).out().has("uniqueId", 0), HasStep.class);
+        assertNumStep(1, 0, g.V(sv[0]).out().barrier(2500).has("uniqueId", 0), NoOpBarrierStep.class);
```
li-boxuan, line 41: Question (not directly related to this line but since you are testing barrier step): is there any difference if your Strategy is put before / after LazyBarrierStrategy?

li-boxuan:
Can you show profile() result for the following cases:
a) optimization applied and there is a result
b) optimization applied and there is no result
c) no optimization applicable (no composite index with unique constraint applicable)

Can you also try if `graph.traversal().withoutStrategies(AdjacentVertexHasIdOptimizerStrategy)` works? Hopefully, it should exclude this strategy in traversal (TBH I never tried), which means users can tune their queries by turning off this strategy temporarily if needed (especially when they know there are very few neighbors). Then we can create a doc explaining how this strategy works, when it works well, and when it doesn't work well. This could be a follow-up issue.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+        return determinedId;
+    }
+
+    @Override
+    protected FilterStep<Edge> makeFilterByAdjacentIdStep(Traversal.Admin<?, ?> traversal, HasStep<?> originalStep) {
+        Traversal.Admin<?,Long> filterTraversal = generateFilter(traversal, originalStep);
+        long adjacentId = evaluateFilter(filterTraversal);
+        if (adjacentId == -1) {
+            return new NoneStep<>(traversal);
```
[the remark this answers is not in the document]
rngcntr: Yes of course

janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategyTest.java
rngcntr: I had to ensure that barrier steps between the `out()` step and the `has()` step don't break the strategy, because I want my generated filter step to be merged into a preceeding vertex step. This should happen no matter if there is a `barrier()` in between these steps, so I tested it.

janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategyTest.java
```diff
@@ -0,0 +1,69 @@
...
+        assertNumStep(1, 1, g.V(sv[0]).barrier(2500).out().barrier(2500).has("uniqueId", 0), NoOpBarrierStep.class);
+        assertNumStep(1, 0, g.V(sv[0]).out().barrier(2500).has("uniqueId", 0), HasStep.class);
+        assertNumStep(1, 1, g.V(sv[0]).out().has("id", 0), HasStep.class);
+        assertNumStep(1, 0, g.V(sv[0]).out().has("id", 0).has("uniqueId", 0), HasStep.class);
+        assertNumStep(0, 0, g.V(sv[0]).out().has("uniqueId", 10000), HasStep.class);
```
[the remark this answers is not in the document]
rngcntr: Nice idea

rngcntr:
As I [mentioned](https://github.com/JanusGraph/janusgraph/pull/2189#discussion_r501668591) there is no clean way to include optimization timings into the profile output.

a) `g.V(sv[0]).out().has("uniqueId", 0).profile().next()`
```
Traversal Metrics
GraphStep(vertex,[v[32792]])                                           1           1           1.700    26.02
JanusGraphVertexStep([~adjacent.eq(4232)])                             1           1           0.894    13.68
    \_condition=(EDGE AND visibility:normal)
    \_orders=[]
    \_isFitted=false
    \_isOrdered=true
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8009c75e
    \_vertices=1
  optimization                                                                                 0.019
EdgeVertexStep(IN)                                                     1           1           3.940    60.30
                                            >TOTAL                     -           -           6.535        -
```
b) `g.V(sv[0]).out().has("uniqueId", -1).profile().next()`
```
Traversal Metrics
GraphStep(vertex,[v[32792]])                                           1           1           0.082     2.73
JanusGraphVertexStep(OUT,edge)                                       100         100           2.579    84.99
    \_condition=(EDGE AND visibility:normal)
    \_orders=[]
    \_isFitted=false
    \_isOrdered=true
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8009c75e
    \_vertices=1
  optimization                                                                                 0.009
NoneStep                                                                                       0.336    11.09
EdgeVertexStep(IN)                                                                             0.036     1.20
                                            >TOTAL                     -           -           3.035        -
```
c) `g.V(sv[0]).out().has("id", 0).profile().next()`
```
Traversal Metrics
GraphStep(vertex,[v[32792]])                                           1           1           0.173     0.89
JanusGraphVertexStep(OUT,vertex)                                     100         100           2.153    11.12
    \_condition=(EDGE AND visibility:normal)
    \_orders=[]
    \_isFitted=false
    \_isOrdered=true
    \_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@8009c75e
    \_vertices=1
  optimization                                                                                 0.013
HasStep([id.eq(1)])                                                    1           1          17.047    87.99
                                            >TOTAL                     -           -          19.374        -
```
I've also added a test case to test the behavior of `withoutStrategies`.

One more thing: The code coverage reports you mentioned are not really helpful to me. When I look at [this page](https://codecov.io/gh/JanusGraph/janusgraph/tree/9dd66bb6718e6d255fffc5518b2d5cb359a8ed9a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy), it tells me that I have one line only partially covered in my strategy class. But once I click on the file to see which line it is, it throws an error. Same for any other file.

li-boxuan:
> One more thing: The code coverage reports you mentioned are not really helpful to me. When I look at this page, it tells me that I have one line only partially covered in my strategy class. But once I click on the file to see which line it is, it throws an error. Same for any other file.

I could open the link and opened the files. Maybe there was server error on codecov side when you tried? Btw, https://codecov.io/gh/JanusGraph/janusgraph/commit/37ba127ce4df64f4c1740c39e545135cb73f89c4 this should show all codecov of your changes in a single page. The report seems pretty useful. For example, "remove obsolete NoOpBarrier" part in replaceSequenceV2EthenE2VthenID method is not tested.

janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategyTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void testWithAndWithoutStrategy() {
+        makeSampleGraph();
+        assertNumStep(1, 0, g.V(sv[0]).out().has("uniqueId", 0), HasStep.class);
+        assertNumStep(1, 1, graph.traversal().withoutStrategies(AdjacentVertexHasUniquePropertyOptimizerStrategy.class).V(sv[0]).out().has("uniqueId", 0), HasStep.class);
+    }
+
+    @Test
+    public void testWithBackendAccess() {
```
li-boxuan, line 44: What about the case you mentioned in the https://github.com/JanusGraph/janusgraph/pull/2189#issue-472481806, i.e.,

g.V().has('user', 'name', 'alice').out('knows').as('x').in('knows').has('user', 'name', 'bob').select('x')

li-boxuan:
> As I mentioned there is no clean way to include optimization timings into the profile output.

Yeah, got it. Just wanted to see if the profile output makes sense to users. They look good to me!

janusgraph-test/src/test/java/org/janusgraph/graphdb/tinkerpop/optimize/AdjacentVertexHasUniquePropertyOptimizerStrategyTest.java
rngcntr: That is an equivalent example which is just more readable because of the given names. I don't even remember why I used step labels for that example, as they only add complexity. Probably did so in order to point out that this part can be nested in arbitrary traversals.

li-boxuan (approved):

farodin91 (approved):
LGTM! Thank you @rngcntr. Do you like to rebase before merging?

li-boxuan:
@rngcntr Can you create a follow-up issue for profiling the optimization phase? It would be helpful if you could explain the approach(es) you tried and why it was hard to profile it, as a hint for whoever wants to tackle this.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionUtil.java
```diff
@@ -25,4 +40,29 @@ public static boolean indexCoversOrder(MixedIndexType index, OrderList orders) {
         }
         return true;
     }
+
+    public static Set<IndexType> getMatchingIndexes(MultiCondition<JanusGraphElement> conditions) {
```
porunov, line 44: Nitpick: The result of this method is always filtered. Instead of filling `Set` with values and then removing them, we could just pass filter (or predicate) function to this method and filter it directly here without conducting additional pass to filter the resulting set afterwards.

rngcntr: I see what you mean. Do you think it's worth creating an issue to rewrite the method?

rngcntr:
@li-boxuan Done! #2335

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionUtil.java
porunov: I think issue could be helpful so that we don't forget to change it later

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -0,0 +1,141 @@
...
+        if (TraversalHelper.onGraphComputer(traversal) || !traversal.getGraph().isPresent()) {
+            return;
+        }
+
+        Graph graph = traversal.getGraph().get();
+        StandardJanusGraph janusGraph = graph instanceof StandardJanusGraph
+            ? (StandardJanusGraph) graph
+            : ((StandardJanusGraphTx) graph).getGraph();
+
+        if (!janusGraph.getConfiguration().optimizerBackendAccess()) {
```
porunov, line 84: I think this check could be placed on a more high level. Instead of always triggering `AdjacentVertexHasUniquePropertyOptimizerStrategy.java` for each query we can just skip adding this optimization strategy when this check is disabled in JanusGraph. When `janusGraph.getConfiguration().optimizerBackendAccess()` is `false` we don't even need to add this optimization to our life-cycle. Unless I miss something

porunov: I.e. I see that `TraversalStrategies` are registered in a static module in `StandardJanusGraph` but I think we could refactor that later. I guess we need just a separate `refactor` issue for that

porunov: Created #2338

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionUtil.java
porunov: Created #2339

PR #2200, opened 2020-09-05
Add strategy that optimize mixed index lookup followed by count step

### LINKED ISSUE #874: Add strategy that will optimize mixed index lookup followed by a count
I think we can make use of the work accomplished in #318 to optimize queries with the following structure `g.V().has("text", textFuzzy("foo")).count()`. The strategy should be able to replace the default counting behavior with one that uses the count retrieved from the index provider.

li-boxuan:
This adds JanusGraphMixedIndexCountStrategy which leverages mixed index
to return the total count if applicable. The implementation idea is to
leverage the partial result of JanusGraphStepStrategy and JanusGraphStep
which handle folding and building a global graph centric query.

Closes #874

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -76,6 +76,9 @@ import org.janusgraph.graphdb.query.index.ApproximateIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.index.BruteForceIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.index.ThresholdBasedIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.profile.QueryProfiler;
+import org.janusgraph.graphdb.tinkerpop.optimize.step.JanusGraphMixedIndexCountStep;
+import org.janusgraph.graphdb.tinkerpop.optimize.step.JanusGraphStep;
+import org.janusgraph.graphdb.tinkerpop.optimize.strategy.JanusGraphMixedIndexCountStrategy;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.graphdb.types.ParameterType;
 import org.janusgraph.graphdb.types.StandardEdgeLabelMaker;
@@ -1124,6 +1127,7 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         if (indexFeatures.supportNotQueryNormalForm()) {
             newTx();
             assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).count().next());
+            assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).toList().size());
             Metrics mMixedOr = tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val"))
                 .profile().next().getMetrics(0);
             assertEquals("Or(JanusGraphStep([],[name.eq(bob)]),JanusGraphStep([],[prop.eq(val)]))", mMixedOr.getName());
@@ -1150,6 +1154,7 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         // satisfied by two graph-centric queries, one is mixed index query and the other one is full scan
         newTx();
         assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).count().next());
+        assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).toList().size());
         Metrics mMixedOr = tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20))
             .profile().next().getMetrics(0);
         assertEquals("Or(JanusGraphStep([],[name.eq(bob)]),JanusGraphStep([],[age.eq(20)]))", mMixedOr.getName());
@@ -1202,6 +1207,7 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         // satisfied by a single graph-centric query which is satisfied by a single mixed index query
         newTx();
         assertEquals(1, tx.traversal().V().has("name", "bob").has("prop", "val").count().next());
+        assertEquals(1, tx.traversal().V().has("name", "bob").has("prop", "val").toList().size());
         Metrics mMixedAnd = tx.traversal().V().has("name", "bob").has("prop", "val")
             .profile().next().getMetrics(0);
         assertEquals("JanusGraphStep([],[name.eq(bob), prop.eq(val)])", mMixedAnd.getName());
@@ -1228,6 +1234,7 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         // satisfied by a single graph centric query which is satisfied by union of two mixed index queries
         newTx();
         assertEquals(1, tx.traversal().V().has("name", "bob").has("prop", "val").count().next());
+        assertEquals(1, tx.traversal().V().has("name", "bob").has("prop", "val").toList().size());
         final Metrics mMixedAnd2 = tx.traversal().V().has("name", "bob").has("prop", "val")
             .has("desc", Text.textContains("coding")).profile().next().getMetrics(0);
         assertEquals("JanusGraphStep([],[name.eq(bob), prop.eq(val), desc.textContains(coding)])", mMixedAnd2.getName());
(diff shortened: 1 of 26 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/IndexQueryBuilder.java
```diff
@@ -152,6 +184,73 @@ public String getUnknownKeyName() {
...
+        if (Contain.IN.equals(constraint.getPredicate())) {
+            queryBuilder.append("(");
+            for (Object value : (List) constraint.getValue()) {
+                queryBuilder.append("\"" + value + "\"");
+            }
+            queryBuilder.append(")");
+        } else if (Text.CONTAINS.equals(constraint.getPredicate()) || Cmp.EQUAL.equals(constraint.getPredicate())) {
+            queryBuilder.append("\"" + constraint.getValue() + "\"");
+        } else {
+            // TODO: we should support all types as long as a single mixed query can be constructed
```
li-boxuan, line 249: Now I think constructing a direct index query might be the wrong direction. Using existing `Stream<String> query(IndexQuery query, KeyInformation.IndexRetriever information, BaseTransaction tx)` in IndexProvider and appending `count` seems better.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphMixedIndexCountStep.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public MixedIndexCountQuery buildGlobalGraphIndexQuery(final GraphStep originalGraphStep, final JanusGraphTransaction tx) {
+        optProfiler = new TP3ProfileWrapper(new MutableMetrics(this.getId() + "optimization", "optimization"));
+        MixedIndexCountQueryBuilder countQueryBuilder = (MixedIndexCountQueryBuilder) tx.mixedIndexCountQuery();
+        List<PredicateCondition<String, JanusGraphElement>> constraints = new ArrayList<>();
+        for (final HasContainer condition : hasContainers) {
+            constraints.add(new PredicateCondition<>(condition.getKey(), JanusGraphPredicateUtils.convert(condition.getBiPredicate()), condition.getValue()));
+        }
+        return countQueryBuilder.constructIndex(optProfiler, constraints,
```
rngcntr, line 82: I like the Idea that the step itself performs the time consuming task of index selection, which makes it possible to track the work done in the optimization phase. Great idea!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphMixedIndexCountStrategy.java
```diff
@@ -0,0 +1,147 @@
...
+            return;
+        }
+        final JanusGraphMixedIndexCountStep<?> directQueryCountStep = new JanusGraphMixedIndexCountStep<>(traversal);
+        // we shouldn't really fold in, but only check it, because we don't want to break the original traversal,
+        // as this optimizer might fail
+        if (!foldInHasContainer(originalGraphStep, directQueryCountStep, traversal, traversal)) {
+            return;
+        }
+        final JanusGraphTransaction tx = JanusGraphTraversalUtil.getTx(traversal);
+        MixedIndexCountQuery mixedIndexCountQuery = directQueryCountStep.buildGlobalGraphIndexQuery(originalGraphStep, tx);
```
rngcntr, line 79: It took me about half an hour to figure out that _this_ is actually the line that checks if the optimization should be applied at all. I tried to find the line that clearly differentiates between mixed and composite indexes. The fact that this is buried deeply inside the index selection part is a bit of an obfuscation of what actually happens here.
Now that it became clear to me, it would be good to at least point out exactly what happens here.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/JanusGraphMixedIndexCountStrategy.java
```diff
@@ -0,0 +1,147 @@
...
+        if (mixedIndexCountQuery != null) {
+            directQueryCountStep.setMixedIndexCountQuery(mixedIndexCountQuery);
+            removeHasContainerSteps(originalGraphStep, directQueryCountStep, traversal);
+        }
+    }
+
+    private boolean hasCountGlobalStep(final GraphStep originalGraphStep) {
+        Step<?, ?> currentStep = originalGraphStep.getNextStep();
+        while (currentStep instanceof IdentityStep || currentStep instanceof NoOpBarrierStep || currentStep instanceof HasContainerHolder) {
```
rngcntr, line 88: These `instanceof` chains are used quite often so it would be better to move the check to a separate method. This would also give these steps a name that is more readable. For example `while (isEligibleToSkip(currentStep)) {...}`

rngcntr (commented):
I took a look at your PR and really can't say much about it. It touches many components that I was unaware of before so I can't tell to which extent the introduced changes are appropriate. Left some comments though

rngcntr:
One more thing: Have you measured the performance impact as you did for your previous PRs?

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1112,6 +1113,99 @@ public void testIndexParameters() {
...
+    public void testMixedIndexQueryFollowedByCount() {
+        final int numV = 1000;
+        long total;
+        TraversalMetrics metrics;
+        final String[] strings = {"Uncle Berry has a farm", "and on his farm he has five ducks", "ducks are beautiful animals", "the sky is very blue today"};
+        setupChainGraph(numV, strings, true);
+        clopen();
+
+        /* ===================================================
+        has(key, value) followed by count() can be optimized */
```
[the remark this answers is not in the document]
li-boxuan: Totally. I need to add more test cases.

porunov:
@li-boxuan Do you have plans to work on this PR? Would be great to see this feature as many users are not aware of direct index queries

li-boxuan:
@porunov Thanks for reminder! I already forgot about this PR 😅 I will try to find some time to work on it this weekend, but please don’t count it as a blocker for 0.6.0 release.

li-boxuan:
Based on https://github.com/JanusGraph/janusgraph/pull/2684/

porunov:
@li-boxuan Looks like "Reduce redundant global centric query build" commit is still in this PR

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1523,6 +1525,203 @@ public void testIndexParameters() {
...
+
+        // vertex
+        total = numV / strings.length * 2;
+        assertEquals(total, graph.indexQuery(VINDEX, "v.text:ducks").vertexTotals());
+        TraversalMetrics profile = graph.traversal().V().has("text", Text.textContains("ducks")).count().profile().next();
+        final String xtextKey = "xtext";
```
porunov, line 1557: To pass Solr tests it probably should be:
```
final String xtextKey = getStringField("xtext");
```

Maybe in some other places as well. I didn't check the whole test yet

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexCountStrategy.java
```diff
@@ -0,0 +1,121 @@
...
+        TraversalHelper.replaceStep(originalGraphStep, mixedIndexCountStep, traversal);
+    }
+
+    public static JanusGraphMixedIndexCountStrategy instance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public Set<Class<? extends ProviderOptimizationStrategy>> applyPrior() {
+        return Collections.singleton(JanusGraphStepStrategy.class);
```
porunov, line 119: Looks like this one can actually be a constant (i.e. `private final static`)

porunov (approved):
LGTM. Thank you @li-boxuan for such a great improvement! No need to use direct index queries for count queries anymore :)
I have only one very small nitpick which not necessary to be resolved.

li-boxuan:
@rngcntr Do you want to review again?

li-boxuan:
Using the following code to do a simple performance test (single Cassandra + single Elasticsearch)

```java
public class JanusGraphCqlApp {
    public static void main(String[] args) {
        Map<String, String> map = new HashMap() {{
            put("storage.backend", "cql");
            put("storage.hostname", "127.0.0.1");
            put("storage.cql.keyspace", "janusgraph");
            put("index.search.backend", "elasticsearch");
            put("index.search.hostname", "127.0.0.1");
        }};
        MapConfiguration config = new MapConfiguration(map);
        config.setListDelimiterHandler(new DefaultListDelimiterHandler(','));
        JanusGraph graph = JanusGraphFactory.open(config);
        JanusGraphManagement mgmt = graph.openManagement();
        PropertyKey stringKey = mgmt.makePropertyKey("name").dataType(String.class).make();
        mgmt.buildIndex("namev", Vertex.class).addKey(stringKey, Mapping.STRING.asParameter())
            .buildMixedIndex("search");
        mgmt.commit();
            for (int j = 0; j < 10000; j++) {
                graph.addVertex("name", "value");
            }
            graph.tx().commit();
        }

        // query
        long startTime = System.currentTimeMillis();
        long count = graph.traversal().V().has("name", "value").count().next();
        System.out.println("count = " + count + ", time spent (ms) = " + (System.currentTimeMillis() - startTime));
        graph.close();
    }
}
```

master:

count = 1000000, time spent (ms) = 318092
count = 1000000, time spent (ms) = 399118
count = 1000000, time spent (ms) = 315376


with this PR:

count = 1000000, time spent (ms) = 181
count = 1000000, time spent (ms) = 210
count = 1000000, time spent (ms) = 172

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1523,6 +1529,202 @@ public void testIndexParameters() {
...
+
+        profile = graph.traversal().V().has("age", 0).count().profile().next();
+        assertFalse(profile.getMetrics(0).getName().contains(JanusGraphMixedIndexCountStep.class.getSimpleName()));
+        assertTrue(profile.getMetrics(0).getName().contains(JanusGraphStep.class.getSimpleName()));
+        assertEquals(5, graph.traversal().V().has("age", 0).count().next());
+        profile = graph.traversal().V().has("age").count().profile().next();
+        assertFalse(profile.getMetrics(0).getName().contains(JanusGraphMixedIndexCountStep.class.getSimpleName()));
+        assertTrue(profile.getMetrics(0).getName().contains(JanusGraphStep.class.getSimpleName()));
+        assertEquals(10, graph.traversal().V().has("age").count().next());
+
```
porunov, line 1725: I was thinking about more test cases. Can we add 2 more test cases?
`has("nonIndexedProperty").has("indexedProperty").count()`
`has("indexedProperty").has("nonIndexedProperty").count()`
What do you think about it? Maybe we already checking these cases in other tests. If so, just ignore this comment 
the above 2 queries are computed the same but they should not be optimized with Mixed Query Count optimization because the whole chain of 'has' steps cannot be answered by a single mixed index.

porunov (approved):

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1124,6 +1126,7 @@ public void testGraphCentricQueryProfiling() {
         if (indexFeatures.supportNotQueryNormalForm()) {
             newTx();
             assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).count().next());
+            assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).toList().size());
```
rngcntr, line 1129: What is the reason to change these queries from `count().next()` to `toList().size()`? I see that the result should be equivalent, but therefore I don't see the necessity to change the query.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexCountStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @param originalGraphStep
+     * @param mixedIndexCountStep
+     * @param traversal
+     */
+    private void applyMixedIndexCountStep(final GraphStep originalGraphStep, final JanusGraphMixedIndexCountStep mixedIndexCountStep,
+                                          final Traversal.Admin<?, ?> traversal) {
+        Step<?, ?> currentStep = originalGraphStep.getNextStep();
+        while (true) {
+            if (currentStep instanceof HasContainerHolder || currentStep instanceof CountGlobalStep) {
```
rngcntr, line 104: I stumbled upon this logic and was wondering if it is correct. For example, a query with `count().count()` (which by the way is completely bonkers) should always return 1, but would be reduced to one `JanusGraphMixedIndexCountStep` here.

rngcntr (commented):
Hi @li-boxuan! I reviewed your changes again and I think I would feel better releasing this if there were also tests combining mixed and composite indexes. Other than that, there are two questions that came to my mind:

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1523,6 +1529,213 @@ public void testIndexParameters() {
...
+        assertTrue(profile.getMetrics(0).getName().contains(JanusGraphStep.class.getSimpleName()));
+        assertEquals(10, graph.traversal().V().has("age").count().next());
+
+        /* =============================================================================
+        count() cannot be optimized if a single mixed index cannot meet all conditions */
+
+        assertEquals(10, graph.traversal().V().has("name").has("age").count().next());
+        profile = graph.traversal().V().has("name").has("age").count().profile().next();
+        assertFalse(profile.getMetrics(0).getName().contains(JanusGraphMixedIndexCountStep.class.getSimpleName()));
+        assertEquals(5, graph.traversal().V().has("name").has("age", 0).count().next());
```
li-boxuan, line 1732: @rngcntr This covers the case "tests combining mixed and composite indexes". "age" is indexed by a composite index while "name" is indexed by a mixed index.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
li-boxuan: Now that count() queries are optimized by `JanusGraphMixedIndexCountStrategy`, I just add this extra check to ensure correctness when JanusGraphMixedIndexCountStrategy does not apply.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexCountStrategy.java
li-boxuan: That is a great point! I don't even know `count().count()` is legal. I'll check if the result is still correct in this case.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
rngcntr: Oh, I overlooked that one. Well in this case, just ignore my comment ;)

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexCountStrategy.java
rngcntr: Should be legal because `count()` simply produces a single traverser with a `count` value. This traverser itself can also be counted even if that makes no sense.
In addition, I am not entirely sure if there can exist other `HasContainerHolder` steps between the `JanusGraphStep` and the `CountGlobalStep`. For example, `JanusGraphStep` itself is a `HasContainerHolder` and so may be other steps. Let me take something like this as an example: `g.V().V().count()`. Since both `V()` are represented by `JanusGraphStep`, both satisfy the condition for this strategy.

li-boxuan: I rewrote this method and added a few test cases. Let me know if you come up with other corner cases.

porunov:
@li-boxuan there is a checkstyle error due to unused import
```
Error:  /home/runner/work/janusgraph/janusgraph/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexCountStrategy.java:20:8: Unused import - org.apache.tinkerpop.gremlin.process.traversal.step.HasContainerHolder. [UnusedImports]
```

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1523,6 +1529,223 @@ public void testIndexParameters() {
...
+        profile = graph.traversal().V().has("name").has("age", 0).count().profile().next();
+        assertFalse(profile.getMetrics(0).getName().contains(JanusGraphMixedIndexCountStep.class.getSimpleName()));
+        profile = graph.traversal().V().has("age").has("name").count().profile().next();
+        assertFalse(profile.getMetrics(0).getName().contains(JanusGraphMixedIndexCountStep.class.getSimpleName()));
+
+        /* ==========================
+        verify other special cases */
+
+        assertEquals(1, graph.traversal().V().has("name").count().count().next());
+        assertEquals(numV, graph.traversal().V().V().has("text").count().next());
```
rngcntr, line 1742: I don't think this would be the expected behavior for chained `V()` calls. Instead, `V()` should spawn a new traverser at every single vertex for each incoming traverser. Thus, for a graph with `100` vertices, `g.V().V().count()` would return `10.000`. With this test however, you seem to verify that the result is `100`.

I suggest to test the correctness by comparing the result of `g.[...]` to `g.withoutStrategies(JanusGraphMixedIndexCountStrategy.class).[...]`. That would make sure that under no circumstances the optimization affects the output result. In my opinion, this should even be a considered standard procedure when introducing new optimization strategies. (Not blaming you, I skipped this kind of verification for some of my optimization strategies, too)

li-boxuan: Great catch! I'll fix this.

li-boxuan: Now if JanusGraphStep is not the start step, JanusGraphMixedIndexCountStrategy won't be applied.

rngcntr: That looks much better, thank you @li-boxuan

rngcntr (approved):

PR #2202, opened 2020-09-08
Reorganize optimization strategies

rngcntr:
As more and more optimization strategies are created, the wish for better organization ­— especially in the testing area — emerged. This PR restructures the current state by applying the following changes:

1. Split optimization strategies, custom steps and optimization utilities into separate packages. These are:
  - `org.janusgraph.graphdb.tinkerpop.optimize.strategy`
  - `org.janusgraph.graphdb.tinkerpop.optimize.step`
  - `org.janusgraph.graphdb.tinkerpop.optimize`
2. Split up the tests. Before, most of the tests were bundled in a single Method within `JanusGraphTest` and were therefore executed for every backend. Since most of the strategies are completely agnostic of the backend, I moved them from `janusgraph-backend-testutils` to `janusgraph-test`. Furthermore, every strategy now has it's own test class.
3. One test which had to remain in `JanusGraphTest` because of it's backend dependency has been renamed accordingly.

-----

### For all changes:

### For code changes:

### Note:


rngcntr:
Updated this PR with the changes from #2209

janusgraph-backend-testutils/src/main/java/org/janusgraph/testutil/JanusGraphAssert.java
```diff
@@ -81,4 +91,57 @@ private static boolean isEmpty(Object obj) {
...
+        final TraversalStrategies optimizations = new DefaultTraversalStrategies();
+        for (final TraversalStrategy<?> strategy : optimizationStrategies) {
+            optimizations.addStrategies(strategy);
+        }
+
+        originalTraversal.asAdmin().setStrategies(optimizations);
+        originalTraversal.asAdmin().applyStrategies();
+
+        assertEquals(expectedTraversal.asAdmin().getSteps().toString(),
+            originalTraversal.asAdmin().getSteps().toString());
```
li-boxuan, line 128: Is there any way to compare steps other than toString()?

janusgraph-backend-testutils/src/main/java/org/janusgraph/testutil/JanusGraphAssert.java
```diff
@@ -81,4 +91,57 @@ private static boolean isEmpty(Object obj) {
...
     }
 
+    public static void assertNumStep(int expectedResults, int expectedSteps, GraphTraversal traversal, Class<? extends Step>... expectedStepTypes) {
+        int num = 0;
+        while (traversal.hasNext()) {
+            traversal.next();
+            num++;
+        }
+
+        assertEquals(expectedResults, num);
```
li-boxuan, line 101: Do we have to iterate and count? What about this:
```suggestion
        assertEquals(expectedResults, traversal.toList().size());
```

janusgraph-backend-testutils/src/main/java/org/janusgraph/testutil/JanusGraphAssert.java
```diff
@@ -81,4 +91,57 @@ private static boolean isEmpty(Object obj) {
         throw new IllegalArgumentException("Cannot determine size of: " + obj);
     }
 
+    public static void assertNumStep(int expectedResults, int expectedSteps, GraphTraversal traversal, Class<? extends Step>... expectedStepTypes) {
```
li-boxuan, line 94: nitpick: Can you add Javadoc for this method? The method name itself is not clear enough to understand its usage.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -4058,24 +4037,7 @@ public void testTinkerPopCardinality() {
     }
 
     @Test
-    public void testTinkerPopOptimizationStrategies() {
-        PropertyKey id = mgmt.makePropertyKey("id").cardinality(Cardinality.SINGLE).dataType(Integer.class).make();
-        PropertyKey weight = mgmt.makePropertyKey("weight").cardinality(Cardinality.SINGLE).dataType(Integer.class).make();
-
-        mgmt.buildIndex("byId", Vertex.class).addKey(id).buildCompositeIndex();
-        mgmt.buildIndex("byWeight", Vertex.class).addKey(weight).buildCompositeIndex();
-        mgmt.buildIndex("byIdWeight", Vertex.class).addKey(id).addKey(weight).buildCompositeIndex();
```
li-boxuan, line 4067: This test involves composite indexes, which is worth testing against different backends IMO.

li-boxuan (commented):
Thanks for the code clean-up! Since this PR almost has no functional change, I am happy with it. However, I have some concerns about the reduction of backend storage involved in testing (see my comment below).

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
rngcntr: I never thought of it that way. My idea of it is more like: If the backends are well tested, their functioning can be assumed equivalent. Therefore it's the scope of the index tests to ensure that the backend fulfills the specification. Based on that the optimization tests can then assume a functionally correct index.

janusgraph-backend-testutils/src/main/java/org/janusgraph/testutil/JanusGraphAssert.java
rngcntr: I didn't find one. If you (or anyone) knows how to do it better, please let me know :)

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
rngcntr: Are you ok with keeping this as it is? And if so, would you please review the PR again? Thanks!

li-boxuan: That's a fair assumption. I am okay with it but I'd like to leave this conversation open in case anyone has any other opinion.

janusgraph-backend-testutils/src/main/java/org/janusgraph/testutil/JanusGraphAssert.java
li-boxuan: What about simply doing `expectedTraversal.asAdmin().getSteps().equals(originalTraversal.asAdmin().getSteps())`?

rngcntr: IIRC there was a problem that equality on steps does not work as expected because of some == / equals discrepancies. I can look at that again and then reply back

rngcntr:
Merged with lazy consensus

Issue #2033, opened 2020-03-08
Remove Guava dependency


porunov:
Guava is often binary  incompatible with previous versions that is why it produces some troubles when some dependencies depend on one guava version and some dependencies on another guava version.

The issue can be split to the next subtasks:

**Direct guava usage removal**:
- Move all usage of guava to Java 8 Streams API or direct Java code.

**Dependencies update**:
- Resolve #1510 (Cassandra driver `4.x` doesn't depend on guava)
- Resolve #1737 (Cassandra `2.x` depends on guava)

**Fixing dependencies (should be fixed not on our side but on libraries which we depend on)**:
- remove guava from `org.apache.tinkerpop:hadoop-gremlin 2.x`
- remove guava from `org.apache.hadoop:hadoop-common 2.x`
- remove guava from `org.apache.hadoop:hadoop-hdfs 2.x`
- remove guava from `org.apache.solr:solr-test-framework 7.x`
- remove guava from `airlift:airline`

PR #2218, opened 2020-09-20
Make iterators auto-closeable and fix SubqueryIterator close bug

### LINKED ISSUE #2217: SubqueryIterator does not close properly
When SubqueryIterator finishes its job, it does two things:

1) stop the timer of the profiler
2) put results into the transaction-wise cache

as shown in the below snippet:

https://github.com/JanusGraph/janusgraph/blob/6c99379352799c6bd3aff80fd23f9e3561ced2ff/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java#L74-L81

However, this does not work if the enclosed iterator never reaches its end (`!elementIterator.hasNext()`). Reasons could be: user quits the iteration, or more likely when `ResultsSetIterator` collects enough results and just stops calling hasNext() of LimitAdjustingIterator, which in turn calls hasNext() of SubqueryIterator.

- Version: 0.5.2, 0.6.0-SNAPSHOT
- Storage Backend: doesn't matter
- Mixed Index Backend: doesn't matter
- Steps to Reproduce: 

Using 0.5.2:
```
gremlin> graph = JanusGraphFactory.open("inmemory")
==>standardjanusgraph[inmemory:[127.0.0.1]]
gremlin>
gremlin> mgmt = graph.openManagement()
==>org.janusgraph.graphdb.database.management.ManagementSystem@60990e5c
gremlin> nameKey = mgmt.makePropertyKey('name').dataType(String.class).make()
==>name
gremlin> mgmt.buildIndex("name", Vertex.class).addKey(nameKey).buildCompositeIndex()
==>name
gremlin> mgmt.commit()
==>null
gremlin>
gremlin> g = graph.traversal()
==>graphtraversalsource[standardjanusgraph[inmemory:[127.0.0.1]], standard]
gremlin> for (i = 0; i < 100; i++) {
......1> g.addV().property("name", "value").iterate();
......2> }
==>null
gremlin> graph.tx().commit()
==>null
gremlin>
gremlin> g.V().has("name", "value").limit(10).profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                   10          10           7.233   100.00
    \_condition=(name = value)
    \_orders=[]
    \_limit=10
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@20
    \_index=name
  optimization                                                                                 0.040
  optimization                                                                                 0.659
  backend-query                                                                                0.000
    \_query=name:multiKSQ[1]@20
    \_limit=20
                                            >TOTAL                     -           -           7.233        -
gremlin> g.V().has("name", "value").limit(10).profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                   10          10           0.541   100.00
    \_condition=(name = value)
    \_orders=[]
    \_limit=10
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@20
    \_index=name
  optimization                                                                                 0.013
  optimization                                                                                 0.109
  backend-query                                                                                0.000
    \_query=name:multiKSQ[1]@20
    \_limit=20
                                            >TOTAL                     -           -           0.541        -
```

As shown in the previous example, there are two problems:
1) There is no timing for `backend-query`. The `Time` column simply shows zero. This is because the timer is not properly stopped.
2) The second query does not hit the cache. This is because the first query did not manage to put results into the cache.

0.6.0-SNAPSHOT (6c9937935) has the same problem:

```
gremlin> g.V().has("name", "value").limit(10).profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                   10          10           7.471   100.00
    \_condition=(name = value)
    \_orders=[]
    \_limit=10
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@10
    \_index=name
  optimization                                                                                 0.029
  optimization                                                                                 0.709
  backend-query                                                                                0.000
    \_query=name:multiKSQ[1]@10
    \_limit=10
                                            >TOTAL                     -           -           7.471        -
gremlin> g.V().has("name", "value").limit(10).profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                   10          10           0.536   100.00
    \_condition=(name = value)
    \_orders=[]
    \_limit=10
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@10
    \_index=name
  optimization                                                                                 0.016
  optimization                                                                                 0.116
  backend-query                                                                                0.000
    \_query=name:multiKSQ[1]@10
    \_limit=10
                                            >TOTAL                     -           -           0.536        -
```

These problems, however, don't appear when you don't have a `limit` step (which means ResultsSetIterator does not stop iterating early).

```
gremlin> g.V().has("name", "value").profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                  100         100           5.930   100.00
    \_condition=(name = value)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@2000
    \_index=name
  optimization                                                                                 0.017
  optimization                                                                                 0.164
  backend-query                                                      100                       6.052
    \_query=name:multiKSQ[1]@2000
    \_limit=2000
                                            >TOTAL                     -           -           5.930        -
gremlin> g.V().has("name", "value").profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                  100         100           1.907   100.00
    \_condition=(name = value)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@2000
    \_index=name
  optimization                                                                                 0.013
  optimization                                                                                 0.154
                                            >TOTAL                     -           -           1.907        -
```

li-boxuan:
Changes:
1. Make a few iterators auto-closeable. Now traversal.close() can be delegated to `SubqueryIterator`.
2. Extract some logic from MultiDistinctOrderedIterator and use new MultiDistinctUnorderedIterator when orders are not given. This makes the code cleaner and reduces the overhead of creating & maintaining complex data structures.
3. Make a few iterators extend CloseableAbstractIterator which in turn leverages Guava Iterator's `computeNext` method. Now, these iterators don't need to explicitly write `hasNext` and `next`.
4. Fix SubqueryIterator bug so that profiler is correctly stopped and results can be cached appropriately.

This PR does not:
1. Make all iterators auto-closeable. It only refactors iterators on the traversal -> SubqueryIterator chain so that traversal.close() can be delegated.
2. Refactor all iterators using `computeNext` method. Some iterators have complex internal logic thus I would like to leave them as they are in this PR. Generally, all iterators using `nextInternal` method could be rewritten using `computeNext` which is much cleaner and avoids eager-computing (fetches the first result in the constructor). 

Fixes #2217

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
@@ -17,6 +17,7 @@ package org.janusgraph.graphdb.query;
 import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.janusgraph.core.JanusGraphElement;
 
+import java.util.Iterator;
 import java.util.NoSuchElementException;
 
 /**
@@ -27,7 +28,7 @@ import java.util.NoSuchElementException;
  */
 public class ResultSetIterator<R extends JanusGraphElement> implements CloseableIterator<R> {
 
-    private final CloseableIterator<R> iterator;
+    private final Iterator<R> iterator;
     private final int limit;
 
     private R current;
@@ -35,7 +36,7 @@ public class ResultSetIterator<R extends JanusGraphElement> implements Closeable
     private int count;
 
 
-    public ResultSetIterator(CloseableIterator<R> inner, int limit) {
+    public ResultSetIterator(Iterator<R> inner, int limit) {
         this.iterator = inner;
         this.limit = limit;
         count = 0;
@@ -80,11 +81,11 @@ public class ResultSetIterator<R extends JanusGraphElement> implements Closeable
 
     @Override
     public void close() {
-        iterator.close();
+        CloseableIterator.closeIterator(iterator);
     }
 
     public static<R extends JanusGraphElement> Iterable<R> wrap(final Iterable<R> inner, final int limit) {
-        return () -> new ResultSetIterator<>(CloseableIterator.asCloseable(inner.iterator()), limit);
+        return () -> new ResultSetIterator<>(inner.iterator(), limit);
     }
 
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
@@ -14,9 +14,9 @@
 
 package org.janusgraph.graphdb.query;
 
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.janusgraph.core.JanusGraphElement;
 
-import java.util.Iterator;
 import java.util.NoSuchElementException;
 
 /**
@@ -25,9 +25,9 @@ import java.util.NoSuchElementException;
  *
  * @author Matthias Broecheler (me@matthiasb.com)
  */
-public class ResultSetIterator<R extends JanusGraphElement> implements Iterator<R> {
+public class ResultSetIterator<R extends JanusGraphElement> implements CloseableIterator<R> {
 
-    private final Iterator<R> iterator;
+    private final CloseableIterator<R> iterator;
     private final int limit;
 
     private R current;
@@ -35,7 +35,7 @@ public class ResultSetIterator<R extends JanusGraphElement> implements Iterator<
     private int count;
 
 
-    public ResultSetIterator(Iterator<R> inner, int limit) {
+    public ResultSetIterator(CloseableIterator<R> inner, int limit) {
         this.iterator = inner;
         this.limit = limit;
         count = 0;
@@ -53,6 +53,8 @@ public class ResultSetIterator<R extends JanusGraphElement> implements Iterator<
         R r = null;
         if (count < limit && iterator.hasNext()) {
             r = iterator.next();
+        } else {
+            close();
         }
         return r;
     }
@@ -76,8 +78,13 @@ public class ResultSetIterator<R extends JanusGraphElement> implements Iterator<
             throw new UnsupportedOperationException();
     }
 
+    @Override
+    public void close() {
+        iterator.close();
+    }
+
     public static<R extends JanusGraphElement> Iterable<R> wrap(final Iterable<R> inner, final int limit) {
-        return () -> new ResultSetIterator<>(inner.iterator(),limit);
+        return () -> new ResultSetIterator<>(CloseableIterator.asCloseable(inner.iterator()), limit);
     }
 
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
@@ -1,0 +1,62 @@
+// Copyright 2020 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.util;
+
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+public class MultiDistinctUnorderedIterator<E> extends CloseableAbstractIterator<E> {
+
+    private final Set<E> allElements = new HashSet<E>();
+    private final CloseableIterator<E> iterator;
+    private final int limit;
+    private long count;
+
+    public MultiDistinctUnorderedIterator(final int lowLimit, final int highLimit, final List<CloseableIterator<E>> iterators) {
+        Objects.requireNonNull(iterators);
+        iterator = IteratorUtils.concat(iterators);
+        limit = highLimit;
+
+        long i = 0;
+        while (i < lowLimit && hasNext()) {
+            next();
+            i++;
+        }
+    }
+
+    @Override
+    protected E computeNext() {
+        if (count < limit) {
+            while (iterator.hasNext()) {
+                E elem = iterator.next();
+                if (allElements.add(elem)) {
+                    count++;
+                    return elem;
+                }
+            }
+        }
+        close();
+        return endOfData();
+    }
+
+    @Override
+    public void close() {
+        iterator.close();
+    }
+}
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
@@ -17,6 +17,7 @@ package org.janusgraph.graphdb.util;
 import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -28,9 +29,9 @@ public class MultiDistinctUnorderedIterator<E> extends CloseableAbstractIterator
     private final int limit;
     private long count;
 
-    public MultiDistinctUnorderedIterator(final int lowLimit, final int highLimit, final List<CloseableIterator<E>> iterators) {
+    public MultiDistinctUnorderedIterator(final int lowLimit, final int highLimit, final List<Iterator<E>> iterators) {
         Objects.requireNonNull(iterators);
-        iterator = IteratorUtils.concat(iterators);
+        iterator = CloseableIteratorUtils.concat(iterators);
         limit = highLimit;
 
         long i = 0;
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
@@ -22,10 +22,10 @@ import java.util.Objects;
 
 public class MultiIterator<E> extends CloseableAbstractIterator<E> {
 
-    private Iterator<CloseableIterator<E>> listIterator;
-    private CloseableIterator<E> currentIterator;
+    private Iterator<Iterator<E>> listIterator;
+    private Iterator<E> currentIterator;
 
-    public MultiIterator(final List<CloseableIterator<E>> iterators) {
+    public MultiIterator(final List<Iterator<E>> iterators) {
         Objects.requireNonNull(iterators);
         listIterator = iterators.iterator();
     }
@@ -38,7 +38,7 @@ public class MultiIterator<E> extends CloseableAbstractIterator<E> {
             } else if (currentIterator.hasNext()) {
                 return currentIterator.next();
             } else {
-                currentIterator.close();
+                CloseableIterator.closeIterator(currentIterator);
                 currentIterator = null;
             }
         }
@@ -51,7 +51,7 @@ public class MultiIterator<E> extends CloseableAbstractIterator<E> {
             if (currentIterator == null) {
                 currentIterator = listIterator.next();
             } else {
-                currentIterator.close();
+                CloseableIterator.closeIterator(currentIterator);
                 currentIterator = null;
             }
         }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
@@ -1,0 +1,59 @@
+// Copyright 2020 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.util;
+
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Objects;
+
+public class MultiIterator<E> extends CloseableAbstractIterator<E> {
+
+    private Iterator<CloseableIterator<E>> listIterator;
+    private CloseableIterator<E> currentIterator;
+
+    public MultiIterator(final List<CloseableIterator<E>> iterators) {
+        Objects.requireNonNull(iterators);
+        listIterator = iterators.iterator();
+    }
+
+    @Override
+    protected E computeNext() {
+        while (currentIterator != null || listIterator.hasNext()) {
+            if (currentIterator == null) {
+                currentIterator = listIterator.next();
+            } else if (currentIterator.hasNext()) {
+                return currentIterator.next();
+            } else {
+                currentIterator.close();
+                currentIterator = null;
+            }
+        }
+        return endOfData();
+    }
+
+    @Override
+    public void close() {
+        while (currentIterator != null || listIterator.hasNext()) {
+            if (currentIterator == null) {
+                currentIterator = listIterator.next();
+            } else {
+                currentIterator.close();
+                currentIterator = null;
+            }
+        }
+    }
+}
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
@@ -35,7 +35,7 @@ import com.google.common.cache.Cache;
 /**
  * @author davidclement90@laposte.net
  */
-public class SubqueryIterator implements Iterator<JanusGraphElement>, AutoCloseable {
+public class SubqueryIterator extends CloseableAbstractIterator<JanusGraphElement> {
 
     private final JointIndexQuery.Subquery subQuery;
 
@@ -72,25 +72,28 @@ public class SubqueryIterator implements Iterator<JanusGraphElement>, AutoClosea
     }
 
     @Override
-    public boolean hasNext() {
-        if (!elementIterator.hasNext() && currentIds != null) {
-            indexCache.put(subQuery, currentIds);
+    protected JanusGraphElement computeNext() {
+        if (elementIterator.hasNext()) {
+            return elementIterator.next();
+        }
+        close();
+        return endOfData();
+    }
+
+    /**
+     * Close the iterator, stop timer and update profiler.
+     * Put results into cache if the underlying elementIterator is exhausted.
+     */
+    @Override
+    public void close() {
+        if (isTimerRunning) {
+            assert currentIds != null;
+            if (!elementIterator.hasNext()) {
+                indexCache.put(subQuery, currentIds);
+            }
+            profiler.setResultSize(currentIds.size());
             profiler.stopTimer();
             isTimerRunning = false;
-            profiler.setResultSize(currentIds.size());
-        }
-        return elementIterator.hasNext();
-    }
-
-    @Override
-    public JanusGraphElement next() {
-        return this.elementIterator.next();
-    }
-
-    @Override
-    public void close() throws Exception {
-        if (isTimerRunning) {
-            profiler.stopTimer();
         }
     }
 
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
@@ -15,6 +15,7 @@
 package org.janusgraph.graphdb.query;
 
 import com.google.common.base.Preconditions;
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 
 import java.util.Iterator;
 import java.util.NoSuchElementException;
@@ -32,13 +33,13 @@ import java.util.NoSuchElementException;
  *
  * @author Matthias Broecheler (me@matthiasb.com)
  */
-public abstract class LimitAdjustingIterator<R> implements Iterator<R> {
+public abstract class LimitAdjustingIterator<R> implements CloseableIterator<R> {
 
     private final int maxLimit;
     private int currentLimit;
     private int count;
 
-    private Iterator<R> iterator;
+    private CloseableIterator<R> iterator;
 
     /**
      * Initializes this iterator with the current limit and the maximum number of elements that may be retrieved from the
@@ -61,7 +62,7 @@ public abstract class LimitAdjustingIterator<R> implements Iterator<R> {
      * @param newLimit
      * @return
      */
-    public abstract Iterator<R> getNewIterator(int newLimit);
+    public abstract CloseableIterator<R> getNewIterator(int newLimit);
 
     @Override
     public boolean hasNext() {
@@ -97,5 +98,12 @@ public abstract class LimitAdjustingIterator<R> implements Iterator<R> {
     @Override
     public void remove() {
         throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void close() {
+        if (iterator != null) {
+            iterator.close();
+        }
     }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
@@ -39,7 +39,7 @@ public abstract class LimitAdjustingIterator<R> implements CloseableIterator<R> 
     private int currentLimit;
     private int count;
 
-    private CloseableIterator<R> iterator;
+    private Iterator<R> iterator;
 
     /**
      * Initializes this iterator with the current limit and the maximum number of elements that may be retrieved from the
@@ -62,7 +62,7 @@ public abstract class LimitAdjustingIterator<R> implements CloseableIterator<R> 
      * @param newLimit
      * @return
      */
-    public abstract CloseableIterator<R> getNewIterator(int newLimit);
+    public abstract Iterator<R> getNewIterator(int newLimit);
 
     @Override
     public boolean hasNext() {
@@ -101,9 +101,7 @@ public abstract class LimitAdjustingIterator<R> implements CloseableIterator<R> 
     }
 
     @Override
-    public void close() {
-        if (iterator != null) {
-            iterator.close();
-        }
+    public void close(){
+        CloseableIterator.closeIterator(iterator);
     }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/QueryProcessor.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/QueryProcessor.java
@@ -19,9 +19,11 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.janusgraph.core.QueryException;
 import org.janusgraph.core.JanusGraphElement;
 import org.janusgraph.graphdb.query.profile.QueryProfiler;
+import org.janusgraph.graphdb.util.IteratorUtils;
 
 import java.util.*;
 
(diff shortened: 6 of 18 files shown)


li-boxuan:
Testing result (in comparison with https://github.com/JanusGraph/janusgraph/issues/2217#issue-705069352):

```
gremlin> graph = JanusGraphFactory.open("inmemory")
==>standardjanusgraph[inmemory:[127.0.0.1]]
gremlin>
gremlin> mgmt = graph.openManagement()
==>org.janusgraph.graphdb.database.management.ManagementSystem@2c7db926
gremlin> nameKey = mgmt.makePropertyKey('name').dataType(String.class).make()
==>name
gremlin> mgmt.buildIndex("name", Vertex.class).addKey(nameKey).buildCompositeIndex()
==>name
gremlin> mgmt.commit()
==>null
gremlin>
gremlin> g = graph.traversal()
==>graphtraversalsource[standardjanusgraph[inmemory:[127.0.0.1]], standard]
gremlin> for (i = 0; i < 100; i++) {
......1> g.addV().property("name", "value").iterate();
......2> }
==>null
gremlin> graph.tx().commit()
==>null
gremlin>
gremlin> g.V().has("name", "value").limit(10).profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                   10          10           7.810   100.00
    \_condition=(name = value)
    \_orders=[]
    \_limit=10
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@10
    \_index=name
  optimization                                                                                 0.030
  optimization                                                                                 0.887
  backend-query                                                       10                       2.667
    \_query=name:multiKSQ[1]@10
    \_limit=10
                                            >TOTAL                     -           -           7.810        -
gremlin> g.V().has("name", "value").limit(10).profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                   10          10           0.569   100.00
    \_condition=(name = value)
    \_orders=[]
    \_limit=10
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@10
    \_index=name
  optimization                                                                                 0.015
  optimization                                                                                 0.123
                                            >TOTAL                     -           -           0.569        -
gremlin> g.V().has("name", "value").profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                  100         100           4.537   100.00
    \_condition=(name = value)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@1000
    \_index=name
  optimization                                                                                 0.016
  optimization                                                                                 0.121
  backend-query                                                      100                       4.607
    \_query=name:multiKSQ[1]@1000
    \_limit=1000
                                            >TOTAL                     -           -           4.537        -
gremlin> g.V().has("name", "value").profile()
==>Traversal Metrics
JanusGraphStep([],[name.eq(value)])                                  100         100           1.753   100.00
    \_condition=(name = value)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@1000
    \_index=name
  optimization                                                                                 0.016
  optimization                                                                                 0.111
                                            >TOTAL                     -           -           1.753        -
```

farodin91:
Could you retarget this for JanusGraph 0.5?

li-boxuan:
> Could you retarget this for JanusGraph 0.5?

Done. Just curious, what is the benefit of targeting at 0.5 and merging 0.5 to master later, in comparison with targeting at master and cherry-picking it to 0.5 later?

farodin91:
In the past, we basically just used merging instead of cherry picking.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -73,11 +74,8 @@ public SubqueryIterator(JointIndexQuery.Subquery subQuery, IndexSerializer index
 
     @Override
     public boolean hasNext() {
-        if (!elementIterator.hasNext() && currentIds != null) {
```
rngcntr, line 76: What happened to the `currentIds != null` check here? Before your change, the `SubQueryIterator` was also stopped if `currentIds` was `null`, even if the `elementIterator` had a next element.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -88,9 +86,13 @@ public JanusGraphElement next() {
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() {
         if (isTimerRunning) {
+            assert currentIds != null;
+            indexCache.put(subQuery, currentIds);
```
rngcntr, line 92: Not sure: The encapsulation of all the code within an `if (isTimerRunning)` suggests that calling close() with an already stopped timer is also a valid operation. Does it make sense to restrict the cache update to the case where the timer is still running?
From my understanding, there should be no real difference since after the timer has stopped, `currentIds` should not be altered anymore. But still I would be more comfortable if the if block only encapsulates operations that are invalid if the timer is stopped.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
```diff
@@ -36,7 +37,7 @@
 
 
     public ResultSetIterator(Iterator<R> inner, int limit) {
```
rngcntr, line 39: I don't have a strong opinion on this but if the `ResultSetIterator` itself is a `CloseableIterator`, would it make sense to enforce it's inner iterator to be closeable, too?

li-boxuan: I actually indeed tried that at first, but I ended up changing a lot more classes, and then I gave up and kept it like this. Not because that is undoable, but because I want to keep initial changes simple. If people think that is the better way to go, I'll do that and amend this commit.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
rngcntr: You moved the lines inside this `if` statement to a separate function named `close` and now call this function instead of executing the lines directly. But the condition inside the `if` has changed. It was
```java
if (!elementIterator.hasNext() && currentIds != null) {
   ...
}
```
before and now it is
```java
if (!elementIterator.hasNext()) {
   ...
}
```

You removed the ``&& currentIds != null` part.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
rngcntr: Understood. As I said, I'm also happy with your solution but if someone has a different oppinion, please tell :smile:

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
li-boxuan: Got your question. What confused me was the sentence:

> Before your change, the SubQueryIterator was also stopped if currentIds was null, even if the elementIterator had a next element.

Anyway, I removed this `currentIds != null` condition because it is not needed anymore. In the old logic, it is needed, because `currentIds == null` means cached response is used, in which case there is no timer running & no results fetched from backend. 

After my changes, I moved the logic into the `close()` method. In that method, nothing would be done if `isTimerRunning` evaluates to false (note that `isTimerRunning == true` and `currentIds != null` are equivalent), so the effect remains the same.

rngcntr (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -88,9 +86,13 @@ public JanusGraphElement next() {
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() {
```
porunov, line 89: I didn't spend much time checking the logic behind this class but from the first glance both solutions look error prone to me.
The first solution binds the logic to `hasNext` to add new elements to the cache (no one said that `hasNext` should always be executed before `next` or `close` methods, thus the solution is error prone).
A new solution binds logic to `close` method which might be a little bit better but still looks error prone to me (if we don't call `close` method at all). 
I think it would be more correct to bind that logic to `next` method because only `next` method is responsible for retrieving new elements. Without calling `next` we can't collect necessary elements for our cache.
I didn't had much time to think about another solution but what if we use a separate method, something like:
```
private void cacheQueryIfNoMoreElementsAvailable(){
    if (!elementIterator.hasNext() && currentIds != null) {
        indexCache.put(subQuery, currentIds);
        profiler.setResultSize(currentIds.size());
        profiler.stopTimer();
        isTimerRunning = false;
    }
}
```
Than our next method could look like:
```
public JanusGraphElement next() {
    JanusGraphElement element = this.elementIterator.next();
    cacheQueryIfNoMoreElementsAvailable();
    return element;
}
```
In addition we can execute this method in the end of `SubqueryIterator` constructor just for cases when an index contains 0 elements.
That said, I didn't check the above soltuion and didn't think too much about it, thus I may missed something. Nevertheless, I would like to hear some opinions regarding it.

li-boxuan: Nice idea. I think stopping the timer in the close() method still would be useful in the try-with-resources statement, or when the user quits the iteration (although I cannot think of such a case where profiling is used). I think I might let next() method call hasNext() method, which in turn calls close() method when results are exhausted.

It also comes into my mind that the current solution has a bug: when the iteration is closed before iterating is over, we shouldn't put results into the cache, since the results are probably incomplete.

porunov: Agree, so I assume the resulting code could look something like:
```
public class SubqueryIterator implements CloseableIterator<JanusGraphElement> {

    private final JointIndexQuery.Subquery subQuery;

    private final Cache<JointIndexQuery.Subquery, List<Object>> indexCache;

    private final Iterator<? extends JanusGraphElement> elementIterator;

    private List<Object> currentIds;

    private QueryProfiler profiler;

    private boolean isTimerRunning;

    public SubqueryIterator(JointIndexQuery.Subquery subQuery, IndexSerializer indexSerializer, BackendTransaction tx,
            Cache<JointIndexQuery.Subquery, List<Object>> indexCache, int limit,
            Function<Object, ? extends JanusGraphElement> function, List<Object> otherResults) {
        this.subQuery = subQuery;
        this.indexCache = indexCache;
        final List<Object> cacheResponse = indexCache.getIfPresent(subQuery);
        final Stream<?> stream;
        if (cacheResponse != null) {
            stream = cacheResponse.stream();
        } else {
            try {
                currentIds = new ArrayList<>();
                profiler = QueryProfiler.startProfile(subQuery.getProfiler(), subQuery);
                isTimerRunning = true;
                stream = indexSerializer.query(subQuery, tx).peek(r -> currentIds.add(r));
            } catch (final Exception e) {
                throw new JanusGraphException("Could not call index", e.getCause());
            }
        }
        elementIterator = stream.filter(e -> otherResults == null || otherResults.contains(e)).limit(limit).map(function).map(r -> (JanusGraphElement) r).iterator();
        cacheQueryIfNoMoreElementsAvailable();
    }

    @Override
    public boolean hasNext() {
        return elementIterator.hasNext();
    }

    @Override
    public JanusGraphElement next() {
        JanusGraphElement element = this.elementIterator.next();
        cacheQueryIfNoMoreElementsAvailable();
        return element;
    }

    @Override
    public void close() {
        if (isTimerRunning) {
            profiler.stopTimer();
        }
    }

    private void cacheQueryIfNoMoreElementsAvailable(){
        if (!elementIterator.hasNext() && currentIds != null) {
            indexCache.put(subQuery, currentIds);
            profiler.setResultSize(currentIds.size());
            profiler.stopTimer();
            isTimerRunning = false;
        }
    }
}
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
```diff
@@ -73,24 +75,37 @@ public SubqueryIterator(JointIndexQuery.Subquery subQuery, IndexSerializer index
...
-            profiler.setResultSize(currentIds.size());
+        final boolean hasNext = elementIterator.hasNext();
+        if (!hasNext) {
+            if (currentIds != null) {
+                indexCache.put(subQuery, currentIds);
+            }
+            close();
         }
-        return elementIterator.hasNext();
+        return hasNext;
```
porunov, line 85: Don't forget to cache a query when the result is empty and nobody calls `hasNext` or `next` or `close`.

li-boxuan: I am a bit against doing so because generally, I want to avoid running the real (and maybe costly) computation in the constructor itself, especially when nobody calls `hasNext` or `next` afterward. In this class, the iterator comes from an unfiltered iterator which intersects with `otherResults`. The intersection can be expensive. Thus, if nobody uses this iterator afterward, the "pre-computation" will be wasted.

That being said, in this class, we know the underlying stream is not lazy, which means in the constructor, the underlying index query already happens. Probably we want to cache all results in the constructor even if nobody uses iterator, or the iterator is used but not exhausted... However, if we don't consider the internal implementation of `stream = indexSerializer.query(subQuery, tx).peek(r -> currentIds.add(r));`, we should consider it as lazy and avoid eager fetching if no one calls `hasNext` or `next` IMO.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
li-boxuan: I made them auto-closeable as well. To correctly delegate Traversal.close(), I ended up refactoring/modifying many iterator classes.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java
porunov: Great points! Totally makes sense.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
```diff
@@ -0,0 +1,63 @@
...
+            i++;
+        }
+    }
+
+    @Override
+    protected E computeNext() {
+        if (count < limit) {
+            while (iterator.hasNext()) {
+                E elem = iterator.next();
+                if (!allElements.contains(elem)) {
```
rngcntr, line 48: Just an idea: For large sets it can be quite inefficient to perform this check, because hash sets are not cache friendly at all. So before this check, one could query a bloom filter. If the filter says that this element is new, the check can be skipped.
On the downside, we would still need to collect the elements in `allElements` so insertions would still consume as much time as before. Therefore I'm not sure how significant a change like this would be.

This is just a general consideration and has nothing to do with your PR. So feel free to dismiss it :smile:

rngcntr (approved):
LGTM

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public MultiIterator(final List<CloseableIterator<E>> iterators) {
+        Objects.requireNonNull(iterators);
+        this.iterators = iterators;
+    }
+
+    @Override
+    protected E computeNext() {
+        while (current < iterators.size()) {
+            CloseableIterator<E> iterator = iterators.get(current);
```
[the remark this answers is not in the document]
li-boxuan: That would be a bit more complicated because it requires two instance variables: a current iterator and an iterator of iterators. May I know why you think one should avoid calling `get(index)` on `List`?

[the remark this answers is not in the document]
porunov: I guess that's because `List` isn't guaranteed to be `ArrayList` which supports Random Access in `O(1)` time. Thus, someone may for some reason pass `LinkedList` to the constructor of this class. If so, `get` operation will cost `O(n)` time instead of `O(1)` time. Making this loop to be `O(n^2)` time. That said, it depends on `List` implementation which is used here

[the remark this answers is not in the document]
porunov: I think that using a current iterator and an iterator of iterator is a more safer approach here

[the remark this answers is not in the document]
rngcntr: Thanks @porunov, that's exactly what I wanted to point out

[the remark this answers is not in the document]
rngcntr: However, I would prefer not to enforce `ArrayList`. Using an iterator over an arbitrary list should be equally es fast but maintains more flexibility. Especially in case we decide to switch back to `List` for some reason in the future.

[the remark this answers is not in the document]
porunov: @rngcntr Agree with your points. Using iterator will have more sense here.

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
li-boxuan: Sounds interesting. My intuition is that would be even slower, but worth giving it a try. If I got some time I'll do some experiments and probably open up a new issue.

porunov: Also, a small nitpick, why not to add `elem` to the set instead of checking for its existence and then adding it to the set?
I.e. the next code is 1 line shorted and doesn't call `contains` operation:
```
if(allElements.add(elem)){
  count++;
  return elem;
}
```
That said, I didn't check if bloom filter is actually faster for this logic or not. Not sure about that

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Objects;
+
+public class MultiIterator<E> extends CloseableAbstractIterator<E> {
+
+    private Iterator<CloseableIterator<E>> listIterator;
+    private CloseableIterator<E> currentIterator;
+
+    public MultiIterator(final List<CloseableIterator<E>> iterators) {
+        Objects.requireNonNull(iterators);
+        listIterator = iterators.iterator();
```
porunov, line 30: I've seen you used `IteratorUtils.concat(iterators);`. Will it be easier to manage `MultiIterator` with `IteratorUtils.concat(iterators);` or did you create it for some other purposes (like better performance, less overhead or something like that)?

li-boxuan: In fact that is the reason why `MultiIterator` was created at the first place:

```java
public static <T> CloseableIterator<T> concat(final List<CloseableIterator<T>> iterators) {
    return new MultiIterator<T>(iterators);
}
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiDistinctUnorderedIterator.java
rngcntr: @porunov With your idea implemented, I'm now pretty sure a bloom filter would just cause overhead here. Even with one, either `contains` or `add` would be called anyway, so I don't see an improvement there anymore.

rngcntr (approved):

li-boxuan:
@porunov Would it be possible for you to help restart failed Travis tests? I prefer not to amend the commit without any change just to restart Travis which wastes cycles.

porunov:
@li-boxuan restarted. Sorry for that, we have slight permission problems with TravisCI (not all committers have "restart" access which is wrong) but we are trying to move out from it to GitHub Actions. Each committer will have a proper access when we switch CI.

porunov:
I will review this PR on Sunday / Monday

porunov (commented):
From the issue description:
> However, this does not work if the enclosed iterator never reaches its end (!elementIterator.hasNext()). Reasons could be: user quits the iteration, or more likely when ResultsSetIterator collects enough results and just stops calling hasNext() of LimitAdjustingIterator, which in turn calls hasNext() of SubqueryIterator.

Does this PR solves the problem when a user just quits the iteration without reaching the limit?
Do you know how should `SubqueryIterator` behave in such situation? Do we need to stop timer?
I think, maybe we should stop a timer when GC removes the object (i.e. when we are sure the user will not iterate anymore)? I.e.:
```
@Override
protected void finalize(){
  if(isTimerRunning){
    profiler.stopTimer();
  }
}
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/MultiIterator.java
porunov: Yap, I missed that. Now I see the purpose of this class. I'm just thinking if we can somehow reuse already existent general functionality instead of introducing our own implementation and then support it.
I know we have #2033 but I don't see it to be resolved anytime soon. Thus, I think, in case we don't have better alternatives for some functionality, it's okay to use that functionality from Guava.
For example, Guava already has this `concat` functionality in `Iterators`. So, basically you would be able to do something like this:
```
public static <T> CloseableIterator<T> concat(final List<CloseableIterator<T>> iterators) {
    return Iterators.concat(iterators.iterator());
}
```
Instead of supporting this `MultiIterator`, but I see that `MultiIterator` not just concat iterators as `Iterators.concat` does, but also closes iterators which don't have any value anymore.
So, I was thinking, if it is right. I'm not sure yet. Maybe it is a correct implementation but for me it's a little bit confusing.
My confusion is that it's not clear, when exactly we trigger `close` for `CloseableIterator`. Looking in this PR I see that some Iterators automatically trigger `close` after the last element they read but some Iterators don't trigger this logic and expect external logic to trigger `close` for them (like `LimitAdjustingIterator`).
I need to look at this PR a little bit more deeply. I have a small feeling that we can remove our `concat` implementation and if each Iterator which implements `CloseableIterator` would close itself automatically, then we don't need to pass `CloseableIterator` iterator everywhere and we don't need to wrap `Iterator` as `CloseableIterator` when it doesn't have anything to close. That said, I could be completely wrong, I just didn't check the full PR yet.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ResultSetIterator.java
porunov: @rngcntr Could be great if you could describe your thoughts a little bit, why should we force to accept `CloseableIterator` and not just `Iterator`. I'm still trying to review this PR, I guess I may miss some ideas / code style / patterns / etc. begind this logic. Also, I posted my thoughts here: https://github.com/JanusGraph/janusgraph/pull/2218#discussion_r537111253 . I may totally be wrong because I didn't research it and didn't spend much time reviewing this PR but for some reason, I feel that we can achieve better results if we don't enforce `CloseableIterator` to be an input and just force `CloseableIterator` to be be automatically closed when the last element is read.

Going in Javadoc to `java.io.Closeable`, it says:  
> Closes this stream and releases any system resources associated with it. If the stream is already closed then invoking this method has no effect.

We do implement safe `close` method in `SubqueryIterator` and this method is automatically called when we read last element from `elementIterator`, so, I'm not sure why we are forcing `CloseableIterator` to be used everywhere even so it should be used just in a couple of places. 
Could be great to hear some thoughts about it.

li-boxuan:
@porunov I am trying to answer a few doubts of yours together:

## Make Iterators auto-closeable

> Does this PR solves the problem when a user just quits the iteration without reaching the limit

Yes, as long as the user puts the iterator in a try-with-resources clause. For example, a common use case is:

```java
try (GraphTraversal<Vertex, Vertex> query = g.V()) {
    // do the query
}
```

Anything implements `AutoCloseable` will be invoked at the end of the statement (https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html).

See

https://github.com/apache/tinkerpop/blob/eb0813e007e5c3218742e19f955cb3db8026c29b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java#L285-L294

By making all iterators autocloseble (at least all iterators on the traversal -> SubqueryIterator chain), the SubqueryIterator can be closed externally even if there is an exception or the user wants to quit the iteration early (again, provided the user is using try-with-resources statement).

> For example, Guava already has this concat functionality in Iterators

JanusGraph was indeed using Guava's concat functionality. However, Google's Guava library does not have much support for auto-closeable. Not sure about the reason, but my guess is: 1) The try-with-resources clause comes out with jdk7 which was released in 2011, but Iterators support has been there in Guava since [v2.0](https://github.com/google/guava/releases/tag/v2.0), which was released in 2010.  2) it is not a common need for an iterator to release resources.

If users never use the try-with-resources clause, AND they never close the GraphTraversal iterator explicitly, then this part (make iterators auto-closeable) is not needed.

In fact, as of now, the close method does not do many meaningful things. The changes are more of an enhancement to comply with the TinkerPop interface.

## Fix SubqueryIterator Bug

This is related to the last thing we talked about, but can also be an independent PR (in fact, this was what this PR did originally).

There are three scenarios:
1) SubqueryIterator ONLY stops the profiler when all underlying results are fetched (which is common when the `limit` step is not used)
2) Apart from the last scenario, SubqueryIterator can also stop the profiler when all underlying resources within the limit are fetched (no matter if the `limit` step is used or not).
3) Apart from the last two scenarios, SubqueryIterator can also stop the profiler when the user explicitly closes the GraphTraversal iterator, or the user uses traversal in a try-with-resources clause.

Scenario (1) is what we have before this PR. See https://github.com/JanusGraph/janusgraph/issues/2217#issue-705069352
Scenario (2) is what this PR originally aimed for. Changes in SubqueryIterator.java alone (almost) suffice this scenario.
Scenario (3) is what this PR now aims for. Now Traversal::close method can be delegated to JanusGraph internal iterators.

Regarding the finalize method: there are many articles explaining why this should not be used to release resources. The simplest reason is when the finalize method is called is out of control.

porunov:
@li-boxuan Thanks for the description. I didn't know we are now extending the API of iterators which will be used by users. 
That said, I didn't fully get t your thoughts about 'finalize' method. You said that there are many articles which don't suggest to use finalize statements to release resources. Well, that's true, but as far as I know they talk more about relaying on 'finalize' method to release the resources. I.e., your logic should never expect that 'finalize' method will ever be called but it doesn't mean that 'finalize' method is useless and have no any purpose.
'finalize' method might be a good tool to protect your application from resource leaks and / or additional logic.
Let's imagine, user doesn't use try with resources and doesn't explicitly call 'close' and doesn't iterate through all elements. The user just drops usage of the iterator and that's it. So, in such case, the iterator becomes available for GC to be removed. In case GC removes the object which has still opened profile timer, do we need to close it or not?
If not - we don't need to use 'finalize' method. If yes - we need to use 'finalize' method.
I just didn't go deep in this profile timer and don't know if it is OK to leave it opened (will we have any resource leaks or strange behavior if we don't close it ever?)

porunov:
@li-boxuan What if we use TinkerPop's method `CloseableIterator.closeIterator(iterator)` instead of forcing all iterators to be closable? Not a suggestion but just a thought

li-boxuan:
> Let's imagine, user doesn't use try with resources and doesn't explicitly call 'close' and doesn't iterate through all elements. The user just drops usage of the iterator and that's it. So, in such case, the iterator becomes available for GC to be removed. In case GC removes the object which has still an open profile timer, do we need to close it or not?

@porunov Nope. It is totally fine to leave the profiler running. No resource is held by the profiler. It simply does not make sense to stop the timer in finalize(), from the user's perspective. If the timer is not stopped, the profiler returns the wrong result - that's it. Sorry I wasn't clear earlier.

> What if we use TinkerPop's method CloseableIterator.closeIterator(iterator) instead of forcing all iterators to be closable?

I tried this earlier when I wanted to delegate the close() method from TinkerPop to SubqueryIterator. I couldn't make it work because we were using guava utilities like `Iterators.filter`, `Iterators.concat` in many places. Thus this PR.

porunov:
@li-boxuan I pushed 1 commit to your branch. That said, I didn't test it at all and made changes very quickly, so it might actually not work.
My thoughts above were about `asClosable` wrappings. We always wrap all iterators into Closable iterators even when they are not closable. So, I just replaced that logic of `asClosable` wrappings to `closeIterator` logic which will close an iterator if it is closable.
I definitely don't insist on my commit. I just wanted to show my thoughts around it, so we can just remove my commit from your branch if it not suits.
What do you think about it?
https://github.com/JanusGraph/janusgraph/pull/2218/commits/a008d965a0a2e62407ffee118cd128bdc0ef3d86

li-boxuan:
@porunov  Thank you, I like your idea! The code looks cleaner now. I have tested locally and verified GraphTraversal::close method can be delegated to SubqueryIterator correctly.

Btw how did you manage to push to my local fork? I am really curious about it.

porunov:
@li-boxuan Thanks! I will do the final review pass shortly.

> Btw how did you manage to push to my local fork? I am really curious about it.

You allowed edits from maintainers to your local branch when you were creating a PR.
Here this check-box: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2232, opened 2020-10-11
Fix cached relation update hashcode change bug

### LINKED ISSUE #2138: Remove relation(Edge) not success in StandardJanusGraphTx
Please post all questions and issues on [janusgraph-users](https://groups.google.com/forum/#!forum/janusgraph-users)
before opening a GitHub issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

Please include configurations and logs if available.

- Version: 0.5.2
- Storage Backend:hbase 
- Mixed Index Backend:
- Mailing list Thread URL:
- Steps to Reproduce: 
1. get edge (alias e1) by  traversal().V(id).outE(label) 
2. modify e1 property  
3. do something with graph 
4. remvoe e1
5. commit it

such as :
```
 @Test
    public void cc() {
        Edge next = graph.traversal().V().has("_v", 1)
                .outE("t_9")
                .next();
        next.property("_p",2);
        next.remove();
        
        graph.traversal().V().has("_v",1)
                .next().remove();
        graph.tx().commit();
    }
```

Maybe：
HashCode of edge is id : 
```
@Override
    public int hashCode() {
        return Long.hashCode(getCompareId());
    }
```
The container addedRelations in StandardJanusGraphTx:
```
  //######## Data structures that keep track of new and deleted elements
    //These data structures cannot release elements, since we would loose track of what was added or deleted
    /**
     * Keeps track of all added relations in this transaction
     */
    private final AddedRelationsContainer addedRelations;
```

when modify edge(CacheEdge) property, new edge  is added  in  addedRelations with new id , but the new edge set id to old id after add in addedRelations : 
```
private synchronized InternalRelation update() {
        StandardEdge copy = new StandardEdge(super.longId(), edgeLabel(), getVertex(0), getVertex(1), ElementLifeCycle.Loaded);
        copyProperties(copy);
        copy.remove();

        StandardEdge u = (StandardEdge) tx().addEdge(getVertex(0), getVertex(1), edgeLabel());
        if (type.getConsistencyModifier()!=ConsistencyModifier.FORK) u.setId(super.longId());
        u.setPreviousID(super.longId());
        copyProperties(u);
        setId(u.longId());
        return u;
    }
``` 
notice : hashcode is changed because of id is changed 

next step, i remove this edge ,the addedRelations.remove(relation) is trigged,but it can not be removed because of hashcode is changed
Method (StandardJanusGraphTx)as follow : 
```
public void removeRelation(InternalRelation relation) {
        Preconditions.checkArgument(!relation.isRemoved());
        relation = relation.it();
        for (int i = 0; i < relation.getLen(); i++)
            verifyWriteAccess(relation.getVertex(i));

        //Delete from Vertex
        for (int i = 0; i < relation.getLen(); i++) {
            relation.getVertex(i).removeRelation(relation);
        }
        //Update transaction data structures
        if (relation.isNew()) {
            addedRelations.remove(relation);  // remove it from ObjectHashSet
            if (TypeUtil.hasSimpleInternalVertexKeyIndex(relation)) newVertexIndexEntries.remove((JanusGraphVertexProperty) relation);
        } else {//.....}
``` 

### Stack Trace (if you have one)

```
java.lang.IllegalArgumentException: null
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:108)
	at org.janusgraph.graphdb.database.StandardJanusGraph.prepareCommit(StandardJanusGraph.java:559)
	at org.janusgraph.graphdb.database.StandardJanusGraph.commit(StandardJanusGraph.java:731)
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.commit(StandardJanusGraphTx.java:1425)
	at org.janusgraph.graphdb.tinkerpop.JanusGraphBlueprintsGraph$GraphTransaction.doCommit(JanusGraphBlueprintsGraph.java:297)
	at org.apache.tinkerpop.gremlin.structure.util.AbstractTransaction.commit(AbstractTransaction.java:104)

```

li-boxuan:
When a CacheEdge or a CacheVertexProperty is updated, JanusGraph
creates a new edge/property with auto-generated longId, and then overrides
it with the old id. Since AddedRelationsContainer implementation
relies on the longId as part of hash key, creating a new edge/property
with new longId can lead to a commit exception. This bug appears
when a vertex property or an edge is updated and then deleted in the
same transaction.

Fixes #2138

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
@@ -87,9 +87,9 @@ public class CacheEdge extends AbstractEdge {
         copyProperties(copy);
         copy.remove();
 
-        StandardEdge u = (StandardEdge) tx().addEdge(getVertex(0), getVertex(1), edgeLabel());
-        if (type.getConsistencyModifier()!=ConsistencyModifier.FORK) u.setId(super.longId());
-        u.setPreviousID(super.longId());
+        Long id = type.getConsistencyModifier() != ConsistencyModifier.FORK ? longId() : null;
+        StandardEdge u = (StandardEdge) tx().addEdge(id, getVertex(0), getVertex(1), edgeLabel());
+        u.setPreviousID(longId());
         copyProperties(u);
         setId(u.longId());
         return u;
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheVertexProperty.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheVertexProperty.java
@@ -71,8 +71,8 @@ public class CacheVertexProperty extends AbstractVertexProperty {
         copyProperties(copy);
         copy.remove();
 
-        StandardVertexProperty u = (StandardVertexProperty) tx().addProperty(getVertex(0), propertyKey(), value());
-        if (type.getConsistencyModifier()!= ConsistencyModifier.FORK) u.setId(longId());
+        Long id = type.getConsistencyModifier() != ConsistencyModifier.FORK ? longId() : null;
+        StandardVertexProperty u = (StandardVertexProperty) tx().addProperty(getVertex(0), propertyKey(), value(), id);
         u.setPreviousID(longId());
         copyProperties(u);
         return u;
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -162,6 +162,7 @@ import org.janusgraph.graphdb.query.graph.GraphCentricQueryBuilder;
 import org.janusgraph.graphdb.query.profile.QueryProfiler;
 import org.janusgraph.graphdb.query.profile.SimpleQueryProfiler;
 import org.janusgraph.graphdb.query.vertex.BasicVertexCentricQueryBuilder;
+import org.janusgraph.graphdb.relations.CacheVertexProperty;
 import org.janusgraph.graphdb.relations.RelationIdentifier;
 import org.janusgraph.graphdb.schema.EdgeLabelDefinition;
 import org.janusgraph.graphdb.schema.PropertyKeyDefinition;
@@ -296,6 +297,38 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
         assertCount(1, graph.query().vertices());
         assertCount(1, graph.query().has(nameUniqueVertexPropertyName, "v1").vertices());
         assertCount(0, graph.query().has(nameUniqueVertexPropertyName, "v2").vertices());
+    }
+
+    /**
+     * Update properties of vertex property, vertex, and edge, and then remove the vertex and edge in the same transaction
+     */
+    @Test
+    public void testUpdateThenRemove() {
+        EdgeLabel forkConnect = mgmt.makeEdgeLabel("fork-connect").make();
+        mgmt.setConsistency(forkConnect, ConsistencyModifier.FORK);
+        mgmt.commit();
+
+        Vertex v = graph.traversal().addV().property("_v", 1).next();
+        v.property("_v").property("flag", false);
+        Vertex v2 = graph.traversal().addV().property("_v", 2).next();
+        Edge edge = v.addEdge("connect", v2, "_p", 1);
+        Edge forkEdge = v.addEdge("fork-connect", v2, "_p", 1);
+        graph.tx().commit();
+
+        // update properties and then remove vertices & edges
+        JanusGraphVertexProperty p = (JanusGraphVertexProperty) graph.traversal().V(v).properties("_v").next();
+        p.property("flag", true);
+        edge = graph.traversal().E(edge).next();
+        edge.property("_p", 2);
+        forkEdge = graph.traversal().E(forkEdge).next();
+        forkEdge.property("_p", 2);
+        // FIXME: https://github.com/JanusGraph/janusgraph/issues/1981
+        // v2 = graph.traversal().V().has("_v", 2).next();
+        // v2.property("_v", 3);
+        graph.traversal().V().drop().iterate();
+        graph.tx().commit();
+        assertFalse(graph.traversal().V().hasNext());
+        assertFalse(graph.traversal().E().hasNext());
     }
 
     /**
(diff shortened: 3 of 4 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
```diff
@@ -87,9 +87,12 @@ private synchronized InternalRelation update() {
         copyProperties(copy);
         copy.remove();
 
-        StandardEdge u = (StandardEdge) tx().addEdge(getVertex(0), getVertex(1), edgeLabel());
-        if (type.getConsistencyModifier()!=ConsistencyModifier.FORK) u.setId(super.longId());
-        u.setPreviousID(super.longId());
+        Long id = null;
+        if (type.getConsistencyModifier() != ConsistencyModifier.FORK) {
+            id = longId();
+        };
```
porunov, line 93: Nitpick. Don't assign `null` to the variable when it isn't needed to be assigned. I.e. change the code to:
```
Long id;
if (type.getConsistencyModifier() != ConsistencyModifier.FORK) {
  id = longId();
} else {
  id = null;
}
```
I'm also OK with ternary operator if `if{}else{}` isn't suitable. I.e.:
```
Long id = (type.getConsistencyModifier() != ConsistencyModifier.FORK) ? longId() : null;
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheVertexProperty.java
```diff
@@ -71,8 +71,11 @@ private synchronized InternalRelation update() {
         copyProperties(copy);
         copy.remove();
 
-        StandardVertexProperty u = (StandardVertexProperty) tx().addProperty(getVertex(0), propertyKey(), value());
-        if (type.getConsistencyModifier()!= ConsistencyModifier.FORK) u.setId(longId());
+        Long id = null;
+        if (type.getConsistencyModifier() != ConsistencyModifier.FORK) {
+            id = longId();
+        };
```
porunov, line 77: Same here. Don't assign `null` when it isn't needed

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
li-boxuan: Done, but may I know why?

li-boxuan:
I triggered Travis three times by re-push, but every time it failed for a different reason. I'll stop doing re-push for now.

farodin91:
If you like you can flag these failing tests, using `@FlakyTest` instead of `@Test`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
porunov: Thanks! Assigning value to a variable which we know can be changed in the next operation and which isn't used before that operation is a redundant operation.

porunov (approved):
LGTM. Thank you @li-boxuan !

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/CacheEdge.java
li-boxuan: I see. Are you speaking from a code style perspective? Then that would be make some sense to me. I would be surprised and interested if there is actually difference in runtime performance.

porunov: I don't know if JIT optimizes such cases or not, but from bytecode perspective there is a very tiny difference. It's almost not noticeable and most likely won't affect performance in real life but there is a difference from operations amount. I don't want to check the byte code of this class but lets see the bytecode difference in much smaller use case:
```
/*
0: aconst_null
1: astore_1
2: aload_0
3: arraylength
4: iconst_5
5: if_icmple     13
8: iconst_5
9: invokestatic  #2                  // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
12: astore_1
13: getstatic     #3                  // Field java/lang/System.out:Ljava/io/PrintStream;
16: aload_1
17: invokevirtual #4                  // Method java/io/PrintStream.println:(Ljava/lang/Object;)V
20: return
 */
public static void main(String[] args) {

    Integer s = null;

    if(args.length > 5){
        s = 5;
    }

    System.out.println(s);
}

/*
0: aload_0
1: arraylength
2: iconst_5
3: if_icmple     14
6: iconst_5
7: invokestatic  #2                  // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
10: astore_1
11: goto          16
14: aconst_null
15: astore_1
16: getstatic     #3                  // Field java/lang/System.out:Ljava/io/PrintStream;
19: aload_1
20: invokevirtual #4                  // Method java/io/PrintStream.println:(Ljava/lang/Object;)V
23: return
 */
public static void main(String[] args) {

    Integer s;

    if(args.length > 5){
        s = 5;
    } else {
        s = null;
    }

    System.out.println(s);
}
```
In the first main method there are 13 operations in the main method. In the second main method we have 14 operations in the main method. We might assume that the second method will be executed slower than the first method because it has more operations but if we read them we clearly see that it's the opposite way. 
In case `args.length > 5` is `false` than we will execute 10 operations in the first method and 10 operations in the second method. So it's a tie when the if statement is `false`.
In case `args.length > 5` is `true` than we will execute 13 operations in the first method and 12 operations in the second method (notice `goto 16` operation). So in case `if` statement is `true` the second method will perform 1 operation less than the first method.

In conclusion: JIT compiler *may* optimize such use cases but I'm not sure about that. Comparing the performance for such a small difference for modern apps is a waste of time. In the real world most likely we won't see any performance difference in such a small use cases but it doesn't mean that we shouldn't care about redundant operations at all because they *may* decrease the performance and they just increase the code complexity. 
That said, I don't push to use such style and I'm open for other opinions in case you have arguments to use the opposite style, I'm open to discuss that

li-boxuan: Very interesting analysis! I am also not sure if JIT compiler will optimize this. I agree that comparing the performance for such a small difference is not valuable, but since you have the byte code pasted, I'd also like to add some points:

> In case args.length > 5 is true than we will execute 13 operations in the first method and 12 operations in the second method (notice goto 16 operation). So in case if statement is true the second method will perform 1 operation less than the first method.

I agree the second method seems to save one operation, but I doubt if it will be actually faster than the first method. Reasons are:

1) goto operation MIGHT be slower than a normal operation since a jump probably does not leverage cache spatial locality well.
2) CPU branch prediction MIGHT suffer more on goto operation when branch prediction fails. This is just my conjecture though.

porunov: Very good points!

PR #2237, opened 2020-10-13
Issue 2236: Remove graph and traversal source from script engine bind…

### LINKED ISSUE #2236: JanusGraphManager leaks dropped graph instances
The `JanusGraphManager` currently does not clean up the Gremlin script engine bindings for graphs that have been dropped via the `ConfiguredGraphFactory`. These graphs will persist in memory because they are referenced in the script engine bindings until the JanusGraph instance has been restarted. This can be reproduced by creating and then dropping a graph. An inspection of a heap dump post drop will show that the graph instance is still held in memory despite it being deleted.

twilmes:
This fixes #2236 by removing graph traversal sources and graphs when they are dropped/removed by the `ConfiguredGraphFactory`.

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
@@ -139,6 +139,7 @@ public class ConfiguredGraphFactory {
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
         Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
         final Graph graph = jgm.removeGraph(graphName);
+        jgm.removeTraversalSource(toTraversalSourceName(graphName));
         if (null != graph) graph.close();
         return (JanusGraph) graph;
     }
@@ -243,7 +244,9 @@ public class ConfiguredGraphFactory {
     private static void removeGraphFromCache(final JanusGraph graph) {
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
         Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
-        jgm.removeGraph(((StandardJanusGraph) graph).getGraphName());
+        final String graphName = ((StandardJanusGraph) graph).getGraphName();
+        jgm.removeGraph(graphName);
+        jgm.removeTraversalSource(toTraversalSourceName(graphName));
         final ManagementSystem mgmt = (ManagementSystem) graph.openManagement();
         mgmt.evictGraphFromCache();
         mgmt.commit();
@@ -287,6 +290,15 @@ public class ConfiguredGraphFactory {
     public static Map<String, Object> getTemplateConfiguration() {
         final ConfigurationManagementGraph configManagementGraph = getConfigGraphManagementInstance();
         return configManagementGraph.getTemplateConfiguration();
+    }
+
+    /**
+     * Get traversal source name given a graph name.
+     * @param graphName
+     * @return A traversal source name for the provided graph name
+     */
+    public static String toTraversalSourceName(String graphName){
+        return graphName + "_traversal";
     }
 }
 
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java
@@ -14,6 +14,7 @@
 
 package org.janusgraph.graphdb.management;
 
+import org.apache.tinkerpop.gremlin.jsr223.GremlinScriptEngineManager;
 import org.janusgraph.graphdb.database.StandardJanusGraph;
 import org.janusgraph.core.JanusGraphFactory;
 import org.janusgraph.core.ConfiguredGraphFactory;
@@ -170,6 +171,8 @@ public class JanusGraphManager implements GraphManager {
     @Override
     public TraversalSource removeTraversalSource(String tsName) {
         if (tsName == null) return null;
+        // Remove the traversal source from the script engine bindings so that closed/dropped graph instances do not leak
+        removeGremlinScriptEngineBinding(tsName);
         return traversalSources.remove(tsName);
     }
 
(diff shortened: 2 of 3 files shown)


janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
```diff
@@ -139,6 +139,7 @@ public static JanusGraph close(String graphName) throws Exception {
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
         Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
         final Graph graph = jgm.removeGraph(graphName);
+        jgm.removeTraversalSource(graphName + "_traversal");
```
angadsalaria, line 142: might be a good idea to keep `_traversal` here and in line 249 below as a referenced constant.
Seems to imply usage of convention used in multiple places as a "magic string". In which case, changing in 1 place will break overall & dependent functions.
Also referenced [here](https://github.com/JanusGraph/janusgraph/blob/ac20052335506012e9b721b9897fef4b6100026e/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java#L267) in the master branch.

porunov: Agree with @angadsalaria .
We can also make a static method which converts `graphName` to Traversal Source name. Something like:
```
public static String toTraversalSourceName(String graphName){
    return graphName + "_traversal";
}
```
and then use that method to convert `graphName` to `traversalSourceName`.
I.e.:
```
jgm.removeTraversalSource(toTraversalSourceName(graphName));
```

janusgraph-core/src/main/java/org/janusgraph/core/ConfiguredGraphFactory.java
```diff
@@ -243,7 +244,9 @@ private static void removeGraphFromCache(final String graphName) {
     private static void removeGraphFromCache(final JanusGraph graph) {
         final JanusGraphManager jgm = JanusGraphManagerUtility.getInstance();
         Preconditions.checkNotNull(jgm, JANUS_GRAPH_MANAGER_EXPECTED_STATE_MSG);
-        jgm.removeGraph(((StandardJanusGraph) graph).getGraphName());
+        final String graphName = ((StandardJanusGraph) graph).getGraphName();
+        jgm.removeGraph(graphName);
+        jgm.removeTraversalSource(graphName + "_traversal");
```
porunov, line 249: Same here, if we would have a dedicated method to convert `graphName` to `traversalSourceName` we could change this line to:
```
jgm.removeTraversalSource(toTraversalSourceName(graphName));
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java
```diff
@@ -252,6 +254,8 @@ public Graph openGraph(String gName, Function<String, Graph> thunk) {
     @Override
     public Graph removeGraph(String gName) {
         if (gName == null) return null;
+        // Remove the graph from the script engine bindings so that closed/dropped graph instances do not leak
+        removeBinding(gName);
```
porunov, line 258: Shouldn't we convert `gName` to `tsName`? I.e. `removeBinding(gName + "_traversal");`

porunov (commented):
Thank you @twilmes !
I just have several small comments, otherwise LGTM

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java
```diff
@@ -270,5 +274,13 @@ private void updateTraversalSource(String graphName, Graph graph, GremlinExecuto
         graphManager.putTraversalSource(traversalName, traversalSource);
     }
 
+    private void removeBinding(String key) {
+        if (null != gremlinExecutor) {
+            final Bindings bindings = gremlinExecutor.getScriptEngineManager().getBindings();
+            if (bindings.get(key) != null) {
+                gremlinExecutor.getScriptEngineManager().getBindings().remove(key);
+            }
```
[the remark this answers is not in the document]
rngcntr: `Bindings.get(key) == null`  is ambiguous. From the [docs](https://docs.oracle.com/javase/7/docs/api/javax/script/Bindings.html#get(java.lang.Object)):

> [...] if this map contains a mapping from a key k to a value v such that (key==null ? k==null : key.equals(k)), then this method returns v; otherwise it returns null. 

Therefore `!=null` means that the binding contains a key which is also not `null`. However, I did not research how a `null` entry in a `Binding` arises and which meaning it has in the context of the `GremlinExecutor` and therefore I'm not sure if such an entry should remain here or not. But from my naive understanding, @porunov is right here.

li-boxuan:
ping @twilmes any updates?

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java
twilmes: In this case, we need to refer the graph itself so I believe it should be `gName`.

twilmes:
Thanks for the review, I've pushed an update to remove duplication and added a unit test.

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java
porunov: You right, thanks!

porunov (approved):
LGTM. Thank you @twilmes !

janusgraph-test/src/test/java/org/janusgraph/core/ConfiguredGraphFactoryTest.java
```diff
@@ -278,5 +286,54 @@ public void ensureCallingGraphCloseResultsInNewGraphReferenceOnNextCallToOpen()
             ConfiguredGraphFactory.close("graph1");
         }
     }
+
+    @Test
+    public void dropShouldCleanUpTraversalSourceAndBindings() throws Exception {
```
li-boxuan, line 291: This test seems to be flaky. I encountered "ConfiguredGraphFactoryTest.dropShouldCleanUpTraversalSourceAndBindings:312 expected: <true> but was: <false>" twice and had to retry to make it pass.

PR #2242, opened 2020-10-24
Fix VertexIDAssigner bug when ids-flush is disabled

li-boxuan:
When ids-flush is disabled, ids are re-assigned even if elements
already have ids, e.g. update of existing elements. This commit
fixes this bug and does not re-assign ids.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idassigner/VertexIDAssigner.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idassigner/VertexIDAssigner.java
@@ -297,7 +297,7 @@ public class VertexIDAssigner implements AutoCloseable {
 
     private void assignID(final InternalElement element, final long partitionIDl, final IDManager.VertexIDType userVertexIDType) {
         Preconditions.checkNotNull(element);
-        Preconditions.checkArgument(!element.hasId());
+        if (element.hasId()) return;
         Preconditions.checkArgument((element instanceof JanusGraphRelation) ^ (userVertexIDType!=null));
         Preconditions.checkArgument(partitionIDl >= 0 && partitionIDl < partitionIdBound, partitionIDl);
         final int partitionID = (int) partitionIDl;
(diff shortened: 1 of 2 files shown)


li-boxuan:
The bug can be reproduced using the test case in this PR.

janusgraph-test/src/test/java/org/janusgraph/graphdb/idmanagement/VertexIDAssignerTest.java
```diff
@@ -111,6 +114,27 @@ private JanusGraph getInMemoryGraph(boolean allowSettingVertexId, boolean idsFlu
...
 
+    @Test
+    public void testDisableIdsFlush() {
+        final JanusGraph graph = getInMemoryGraph(false, false, 2);
+        JanusGraphVertex v1 = graph.addVertex();
+        JanusGraphVertex v2 = graph.addVertex();
+        Edge e = v1.addEdge("knows", v2);
+        e.property("prop", "old");
+        graph.tx().commit();
+        Assert.assertEquals("old", graph.traversal().E().next().property("prop").value());
```
farodin91, line 125: Could you use the assert from Junit5 instead of Junit4

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idassigner/VertexIDAssigner.java
```diff
@@ -297,7 +297,7 @@ private long getPartitionID(final InternalVertex v) {
 
     private void assignID(final InternalElement element, final long partitionIDl, final IDManager.VertexIDType userVertexIDType) {
         Preconditions.checkNotNull(element);
-        Preconditions.checkArgument(!element.hasId());
+        if (element.hasId()) return;
```
li-boxuan, line 300: Adding some explanation:
https://github.com/JanusGraph/janusgraph/blob/8b2dd99b5dc4a9d4b29be168769aacd059d8a12e/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java#L674-L685

When `hasAssignIDsImmediately` is disabled, `idAssigner::assignIDs` is called, which in turn (skipping a few function calls within `VertexIDAssigner`) calls this `assignId` method.

If the given `element` is an existing element, e.g. an existing edge, then this `Preconditions.checkArgument(!element.hasId())` will fail. Thus, we should not throw an exception here. Moreover, we should `return` immediately, as we do not want to assign a new ID to it.

porunov (approved):
LGTM. Thank you @li-boxuan !
Test fails without the fix (as expected).

farodin91:
@li-boxuan Did you forget to push to master?

li-boxuan:
@farodin91 Thanks for the reminder. I just pushed it to my fork repository. Will push to janusgraph/master once the build passes.

PR #2244, opened 2020-11-01
Bug fixes: relations cannot be removed after update

### LINKED ISSUE #1981: Modified vertex property is not removed after remove operation
If the property is modified and then removed in the same transaction, it won't be removed in the storage backend.

- Version: 0.5.0-SNAPSHOT
- Storage Backend: ScyllaDB 3.2.1 (not important)
- Mixed Index Backend: ElasticSearch 6.8.6 (not important)
- Mailing list Thread URL: https://groups.google.com/forum/#!topic/janusgraph-users/c4AAydzlEV4
- Steps to Reproduce: 

```
@Test
public void testRemoveVertexPropertyAfterModification() {

    Vertex v = graph.addVertex();
    v.property("name", "name1");
    graph.tx().commit();

    v = graph.traversal().V(v.id()).next();
    v.property("name", "name2");
    v.property("name").remove();
    graph.tx().commit();

    assertFalse(graph.traversal().V(v.id()).values("name").hasNext());
}
```

As seen from the example above, the `name` property isn't removed even so we called `remove` before the commit.

li-boxuan:
This commit includes fixes for three bugs:

1) Caused by the optimization in StandardJanusGraphTx::addProperty, which saves a database
read (and delete) operation for vertex property upsert. If a vertex property (or the vertex
as a whole) is removed after the property is modified in the same transaction, then the old
value in database will not be deleted after the transaction commits. Reported by #1981.

2) Caused by CacheVertexProperty::remove method which does nothing if it finds the given
vertex property is already in the deletedRelations map. If the user deletes a vertex property
after they update a nested property of that vertex property in the same transaction, then
the old value will be removed from database but the new value will be saved into database
after the transaction commits.

3) Caused by CacheEdge::remove method which does nothing if it finds the given edge is
already in the deletedRelations map. If the user deletes an edge after they update a property
of that edge in the same transaction, then the old value will be removed from database but the
new value will be saved into database after the transaction commits.
-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphVertexProperty.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphVertexProperty.java
@@ -17,6 +17,8 @@ package org.janusgraph.core;
 
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 
+import java.util.function.Consumer;
+
 /**
  * JanusGraphProperty is a {@link JanusGraphRelation} connecting a vertex to a value.
  * JanusGraphProperty extends {@link JanusGraphRelation}, with methods for retrieving the property's value and key.
@@ -43,6 +45,18 @@ public interface JanusGraphVertexProperty<V> extends JanusGraphRelation, VertexP
     @Override
     default PropertyKey propertyKey() {
         return (PropertyKey)getType();
+    }
+
+    static Consumer<JanusGraphVertexProperty> getRemover(VertexProperty.Cardinality cardinality, Object value) {
+        if (cardinality == VertexProperty.Cardinality.single) {
+            return JanusGraphElement::remove;
+        } else {
+            return p -> {
+                if (p.value().equals(value)) {
+                    p.remove();
+                }
+            };
+        }
     }
 
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/StandardVertexProperty.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/StandardVertexProperty.java
@@ -17,6 +17,10 @@ package org.janusgraph.graphdb.relations;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+import org.janusgraph.core.JanusGraphElement;
+import org.janusgraph.core.JanusGraphRelation;
+import org.janusgraph.core.JanusGraphVertexProperty;
 import org.janusgraph.core.PropertyKey;
 import org.janusgraph.graphdb.internal.ElementLifeCycle;
 import org.janusgraph.graphdb.internal.InternalVertex;
@@ -25,6 +29,7 @@ import org.janusgraph.graphdb.types.system.ImplicitKey;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Consumer;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
@@ -37,13 +42,21 @@ public class StandardVertexProperty extends AbstractVertexProperty implements St
         this.lifecycle = lifecycle;
     }
 
-    //############## SAME CODE AS StandardEdge #############################
-
     private static final Map<PropertyKey, Object> EMPTY_PROPERTIES = ImmutableMap.of();
 
+    private boolean isUpsert;
     private byte lifecycle;
     private long previousID = 0;
     private volatile Map<PropertyKey, Object> properties = EMPTY_PROPERTIES;
+
+    /**
+     * Mark this property as 'upsert', i.e. the old property value is not read from DB and marked as
+     * deleted in the transaction
+     * @param upsert
+     */
+    public void setUpsert(final boolean upsert) {
+        isUpsert = upsert;
+    }
 
     @Override
     public long getPreviousID() {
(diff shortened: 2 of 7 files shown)


li-boxuan (commented):
Added some comments to explain the changes:

janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphVertexProperty.java
```diff
@@ -45,4 +47,16 @@ default PropertyKey propertyKey() {
         return (PropertyKey)getType();
     }
 
+    static Consumer<JanusGraphVertexProperty> getRemover(VertexProperty.Cardinality cardinality, Object value) {
+        if (cardinality == VertexProperty.Cardinality.single) {
+            return JanusGraphElement::remove;
```
porunov, line 52: I don't know much about how lambda are compiled but if it is changed to `new Consumer<JanusGraphVertexProperty> {....}` shouldn't this consumer be static? I don't think we need to create a new instance of this consumer because it doesn't reuse any other parameters.
I.e., I think it would be better to create a static consumer like:
```
private static Consumer<JanusGraphVertexProperty> SINGLE_PROPERTY_REMOVER_CONSUMER = JanusGraphElement::remove; 
```
and reuse `SINGLE_PROPERTY_REMOVER_CONSUMER` here. That said, if you know that `JanusGraphElement::remove` will be compiled to static consumer, I'm totally good with it.

porunov (commented):
Great work @li-boxuan ! Thank you!
I made a very quick pass and I have a small question below

janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphVertexProperty.java
li-boxuan: Looks like it is compiled to static consumer:

```java
boolean sameObject = false;
Consumer<JanusGraphVertexProperty> c1 = null;
for (int i = 0; i < 2; i++) {
    Consumer<JanusGraphVertexProperty> c2 = new Consumer<JanusGraphVertexProperty>() {
        @Override
        public void accept(JanusGraphVertexProperty p) {
            p.remove();
        }
    };
    if (c1 == null) c1 = c2;
    else sameObject = c1 == c2; // false
}
```

when using method reference or lambda function, result is different:

```java
boolean sameObject = false;
Consumer<JanusGraphVertexProperty> c1 = null;
for (int i = 0; i < 2; i++) {
    Consumer<JanusGraphVertexProperty> c2 = JanusGraphElement::remove;
    if (c1 == null) c1 = c2;
    else sameObject = c1 == c2; // true
}
```

Example modified from https://stackoverflow.com/questions/23983832/is-method-reference-caching-a-good-idea-in-java-8/23991339#23991339

porunov: Thank you for checking that!

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -798,14 +798,13 @@ public JanusGraphVertexProperty addProperty(VertexProperty.Cardinality cardinali
 //                }
 //            }
 
+            long propId = id == null ? IDManager.getTemporaryRelationID(temporaryIds.nextID()) : id;
+            StandardVertexProperty prop = new StandardVertexProperty(propId, key, (InternalVertex) vertex, normalizedValue, ElementLifeCycle.New);
+            if (config.hasAssignIDsImmediately() && id == null) graph.assignID(prop);
+
```
porunov, line 804: Why did you move this logic from bottom to up?

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/StandardVertexProperty.java
```diff
@@ -101,6 +114,11 @@ public synchronized void remove() {
         if (!ElementLifeCycle.isRemoved(lifecycle)) {
             tx().removeRelation(this);
             lifecycle = ElementLifeCycle.update(lifecycle, ElementLifeCycle.Event.REMOVED);
+            if (isUpsert) {
+                VertexProperty.Cardinality cardinality = ((PropertyKey) type).cardinality().convert();
+                Consumer<JanusGraphVertexProperty> propertyRemover = JanusGraphVertexProperty.getRemover(cardinality, value());
+                getVertex(0).query().types(type.name()).properties().forEach(propertyRemover);
```
porunov, line 120: I'm good with `getVertex(0)` but maybe it's better to use `element()`?

porunov (commented):
Just a couple more questions

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -816,7 +815,10 @@ public JanusGraphVertexProperty addProperty(VertexProperty.Cardinality cardinali
                 if ( (!config.hasVerifyUniqueness() || ((InternalRelationType)key).getConsistencyModifier()!=ConsistencyModifier.LOCK) &&
                         !TypeUtil.hasAnyIndex(key) && cardinality==keyCardinality.convert()) {
                     //Only delete in-memory so as to not trigger a read from the database which isn't necessary because we will overwrite blindly
-                    ((InternalVertex) vertex).getAddedRelations(p -> p.getType().equals(key)).forEach(propertyRemover);
+                    //We need to label the new property as "upsert", so that in case property deletion happens, we not only delete this new
+                    //in-memory property, but also read from database to delete the old value (if exists)
+                    ((InternalVertex) vertex).getAddedRelations(p -> p.getType().equals(key)).forEach(p -> propertyRemover.accept((JanusGraphVertexProperty) p));
+                    prop.setUpsert(true);
```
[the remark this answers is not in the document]
li-boxuan: A property is always "new" when you update its value. You cannot directly set a value to a property object; instead, you can only attach a key-value pair to a vertex object, i.e. `v.property("key", "value")`. 

Thus, the `prop` here is defined by:
`StandardVertexProperty prop = new StandardVertexProperty(propId, key, (InternalVertex) vertex, normalizedValue, ElementLifeCycle.New);` which is always "new".

I believe the scenario you mentioned is already handled as part of this PR, but I am adding a test case to be sure.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
li-boxuan: because we need to set `prop.setUpsert(true);` later

porunov (approved):
Great work @li-boxuan ! LGTM.
Just need to wait for TravisCI build to pass

porunov:
@li-boxuan Don't forget to merge this fix to `master` branch as well

PR #2061, opened 2020-03-26
Add support for graph binary

farodin91:
Fixes #1564

Signed-off-by: Jan Jansen <jan.jansen@gdata.de>

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there an issue associated with this PR? Is it referenced in the commit message?
- [x] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?
- [ ] If this PR is a documentation-only change, have you added a `[doc only]`
  tag to the first line of your commit message to avoid spending CPU cycles in
  Travis CI when no code, tests, or build configuration are modified?

### Note:
Please ensure that once the PR is submitted, you check Travis CI for build issues and submit an update to your PR as soon as possible.

PR #2253, opened 2020-11-18
Add serializiation fallback for older janusgraph drivers

farodin91:
Follow up for #2061 to reduce breaking changes between version 0.5 and 0.6.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -85,10 +85,12 @@ on Cassandra's JVM configurations, Elasticsearch and the old gremlin-server.sh.
 ##### Serialization of JanusGraph predicates has changed
 
 The serialization of JanusGraph predicates has changed in this version for both 
-GraphSON and Gryo. It is therefore necessary to update both the client and 
-the server to this version in parallel as the server will not be able to 
-deserialize a JanusGraph predicate that was serialized by a client prior 
-to version 0.6.0 once it was updated to version 0.6.0.
+GraphSON and Gryo. The newest version of the JanusGraph Driver requires a JanusGraph 
+Server version of 0.6.0 and above. The server includes a fallback for clients with an 
+older driver to make the upgrade to version 0.6.0 easier. This means that the server 
+can be upgraded first without having to update all clients at the same time. The 
+fallback will however be removed in a future version of JanusGraph so clients should 
+also be upgraded.
 
 ##### GraphBinary is now supported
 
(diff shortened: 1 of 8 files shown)


docs/changelog.md
```diff
@@ -85,10 +85,9 @@ on Cassandra's JVM configurations, Elasticsearch and the old gremlin-server.sh.
 ##### Serialization of JanusGraph predicates has changed
 
 The serialization of JanusGraph predicates has changed in this version for both 
-GraphSON and Gryo. It is therefore necessary to update both the client and 
-the server to this version in parallel as the server will not be able to 
-deserialize a JanusGraph predicate that was serialized by a client prior 
-to version 0.6.0 once it was updated to version 0.6.0.
+GraphSON and Gryo. The newest version of the JanusGraph Driver requires a JanusGraph 
+Server version of 0.6.0 and above. The server includes a fallback for older driver,
```
FlorianHockmann, line 89: I would formulate the last sentence slightly differently and expand it a bit to better explain how the upgrade can be performed now:

> The server includes a fallback for clients with an older driver to make the upgrade to version 0.6.0 easier. This means that the server can be upgraded first without having to update all clients at the same time. The fallback will however be removed in a future version of JanusGraph so clients should also be upgraded.

FlorianHockmann (commented):
Apart from my comment for the changelog, I wonder how we could test this. Given that we would need a client from an earlier version to test this, we can probably only test it manually, right? Did you perform a quick test to verify that GraphSON and Gryo still work with an older driver?

farodin91:
> Apart from my comment for the changelog, I wonder how we could test this. Given that we would need a client from an earlier version to test this, we can probably only test it manually, right? Did you perform a quick test to verify that GraphSON and Gryo still work with an older driver?

I will check to build a test.

farodin91:
I found a pretty simple solution, see https://github.com/JanusGraph/janusgraph/pull/2253/files#diff-397bb0e0cc2bcb59b56de6da870cbfcade8f40dbf212dfdee84c55e9cba73180R118-R169.

janusgraph-driver/src/test/java/org/janusgraph/graphdb/tinkerpop/JanusGraphIoRegistryTest.java
```diff
@@ -115,6 +115,58 @@ public void testJanusGraphPredicatesAsGryo() throws SerializationException {
 
     }
 
+    @Test
+    public void testOldFormatJanusGraphPredicatesAsGryo() throws SerializationException {
+        Graph graph = EmptyGraph.instance();
+        GraphTraversalSource g = graph.traversal();
+
+        // Janus Graph Geo, Text Predicates
```
FlorianHockmann, line 123: (nitpick) The comment doesn't make much sense as there are no Geo predicates here and only one text predicate.

janusgraph-driver/src/test/java/org/janusgraph/graphdb/tinkerpop/JanusGraphIoRegistryTest.java
```diff
@@ -115,6 +115,58 @@ public void testJanusGraphPredicatesAsGryo() throws SerializationException {
...
+        Bytecode oldBytecode = serializeByteCodeAfterDeserializeAsGryo(g.V().has("name", new P<>(Text.CONTAINS, "test")));
```
FlorianHockmann, line 124: Nice and easy solution!

FlorianHockmann (approved):

Issue #1457, opened 2019-03-01
Flaky Cassandra test: JanusGraphTest.simpleLogTest


FlorianHockmann:
```java
[ERROR] Failures: 
[ERROR]   AstyanaxGraphTest>JanusGraphTest.simpleLogTest:4133->JanusGraphTest.simpleLogTest:4427 expected: <5> but was: <6>
```

Job was: 
```
MODULE='cassandra' ARGS='-Dtest=**/graphdb/astyanax/*'
```

FlorianHockmann:
This test just also failed for Thrift:

```java
[ERROR] Failures: 
[ERROR]   ThriftGraphTest>JanusGraphTest.simpleLogTest:4190->JanusGraphTest.simpleLogTest:4484 expected: <5> but was: <6>
```

in job:

```
MODULE='cassandra' ARGS='-Dtest=**/graphdb/thrift/* -Dtest.skip.ordered=true -Dtest.skip.ssl=true -Dtest.skip.serial=true'
```

FlorianHockmann:
simpleLogTestWithFailure also seems to be affected [which is nearly the same test](https://github.com/JanusGraph/janusgraph/blob/11e40fca4c3a9584163f36bf3aa5bacd232ed1a7/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java#L4850):

```java
    @Test
    public void simpleLogTest() throws InterruptedException {
        simpleLogTest(false);
    }

    @Test
    public void simpleLogTestWithFailure() throws InterruptedException {
        simpleLogTest(true);
    }
```

It failed for CQL for a PR on `master`:

```
Error:  Failures: 
Error:    CQLGraphCacheTest>JanusGraphTest.simpleLogTestWithFailure:4851->JanusGraphTest.simpleLogTest:5137 expected: <1> but was: <2>
[INFO] 
Error:  Tests run: 419, Failures: 1, Errors: 0, Skipped: 2
```

PR #2257, opened 2020-11-23
Replace TravisCI Standard tests

### LINKED ISSUE #2260: Use GitHub Actions instead of TravisCI
https://groups.google.com/g/janusgraph-dev/c/02rWKvC9LwU

- [x] Standard Tests
  - [x] most tests
  - [x]  solr
- [x] Full build
  - [x] cql tests
  - [x] flag
- [x] Release builds
- [x] Documentation 
- [x] TinkerPop tests
- [ ] Style and license enforcement 
- [ ] Find a solution for coverity scan

farodin91:
Fixes #2260

Have be done in follow up PR's
- [ ] Replace release tests
- [ ] Replace documentation build and deployment
- [ ] Full build

-----

### For all changes:

### For code changes:

### Note:


janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/berkeleyje/BerkeleyGraphTest.java
```diff
@@ -112,19 +113,14 @@ public void testConcurrentConsistencyEnforcement() {
         //Do nothing TODO: Figure out why this is failing in BerkeleyDB!!
     }
 
-    @Test
+    @FlakyTest
```
FlorianHockmann, line 116: Is this test now suddenly flaky in GitHub Actions?

.github/workflows/standard-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+# limitations under the License.
+
+name: Standard Tests
+
+on:
+  pull_request:
+    paths-ignore:
+    - 'docs/**'
+  push: 
+    branches: [ master, v* ]
```
FlorianHockmann, line 22: Don't we also want to have tests for feature branches?

.github/workflows/standard-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - run: mvn clean install --projects janusgraph-${{ matrix.module }} --also-make 
+          -DskipTests=true -Dmaven.javadoc.skip=true --batch-mode ${{ matrix.install-args }}
+      - run: mvn verify --projects janusgraph-${{ matrix.module }} ${{ matrix.args }}
```
FlorianHockmann, line 112: You removed the `Pcoverage` parameter here if I'm not mistaken. Don't we need that for _codecov_: https://codecov.io/gh/JanusGraph/janusgraph ?

FlorianHockmann (commented):
Thanks for tackling this so quickly, @farodin91! The changes also look good to me, I just added three questions.

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/berkeleyje/BerkeleyGraphTest.java
farodin91: The test hit a race condition on the github actions consistently.

.github/workflows/standard-tests.yml
farodin91: changed

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/berkeleyje/BerkeleyGraphTest.java
li-boxuan: Do you have a link for the failure message? I wonder why it hits a race condition on GitHub Actions but not Travis.

farodin91: It seems that to be that GitHub Actions is a bit slower. You have to reduce the time of timeout which is not possible.

farodin91:
I did a split up three parts to reduce of re run all builds.

pom.xml
```diff
@@ -1501,4 +1501,4 @@
             </build>
         </profile>
     </profiles>
-</project>
+</project>
```
li-boxuan, line 1504: this can be reverted

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -19,9 +19,6 @@ to ensure the following steps have been taken:
 
 ### For documentation related changes:
 - [ ] Have you ensured that format looks appropriate for the output in which it is rendered?
-- [ ] If this PR is a documentation-only change, have you added a `[doc only]`
-  tag to the first line of your commit message to avoid spending CPU cycles in
-  Travis CI when no code, tests, or build configuration are modified?
 
 ### Note:
 Please ensure that once the PR is submitted, you check Travis CI for build issues and submit an update to your PR as soon as possible.
```
li-boxuan, line 24: Travis CI should be replaced by GitHub Actions

li-boxuan (commented):
Changes look good to me in general! I didn't check the workflow configs but I suppose they are just more or less the same as Travis' jobs.

I see your branch is `github-actions-step-1`. Is there anything left?

pom.xml
farodin91: Yes and no. I can revert it, which would increase the build since is wrongly filled. The cache is selected by the hash of all pom's.

farodin91:
@li-boxuan I just replaced the Standard Tests. Missing is full build, docs, and release build.

.github/PULL_REQUEST_TEMPLATE.md
farodin91: Do we need this not anymore? The pr can't be merge with hit all checks.

li-boxuan: I thought the plan was to abandon Travis and use GitHub Actions only... if we need both then we can change it into "check Travis CI and GitHub Actions" here I guess.

farodin91: I'm working on the next PR already to add release tests and the docs tests to GitHub. Just for a short time we have Travis and GitHub

amcp: maybe you can add a TODO here that you can remove when the transition is complete?

.github/workflows/ci-backend.yml
```diff
@@ -0,0 +1,98 @@
...
+on:
+  pull_request:
+    paths-ignore:
+    - 'docs/**'
+  push:
+    paths-ignore:
+    - 'docs/**'
+
+env:
+  ES_JAVA_OPTS: "-Xms256m -Xmx512m"
```
amcp, line 26: is this enough memory?

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLDistributedStoreManagerTest.java
```diff
@@ -54,6 +55,7 @@ public void tearDown() throws BackendException {
     @Override
     @Test
     @FeatureFlag(feature = JanusGraphFeature.OrderedScan)
+    @Disabled(value = "GitHub Actions do run inside of a container")
```
amcp, line 58: if you need to make this change, seems like the new build process (Github Action) is breaking this test. Please add a comment describing why this should be disabled or fix the test so that it does not need to be disabled.

amcp (changes requested):
minor comments

.github/PULL_REQUEST_TEMPLATE.md
farodin91: I added a list of parts that are missing in this pr.

.github/workflows/ci-backend.yml
farodin91: This is copied from TravisCI.

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLDistributedStoreManagerTest.java
farodin91: GitHub isn't breaking test. Test is just designed to run inside of container. 

For me, it's also not 100% clear. What test tries to test!

li-boxuan:
I see several "Check failure" Annotations on https://github.com/JanusGraph/janusgraph/actions/runs/390439092 Do you know what that means and why there is "Check failure" while the CI is still successful?

Another question I have is, why do we have "pull_request" checks and "push" checks?

I never used GitHub Actions before, so probably answers to my questions are dumb 😅

farodin91:
> I see several "Check failure" Annotations on https://github.com/JanusGraph/janusgraph/actions/runs/390439092 Do you know what that means and why there is "Check failure" while the CI is still successful?

These annotations are log lines. They seems be parsed magically by github actions.

> Another question I have is, why do we have "pull_request" checks and "push" checks? They look identical to me

- push is for local branches such as master
- pr...

> I never used GitHub Actions before, so probably my questions are dumb sweat_smile

it is also the first time, i play around with GitHub Actions.

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLDistributedStoreManagerTest.java
```diff
@@ -54,6 +55,7 @@ public void tearDown() throws BackendException {
     @Override
     @Test
     @FeatureFlag(feature = JanusGraphFeature.OrderedScan)
+    @Disabled(value = "Test can't run inside of a container")
```
porunov, line 58: I didn't check it but could you maybe give a short explanation why can't we execute this test?
As I understand we can execute it in TravisCI, so why can't we execute it in GitHub Actions?

janusgraph-cql/src/test/java/org/janusgraph/graphdb/cql/CQLGraphTest.java
```diff
@@ -138,4 +139,10 @@ public void testCustomConfigUsedByTx() {
                 .get(WRITE_CONSISTENCY));
         tx.rollback();
     }
+
+    @RepeatedIfExceptionsTest(repeats = 3)
+    @Override
+    public void simpleLogTest() throws InterruptedException{
+        super.simpleLogTest();
+    }
```
porunov, line 147: Could be great if you could explain why you Overwrided the `super` test to make it Flacky for CQL

CONTRIBUTING.md
```diff
@@ -169,18 +169,6 @@ you've signed the contribution.
...
-> build time for the other changes which actually change the code.
->
-> The tests actually run twice for each PR:
-> 
-> * when the PR is submitted for review
-> * when the PR is merged to the base branch
->
-> Having [doc only] in the commit skips the first one, but the merge commit also
-> needs it, so having it in the title (first line of commit) helps it easily
-> propagate to both places.
```
porunov, line 183: Should we maybe support `[doc only]` in GitHub actions and skip tests if someone specifies it?
https://github.com/marketplace/actions/skip-based-on-commit-message

porunov (commented):
Thank you @farodin91 ! I have several questions

janusgraph-berkeleyje/src/test/java/org/janusgraph/graphdb/berkeleyje/BerkeleyGraphTest.java
porunov: I think this test is executed faster in GitHub Actions, thus it doesn't hit `timeout` as expected by the test. That is why it fails. I think we need to resolve this commit because it is now Disabled in this PR (which I think is right because the test is very flaky).

porunov:
Looks like GitHub plans to add possibility to re-run failed jobs in the future but this feature isn't available right now. Maybe it's better to move all tests which might hit a Flaky test to a separate workflow? If so, we will be able to re-execute them quickly. That said, I'm not sure if that's a good idea or not. I'm also good to have several common groups of workflows (as you did right now).
This comment is just my thoughts and not a change request.

CONTRIBUTING.md
farodin91: I think we should wait until all functions are moved to GitHub Actions. I don't think we need for the doc only for the most cases. The path ignore will work in the most cases.

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLDistributedStoreManagerTest.java
farodin91: I tried to make the skip description independently of GitHub Actions. The test isn't build to run inside of container which is normal case in GitHub Actions and not on the TravisCI side.

janusgraph-cql/src/test/java/org/janusgraph/graphdb/cql/CQLGraphTest.java
farodin91: https://github.com/JanusGraph/janusgraph/issues/1457

farodin91: It was never marked as flakytest. I often did just restart the failing tests on the TravisCI instead of marking Thema as flaky.

farodin91:
You can't move flaky tests into one workflow. They are good distribute over multiple test classes and back ends.

For The failing hbase tests, I have no idea. It fails often during downloading of an dependency.

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLDistributedStoreManagerTest.java
porunov: Do you have an error log of this test? I just want to understand why it can't be executed inside a container. What are the limitations of it? Is it related to networks? GitHub Actions allow to configure networks. I'm not sure I understand why this test can't be executed inside the container. Does it have some specific requirements which are not accessible inside a container?

CONTRIBUTING.md
porunov: What do you mean by `path ignore`? I'm newbie in GitHub Actions so don't know much yet.
I though that `[doc only]` would be useful if we are changing docs only and we don't want to run the whole build. It would save us some workers from unnecessary runs. That said, I'm fine dropping `[doc only]` support if it is complicated to add that feature right now. We may just open another task for this feature.

porunov:
> You can't move flaky tests into one workflow. They are good distribute over multiple test classes and back ends.

I didn't mean to move them into one workflow. I mean that we could create a separate workflow for each test suit which often fails.
For example, in Travis we have 25 Test suits (excluding `[full build]`). About half of them are more or less stable and mostly never fail due to Flaky tests but other tests (like cql, hbase modules) fail often. So, I was thinking about creating 10-15 different workflows in GitHub actions. Each workflow would have only 1 test suit. Right now your created 3 workflows. I'm totally fine with 3 workflows but was thinking if we split test suits even more (10-15 workflows).
That said, it's just my thoughts and I didn't verify if it is a good idea.

janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLDistributedStoreManagerTest.java
farodin91: It will return remote instead of local. The reason is testcontainers fallback to the gateway if testcontainers is running inside a container. This leads to a non localhost address.

CONTRIBUTING.md
farodin91: GitHub actions can decide bases on the changes if it runs a workflow or not. For us the standard tests will never run if only things are changed inside of the docs directory.

porunov: Oh cool. I didn't know about this feature. If so, you are right. We don't need `[doc only]`

farodin91:
I found a solution for the hbase build issues.

farodin91:
I currently trying to improve the cache mechanism.

farodin91:
I'll close this PR for now. I'll try find a solution for the caching issue. In the mean time, I'll open a pr for the documentation and the release tests which a bit easier to handle.

PR #2258, opened 2020-11-26
Fixes #2255: leaked transactions in ConsistentKeyLocker

### LINKED ISSUE #2255: Leaked transactions in ConsistentKeyLocker
Transactions created in ConsistentKeyLocker.overrideTimestamp() are neither committed or rolled back by the callers. This results in leaked transactions. ConsistentKeyLocker needs to be updated to commit or rollback the created transactions.
 
Relevant code from ConsistentKeyLocker.java given below.

https://github.com/JanusGraph/janusgraph/blob/9ba0b2d850066a32551f9da4af0780f643288376/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java#L549-L554

https://github.com/JanusGraph/janusgraph/blob/9ba0b2d850066a32551f9da4af0780f643288376/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java#L384-L388

https://github.com/JanusGraph/janusgraph/blob/9ba0b2d850066a32551f9da4af0780f643288376/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java#L400-L403

https://github.com/JanusGraph/janusgraph/blob/9ba0b2d850066a32551f9da4af0780f643288376/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java#L535-L539

mneethiraj:
#2225: Fixed transaction leaks in ConsistentKeyLocker. Updated unit tests to detect leaked transaction.

(cherry picked from commit 12718473fae6e9172e292cfcd61de175d22b2577)
(cherry picked from commit 3e16a8fdd8dc81a47529cf0950d60c7ed3debdd3)

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
@@ -382,13 +382,20 @@ public class ConsistentKeyLocker extends AbstractLocker<ConsistentKeyLockStatus>
         final Timer writeTimer = times.getTimer().start();
         StaticBuffer newLockCol = serializer.toLockCol(writeTimer.getStartTime(), rid, times);
         Entry newLockEntry = StaticArrayEntry.of(newLockCol, zeroBuf);
+        StoreTransaction newTx = null;
         try {
-            final StoreTransaction newTx = overrideTimestamp(txh, writeTimer.getStartTime());
+            newTx = overrideTimestamp(txh, writeTimer.getStartTime());
+
             store.mutate(key, Collections.singletonList(newLockEntry),
                 null == del ? KeyColumnValueStore.NO_DELETIONS : Collections.singletonList(del), newTx);
+
+            newTx.commit();
+            newTx = null;
         } catch (BackendException e) {
             log.debug("Lock write attempt failed with exception", e);
             t = e;
+        } finally {
+            rollbackIfNotNull(newTx);
         }
         writeTimer.stop();
 
@@ -398,11 +405,18 @@ public class ConsistentKeyLocker extends AbstractLocker<ConsistentKeyLockStatus>
     private WriteResult tryDeleteLockOnce(StaticBuffer key, StaticBuffer col, StoreTransaction txh) {
         Throwable t = null;
         final Timer delTimer = times.getTimer().start();
+        StoreTransaction newTx = null;
         try {
-            final StoreTransaction newTx = overrideTimestamp(txh, delTimer.getStartTime());
+            newTx = overrideTimestamp(txh, delTimer.getStartTime());
+
             store.mutate(key, ImmutableList.of(), Collections.singletonList(col), newTx);
+
+            newTx.commit();
+            newTx = null;
         } catch (BackendException e) {
             t = e;
+        } finally {
+            rollbackIfNotNull(newTx);
         }
         delTimer.stop();
 
@@ -533,9 +547,13 @@ public class ConsistentKeyLocker extends AbstractLocker<ConsistentKeyLockStatus>
     protected void deleteSingleLock(KeyColumn kc, ConsistentKeyLockStatus ls, StoreTransaction tx) {
         List<StaticBuffer> deletions = ImmutableList.of(serializer.toLockCol(ls.getWriteTimestamp(), rid, times));
         for (int i = 0; i < lockRetryCount; i++) {
+            StoreTransaction newTx = null;
             try {
-                StoreTransaction newTx = overrideTimestamp(tx, times.getTime());
+                newTx = overrideTimestamp(tx, times.getTime());
                 store.mutate(serializer.toLockKey(kc.getKey(), kc.getColumn()), ImmutableList.of(), deletions, newTx);
+
+                newTx.commit();
+                newTx = null;
                 return;
             } catch (TemporaryBackendException e) {
                 log.warn("Temporary storage exception while deleting lock", e);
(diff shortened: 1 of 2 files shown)


li-boxuan (approved):

mneethiraj:
Boxuan,

 

Thanks for reviewing and approving the pull request. Is there anything else needed from me before this can be merged?

 

Thanks,

Madhan

li-boxuan:
Hi Madhan, there is nothing pending from your side. According to https://docs.janusgraph.org/development/#review-then-commit-rtc, we need one more approval from another committer. If there is no review in the next two days, I will merge your PR following the lazy consensus.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
```diff
@@ -382,13 +382,30 @@ private WriteResult tryWriteLockOnce(StaticBuffer key, StaticBuffer del, StoreTr
...
+            newTx.commit();
+            newTx = null;
         } catch (BackendException e) {
             log.debug("Lock write attempt failed with exception", e);
             t = e;
+        } finally {
+            if (newTx != null) {
+                try {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Transaction is still open! Rolling back", new Throwable());
```
porunov, line 401: Why do we adding `new Throwable()` to the log? I guess, it isn't useful because it doesn't have any information. Instead, I would log `toString()` of the transaction, which may have more useful information. I.e. something like:
```
if (log.isDebugEnabled()) {
    log.debug("Transaction is still open. Rolling back the transaction: "+tx.toString());
}
```

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
```diff
@@ -382,13 +382,30 @@ private WriteResult tryWriteLockOnce(StaticBuffer key, StaticBuffer del, StoreTr
...
+                    }
+
+                    newTx.rollback();
+                } catch (Throwable excp) {
+                    log.debug("Transaction rollback failed", excp);
```
porunov, line 406: In this situation the transaction will leak. I think we should log here `error` with the explanation that the above transaction is leaked. Of course it is better to somehow track it and let the logic somehow handle the transaction but I guess it makes the implementation slightly harder, so I'm totally fine to just log an error here and add a `TODO` message to actually resolve it later.
Something like:
```
//TODO: We need to handle the transaction newTx which potentially may leak into memory
log.error("Failed to rollback the transaction "+newTx.toString()+". The transaction now may be leaked into memory.", e);
```

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
```diff
@@ -398,11 +415,28 @@ private WriteResult tryWriteLockOnce(StaticBuffer key, StaticBuffer del, StoreTr
...
+                }
+            }
```
porunov, line 439: We have multiple same `finally` blocks. I propose to move the same code into a separate method. I.e.:
```
private void rollbackIfNotNull(BaseTransaction tx){
    if (tx != null) {
        try {
            if (log.isDebugEnabled()) {
                log.debug("Transaction is still open. Rolling back the transaction: "+tx.toString());
            }
            tx.rollback();
        } catch (Throwable e) {
            //TODO: We need to handle the transaction newTx which potentially may leak into memory
            log.error("Failed to rollback the transaction "+tx.toString()+". The transaction now may be leaked into memory.", e);
        }
    }
}
```

After that we can change all places where you added `finally` like the next:
```
finally {
  rollbackIfNotNull(newTx);
}
```

porunov (changes requested):
Thank you @mneethiraj for the contribution! I have several comments below

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
mneethiraj: Throwable in this debug log statement will print the stack, which could be useful in troubleshooting leaked transactions. Let me know if this should be removed.

I agree on including tx.toString() in the log. Will update shortly for this.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
mneethiraj: I agree. Will update the patch shortly. Thanks!

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
mneethiraj: I agree. Will update the patch shortly. Thanks!

janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
porunov: Good point. Agree about stack trace. It might be useful

porunov (commented):
The code LGTM but could you please squash you commits into one commit?
You can use `git rebase -i a22a62e01271d1e3bdeeb6f698a3a3fed641a7c1` where you will need to pick one commit and squash another commit or you can use any other technique to squash commits which you are convenient with

porunov (approved):
LGTM. Thank you @mneethiraj for the contribution!

PR #2264, opened 2020-12-06
Rewrite testConcurrentConsistencyEnforcement to properly test lock contention

li-boxuan:
The old test fails to test the locking mechanism because one thread runs so fast such that other threads often fail before they commit. To properly test lock contention, we need to let them commit together. Note that this test still can only test local lock contention (threads running on the same process), but not the case where transactions running on different JVMs compete for locks.

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -122,6 +122,7 @@ import org.janusgraph.diskstorage.configuration.ConfigElement;
 import org.janusgraph.diskstorage.configuration.ConfigOption;
 import org.janusgraph.diskstorage.configuration.WriteConfiguration;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.ScanMetrics;
+import org.janusgraph.diskstorage.locking.PermanentLockingException;
 import org.janusgraph.diskstorage.log.Log;
 import org.janusgraph.diskstorage.log.Message;
 import org.janusgraph.diskstorage.log.MessageReader;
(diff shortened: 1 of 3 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -2959,17 +2960,31 @@ private void executeLockConflictingTransactionJobs(JanusGraph graph, Transaction
                 tx2.commit();
                 fail("Storage backend does not abort conflicting transactions");
             } catch (JanusGraphException ignored) {
+                Throwable rootCause = ignored.getCause().getCause();
+                assertTrue(rootCause instanceof PermanentLockingException);
+                assertTrue(rootCause.getMessage().contains("Expected value mismatch for"));
             }
         } else {
             try {
```
farodin91, line 2968: It would cool to replace these lines with `assertThrows`.

li-boxuan: AFAIK, `assertThrows` can only assert the outmost exception but not inner exception(s).

farodin91:
```
JanusGraphException janusGraphException = assertThrows(JanusGraphException.class, () -> tx2.commit());
Throwable rootCause = janusGraphException.getCause().getCause();
assertTrue(rootCause instanceof PermanentLockingException);
assertTrue(rootCause.getMessage().contains("Expected value mismatch for"));
```

farodin91: This was my idea.

farodin91 (approved):
LGTM! Thank you @li-boxuan.

Issue #1868, opened 2019-10-24
Not step doesn't look in Index


jakubclark:
In JanusGraph `.not(__.has("pname", "value"))` after an or step, is not processed correctly by the query planner.
This issue was first described here: https://github.com/JanusGraph/janusgraph/issues/163#issuecomment-343517162.

For example:

The query bellow does not use an index, since"query.force-index" is set to "true" and thus, results in an exception:

```
g.V().or(__.has("pname1","value1"),__.has("pname2","value2")).not(__.has("props","value3"))
```

The query bellow works just fine:

```
g.V().or(__.has("pname1","value1"),__.has("pnam2","value2")).has("props",P.neq("value3"))
```

Those queries are semantically the same. But the execution is not the same.
The first version tries to do a full-graph scan, which is disabled in our setup.
The second query using `P.neq` is able to use the index.

The `Not` step does not use the index, if it occurs just after an `Or` step.

We have a mixed index with all the properties used in the query.

- Version: `0.3.2`
- Storage Backend: Cassandra/ScyllaDB
- Mixed Index Backend: Elasticsearch
- Steps to Reproduce: Execute the following query in the Gremlin console:

```
g.V().or(__.has("pname1","value1"),__.has("pname2","value2")).not(__.has("props","value3"))
```

### Stack Trace

```
gremlin> g.V().or(__.has("pname1","value1"),__.has("pname2","value2")).not(__.has("props","value3"))
Could not find a suitable index to answer graph query and graph scans are disabled: [()]:VERTEX
Type ':help' or ':h' for help.
Display stack trace? [yN]y
org.apache.tinkerpop.gremlin.jsr223.console.RemoteException: Could not find a suitable index to answer graph query and graph scans are disabled: [()]:VERTEX
	at org.apache.tinkerpop.gremlin.console.jsr223.DriverRemoteAcceptor.submit(DriverRemoteAcceptor.java:178)
	at org.apache.tinkerpop.gremlin.console.GremlinGroovysh.execute(GremlinGroovysh.groovy:99)
	at org.codehaus.groovy.tools.shell.Shell.leftShift(Shell.groovy:122)
	at org.codehaus.groovy.tools.shell.ShellRunner.work(ShellRunner.groovy:95)
	at org.codehaus.groovy.tools.shell.InteractiveShellRunner.super$2$work(InteractiveShellRunner.groovy)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #2265, opened 2020-12-06
Fix neq and null condition query behavior

### LINKED ISSUE #2205: neq behaviour differs from TinkerGraph
TinkerPop added a test which now fails with JanusGraph. 
In our case we assume that if there is no property with the specified name in `has` step with `neq` predicate than the result is `true` but TinkerGraph assumes that if there is no such property the result should always be `false`.

The related discussion with @spmallette: https://github.com/apache/tinkerpop/commit/76fc5c23e72b7ad9faafb1faab836ce1ac1259d2#r42228711

As we need to alight with TinkerPop, I believe we need to change the behaviour on neq step to preserve compatibility.

- Version: 0.5.2, 0.6.0-SNAPSHOT
- Storage Backend: All
- Mixed Index Backend: N/A
- Mailing list Thread URL: https://groups.google.com/g/gremlin-users/c/DbuJ0Drt0Ec/m/fg6WJ_SwAwAJ
- Steps to Reproduce: 

```
gremlin> graph=JanusGraphFactory.open("inmemory")
==>standardjanusgraph[inmemory:[127.0.0.1]]
gremlin> graph.addVertex()
==>v[4344]
gremlin> graph.traversal().V().has("p", neq("v"))
18:51:57 WARN  org.janusgraph.graphdb.transaction.StandardJanusGraphTx  - Query requires iterating over all vertices [()]. For better performance, use indexes
==>v[4344]
gremlin>
gremlin> graph=TinkerGraph.open()
==>tinkergraph[vertices:0 edges:0]
gremlin> graph.addVertex()
==>v[0]
gremlin> graph.traversal().V().has("p", neq("v"))
gremlin> 
```

li-boxuan:
Fixes #2205

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -5792,16 +5792,100 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
 
     //................................................
 
+    @Test
+    public void testNeqQuery() {
+        makeKey("p2", String.class);
+        PropertyKey p3 = makeKey("p3", String.class);
+        mgmt.buildIndex("composite", Vertex.class).addKey(p3).buildCompositeIndex();
+        finishSchema();
 
+        tx.addVertex();
+        tx.commit();
+        newTx();
+
+        // property not registered in schema
+        assertFalse(tx.traversal().V().has("p1", P.neq("v")).hasNext());
+        assertFalse(tx.traversal().V().has("p1", P.neq(null)).hasNext());
+        // property registered in schema
+        assertFalse(tx.traversal().V().has("p2", P.neq("v")).hasNext());
+        assertFalse(tx.traversal().V().has("p2", P.neq(null)).hasNext());
+        // property registered in schema and has composite index
+        assertFalse(tx.traversal().V().has("p3", P.neq("v")).hasNext());
+        assertFalse(tx.traversal().V().has("p3", P.neq(null)).hasNext());
+    }
+
+    /**
+     * The behaviour of has(p, null) deviates from TinkerGraph. Since JanusGraph does not support null values,
+     * has(p, null) indicates hasNot(p). In other words, an absent property implicitly implies null value for that property.
+     */
+    @Test
+    public void testHasNullQuery() {
+        makeKey("p2", String.class);
+        PropertyKey p3 = makeKey("p3", String.class);
+        mgmt.buildIndex("composite", Vertex.class).addKey(p3).buildCompositeIndex();
+        finishSchema();
+
+        tx.addVertex();
+        tx.commit();
+        newTx();
+
+        // property not registered in schema
+        assertTrue(tx.traversal().V().has("p1", (Object) null).hasNext());
+        // property registered in schema
+        assertTrue(tx.traversal().V().has("p2", (Object) null).hasNext());
+        // property registered in schema and has composite index
+        assertTrue(tx.traversal().V().has("p3", (Object) null).hasNext());
+    }
+
+    /**
+     * The behaviour of hasNot(p) is straight-forward: hasNot(p) means it does not have such property p.
+     * Note that hasNot(p, value) (which is a JanusGraph API rather than gremlin API) is a bit tricky and it is equivalent
+     * to has(p, neq(value)). Therefore, hasNot(p, null) means has(p, neq(null)) which is equivalent to has(p).
+     */
     @Test
     public void testHasNot() {
-        JanusGraphVertex v1, v2;
-        v1 = graph.addVertex();
+        makeKey("p2", String.class);
+        PropertyKey p3 = makeKey("p3", String.class);
+        mgmt.buildIndex("composite", Vertex.class).addKey(p3).buildCompositeIndex();
 
-        v2 = graph.query().hasNot("abcd").vertices().iterator().next();
-        assertEquals(v1, v2);
-        v2 = graph.query().hasNot("abcd", true).vertices().iterator().next();
-        assertEquals(v1, v2);
+        tx.addVertex();
+        tx.commit();
+        newTx();
+
+        // property not registered in schema
+        assertTrue(tx.traversal().V().hasNot("p1").hasNext());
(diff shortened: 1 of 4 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -5792,18 +5792,6 @@ private static void assertElementOrderForWithoutIndex(StandardJanusGraph graph)
...
-    @Test
-    public void testHasNot() {
-        JanusGraphVertex v1, v2;
-        v1 = graph.addVertex();
-
-        v2 = graph.query().hasNot("abcd").vertices().iterator().next();
-        assertEquals(v1, v2);
-        v2 = graph.query().hasNot("abcd", true).vertices().iterator().next();
-        assertEquals(v1, v2);
-    }
```
porunov, line 5805: Why this test was removed from `JanusGraphTest.java`? Tests in `JanusGraphTest.java` are executed against different storages backends. I guess, it would be better to just fix this test instead of removing it

janusgraph-driver/src/main/java/org/janusgraph/core/attribute/Cmp.java
```diff
@@ -73,10 +73,13 @@ public boolean isValidCondition(Object condition) {
...
+            if (value == null) {
+                // To align with TinkerPop behaviour, if an element does not have property p, then has(p, neq(anything))
+                // should always evaluate to false. Note that JanusGraph does not support null value, so the "value == null"
+                // here implies the element does not have such property.
+                return false;
             } else {
-                return !condition.equals(value);
+                return !value.equals(condition);
             }
         }
```
porunov, line 84: I guess, slightly simplified version would be:
```
@Override
public boolean test(Object value, Object condition) {
    // To align with TinkerPop behaviour, if an element does not have property p, then has(p, neq(anything))
    // should always evaluate to false. Note that JanusGraph does not support null value, so the "value == null"
    // here implies the element does not have such property.
    return value != null && !value.equals(condition);
}
```

porunov:
Other than above two comments the code LGTM. Checked that added tests fail without changes with some index backends (as expected).

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
li-boxuan: It is moved to `JanusGraphIndexTest` because I want to enforce the existence of indexes does not alter the behavior. Example: https://github.com/JanusGraph/janusgraph/issues/1868#issuecomment-694933702

porunov: What about behaviour without any index? As I understand, this logic was tested with use cases when there is no necessary indices

li-boxuan: That is also tested. See [testHasNot](https://github.com/JanusGraph/janusgraph/pull/2265/files#diff-e22cf3f57c2f67c294951e7c0318ad14f06ae190ec0237e8076784198ddb8685R251) in JanusGraphIndexTest.

li-boxuan:
After this PR, tinkerpop version can be upgraded to 3.4.8 without test failure.  https://travis-ci.org/github/li-boxuan/janusgraph/builds/747943340 (ignore first failure which always happens on my Travis)

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
porunov: I don't have strong opinion here. `testHasNot` which you removed was tested against different storage backends (CQL, BerkleyDB, HBase, InMemory). In the new `testHasNot` method we test it with several combinations CQL or BerkleyDB as a storage backend and Lucene / Solr / ElasticSearch as index backend. Thus, now the test is executed with less storage backends. 
As I said, I don't have strong opinion here. If you think it's better to keep all that test logic in one place (`JanusGraphIndexTest`) and don't test against InMemory and HBase I'm fine with it. I thought, maybe it's better to split that test logic between `JanusGraphTest` and `JanusGraphIndexTest`. For example, test cases `property not registered in schema`, `property registered in schema` and `property registered in schema and has composite index` can easily be moved to `JanusGraphTest` where they will be executed against a bigger number of Storage backends whereas `property registered in schema and has mixed index` can be in `JanusGraphIndexTest` where it will be executed against different mixed index backends.

li-boxuan: That's fair. The initial thought was to avoid having similar tests in both classes, but you made a valid point. I'll move some back to JanusGraphTest.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -5792,16 +5792,100 @@ private static void assertElementOrderForWithoutIndex(StandardJanusGraph graph)
...
+        assertTrue(tx.traversal().V().has("p2", (Object) null).hasNext());
+        // property registered in schema and has composite index
+        assertTrue(tx.traversal().V().has("p3", (Object) null).hasNext());
+    }
 
+    /**
+     * The behaviour of hasNot(p) is straight-forward: hasNot(p) means it does not have such property p.
+     * Note that hasNot(p, value) (which is a JanusGraph API rather than gremlin API) is a bit tricky and it is equivalent
+     * to has(p, neq(value)). Therefore, hasNot(p, null) means has(p, neq(null)) which is equivalent to has(p).
+     */
```
porunov, line 5844: Doesn't `hasNot` is Gremlin API?
https://tinkerpop.apache.org/javadocs/3.4.8/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#hasNot-java.lang.String-

porunov: Ah sorry, missed second parameter

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2266, opened 2020-12-07
Move release tests and doc to GitHub Actions

farodin91:
Part of #2260

-----

### For all changes:

### For code changes:

### Note:


.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,95 @@
...
+      # TODO: change --dockerfile-url=".../v0.5/docs.Dockerfile" to --dockerfile-url=".../master/docs/docs.Dockerfile"
+      - run: sudo ./bin/structor -o janusgraph -r janusgraph 
+            --force-edit-url 
+            --rqts-url="https://raw.githubusercontent.com/janusgraph/janusgraph/master/requirements.txt" 
+            --dockerfile-url="https://raw.githubusercontent.com/janusgraph/janusgraph/v0.5/docs.Dockerfile" 
+            --dockerfile-name="default-docs.Dockerfile" 
+            --menu.js-url="https://raw.githubusercontent.com/janusgraph/janusgraph/master/docs/theme/structor-menu.js.gotmpl" 
+            --exp-branch=master --debug
+      - uses: actions/upload-artifact@v2
+        if: github.ref != 'ref/head/master'
```
porunov, line 84: Does it mean that we now deploy on `master` branch only? What about `v*` branches?
As I understand we can do something like:
```
on:
  push:
    # Deploy when push to `master` branch and all `v*` branches 
    branches:
      - master
      - v*
```
And then we could do something like:
```
push:
    needs: build-doc
    runs-on: ubuntu-20.04
    if: github.event_name == 'push'
    ...
```
I could be wrong, did research GitHub Actions too much yet but as far as I know the above should work and trigger `push` on `master`, `v0.5`, `v0.4`, etc.

farodin91: My idea is to run the deploy docs just on master.

farodin91: Structor solution could run into some issue, if it is running on branches except the master.

porunov: I don't have a strong opinion here but if we run documentation deployment in `master` branch only, it means that we will not update documentation for older versions if we don't merge that commit in `master` branch. That said, we usually merge all commits from `v*` branches into `master` branch, so, in most cases there will be no problem. 
I'm just curious what has changed in GitHub Actions so that it may run in some issue in other branches. Do you know what issue we may have? Do you know why don't we have that issue with TravisCI but we have it with GitHub Actions?

farodin91: The issue exists before. We had to go through all old branch and update the docker image and TravisCI. 

To explain the issue:
The document will be built and deploy for all branches with on single branch which could lead to deploy the documentation with a old design if we deploy the documentation from old branch. Someone forgot to update the structor version in older branch. We had this already sometime, it can be fixed by triggering master branch again to rebuilt the doc.

porunov: Got it. Thank you for the explanation. I think, we could extend it this file later to trigger `master` documentation build when we push to `v*` branches. That said, no need to do it right now. What I think, is that it might be better to trigger this workflow only when we push to the necessary branch. I.e. you can see in this PR that we actually triggered `[deploy-doc]` and we spent more then 2 minutes on that workflow even so we didn't actually push the changes and we didn't actually deployed any documentation for this PR.
Basically, what I think, is that it's better to not trigger this step at all if it shouldn't be triggered.
I would still suggest to do it like this: https://github.com/JanusGraph/janusgraph/pull/2266#discussion_r537396153 . The only thing to change is to remove `- v*` line. In such case, we will save time not downloading Structor and not doing unnecessary job when it should be skipped.

farodin91: We could put an if condition to deploy task. 
You can dispatch events in different workflows or branches.

I would like to see that docs build is built and can be checked.

porunov: I think I was just confused about `[deploy-doc]` step. I thought this step doesn't b build anything and that `[build-doc]` is fully responsible for building documentation. Should we move those parts which build documentation to `[build-doc]` and leave `[deploy-doc]` with logic just to deploy documentation?
If that's problematic or less efficient, I'm fine leaving it as is

farodin91: build-doc builds the documentation without versioning and builds and deploys the doc with versioning.

farodin91: I could change the deploy-doc method to:
```yaml

  deploy-doc:
    runs-on: ubuntu-20.04
    if: github.ref == 'ref/head/master'
    needs: build-doc
    steps: 
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0
      - run: mkdir ./bin
      - run: curl -sfL https://raw.githubusercontent.com/traefik/structor/master/godownloader.sh | bash -s -- -b ./bin ${STRUCTOR_VERSION}
      # TODO: change --dockerfile-url=".../v0.5/docs.Dockerfile" to --dockerfile-url=".../master/docs/docs.Dockerfile"
      - run: sudo ./bin/structor -o janusgraph -r janusgraph 
            --force-edit-url 
            --rqts-url="https://raw.githubusercontent.com/janusgraph/janusgraph/master/requirements.txt" 
            --dockerfile-url="https://raw.githubusercontent.com/janusgraph/janusgraph/v0.5/docs.Dockerfile" 
            --dockerfile-name="default-docs.Dockerfile" 
            --menu.js-url="https://raw.githubusercontent.com/janusgraph/janusgraph/master/docs/theme/structor-menu.js.gotmpl" 
            --exp-branch=master --debug
      - uses: JamesIves/github-pages-deploy-action@3.7.1
        with:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          REPOSITORY_NAME: JanusGraph/docs.janusgraph.org
          BRANCH: master
          FOLDER: site
          SINGLE_COMMIT: true
```

porunov: I don't have strong preference here. I'm fine with either of the solutions. If `deploy-doc` should be executed to verify documentation build works correctly, I'm good with it.

farodin91: I would change it if it has run in the master branch once.

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,95 @@
...
+            ${{ runner.os }}-maven-
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - run: mvn clean install -pl janusgraph-doc ${{ env.BUILD_MAVEN_OPTS }}
+      - run: git diff  --exit-code docs/basics/janusgraph-cfg.md
+      - run: docker build -t doc-site:mkdocs -f docs/docs.Dockerfile ./docs
+      - run: docker run --rm -v $PWD:/mkdocs doc-site:mkdocs mkdocs build
+      - uses: actions/upload-artifact@v2
+        if: github.ref != 'ref/head/master'
```
FlorianHockmann, line 61: These are not uploaded on master because we deploy them there completely?

.travis.yml
```diff
@@ -154,42 +147,6 @@ jobs:
...
-        - echo "Check for changes in configuration";
-          git diff  --exit-code docs/basics/janusgraph-cfg.md;
-        - docker run --rm -v $PWD:/mkdocs doc-site:mkdocs mkdocs build
-
-    - stage: deploy
-      install: skip
-      script: skip
-      before_deploy:
-        - echo "Download documentation generator";
-          curl -sfL https://raw.githubusercontent.com/containous/structor/master/godownloader.sh | bash -s -- -b $GOPATH/bin ${STRUCTOR_VERSION}
```
FlorianHockmann, line 171: (nitpick) The variable `STRUCTOR_VERSION` can now also be removed completely.

.github/workflows/ci-docs.yml
farodin91: We deploy the documentation on the master. I think their is no reason store the artefacts.

FlorianHockmann (approved):

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,90 @@
...
+    runs-on: ubuntu-20.04
+    if: github.ref == 'ref/head/master'
+    needs: build-doc
+    steps: 
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - run: mkdir ./bin
+      - run: curl -sfL https://raw.githubusercontent.com/traefik/structor/master/godownloader.sh | bash -s -- -b ./bin ${STRUCTOR_VERSION}
+      # TODO: change --dockerfile-url=".../v0.5/docs.Dockerfile" to --dockerfile-url=".../master/docs/docs.Dockerfile"
```
porunov, line 76: When should we change this url? Why not to change it right now? This PR is targeting `master` branch

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,90 @@
...
+      - run: sudo ./bin/structor -o janusgraph -r janusgraph 
+            --force-edit-url 
+            --rqts-url="https://raw.githubusercontent.com/janusgraph/janusgraph/master/requirements.txt" 
+            --dockerfile-url="https://raw.githubusercontent.com/janusgraph/janusgraph/v0.5/docs.Dockerfile"
```
porunov, line 80: Should we use `master` here instead of `v0.5`?

.github/workflows/ci-docs.yml
farodin91: The file doesn't exists right now in master. It will be create with this PR.

.github/workflows/ci-docs.yml
farodin91: Same as above.

.github/workflows/ci-docs.yml
porunov: I didn't research structor flow. I see that the file exists: https://raw.githubusercontent.com/janusgraph/janusgraph/master/docs.Dockerfile

Maybe I'm just missing something

porunov: I think I understand. Most likely we will need to change the URL with a separate PR. I'm good with it

porunov (approved):
LGTM. Thank you @farodin91 !

Issue #2268, opened 2020-12-08
LockEntry set TTL


wForget:
**Describe the feature:**

In the org.janusgraph.diskstorage.locking.consistentkey.ConsistentKeyLocker#tryWriteLockOnce method, newLockEntry can set TTL

li-boxuan:
Hi @wForget, can you describe why you think this feature is needed? Is it because you see old stale data not getting purged in your lock store tables? If this is inspired by your problems in your use of JanusGraph, can you provide more details, including the data backend you are using?

JanusGraph has a cleaning service that purges expired locks (https://docs.janusgraph.org/basics/configuration-reference/#storagelock). Do you have it enabled? I suspect that one has some flaws which might lead to stale locks, but not sure if you are facing this problem and trying to solve it by using TTL.

wForget:
> Hi @wForget, can you describe why you think this feature is needed? Is it because you see old stale data not getting purged in your lock store tables? If this is inspired by your problems in your use of JanusGraph, can you provide more details, including the data backend you are using?
> 
> JanusGraph has a cleaning service that purges expired locks (https://docs.janusgraph.org/basics/configuration-reference/#storagelock). Do you have it enabled? I suspect that one has some flaws which might lead to stale locks, but not sure if you are facing this problem and trying to solve it by using TTL.

Hi @li-boxuan, I didn't encounter the problem that the locks were not cleaned. 

I just think that setting TTL can avoid not releasing the lock in time and generating some dirty data.

PR #2270, opened 2020-12-08
LockEntry set TTL

wForget:
#2268 LockEntry set TTL

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
@@ -380,7 +380,8 @@ public class ConsistentKeyLocker extends AbstractLocker<ConsistentKeyLockStatus>
         Throwable t = null;
         final Timer writeTimer = times.getTimer().start();
         StaticBuffer newLockCol = serializer.toLockCol(writeTimer.getStartTime(), rid, times);
-        Entry newLockEntry = StaticArrayEntry.of(newLockCol, zeroBuf);
+        StaticArrayEntry newLockEntry = (StaticArrayEntry) StaticArrayEntry.of(newLockCol, zeroBuf);
+        newLockEntry.setMetaData(EntryMetaData.TTL, (int) lockExpire.getSeconds());
         StoreTransaction newTx = null;
         try {
             newTx = overrideTimestamp(txh, writeTimer.getStartTime());


janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/ConsistentKeyLocker.java
```diff
@@ -380,7 +380,8 @@ private WriteResult tryWriteLockOnce(StaticBuffer key, StaticBuffer del, StoreTr
         Throwable t = null;
         final Timer writeTimer = times.getTimer().start();
         StaticBuffer newLockCol = serializer.toLockCol(writeTimer.getStartTime(), rid, times);
-        Entry newLockEntry = StaticArrayEntry.of(newLockCol, zeroBuf);
+        StaticArrayEntry newLockEntry = (StaticArrayEntry) StaticArrayEntry.of(newLockCol, zeroBuf);
+        newLockEntry.setMetaData(EntryMetaData.TTL, (int) lockExpire.getSeconds());
```
li-boxuan, line 384: I think this can be a nice experimental feature because someone might prefer Database native TTL to the JanusGraph [StandardLockCleanerService](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/diskstorage/locking/consistentkey/StandardLockCleanerService.java) (which can be enabled by [storage.lock.clean-expired](https://docs.janusgraph.org/basics/configuration-reference/#storagelock)). By experimental, I mean this should be configurable, optional, and disabled by default. Btw I don't see the point of having this TTL and StandardLockCleanerService enabled at the same time since both do the same thing.

Reasons I think this should be optional and disabled by default:
1) Not every database supports TTL. If the user wants to use this feature, they should be aware that their data backend supports TTL.
2) Some database supports TTL but it does not work this way. For example, BerkeleyDB supports TTL but only at the hour granular level. If you set the TTL to a short period of time, say, 2 minutes, the JanusGraph-BerkeleyDB adapter will automatically raise the TTL to 1 hour. Users are very likely to be unaware of this and might wonder why "expired" locks are not cleaned. This should be clearly documented in the configuration option.
3) Some database like Cassandra does not recommend using a short TTL. I am not familiar with the TTL implementation of Cassandra, so I am not sure if using a TTL here is a good option.

li-boxuan (commented):
Thank you @wForget for your contribution! I have left some comments below. In addition, I think you need to add some test case(s).

wForget:
> Thank you @wForget for your contribution! I have left some comments below. In addition, I think you need to add some test case(s).

Thank you very much for your patient reply. Please forgive me for not having a deeper understanding of this feature. I'll try to improve this feature to make it configurable and add test cases.

PR #2169, opened 2020-07-14
Upgrade DataStax cassandra driver to 4.9.0 [full build]

porunov:
Mostly based on #1733 .

This PR adds 2 DataStax drivers into `janusgraph-cql`:
- DataStax 4.9.0 is the main driver which is now used.
- DataStax 3.10.2 is the driver which is now used for Hadoop. `CqlBinaryRecordReader` which is used for Hadoop jobs is still using DataStax Cassandra Driver 3.x.y.

Notice, even so this PR adds 2 different versions of DataStax Cassandra Driver in the same module, those drivers are binary incompatible and share different classpath, thus they don't conflict with each other and can be used simultaneously. 
Later we can move CQL Hadoop formats from `janusgraph-cql` to a new module like `janusgraph-hadoop-cql` or something like that. It will allow to include only 4.x version of the driver into classpath for users who use `janusgraph-cql` but don't use `janusgraph-hadoop-cql`.

Tests passed:
- Custom integration tests. Storage backend: ScyllaDB 4.2.0 (cql driver). Index backend: ElasticSearch 7.6.0 (es driver).
- Distribution build. `janusgraph.sh` starts Cassandra and ElasticSearch. `janusgraph-server` successfully connects to storage backends via cql driver.
- janusgraph-full distribution successfully logs different metrics via JMX (https://github.com/JanusGraph/janusgraph/issues/1087#issuecomment-534513530), console, slf4j.
- Embedded janusgraph successfully logs different metrics via JMX, console, slf4j. Also, tried to log new CQL Session Level and Node Level metrics via [JMX exporter](https://github.com/prometheus/jmx_exporter). New CQL Session Level and Node Level metrics to enable are provided via `storage.cql.metrics.session-enabled` and `storage.cql.metrics.node-enabled` respectively. 
- All TinkerPop tests passed. Executed with `mvn clean install -Dtest.skip.tp=false -DskipTests=true`.
- TravisCI `[full build]` passed.

[...]

PR #2290, opened 2020-12-20
Use available page size for CQL result set instead of forced page size [full build]

### LINKED ISSUE #2254: Potential CQL iteration page fetch bug
As described in the DataStax driver docs: https://docs.datastax.com/en/developer/java-driver/3.9/manual/paging/

> Note that setting a fetch size doesn’t mean that Cassandra will always return the exact number of rows, it is possible that it returns slightly more or less results.

Our current implementation assumes that we always return `paginatedResultSize` for all pages except the last one. For the last page it assumes `<=paginatedResultSize`.
Any time we fetched `paginatedResultSize` (which is equal to `storage.page-size`) we are forcing to fetch the next page, as seen here:
https://github.com/JanusGraph/janusgraph/blob/f5001e6210baa15bc136f3224ba1e83521782cba/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java#L458-L461

It **may** potentially lead to unprocessed rows in case Cassandra returned more elements on the fetched page then set in `paginatedResultSize`. It may also potentially lead to skipping some of the elements on the next page in case the current page contains less elements then `paginatedResultSize` (because the internal driver will prefetch the next page before we actually trigger `getResultSet` and thus we will trigger `getResultSet` while we are processing a second page, which will lead to going straight to the 3rd page without processing the rest of the elements on the second page).

That said, I don't confirm this bug yet but label it as **possible** bug because I didn't confirm that in our case DSE driver returns any other amount of elements on intermediate pages than `paginatedResultSize` amount of elements.
I.e. in case DSE driver returns always `paginatedResultSize` for intermediate pages then I don't see any bug. Otherwise, there is a potential bug.

`CQLPagingIterator` was introduced here https://github.com/JanusGraph/janusgraph/pull/1814
So, this potential bug may affect branches `0.3`, `0.4`, `0.5` and `master`.

Also, notice, this bug will be resolved in `master` branch in this PR https://github.com/JanusGraph/janusgraph/pull/2169 because we change `CQLPagingIterator` in that PR to not relay on `storage.page-size` directly and thus we eliminate that potential bug for `master` branch.
We still may want to fix this **potential** bug in `0.5` branch because we still support it.

- Version: 0.5
- Storage Backend: CQL
- Mixed Index Backend: -
- Mailing list Thread URL: -
- Steps to Reproduce: -

porunov:
Changes in this PR will be ignored in `master` branch because this *potential* bug is fixed in `master` branch.
In the issue I described why it is hard to confirm and test this bug. This PR either fixes a *potential* bug or does nothing (lets say just refactors the code).

Fixes #2254

-----

### For all changes:

### For code changes:

### Note:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
@@ -47,10 +47,10 @@ import static org.janusgraph.diskstorage.cql.CQLConfigOptions.COMPACTION_STRATEG
 import static org.janusgraph.diskstorage.cql.CQLTransaction.getTransaction;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -87,9 +87,7 @@ import com.datastax.driver.core.exceptions.UnsupportedFeatureException;
 import com.datastax.driver.core.schemabuilder.Create.Options;
 import com.datastax.driver.core.schemabuilder.TableOptions.CompactionOptions;
 import com.datastax.driver.core.schemabuilder.TableOptions.CompressionOptions;
-import com.google.common.collect.Lists;
 
-import io.vavr.Lazy;
 import io.vavr.Tuple;
 import io.vavr.Tuple3;
 import io.vavr.collection.Array;
@@ -401,7 +399,7 @@ public class CQLKeyColumnValueStore implements KeyColumnValueStore {
         return Try.of(() -> new CQLResultSetKeyIterator(
                 query,
                 this.getter,
-                new CQLPagingIterator(this.storeManager.getPageSize(), () ->
+                new CQLPagingIterator(() ->
                     getKeysRanged.bind()
                         .setToken(KEY_START_BINDING, metadata.newToken(query.getKeyStart().asByteBuffer()))
                         .setToken(KEY_END_BINDING, metadata.newToken(query.getKeyEnd().asByteBuffer()))
@@ -421,7 +419,7 @@ public class CQLKeyColumnValueStore implements KeyColumnValueStore {
         return Try.of(() -> new CQLResultSetKeyIterator(
                 query,
                 this.getter,
-                new CQLPagingIterator(this.storeManager.getPageSize(), () ->
+                new CQLPagingIterator(() ->
                     getKeysAll.bind()
                         .setBytes(SLICE_START_BINDING, query.getSliceStart().asByteBuffer())
                         .setBytes(SLICE_END_BINDING, query.getSliceEnd().asByteBuffer())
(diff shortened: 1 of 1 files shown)


li-boxuan (approved):
LGTM 👍

porunov:
Looks like it messed some tests. If anyone wants to check that, feel free to do that, otherwise I will check that later

li-boxuan:
I am just thinking: can we just use the same fix you have in https://github.com/JanusGraph/janusgraph/pull/2169 ? getAllExecutionInfo method seems available in Datastax Driver 3.x as well.

porunov:
> I am just thinking: can we just use the same fix you have in #2169 ? getAllExecutionInfo method seems available in Datastax Driver 3.x as well.

Unfortunately no. In Datastax Driver 3.x it returns a new ArrayList instead of returning current ArrayList

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -441,43 +439,43 @@ public KeyIterator getKeys(final SliceQuery query, final StoreTransaction txh) t
...
         }
 
         @Override
         public boolean hasNext() {
-            return !currentResultSet.isExhausted();
+            if(availableWithoutFetching<=0){
+                if(currentResultSet.isFullyFetched()){
+                    return false;
+                }
+                fetchResultSet();
```
li-boxuan, line 456: LGTM 👍  I tested locally using the following code which creates and reads 100 thousand vertices. The test passed. I could also see fetchResultSet() is called in hasNext() method.
```java
   @Test
    public void testHugeGraph() {
        int maxK = 100, maxI = 1000;
        for (int k = 0; k < maxK; k++) {
            for (int i = 0; i < maxI; i++) {
                tx.addVertex("prop", k * maxI + i);
                newTx();
            }
        }

        List<Object> props = graph.traversal().V().values("prop").toList();
        assertEquals(maxK * maxI, new HashSet<>(props).size());
    }
```

That being said, I also observed something weird: when I ran this locally, I put a debug point (in Intellij) on this line (line 456) to check if this line is reachable. Once I saw the debug point reached, I immediately removed the debug point and resumed the application. Then the test case failed because `graph.traversal().V().values("prop").toList()` returned fewer results than expected. I could always reproduce it; even when I set `maxI = 10` (that is, one thousand vertices in total). However, the same test case always passes as long as I don't set a debug point. I tried letting the thread sleep for 1 or more seconds here, but the test case still passes. Do you have any thoughts?

porunov: It's a default behaviour in Intellij debugger to execute the debugged line to print `toString()` result to convince (most likely you hovered your mouth over this line and thus it executed the line).
In you example, you set a debug point on line 456, then you reach that line and when you hover your mouth it will execute `fetchResultSet()` to check if it can print anything meaningful. After that your resume your debug and go fetching a new page (without actually processing any elements on this page). Basically, calling `fetchResultSet()` 2 times instead of 1 time.
That is why it contains fewer elements (i.e. 1 page fewer elements). Moreover, it's actually can produce more elements (in case you are debugging the last page in multi-page traversal).
That said, it's just some debugging problems and not related to the code directly. Also it's just my assumption, in theory it could be something else

li-boxuan: Got it thanks, that makes sense!

li-boxuan (approved):

Issue #1361, opened 2019-01-02
Add Tested compatibility for java 1.11


Dalamos89:
The java licence agreement for java > 1.8 has changed therefore it would be great to at least test java 1.11 and add it to the tested compatibility on your release page.

FlorianHockmann:
Just stumbled upon this old issue. It was already implemented in #2929.

Issue #2161, opened 2020-07-10
Failed to start on JDK 11


nnanda2016:
My env:
- MacOS Catalina (v10.15.5)
- Java: OpenJDK 11.0.7
- Janusgraph: 0.5.2

Following the documentation at https://docs.janusgraph.org/getting-started/installation/ I tried to start a local instance. But it failed with following error.

```bash
java -version
openjdk version "11.0.7" 2020-04-14 LTS

./janusgraph.sh start -v
Forking Cassandra...
Running `nodetool statusthrift`.Unrecognized VM option 'UseParNewGC'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

Is it a known bug? Could someone tell me what's the workaround?

li-boxuan:
As the doc suggests,
> In order to run JanusGraph, Java 8 SE is required

Btw gitter and mailing list are more appropriate places to ask questions.

porunov:
The PR to add support fir jdk 11 in Gremlin is just added: https://github.com/apache/tinkerpop/pull/1299 

TinkerPop will bring jdk 11 support in version 3.5.0 (not released yet). We are expecting to add jdk 11 support after the upgrade.

Current JanusGraph versions don't support jdk 11.

i-davydov:
Any update?

porunov:
@i-davydov TinkerPop added jdk 11 support for TinkerPop version 3.5.0 but that version isn't released yet.
You can search for available TinkerPop versions here: https://github.com/apache/tinkerpop/releases

shyamremella:
@porunov - couple of questions 
1.  Is there a potential release date for Janusgraph 0.6 release?
2. what does experimental support for Java 11 stand for? should we not move our java infra to jdk 11 with 0.6 yet or any known issues with it? 
Thank you!

porunov:
> Is there a potential release date for Janusgraph 0.6 release?

We are waiting on TinkerPop 3.5.1 right now. I created a topic for that release here: https://lists.apache.org/thread.html/r5c3e754750a2b20d4d06ba1cae1baaa3c6bd93fcd24f46ef20a5e55b%40%3Cdev.tinkerpop.apache.org%3E
I don't know when TinkerPop 3.5.1 will be released, so can't say for sure when JanusGraph 0.6.0 will be released.
[...]

PR #2321, opened 2020-12-31
Add exprimental support for java 11

farodin91:
Part of #1361

Duplicate a bunch of tests for java 11. It just a start.

-----

### For all changes:

### For code changes:

### DIFF
--- a/.github/workflows/ci-backend-cql.yml
+++ b/.github/workflows/ci-backend-cql.yml
@@ -63,43 +63,71 @@ jobs:
           - module: cql
             args: "-Pcassandra3-byteordered -Dtest=\"**/diskstorage/cql/*\""
             name: byteordered-diskstorage
+            java: 8
           - module: cql
             args: "-Pcassandra3-murmur -Dtest=\"**/diskstorage/cql/*\""
             name: murmur-diskstorage
+            java: 8
           - module: cql
             args: "-Pcassandra3-byteordered -Dtest=\"**/graphdb/cql/*\""
             name: byteordered-graphdb
+            java: 8
           - module: cql
             args: "-Pcassandra3-murmur -Dtest=\"**/graphdb/cql/*\""
             name: murmur-graphdb
+            java: 8
           - module: cql
             args: "-Pcassandra3-murmur -Dtest=\"**/hadoop/*\""
             name: murmur-hadoop
+            java: 8
           - module: cql
             args: "-Pcassandra3-byteordered -Dtest=\"**/core/cql/*\""
             name: byteordered-core
+            java: 8
           - module: cql
             args: "-Pcassandra3-murmur -Dtest=\"**/core/cql/*\""
             name: murmur-core
+            java: 8
           - module: cql
             args: "-Pcassandra3-murmur-ssl -Dtest=\"**/diskstorage/cql/CQLStoreTest.java\""
             name: murmur-ssl
+            java: 8
           - module: cql
             args: "-Pcassandra3-murmur-client-auth -Dtest=\"**/diskstorage/cql/CQLStoreTest.java\""
             name: murmur-client-auth
+            java: 8
+          - module: cql
+            install-args: "-Pjava-11"
+            args: "-Pcassandra3-murmur -Pjava-11 -Dtest=\"**/diskstorage/cql/*\""
+            name: murmur-diskstorage-java-11
+            java: 11
+          - module: cql
+            install-args: "-Pjava-11"
+            args: "-Pcassandra3-murmur -Pjava-11 -Dtest=\"**/graphdb/cql/*\""
+            name: murmur-graphdb-java-11
+            java: 11
+          - module: cql
+            install-args: "-Pjava-11"
+            args: "-Pcassandra3-murmur -Pjava-11 -Dtest=\"**/hadoop/*\""
+            name: murmur-hadoop-java-11
+            java: 11
           - module: cql
             args: "-Pscylladb -Dtest=\"**/diskstorage/cql/*\""
             name: scylladb-diskstorage
+            java: 8
           - module: cql
             args: "-Pscylladb -Dtest=\"**/graphdb/cql/*\""
             name: scylladb-graphdb
+            java: 8
           # FIXME: this takes forever to run
           # - module: cql
           #   args: "-Pscylladb -Dtest=\"**/hadoop/*\""
           #   name: scylladb-hadoop
+          #   java: 8
           - module: cql
             args: "-Pscylladb -Dtest=\"**/core/cql/*\""
             name: scylladb-core
+            java: 8
     steps:
       - uses: actions/checkout@v2
         with:
@@ -112,7 +140,7 @@ jobs:
             ${{ runner.os }}-maven-
       - uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: ${{ matrix.java }}
       - run: mvn clean install --projects janusgraph-${{ matrix.module }} ${{ env.BUILD_MAVEN_OPTS }} ${{ matrix.install-args }}
       - run: mvn verify --projects janusgraph-${{ matrix.module }} ${{ env.VERIFY_MAVEN_OPTS }} ${{ matrix.args }}
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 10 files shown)


farodin91:
I'll move the release test refactor into a different PR.

porunov:
Related PR (TinkerPop 3.5.0-SNAPSHOT is not build for Java 11 but TinkerPop 3.4.x will is on Java 8): https://github.com/apache/tinkerpop/pull/1299
Related issue: #2161

farodin91:
> Related PR (TinkerPop 3.5.0-SNAPSHOT is not build for Java 11 but TinkerPop 3.4.x will is on Java 8): [apache/tinkerpop#1299](https://github.com/apache/tinkerpop/pull/1299)

I think we can add java 11 support with any major issue. The only failing tests are the full build tests with janusgraph.sh and easymock tests.

> Related issue: #2161

I don't think we can easily fix this issue with shipping a java 8 for cassandra in the full build.

farodin91:
Blocked by #2378 2378

farodin91:
@porunov Do you like to review this PR?

porunov:
I'm back from the vacation. I will review it this week

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -159,7 +158,6 @@
 
     final ExecutorService executorService;
 
-    @Resource
```
porunov, line 162: Is it related to Java 11 upgrade or is it just a cleanup?

.github/workflows/ci-backend-cql.yml
```diff
@@ -63,30 +63,54 @@ jobs:
...
+          - module: cql
+            install-args: "-Pjava-11"
+            args: "-Pcassandra3-murmur -Pjava-11 -Dtest=\"**/graphdb/cql/*\""
+            name: murmur-graphdb-java-11
+            java: 11
+          - module: cql
+            install-args: "-Pjava-11"
+            args: "-Pcassandra3-murmur -Pjava-11 -Dtest=\"**/hadoop/*\""
+            name: murmur-hadoop-java-11
+            java: 11
```
porunov, line 113: Why for Java 11 we test only `cassandra3-murmur` and not `cassandra3-byteordered` or `cassandra3-murmur-ssl` or `cassandra3-murmur-client-auth`?

porunov (commented):
@farodin91 Sorry for the delay in review. I have some questions.
Why we add Java 11 tests only to `cql` and `release` workflows and not to other workflows like `hbase`, `core`, etc.?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
farodin91: I don't understand why it is need and starting with java 9 it is moved into an external package

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/CqlESAssemblyIT.java
```diff
@@ -23,7 +23,7 @@
 public class CqlESAssemblyIT extends AbstractJanusGraphAssemblyIT {
 
     @Container
-    private static JanusGraphCassandraContainer cql = new JanusGraphCassandraContainer(true);
+    private static JanusGraphCassandraContainer _cql = new JanusGraphCassandraContainer(true);
```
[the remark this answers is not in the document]
farodin91: marking as accessed parameter

.github/workflows/ci-backend-cql.yml
farodin91: I don't want copy all tests.

farodin91: The same for hbase and the others.

porunov: I believe you don't need to copy / paste all the tests. I think you can just add java version to matrix as a separate parameter instead of included parameter and it should run all versions of java (i.e. 8 and 11) with all tests.
Here is I found an example: https://ncorti.com/blog/howto-github-actions-build-matrix#example-detekt
In that example he is using the next strategy to execute all combinations of Java (8, 11, 14) and OSes:
```
jobs:
  gradle:
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
        jdk: [8, 11, 14]
```

So, I don't think you need to add `java: 8` and `java: 11` to all the tests. Instead, we can just move them to matrix parameters and run all combinations.
If so, we will be able to guarantee that all our tests can be executed with Java 8 and Java 11.

Also, I would suggest to include `[tp-tests]` to this PR so that we could check if everything really works with Java 11.

farodin91: I did it before as matrix. Not everything is compatible with java 11 such as our full tests. I had issue with the huge test groups because can just rerun a single check.

If we would like to do it just as a martix, i think we refactor the maven poms a bit more. 
My idea was to have just some test for java 11, as mark it as experimental.

farodin91: I don't see an advantange to run everything against java 8 and java 11.

xiangqiao123:
1

farodin91:
@porunov any feedback?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
porunov: I see there is `CQLStoreTest` which injects `CqlSession` into `CQLStoreManager`.

```
    @Mock
    private CqlSession session;

    @InjectMocks
    private CQLStoreManager mockManager = new CQLStoreManager(getBaseStorageConfiguration());

```

I assume, it injects using `@Resource` annotation. That said, it looks like all tests passes. So, maybe it can also inject mock using data types comparison of fields of a class. That said, I didn't verify that.
I guess it is safe to be removed but we just need to double check that.

porunov: Okay, quickly looking at Mockito docs, I confirm that the field will be injected by type.

From [Mockito docs](https://javadoc.io/doc/org.mockito/mockito-core/3.8.0/org/mockito/InjectMocks.html): 
```
Field injection; mocks will first be resolved by type (if a single type match injection will happen regardless of the name), then, if there is several property of the same type, by the match of the field name and the mock name.

Note 1: If you have fields with the same type (or same erasure), it's better to name all @Mock annotated fields with the matching fields, otherwise Mockito might get confused and injection won't happen.

Note 2: If @InjectMocks instance wasn't initialized before and have a no-arg constructor, then it will be initialized with this constructor.
```

porunov: I see that the usage of `@Resource` annotation is used only here for the whole project.
I could assume that it could be used by IoC somewhere but we don't use any DI frameworks. Also, it's clearly that we violate DI principle inside the constructor as we already creating a new instance of `session` inside the constructor. Thus, I assume `@Resource` here was used to rewrite `session` inside already created `CQLStoreManager`. Probably by some test framework which was using that.
I think it is safe to be removed. At least I didn't found any usage of `@Resource` here.

docs/changelog.md
```diff
@@ -65,7 +65,7 @@ compile "org.janusgraph:janusgraph-core:0.6.0"
 * Apache Lucene 8.6.0
 * Apache Solr 7.7.2, 8.5.2
 * Apache TinkerPop 3.4.10
-* Java 1.8
+* Java 8, 11
```
porunov, line 68: I'm just not sure that we can put `11` here if it is an experimental feature. I personally didn't test JanusGraph compatibility with Java 11 and we didn't execute all JanusGraph tests against Java 11 yet. Thus, I can't tell if we really tested JanusGraph compatibility with Java 11, unless you did it.
That said, I'm not against adding Java 11 compatibility here, it's just thoughts I have.

Another thought I have is that we should probably be consistent with Java versions here. If we change `Java 1.8` to `Java 8`, we should probably change it everywhere in this file. That said, it's more like a nitpick.

farodin91: I will keep just java 8 as support.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
farodin91: Thank you looking into it.

porunov (approved):
LGTM. Thank you @farodin91 !

porunov:
Fails in HBase probably not relaited to this PR. Looks like we have general problems with HBase testing

farodin91:
I started to work on the hbase issues by trying to use testcontainers for hbase.

PR #2326, opened 2021-01-03
Refactor dist tests

farodin91:
* Split up AbstractJanusGraphAssemblyIT into a generic class for all release tests and suite for default backend tests
* activate single vertex tests
* centralized naming in pom.xml for zip names

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-dist/pom.xml
+++ b/janusgraph-dist/pom.xml
@@ -24,10 +24,14 @@
         <assembly.es.dir>${project.build.directory}/elasticsearch-${elasticsearch.version}</assembly.es.dir>
         <assembly.cassandra.dir>${project.build.directory}/apache-cassandra-${cassandra.version}</assembly.cassandra.dir>
 
+        <packname.standard>janusgraph-${project.version}</packname.standard>
+        <packname.full>janusgraph-full-${project.version}</packname.full>
+
         <!-- Variables for integration testing on release-candidate archive files -->
         <test.expect.dir>${project.build.directory}/test-classes</test.expect.dir>
-        <test.zipfile.path>${project.build.directory}/janusgraph-${project.version}.zip</test.zipfile.path>
-        <test.zipfile-full.path>${project.build.directory}/janusgraph-full-${project.version}.zip</test.zipfile-full.path>
+        <test.zipfile.path>${project.build.directory}/${packname.standard}.zip</test.zipfile.path>
+        <test.zipfile-full.path>${project.build.directory}/${packname.full}.zip</test.zipfile-full.path>
+        <build.full>package</build.full>
 
         <!-- RPM/DEB packaging -->
         <pkg.release>1</pkg.release>
@@ -674,7 +678,7 @@
                         <executions>
                             <execution>
                                 <id>assemble-full-distribution-archive</id>
-                                <phase>package</phase>
+                                <phase>${build.full}</phase>
                                 <goals>
                                     <goal>single</goal>
                                 </goals>
@@ -689,7 +693,7 @@
                                         ${finalName}-${<id /> from assembly descriptor}.${archive
                                         format extension}
                                     -->
-                                    <finalName>janusgraph-full-${project.version}</finalName>
+                                    <finalName>${packname.full}</finalName>
                                     <appendAssemblyId>false</appendAssemblyId>
                                     <tarLongFileMode>gnu</tarLongFileMode>
                                     <descriptors>
@@ -709,14 +713,7 @@
                                 <configuration>
                                     <!--
                                         We set attached=true so that the gpg plugin will
-                                        sign all of our assembly artifacts. We handle
-                                        deployment of the artifacts manually because
-                                        maven-deploy-plugin can't upload files to a flat
-                                        S3 directory; it insists on creating a
-                                        long/packagename/and/artifactid/and/version/
-                                        path at the destination before uploading, and we
-                                        don't want that for our files S3. codehaus's
-                                        wagon:upload plugin comes to our rescue.
+                                        sign all of our assembly artifacts.
                                     -->
                                     <attach>true</attach>
                                     <!--
@@ -724,7 +721,7 @@
                                         ${finalName}-${<id /> from assembly descriptor}.${archive
                                         format extension}
                                     -->
-                                    <finalName>janusgraph-${project.version}</finalName>
+                                    <finalName>${packname.standard}</finalName>
                                     <appendAssemblyId>false</appendAssemblyId>
                                     <tarLongFileMode>gnu</tarLongFileMode>
                                     <descriptors>
--- a/janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
+++ b/janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
@@ -14,25 +14,7 @@
 
 package org.janusgraph.pkgtest;
 
-import static org.apache.tinkerpop.gremlin.driver.ser.AbstractMessageSerializer.TOKEN_IO_REGISTRIES;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertTimeout;
-
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileReader;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.PrintStream;
-import java.io.Writer;
-import java.net.Socket;
-import java.time.Duration;
-import java.util.*;
-
-import org.apache.commons.io.FileUtils;
+import com.google.common.collect.ImmutableMap;
 import org.apache.tinkerpop.gremlin.driver.Cluster;
 import org.apache.tinkerpop.gremlin.driver.MessageSerializer;
 import org.apache.tinkerpop.gremlin.driver.remote.DriverRemoteConnection;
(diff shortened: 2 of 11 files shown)


janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
```diff
@@ -44,115 +26,98 @@
...
-        final boolean debug = false;
-        ImmutableMap<String, String> contextVars = ImmutableMap.of("janusgraphServerConfig", janusgraphServerConfig);
-        unzipAndRunExpect("janusgraph-server-sh.before.expect.vm", contextVars, full, debug);
-        assertTimeout(Duration.ofSeconds(30), () -> {
-            while (!serverListening("localhost",8182)) {
-                Thread.sleep(1000);
-            }
-        });
+    @Test
+    @Tag(TestCategory.FULL_TESTS)
```
li-boxuan, line 62: Question: what is the purpose of this tag and how does it work?

li-boxuan (commented):
Hi, @farodin91 Can you make the commit message a bit more elaborative, or leave a comment explaining what this PR aims to achieve? That would help with the review process.

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
farodin91: This tag is just a preparation for the java 11 to filter out full tests.

farodin91:
@li-boxuan Tried to update the commit message.

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/TestCategory.java
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.pkgtest;
+
+public class TestCategory {
```
li-boxuan, line 17: There is also a TestCategory in janusgraph-backend-testutils module.

farodin91: It is container for constants. I could rename it to PkgTestCategory

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
```diff
@@ -44,111 +26,109 @@
...
-        ZIPFILE_EXTRACTED = ZIPFILE_PATH.substring(0, ZIPFILE_PATH.length() - 4);
-        ZIPFILE_FULL_PATH = props.getProperty("zipfile-full.path");
-        ZIPFILE_FULL_EXTRACTED = ZIPFILE_FULL_PATH.substring(0, ZIPFILE_FULL_PATH.length() - 4);
+    protected abstract String getGraphName();
 
-        Properties p = new Properties();
-        p.put("file.resource.loader.path", EXPECT_DIR);
-        Velocity.init(p);
+    protected Boolean supportsIndex() {
+        return false;
```
li-boxuan, line 53: I am just thinking... shall we make it as true by default? If we make it true, then if someone adds a subclass in the future which **does not** support index, the test case will immediately fail and they would know that they should override this method. On contrary, let's say we make it false by default, then if someone adds a subclass in the future which **does** support index (without noting they should override this method), index tests will be skipped, which is not ideal.

janusgraph-dist/pom.xml
```diff
@@ -684,29 +690,22 @@
...
                                     <!--
                                         We set attached=true so that the gpg plugin will
-                                        sign all of our assembly artifacts. We handle
-                                        deployment of the artifacts manually because
-                                        maven-deploy-plugin can't upload files to a flat
-                                        S3 directory; it insists on creating a
-                                        long/packagename/and/artifactid/and/version/
-                                        path at the destination before uploading, and we
-                                        don't want that for our files S3. codehaus's
-                                        wagon:upload plugin comes to our rescue.
```
li-boxuan, line 701: I am not familiar with this part. Can you explain why you deleted this paragraph?

li-boxuan (commented):
In your commit message you mentioned, "activate single vertex tests". Could you point out what part does this? I think I missed it.

janusgraph-dist/pom.xml
farodin91: I just saw this comment which make no sense. We are not using s3 save some files.

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
farodin91: i will check

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
```diff
@@ -44,111 +26,109 @@
...
-        }
-        finally
-        {
-            if(s != null)
-                try {s.close();}
-                catch(Exception e){}
+    @Test
+    @Tag(PkgTestCategory.FULL_TESTS)
+    public void testSimpleGremlinSessionFull() throws Exception {
+        unzipAndRunExpect("single-vertex.expect.vm", getConfigPath(), getGraphName(), true, false);
```
farodin91, line 64: @li-boxuan This test wasn't activated before.

li-boxuan: I can see "testSimpleGremlinSessionFull" in the old tests. Am I missing something?

farodin91: I will rework the naming
I'm pretty sure the file single-vertex.expect was executed.

farodin91:
@li-boxuan I debug using the github actions.

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
```diff
@@ -44,111 +26,90 @@
...
-        Properties p = new Properties();
-        p.put("file.resource.loader.path", EXPECT_DIR);
-        Velocity.init(p);
-    }
+    protected abstract String getGraphName();
 
-    protected void testSimpleGremlinSession(String graphConfig, String graphToString, boolean full) throws Exception {
-        unzipAndRunExpect("single-vertex.expect.vm", graphConfig, graphToString, full, false);
+    @ParameterizedTest
+    @ArgumentsSource(EnvironmentProvider.class)
```
li-boxuan, line 57: Should this be tagged with "FULL_TESTS"?

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
li-boxuan: > I'm pretty sure the file single-vertex.expect was executed.

You mean "was executed" or "was not executed"? In the old tests, I see a few invocations of method `testSimpleGremlinSession` which calls `unzipAndRunExpect("single-vertex.expect.vm", graphConfig, graphToString, full, false);` My confusion is what exactly was missing before.

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/AbstractJanusGraphAssemblyIT.java
farodin91: I don't think so.

The tag full_tests should be just used in the provider to just run full tests under the condition that the tag full exists.

li-boxuan: If I am understanding correctly, according to the logic in EnvironmentProvider, the parameter `fullTest` will be always false unless it is tagged with `FULL_TESTS`. 

In the old code, however, I can see `single-vertex.expect.vm` is tested with `fullTest` being true & false.
```
    @Test
    public void testSimpleGremlinSession() throws Exception {
        testSimpleGremlinSession("conf/janusgraph-berkeleyje.properties", "berkeleyje", false);
    }

    @Test
    public void testSimpleGremlinSessionFull() throws Exception {
        testSimpleGremlinSession("conf/janusgraph-berkeleyje.properties", "berkeleyje", true);
    }
```

Thus, it seems to me that the new test is only covering the case when `fullTest == false`, while the old test covered both cases. Is this by intention?

farodin91: I add the handle to able to disable the full tests. It should execute same tests currently. I'll check it.

farodin91: I'll remove this logic with selecting tests and add it if need.

farodin91:
@li-boxuan I switched away from parametrized testing solution. Now, I';m able to filter all full tests by searching for `*Full`. We did something similar for integration tests.

li-boxuan:
> Now, I'm able to filter all full tests by searching for *Full. We did something similar for integration tests.

I don't quite understand how that helps/works. Can you elaborate?

farodin91:
> I don't quite understand how that helps/works. Can you elaborate?

I can now use exclude tests in surefire to exclude all test that are running against full zip instead of small step.

li-boxuan (approved):
LGTM 👍

Issue #2339, opened 2021-01-05
Allow IndexSelectionUtil.getMatchingIndexes to accept filter function to filter index types during a Set construction 


porunov:
As discussed here: https://github.com/JanusGraph/janusgraph/pull/2189#discussion_r551305978
We should refactor `IndexSelectionUtil.getMatchingIndexes` to accept filter function as a second argument.

porunov:
Fixed in #2346

PR #2346, opened 2021-01-05
Allow passing a filter function to getMatchingIndexes

rngcntr:
Solve #2339 

As suggested, I added the ability to pass a filter to `getMatchingIndexes`. I also kept the old signature as a wrapper around the new function which just returns the unfiltered result.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionUtil.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionUtil.java
@@ -15,11 +15,12 @@
 package org.janusgraph.graphdb.query.index;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Iterables;
+import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.janusgraph.core.JanusGraphElement;
 import org.janusgraph.core.RelationType;
 import org.janusgraph.graphdb.internal.InternalRelationType;
 import org.janusgraph.graphdb.internal.OrderList;
+import org.janusgraph.graphdb.query.condition.Condition;
 import org.janusgraph.graphdb.query.condition.ConditionUtil;
 import org.janusgraph.graphdb.query.condition.MultiCondition;
 import org.janusgraph.graphdb.query.condition.PredicateCondition;
@@ -31,6 +32,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Predicate;
 
 public class IndexSelectionUtil {
     public static boolean indexCoversOrder(MixedIndexType index, OrderList orders) {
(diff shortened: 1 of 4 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
```diff
@@ -106,15 +106,13 @@ protected boolean isValidStep(HasStep<?> step) {
...
-            .map(index -> (CompositeIndexType) index)
-            .filter(cIndex -> cIndex.getCardinality() == Cardinality.SINGLE)
-            .anyMatch(cIndex -> IndexSelectionUtil.isIndexSatisfiedByGivenKeys(cIndex, givenKeys));
+        // collect all matching unique indexes
+        final Set<IndexType> availableIndexes = IndexSelectionUtil.getMatchingIndexes(conditions,
+            indexType -> indexType.isCompositeIndex()
+                && ((CompositeIndexType) indexType).getCardinality() == Cardinality.SINGLE
+                && IndexSelectionUtil.isIndexSatisfiedByGivenKeys(indexType, givenKeys));
+
+        return !availableIndexes.isEmpty();
```
porunov, line 115: (nitpick)
We don't necessary need to collect all available indices here. All we need is just answer if there is any index. I.e. we don't actually need to always process all `conditions`. Instead it's enough to process up to the first index which matches out filter. After that we can immediately return `true`.

porunov (approved):
LGTM. Thank you @rngcntr !
One nitpick. Not necessary to resolve in this PR.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexHasUniquePropertyOptimizerStrategy.java
rngcntr: Yes that is correct. That was potentially even better in the previous version using Java streams and `anyMatch()`. In that implementation the processing can stop after the first match has been found.

porunov: Well, in the previous version you were still creating the full Set. So, no, it wasn't better in the previous version. It is better now actually. But again, I just pointed it because we may improve it even more (we don't need to create any `Set` here. We can just iterate through `conditions` till we find the first condition which meets our criteria)

porunov: Previously you had:
```
// This takes N time
final Set<IndexType> availableIndexes = IndexSelectionUtil.getMatchingIndexes(conditions);

// This takes N time in the worst scenario
return availableIndexes.stream().filter(IndexType::isCompositeIndex).map(index -> (CompositeIndexType) index).filter(cIndex -> cIndex.getCardinality() == Cardinality.SINGLE).anyMatch(cIndex -> IndexSelectionUtil.isIndexSatisfiedByGivenKeys(cIndex, givenKeys)); 
```
So, previously you had N iterations in the best scenario and N*2 iterations in the worst scenario.

Now you had N iterations in both best and worst scenario, which is better then the previous logic.
What I'm telling is that we can improve best case scenario here. We can reduce the amount of iteration in the best scenario to 1 instead of N. That say, the worst scenario will take N iterations still - which is fine.

Again, I'm not telling that we need to do it in this PR. I'm fine doing it in another PR or changing it in this PR.

rngcntr: The full set was created in the previous version but note every entry of the set was necessarily accessed to check the conditions. In this implementation now, every check is executed before adding the index candidates to the set. Thus, the set is more selective, but more checks are executed overall.

porunov: That's true. As I said, we don't need any `Set` here. We can simply implement loop iteration over `conditions` till the first matched condition (something like `anyMatch` in streams). Thus, our time and space will be more efficient.
I.e. we don't need to use `getMatchingIndexes` here at all

rngcntr: That being said, you are obviously right that in this case, the set can be omitted entirely. My initial idea was to externalize the set creation to a method which can then be used for both use cases. It's a matter of design whether or not we want to duplicate the logic of this method or if we sacrifice that bit of efficiency for maintainability. I have a slight preference for maintainability and more general purpose methods but that's not set in stone.

porunov: @rngcntr Makes sense. In such case, I think it makes sense to revert changes in this method back to not execute `IndexSelectionUtil.isIndexSatisfiedByGivenKeys(indexType, givenKeys)` for each condition

rngcntr: I'm just working on a draft for a method like `existsMatchingIndex` as you suggested earlier, just do see how much duplicate code it would cause. Stand by...

rngcntr: Alright, first draft is pushed. I further externalized some of the logic do avoid duplicate code.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/index/IndexSelectionUtil.java
```diff
@@ -41,22 +42,57 @@ public static boolean indexCoversOrder(MixedIndexType index, OrderList orders) {
...
+        return existsMatchingIndex(conditions, i -> true);
+    }
+
+    public static boolean existsMatchingIndex(MultiCondition<JanusGraphElement> conditions, Predicate<IndexType> filter) {
+        if (conditions == null || filter == null) {
+            return false;
+        }
+
+        Set<Condition<JanusGraphElement>> conditionSet = new HashSet<>();
+        ConditionUtil.traversal(conditions, conditionSet::add);
```
rngcntr, line 84: I'm not quite happy with this way of dealing with `MultiCondition`s. But on first glance, the simplest way of iterating `Condition`s is to aggregate them in a `Set` first.

porunov: I think we can omit processing unnecessary conditions and collect them into a set by using apache lang MutableBoolean instead. Something like:
```
public static boolean existsMatchingIndex(MultiCondition<JanusGraphElement> conditions, Predicate<IndexType> filter) {
    if (conditions == null || filter == null) {
        return false;
    }
    MutableBoolean exists = new MutableBoolean();
    ConditionUtil.traversal(conditions, condition -> {
        if(exists.isFalse()){
            for (IndexType indexCandidate : getKeyIndexesForCondition(condition)) {
                if (filter.test(indexCandidate)) {
                    exists.setTrue();
                    return true;
                }
            }
        }
        return true;
    });
    return exists.booleanValue();
}
```

rngcntr: I thought about that, but the `exists.isFalse()` condition didn't come to my mind. Like!

porunov (approved):
LGTM. Thank you @rngcntr !

Issue #2283, opened 2020-12-14
Indicate vertex-centric index used in profile step


li-boxuan:
Composite indexes and mixed indexes, if used, would show up in the profiler result. However, profile() step does not indicate if and which vertex-centric index is used.

See https://groups.google.com/g/janusgraph-users/c/SqjYfsEInrQ/m/RDWALyDMAgAJ

Related issue: https://github.com/JanusGraph/janusgraph/issues/898

cmilowka:
Thank you for considering this enhancement. 
Datastax, which is sharing a lot of the code with JG, is having something of this kind, referring to:
https://docs.datastax.com/en/dse/6.0/dse-dev/datastax_enterprise/graph/using/useIndexes.html
quote: _"If an index was not used, index-query would be missing from the profile output_" where index-query look like: 
```
**index-query** :
    \_indexType=Materialized
   \_statement=SELECT "personId" FROM "dse60"."person_p_byName" 
      WHERE "name" = ? LIMIT ?; with params (java. lang.String) Emeril Lagasse, (java.lang.Integer) 50000
```

We already have a list of conditions considered for backend-query, sample:
` \_condition=(m_at >= 1589592639360 AND m_to <= 9223372036854775807 AND type[v_StartTime_t])`
what we are probably missing is to have a confirmation about the index that was used to support that
`\_index=MY_INDEX (m_at >= 1589592639360 AND type[v_StartTime_t])`
so conditions for m_to need to be filtered in-memory, and it is why isFitted=false.
Regards, Christopher

li-boxuan:
@cmilowka Thanks for the info. How the new profile step will look like is in https://github.com/JanusGraph/janusgraph/pull/2289. You may take a look and see if it helps you.

I admit it is not as detailed & clear as the DSE Graph version. To achieve that there is more work to be done in the future.

cmilowka:
### Thanks you so much for the effort,... desperate to try that asap. I have a feeling that more questions will pop up, like _why indexes are not used_ ;-) Christopher

li-boxuan:
Fixed by https://github.com/JanusGraph/janusgraph/pull/2360

PR #2360, opened 2021-01-15
Indicate vertex-centric index usage in profiler result

### LINKED ISSUE #898: Add toString to KeySliceQuery for profile
Actually, when you use profile you do not know if your query match a vertex-centric indexes or not because.

> GraphStep(vertex,[3539008])                                            1           1           0.107     0.37
> JanusGraphVertexStep([rating.gt(3.0)])                                 6           6          29.252    99.63
>     \_condition=(rating > 3.0 AND type[battled])
>     \_isFitted=false
>     \_vertices=1
>     **\_query=org.janusgraph.diskstorage.keycolumnvalue.SliceQuery@81b60e15**
>     \_orders=[DESC(time)]
>     \_isOrdered=false
>   optimization                                                                                 0.274
>                                             >TOTAL                     -           -          29.359

li-boxuan:
Closes #898, #2283

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -46,7 +46,19 @@ import java.nio.file.Path;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.EnumMap;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -155,8 +167,6 @@ import org.janusgraph.graphdb.schema.VertexLabelDefinition;
 import org.janusgraph.graphdb.serializer.SpecialInt;
 import org.janusgraph.graphdb.serializer.SpecialIntSerializer;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
-import org.janusgraph.graphdb.types.CompositeIndexType;
-import org.janusgraph.graphdb.types.IndexType;
 import org.janusgraph.graphdb.types.StandardEdgeLabelMaker;
 import org.janusgraph.graphdb.types.StandardPropertyKeyMaker;
 import org.janusgraph.graphdb.types.system.BaseVertexLabel;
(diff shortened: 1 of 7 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -5685,6 +5694,36 @@ public void testNotHas() {
...
+        v.addEdge("friend", o, "time", 100);
+        v.addEdge("friend-no-index", o, "time", 100);
+        tx.commit();
+
+        // satisfied by a single vertex-centric query which is satisfied by one edge query
+        newTx();
+        Metrics mSingleLabel = tx.traversal().V(v).outE("friend").has("time", P.lt(10)).profile().next().getMetrics(1);
+        assertEquals("JanusGraphVertexStep([time.lt(10)])", mSingleLabel.getName());
+        assertTrue(mSingleLabel.getDuration(TimeUnit.MICROSECONDS) > 0);
+        Map<String, String> annotations = new HashMap() {{
```
farodin91, line 5715: mit: Wouldn't be better to ensure the order of annotations? I'm not sure if this is possible.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ElementQuery.java
```diff
@@ -25,7 +26,7 @@
  * @author Matthias Broecheler (me@matthiasb.com)
  */
 
-public interface ElementQuery<R extends JanusGraphElement,B extends BackendQuery<B>> extends Query {
```
farodin91, line 28: Question: what is the behavior of this change? I'm just curious.

farodin91 (approved):
LGTM!

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/ElementQuery.java
li-boxuan: There is no behavior change. `ElementQuery` has two concrete subclasses, `GraphCentricQuery` and `VertexCentricQuery`, both of which have implemented the `ProfileObservable` interface. Thus I made `ElementQuery` itself extend `ProfileObservable`. It's just a code cleanup.

Not sure but I think I made this change in https://github.com/JanusGraph/janusgraph/pull/2289 to avoid explicit casting.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
li-boxuan: It should be technically viable, although non-trivial to compare the order of LinkedHashMaps.
My concern is, https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/ImmutableMetrics.java#L45 provides annotations as a Map, so no order is guaranteed. Although its implementation relies on LinkedHashMap which preserves insertion order, we may not want to rely on it in our test.

farodin91: I don't we should rely on the internal implementation of tinkerpop, if they just return a hashmap. It don't it is critical, all of them are basically just arguments.

PR #2363, opened 2021-01-16
Refactor JanusGraphStep::buildGlobalGraphCentricQuery

li-boxuan:
Rewrite the method to reduce NPath complexity

This was part of https://github.com/JanusGraph/janusgraph/pull/2289 to satisfy Codacy.

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphStep.java
```diff
@@ -107,16 +107,15 @@ else if (this.ids.length > 0) {
     }
 
     private GraphCentricQuery buildGlobalGraphCentricQuery(final JanusGraphTransaction tx) {
-        //If a query have a local offset or have a local order without a global order and if a query have a limit lower than the global different from other query we can not build globalquery
```
li-boxuan, line 110: The comment here contradicts the logic. It should be

> If (a query has a local offset) or (has a local order without a global order) or (has a limit lower than the global and different from another query) then we can not build globalquery

Removing the comment since it does not add extra information but creates confusion.

farodin91 (approved):
LGTM! Thank you.

Issue #2285, opened 2020-12-17
Profiler annotations are wrong when there are multiple backend queries 


li-boxuan:
For confirmed bugs, please report:
- Version: 0.5.2, 0.6.0-SNAPSHOT
- Storage Backend: does not matter
- Mixed Index Backend: does not matter
- Steps to reproduce:
```groovy
graph = JanusGraphFactory.open("inmemory")
mgmt = graph.openManagement()
nameKey = mgmt.makePropertyKey('name').dataType(String.class).make()
ageKey = mgmt.makePropertyKey('age').dataType(Integer.class).make()
mgmt.buildIndex("name", Vertex.class).addKey(nameKey).buildCompositeIndex()
mgmt.buildIndex("age", Vertex.class).addKey(ageKey).buildCompositeIndex()
mgmt.commit()

g = graph.traversal()
g.addV().property("name", "bob").property("age", 60)
graph.tx().commit()
```

When there is more than one backend query, annotations (condition, orders, isFitted, isOrdered, query, index) of the last one overwrites previous annotations.

1. Multiple conditions under Or
```
g.V().or(has("name", "bob"), has("age", 60)).profile()

==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
=============================================================================================================
Or(JanusGraphStep([],[name.eq(bob)]),JanusGraph...                     1           1           5.852   100.00
    \_condition=(age = 60)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@2000
    \_index=age
  optimization                                                                                 0.621
  optimization                                                                                 0.175
  optimization                                                                                 0.129
  backend-query                                                        1                       0.730
    \_query=name:multiKSQ[1]@2000
    \_limit=2000
  backend-query                                                        1                       1.396
    \_query=age:multiKSQ[1]@2000
    \_limit=2000
                                            >TOTAL                     -           -           5.852        -
```
Note the following part   
```
   \_condition=(age = 60)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@2000
    \_index=age
```
only represents the second backend-query.

2. Multiple conditions under And
Same issue:
```
g.V().has("name", "bob").has("age", 60).profile()


==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
=============================================================================================================
[...]

PR #2365, opened 2021-01-17
Enhance GraphCentricQuery profiling

li-boxuan:
1. Indicate "constructGraphCentricQuery" phase in profiler rather than a vague
"optimization" annotation, if applicable.
2. Indicate "GraphCentricQuery" in profiler if applicable. Previously profiling result
is corrupted when a gremlin query is satisfied by multiple graph centric queries.

Note that this PR does not fix https://github.com/JanusGraph/janusgraph/issues/2285 (corruption due to flattening multiple AND/OR groups). That issue is trickier and will be addressed in the next PR.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -19,6 +19,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Sets;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.util.Metrics;
 import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalMetrics;
 import org.apache.tinkerpop.gremlin.structure.T;
 import org.janusgraph.TestCategory;
@@ -83,7 +84,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 import static org.janusgraph.graphdb.JanusGraphTest.evaluateQuery;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.*;
@@ -152,10 +152,13 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
 
     public abstract boolean supportsLuceneStyleQueries();
 
-
     public abstract boolean supportsWildcardQuery();
 
     public abstract boolean supportsGeoPointExistsQuery();
+
+    public String getStringField(String propertyKey) {
+        return propertyKey;
+    }
 
     @Override
     public void open(WriteConfiguration config) {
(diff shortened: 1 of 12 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -989,6 +988,76 @@ public void testConditionalIndexing() {
...
+            query = "[(name = bob AND prop = 100)](1000):mixed";
+        }
+        Map<String, String> annotations = new HashMap() {{
+            put("condition", "(name = bob AND prop = 100)");
+            put("orders", "[]");
+            put("isFitted", "true");
+            put("isOrdered", "true");
+            put("query", query);
+            put("index", "mixed");
+            put("index_impl", "search");
```
farodin91, line 1056: Do you like to test two different index_impl's at same time.

farodin91 (changes requested):
SGTM! Just one idea.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
li-boxuan: That sounds interesting. Do you know if we have any existing test case that involves more than one index backend?

farodin91: I have no idea. I can check for it.

farodin91: We don't have any for test cases for it, as far as i can see.

farodin91: https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchIndexTest.java#L409
 I found something.

li-boxuan: I tried it. Essentially, a gremlin query involving two index queries against two index backends is the same as a gremlin query involving two index queries against the same index backend, from the profiling perspective.

I think it would be good to have multiple index backends involved anyway... Will do that in the next PR (to fix AND/OR flattening issue). Right now the annotations are corrupted when there are multiple backend queries.

li-boxuan: Created https://github.com/JanusGraph/janusgraph/pull/2377 to support multiple index backends in JanusGraphIndexTest.

li-boxuan:
```groovy
graph = JanusGraphFactory.open("conf/janusgraph-cql.properties")
mgmt = graph.openManagement()
nameKey = mgmt.makePropertyKey('name').dataType(String.class).make()
weightKey = mgmt.makePropertyKey('weight').dataType(Integer.class).make()
mgmt.buildIndex("nameIdx", Vertex.class).addKey(nameKey).buildCompositeIndex()
mgmt.buildIndex("weightIdx", Vertex.class).addKey(weightKey).buildCompositeIndex()
mgmt.commit()

g = graph.traversal()
for (i = 0; i < 10000; i++) {
g.addV().property("name", "value" + i % 5).property("weight", i % 100).next()
}
graph.tx().commit()

g.V().or(has("name", "bob"), has("weight", 100)).profile()
```

With this PR:
```
==>Traversal Metrics
Or(JanusGraphStep([],[name.eq(bob)]),JanusGraph...                                             7.233   100.00
  constructGraphCentricQuery                                                                   0.810
  constructGraphCentricQuery                                                                   0.154
  GraphCentricQuery                                                                            2.745
    \_condition=(name = bob)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@1000
    \_index=nameIdx
    backend-query                                                      0                       1.307
    \_query=nameIdx:multiKSQ[1]@1000
    \_limit=1000
  constructGraphCentricQuery                                                                   0.129
  GraphCentricQuery                                                                            1.316
    \_condition=(weight = 100)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@1000
    \_index=weightIdx
    backend-query                                                      0                       0.717
    \_query=weightIdx:multiKSQ[1]@1000
    \_limit=1000
                                            >TOTAL                     -           -           7.233        -
```

Using 0.5.3:
```
==>Traversal Metrics
Or(JanusGraphStep([],[name.eq(bob)]),JanusGraph...                                             8.451   100.00
    \_condition=(weight = 100)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@2000
    \_index=weightIdx
  optimization                                                                                 0.889
  optimization                                                                                 0.191
  optimization                                                                                 0.104
  backend-query                                                        0                       1.242
    \_query=nameIdx:multiKSQ[1]@2000
    \_limit=2000
  backend-query                                                        0                       1.468
    \_query=weightIdx:multiKSQ[1]@2000
    \_limit=2000
                                            >TOTAL                     -           -           8.451        -
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphStep.java
```diff
@@ -293,5 +307,39 @@ public int hashCode() {
...
     }
+
+    private final class ProfiledIterator<E extends JanusGraphElement> extends CloseableAbstractIterator<E> {
+        private final QueryProfiler profiler;
+        private Iterator<E> iterator;
+        private boolean timerRunning;
+
+        private ProfiledIterator(QueryProfiler profiler) {
+            this.profiler = profiler;
+            profiler.startTimer();
```
rngcntr, line 318: Is there a reason the timer should start once the iterator is created? In your case, the measured time frame includes the `builder.iterables(...).iterator()` call in line 179. No matter if the time spent there should or should not be recorded, there is currently no mechanism in place to enforce it either way. Currently, everything that happens between the iterator construction and its use is included in the measurement, which does not intuitively make sense to me.

li-boxuan: Agree that it does not make much intuitive sense. The reason is, a couple of underlying iterators, e.g. SubqueryIterator, eagerly fetches results in their constructors. Thus, the backend query is already executed before the next() or hasNext() is called the first time.

rngcntr: I see. There are two ideas that come to my mind:

1. Start the timer for the `ProfiledIterator` explicitly by calling a method. This indicates more clearly that there is a reason to create the `ProfiledIterator` instance this early.
2. Delay the construction of the underlying iterator, e.g. by passing a `Supplier<Iterator<E>>` as a parameter to `setIterator()` (or the constructor). The supplier can then be evaluated on the first call of `next()` which is also when the timer will be started.

The second solution has the advantage that the timer is automatically started once the iterator is actually accessed for the first time. However, that would make calls of `setIterator()` look a little bit more complex:
```java
iterator.setIterator(() -> (Iterator<E>) builder.iterables(query, graphClass).iterator());
```

But of course, we can also move that to the constructor then:
```java
ProfiledIterator iterator = new ProfiledIterator(profiler, () -> (Iterator<E>) builder.iterables(query, graphClass).iterator());
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/ProfiledIterator.java
```diff
@@ -0,0 +1,56 @@
...
+public class ProfiledIterator<E extends JanusGraphElement> extends CloseableAbstractIterator<E> {
+    private final QueryProfiler profiler;
+    private Iterator<E> iterator;
+    private boolean timerRunning;
+
+    public ProfiledIterator(QueryProfiler profiler, Supplier<Iterator<E>> iteratorSupplier) {
+        this.profiler = profiler;
+        profiler.startTimer();
+        timerRunning = true;
+        iterator = iteratorSupplier.get();
```
rngcntr, line 36: (Nitpick) One more thing: I think you can move these three lines to `computeNext()`. In `computeNext()` we can then start the timer and construct the iterator if `timerRunning == false`.
By doing so, the timer is automatically started, completely independent of when the `ProfiledIterator` is created. One can then call the constructor at any time, but the measurement only starts once the iterator is actually used.

rngcntr (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/ProfiledIterator.java
li-boxuan: In fact, I think it actually makes some sense to start the timer in the construction phase, since it means the control flow has entered the scope of ProfiledIterator.

Moving it to computeNext() method means an additional instance variable and an extra if statement for every computeNext() call.

After all, there is very little profile timing & overhead difference between  both ways.

rngcntr: I agree that there is almost no difference in this case. I just thought about a case, where one would for example prepare multiple iterators and then use them one by one at some point in time later on. but maybe that use case is a little bit constructed and far from real world application.

li-boxuan: That actually sounds like a legitimate case. I have changed the logic.

li-boxuan: I had to revert this piece of logic, otherwise, JanusGraphIndexTest::testOrForceIndexPartialIndex would fail. The reason is when we put the "iteratorSupplier.get()" logic into the computeNext() method, the computation is lazy (delayed). In that test case, if the computation is delayed, there will be no exception for "g.V().or(__.has("name", "Totoro"),__.has("age", 2)).hasNext();" because the first condition uses index and returns a result (thus, the second condition is not yet evaluated). On the contrary, if the computation is not delayed, the second condition will throw an exception because it does not use an index, and force-index is enabled.

In fact, I think this actually demonstrates the benefit of lazy computation, but since this would change the behavior and can have a non-trivial impact, I would rather leave it as a follow-up issue (to discuss further whether we shall do so or not).

rngcntr: So if I understand correctly, using lazy computation would actually cause an even better usage of indexes? Or does it cause a problem (in the sense of incorrect results), that the second condition is not evaluated in this case?
If it really is _that easy_ to enable index usage for more expressions, we should definitely elaborate further in the future.
If you stick with this solution for now, you can remove the `Supplier` and pass the `Iterator` directly.

li-boxuan: > using lazy computation would actually cause an even better usage of indexes? Or does it cause a problem (in the sense of incorrect results), that the second condition is not evaluated in this case?

It does not lead to "better usage of indexes" per se. What happens is, `g.V().or(has("name", "Totoro"),has("age", 2)).hasNext();` consists of two graph-centric queries, each wrapped by a "ProfiledIterator". The two "ProfiledIterator"s are then combined into a "MultiDistinctUnorderedIterator". The first condition returns non-empty results, thus "hasNext()" returns true without the need of evaluating the second condition.

> does it cause a problem (in the sense of incorrect results)

Yes and no. It does change the current behaviour, but I tend to believe this is not incorrect. Let's use the same example ("name" is indexed, "age" is not indexed, “hasForceIndexUsage" is true). The current behaviour is, both `g.V().or(has("name", "Totoro"),has("age", 2)).hasNext();`  and `g.V().or(has("age", 2),has("name", "Totoro")).hasNext();` lead to exception because no index is used. If we do this "lazy computation", then `g.V().or(has("age", 2),has("name", "Totoro")).hasNext();` leads to exception but `g.V().or(has("name", "Totoro"),has("age", 2)).hasNext();` not.

rngcntr: > Nope. As we discussed earlier, to do that, I have to construct ProfiledIterator and start the timer first. 
Oh yes, my bad.

> If we do this "lazy computation", then g.V().or(has("age", 2),has("name", "Totoro")).hasNext(); leads to exception but g.V().or(has("name", "Totoro"),has("age", 2)).hasNext(); not.
Ok I understand. So while it sounds good to me in general to implement lazy computation, we would also have to ensure that the order of sub queries inside an `or` does not affect the outcome, i.e. both use the indexed part first.

I see that this needs a closer look at least and we should postpone it to another PR. Would you please create an issue for that?

li-boxuan: Created https://github.com/JanusGraph/janusgraph/issues/2407

Issue #1522, opened 2019-04-05
CacheVertex leads to inconsistent behavior


kptfh:
[CacheVertex](https://github.com/JanusGraph/janusgraph/blob/d12adfbf083f575fa48860daa37bfbd0e6095369/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/CacheVertex.java#L61) doesn't take into account 'Retriever<SliceQuery, EntryList> lookup' parameter while retrieving results from cache thus we get inconsistency.

li-boxuan:
@kptfh Is it possible for you to provide steps to reproduce?

kptfh:
I opened this issue for 0.3.x version cause there was failing test for Aerospike connector.
Now on 0.5.2 I can't find the corresponding test.
In general there is still the issue of wrong design.
To guarantee consistency you should clean cache on lookup change

PR #2368, opened 2021-01-18
EdgesRetriever introduced to fix issue with CacheVertex inconsistency

kptfh:
Fixes Inconsistency Issue in `CacheVertex`
https://github.com/JanusGraph/janusgraph/issues/1522
It's really dangerous as it was hard to find the root cause.
You can set breakpoint [in CacheVertex](https://github.com/kptfh/janusgraph/blob/feature/fix-cache-vertex/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/CacheVertex.java#L72) and run testVertexCentricQuery in debug and you fill face this issue.


li-boxuan:
Hi, @kptfh May I know why you closed this PR?

Issue #1527, opened 2019-04-10
'StandardScannerExecutor' relies on consistency of 'KeyIterator's


kptfh:
'StandardScannerExecutor' relies on consistency of 'KeyIterator's that run in parallel 'DataPuller's. I.e. expects the same order of keys in all KeyIterators. 
Didn't find any mention of this fact in documentation. Is it correct to put this restriction on 'KeyIterator'? 
Aerospike scan operation doesn't provide same order of keys between different calls and fails on 'OLAPTest'.

kptfh:
Just to be sure that this is root cause of failing OLAPTest tried to sort keys in [AerospikeKeyIterator](https://github.com/Playtika/aerospike-janusgraph-storage-backend/blob/develop/aerospike-storage-backend/src/main/java/com/playtika/janusgraph/aerospike/AerospikeKeyIterator.java) and get this test green.

PR #2383, opened 2021-01-22
KeySlicesIterator introduced

kptfh:
`KeySlicesIterator` introduced to fix issue with key inconsistency while running parallel scans in `StandartScannerExcecutor`

fixes https://github.com/JanusGraph/janusgraph/issues/1527

Without this fix Aerospike (that doesn't guaranty order of keys while running scan operation) can't utilize Janusgraph jobs 
(Reindex, GhostVertexRemover)

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyColumnValueStore.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyColumnValueStore.java
@@ -174,6 +174,19 @@ public interface KeyColumnValueStore {
     // like current getKeys if column-slice is such that it queries for vertex state property
 
     /**
+     * Returns a {@link KeySlicesIterator} over all keys in the store that have one or more columns matching the column-range. Calling {@link KeySlicesIterator#getEntries()}
+     * returns the map of all entries that match the column-range specified by the given queries.
+     * <p>
+     * This method is mandatory for stores which do not guaranty key orders while running parallel scans.
+     *
+     * @param queries
+     * @param txh
+     * @return
+     * @throws org.janusgraph.diskstorage.BackendException
+     */
+    KeySlicesIterator getKeys(MultiSlicesQuery queries, StoreTransaction txh) throws BackendException;
+
+    /**
      * Returns the name of this store. Each store has a unique name which is used to open it.
      *
      * @return store name
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeySlicesIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeySlicesIterator.java
@@ -19,27 +19,29 @@ import org.janusgraph.diskstorage.StaticBuffer;
 import org.janusgraph.diskstorage.util.RecordIterator;
 import org.janusgraph.graphdb.olap.VertexJobConverter;
 
+import java.util.Map;
+
 /**
- * @author Matthias Broecheler (me@matthiasb.com)
+ * @author Sergii Karpenko (sergiy.karpenko@gmail.com)
  */
 
-public interface KeyIterator extends RecordIterator<StaticBuffer> {
+public interface KeySlicesIterator extends RecordIterator<StaticBuffer> {
 
     /**
-     * Returns an iterator over all entries associated with the current
-     * key that match the column range specified in the query.
+     * Returns map of iterators over all entries associated with the current
+     * key that match the column range specified in the queries.
      * <p>
-     * Closing the returned sub-iterator has no effect on this iterator.
+     * Closing any of the returned sub-iterators has no effect on this iterator.
      *
      * Calling {@link #next()} might close previously returned RecordIterators
      * depending on the implementation, hence it is important to iterate over
      * (and close) the RecordIterator before calling {@link #next()} or {@link #hasNext()}.
      *
-     * Important! Entries should be sorted inside iterator.
+     * Important! Entries should be sorted inside iterators.
      * Otherwise {@link VertexJobConverter} will not work correctly
      *
      * @return
      */
-    RecordIterator<Entry> getEntries();
+    Map<SliceQuery, RecordIterator<Entry>> getEntries();
 
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/MultiSlicesQuery.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/MultiSlicesQuery.java
@@ -1,0 +1,51 @@
+// Copyright 2017 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.keycolumnvalue;
+
+import org.janusgraph.diskstorage.StaticBuffer;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Queries for a list of slices of data, each identified by a start points (inclusive) and end points (exclusive).
+ * Returns all {@link StaticBuffer}s that lie in this ranges up to the given limit.
+ * <p>
+ * If a MultiSlicesQuery is marked <i>static</i> it is expected that the result set does not change.
+ *
+ * @author Sergii Karpenko (sergiy.karpenko@gmail.com)
+ */
+
+public class MultiSlicesQuery {
+
+    private final List<SliceQuery> queries;
+
+    public MultiSlicesQuery(List<SliceQuery> queries) {
+        this.queries = queries;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        MultiSlicesQuery that = (MultiSlicesQuery) o;
+        return Objects.equals(queries, that.queries);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(queries);
+    }
+}
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java
@@ -40,6 +40,14 @@ public interface StoreFeatures {
     boolean hasUnorderedScan();
 
     /**
+     * Whether this storage backend supports a consistent key order among different scans.
+     * If it supports ordered scan, it must support consistent key scan.
+     * If it doesn't support ordered scan, it may or may not support consistent key scan.
+     * If the consistent scan is not supported, the backend should support {@link KeyColumnValueStore#getKeys(MultiSlicesQuery, StoreTransaction)}
+     */
+    boolean hasConsistentScan();
+
+    /**
      * Whether this storage backend supports global key scans via
      * {@link KeyColumnValueStore#getKeys(KeyRangeQuery, StoreTransaction)}.
      */
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KCVSProxy.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KCVSProxy.java
@@ -62,6 +62,11 @@ public class KCVSProxy implements KeyColumnValueStore {
     }
 
     @Override
+    public KeySlicesIterator getKeys(MultiSlicesQuery queries, StoreTransaction txh) throws BackendException {
+        return store.getKeys(queries, txh);
+    }
+
+    @Override
     public String getName() {
         return store.getName();
     }
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StandardStoreFeatures.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StandardStoreFeatures.java
@@ -22,6 +22,7 @@ import org.janusgraph.diskstorage.util.time.TimestampProviders;
  */
 public class StandardStoreFeatures implements StoreFeatures {
 
+    private final boolean consistentScan;
     private final boolean unorderedScan;
     private final boolean orderedScan;
     private final boolean multiQuery;
@@ -52,6 +53,11 @@ public class StandardStoreFeatures implements StoreFeatures {
     @Override
     public boolean hasUnorderedScan() {
         return unorderedScan;
+    }
+
+    @Override
+    public boolean hasConsistentScan() {
+        return consistentScan;
     }
 
     @Override
@@ -158,6 +164,7 @@ public class StandardStoreFeatures implements StoreFeatures {
      */
     public static class Builder {
 
+        private boolean consistentScan = true;
         private boolean unorderedScan;
         private boolean orderedScan;
         private boolean multiQuery;
@@ -190,6 +197,7 @@ public class StandardStoreFeatures implements StoreFeatures {
          * the supplied {@code template}.
          */
         public Builder(StoreFeatures template) {
+            consistentScan(template.hasConsistentScan());
             unorderedScan(template.hasUnorderedScan());
             orderedScan(template.hasOrderedScan());
             multiQuery(template.hasMultiQuery());
@@ -215,6 +223,11 @@ public class StandardStoreFeatures implements StoreFeatures {
 
         public Builder optimisticLocking(boolean b) {
             optimisticLocking = b;
+            return this;
+        }
+
+        public Builder consistentScan(boolean consistentScan) {
+            this.consistentScan = consistentScan;
             return this;
         }
 
@@ -324,7 +337,7 @@ public class StandardStoreFeatures implements StoreFeatures {
         }
 
         public StandardStoreFeatures build() {
-            return new StandardStoreFeatures(unorderedScan, orderedScan,
+            return new StandardStoreFeatures(consistentScan, unorderedScan, orderedScan,
                     multiQuery, locking, batchMutation, localKeyPartition,
                     keyOrdered, distributed, transactional, keyConsistent,
                     timestamps, preferredTimestamps, cellLevelTTL,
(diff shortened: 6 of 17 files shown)


janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/MultiThreadsScanner.java
```diff
@@ -0,0 +1,259 @@
...
+            queryResults.put(query,entries);
+        }
+        return new Row(key, queryResults);
+    }
+
+    @Override
+    void join() throws InterruptedException {
+        int i = 0;
+        for (DataPuller dataPuller : pullThreads) {
+            dataPuller.join(10);
```
porunov, line 153: (nitpick) I would recommend placing such constants on top of the class to have better observability of hard-coded constants. That said, I see that this part was already in the previous code, so not necessary to resolve.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/MultiThreadsScanner.java
```diff
@@ -0,0 +1,259 @@
...
+        super(store, rowQueue);
+        this.store = store;
+        this.storeFeatures = storeFeatures;
+        this.storeTx = storeTx;
+        this.queries = queries;
+        this.keyFilter = keyFilter;
+        this.graphConfiguration = graphConfiguration;
+
+        this.dataQueues = new ArrayList<>(queries.size());
+        this.pullThreads = new ArrayList<>(queries.size());
```
porunov, line 86: (nitpick) Previously it was an array and not a collection. Is there a reason why you changed it from non resizable array to resizable array list?

porunov (commented):
I see that current runs of `StandardScannerExecutor` isn't affected because all current backends will have `inconsistentScan = false` by default and thus, the current logic (`MultiThreadsScanner.java`) will be used. As I understand, in Aerospike storage backend `inconsistentScan` will be set to `true` and thus the new logic (`SingleThreadScanner.java`) will be used for Aerospike or other storage backends with no support of keys order while running parallel scans.
Thank you @kptfh for this feature and additional code refactor! The code looks good to me, I have just several very small nitpicks.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/MultiThreadsScanner.java
```diff
@@ -0,0 +1,259 @@
...
+
+        setUp(queries);
+    }
+
+    private void setUp(List<SliceQuery> queries) throws BackendException {
+        int numQueries = queries.size();
+
+        for (int pos = 0; pos < numQueries; pos++) {
+            pullThreads.add(addDataPuller(queries.get(pos), storeTx, pos));
+        }
```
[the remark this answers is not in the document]
kptfh: good catch. fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/MultiThreadsScanner.java
kptfh: reverted from List to array

porunov (approved):
LGTM. Thank you @kptfh !

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/Scanner.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static org.janusgraph.diskstorage.keycolumnvalue.scan.StandardScannerExecutor.Row;
+
+/**
+ * Produces data to {@link BlockingQueue<Row>}
+ * for each key in {@link KeyColumnValueStore}
+ *
+ * @author Sergii Karpenko (sergiy.karpenko@gmail.com)
+ */
+abstract class Scanner {
```
li-boxuan, line 31: Is it possible to rename it to something else? Its name gives me an illusion that StandardScanner should be its subclass, which is not the case.

kptfh: Do you mean StandardScannerExecutor when saying StandardScanner ?

kptfh: From perspective all looks logical StandardScannerExecutor executes Scanner

li-boxuan: Nope, I mean org.janusgraph.diskstorage.keycolumnvalue.scan.StandardScanner

kptfh: I'd rather rename `StandardScanner` to `StandardScanManager` or `StandardScanProcessor` because it's main responsibility is to mange scan jobs rather then run scans 
Other option is to rename `Scanner` to `Collector`

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyIterator.java
```diff
@@ -34,6 +35,9 @@
      * depending on the implementation, hence it is important to iterate over
      * (and close) the RecordIterator before calling {@link #next()} or {@link #hasNext()}.
      *
+     * Important! Entries should be sorted inside iterator.
```
li-boxuan, line 38: Isn't the iterator itself ought to be sorted? From my understanding, the entries returned by the "getEntries()" don't have to be sorted.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/SingleThreadScanner.java
```diff
@@ -0,0 +1,121 @@
...
+ *  Uses one thread for all queries. May be used for {@link KeyColumnValueStore}
+ *  that do not guarantee keys order between different scans (f.e. Aerospike)
+ * @author Sergii Karpenko (sergiy.karpenko@gmail.com)
+ */
+
+class SingleThreadScanner extends Scanner{
+
+    private static final Logger log = LoggerFactory.getLogger(SingleThreadScanner.class);
+
+    private final KeyColumnValueStore store;
```
li-boxuan, line 51: Seems you don't need this since "KeyColumnValueStore store" is declared in its superclass already.

li-boxuan (commented):
Looks good to me in general! Just have a few suggestions on comments.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java
```diff
@@ -39,6 +39,12 @@
      */
     boolean hasUnorderedScan();
 
+    /**
+     * Whether this storage backend supports keys order while running parallel scans
```
li-boxuan, line 43: "Whether this storage backend supports keys order while running parallel scans" seems to suggest when `hasInconsistentScan()` is true, key order is consistent while running parallel scans.

My suggestion is, we could make this method "hasConsistentScan()". Then the doc could be,

> Whether this storage backend supports a consistent key order among different scans. If it supports ordered scan, it must support consistent key scan. If it doesn't support ordered scan, it may or may not support consistent key scan. If the consistent scan is not supported, the backend shall support {@link KeyColumnValueStore#getKeys(MultiSlicesQuery, StoreTransaction)}

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyIterator.java
```diff
@@ -34,6 +35,9 @@
...
+     * Otherwise {@link VertexJobConverter} will not work correctly
```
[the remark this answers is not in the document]
kptfh: It's critical for `VertexJobConverter.isGhostVertex`. This method checks only first entry of each row
```
protected boolean isGhostVertex(long vertexId, EntryList firstEntries) {
        if (idManager.isPartitionedVertex(vertexId) && !idManager.isCanonicalVertexId(vertexId)) return false;

        RelationCache relCache = tx.getEdgeSerializer().parseRelation(
                firstEntries.get(0),true,tx);
        return relCache.typeId != BaseKey.VertexExists.longId();
    }
```

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyIterator.java
kptfh: It's critical for `VertexJobConverter.isGhostVertex`. This method checks only first entry of each row
```
protected boolean isGhostVertex(long vertexId, EntryList firstEntries) {
        if (idManager.isPartitionedVertex(vertexId) && !idManager.isCanonicalVertexId(vertexId)) return false;

        RelationCache relCache = tx.getEdgeSerializer().parseRelation(
                firstEntries.get(0),true,tx);
        return relCache.typeId != BaseKey.VertexExists.longId();
    }
```

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java
kptfh: Sounds perfect!

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/MultiSlicesQuery.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.janusgraph.diskstorage.keycolumnvalue;
+
+import org.janusgraph.diskstorage.StaticBuffer;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Queries for a slices of data identified by a start points (inclusive) and end points (exclusive).
```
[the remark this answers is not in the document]
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/scan/SingleThreadScanner.java
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeySlicesIterator.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+
+/**
+ * @author Sergii Karpenko (sergiy.karpenko@gmail.com)
+ */
+
+public interface KeySlicesIterator extends RecordIterator<StaticBuffer> {
+
+    /**
+     * Returns an iterator over all entries associated with the current
```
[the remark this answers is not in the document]
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyColumnValueStore.java
```diff
@@ -173,6 +173,19 @@
     KeyIterator getKeys(SliceQuery query, StoreTransaction txh) throws BackendException;
     // like current getKeys if column-slice is such that it queries for vertex state property
 
+    /**
+     * Returns a {@link KeySlicesIterator} over all keys in the store that have one or more columns matching the column-range. Calling {@link KeyIterator#getEntries()}
+     * returns the list of all entries that match the column-range specified by the given queries.
+     * <p>
+     * This method is only supported by stores which do not guaranty key orders while running parallel scans.
```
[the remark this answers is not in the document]
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyColumnValueStore.java
```diff
@@ -173,6 +173,19 @@
     KeyIterator getKeys(SliceQuery query, StoreTransaction txh) throws BackendException;
     // like current getKeys if column-slice is such that it queries for vertex state property
 
+    /**
+     * Returns a {@link KeySlicesIterator} over all keys in the store that have one or more columns matching the column-range. Calling {@link KeyIterator#getEntries()}
+     * returns the list of all entries that match the column-range specified by the given queries.
```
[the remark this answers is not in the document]
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyColumnValueStore.java
```diff
@@ -173,6 +173,19 @@
     KeyIterator getKeys(SliceQuery query, StoreTransaction txh) throws BackendException;
     // like current getKeys if column-slice is such that it queries for vertex state property
 
+    /**
+     * Returns a {@link KeySlicesIterator} over all keys in the store that have one or more columns matching the column-range. Calling {@link KeyIterator#getEntries()}
```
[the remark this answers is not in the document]
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java
kptfh: fixed

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/KeyIterator.java
li-boxuan: Sorry I was wrong by saying "entries returned by the "getEntries()" don't have to be sorted". In fact, it has to be sorted because of the [JanusGraph Data Model](https://docs.janusgraph.org/advanced-topics/data-model/): "The cells must be sorted by their columns and a subset of the cells specified by a column range must be efficiently retrievable". Therefore this notice seems a bit redundant to me.

kptfh: For me it was not obvious. Spent a decent amount of time trying to figure out why test is not working. And the root cause was wrong order of entries in row. If you insist I may remove this note. At least I've fixed my test.

kptfh: I'd rather retain comment and add link that you provided

li-boxuan: I think I missed something earlier. When I saw this PR, I presumed your data backend does sort columns natively but does not sort keys. That made perfect sense to me because apparently, the StandardScannerExcecutor assumed key orders are "consistent" (either sorted by value like ByteOrderedPartitioner in Cassandra or sorted by key's hash value like Murmur3Partitioner in Cassandra) among multiple scans.

Now, seems your data backend neither supports key order nor supports column order. In other words, rows are not sorted in any way (neither globally nor at partition level). Then it makes perfect sense why you add this comment because your data backend actually does not support that natively. Am I understanding it correctly? If yes, then it would be very interesting to me, because AFAIK some features actually rely on that assumption (columns are sorted) - that is a bit off-topic, though.

kptfh: Exactly. Need to manually sort column values before iterating on them

kptfh: Actually Aerospike is a KeyValue store but it allows to emulate KeyColumnValue store by using Map structure in place of value.

BTW, after this fix (KeySlicesIterator introduced) all Aerospike tests become green.

li-boxuan: Got it, thanks. Then it would be good to keep this comment. I'll just do another pass before approving this PR.

At the same time, I am just curious: If columns are unsorted, vertex-centric indexes will be useless when it comes to range queries (e.g. `g.V(h).outE('battled').has('time', inside(10, 20)).inV()` as shown in https://docs.janusgraph.org/index-management/index-performance/#vertex-centric-indexes) because this feature relies on the columns to be sorted. Have you used vertex-centric indexes and if so, what performance have you observed? I am asking because if not supporting column order actually does not affect normal use (despite might be inefficient in some scenarios), we should update the data model documentation - currently, it says "The cells must be sorted by their columns".

kptfh: Aerospike connector supports all types of indexes (vertex, edge) and we used this feature on several production services.
The main reason is that Aerospike implementation of this method `KeyColumnValueStore.getSlice(KeySliceQuery query, StoreTransaction txh)` returns sorted entries.
WDYT about put this comment regarding column order on EntryList interface also?

li-boxuan: > The main reason is that Aerospike implementation of this method KeyColumnValueStore.getSlice(KeySliceQuery query, StoreTransaction txh) returns sorted entries.

Got it. I can see the usage of MapOperation::getByKeyRange which supports range query. Earlier I thought you had to retrieve the whole Map to do in-memory filtering for "KeyColumnValueStore.getSlice(KeySliceQuery query, StoreTransaction txh)" calls.

> WDYT about put this comment regarding column order on EntryList interface also

At this moment, I am not 100% sure whether "EntryList getSlice(KeySliceQuery query, StoreTransaction txh)" requires the ordering of columns. Column ordering has been a fundamental requirement but now it looks to me that this is not a hard requirement. I would suggest not doing this right now until we have experiment results showing that the EntryList interface must be sorted by column.

kptfh: The final proposal
WDYT: RecordsCollector  -> RowsCollector ?
It actually collects rows to queue.

kptfh: done

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/StoreFeatures.java
```diff
@@ -39,6 +39,15 @@
      */
     boolean hasUnorderedScan();
 
+    /**
+     * Whether this storage backend supports a consistent key order among different scans.
+     * If it supports ordered scan, it must support consistent key scan.
+     * If it doesn't support ordered scan, it may or may not support consistent key scan.
+     * If the consistent scan is not supported, the backend shall support {@link KeyColumnValueStore#getKeys(MultiSlicesQuery, StoreTransaction)}
+     * {@link KeyColumnValueStore#getKeys(MultiSlicesQuery, StoreTransaction)}.
```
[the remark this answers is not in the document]
kptfh: removed

li-boxuan (approved):
LGTM, thank you!

janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/MultiSlicesQuery.java
```diff
@@ -0,0 +1,51 @@
...
+ * Returns all {@link StaticBuffer}s that lie in this ranges up to the given limit.
+ * <p>
+ * If a MultiSlicesQuery is marked <i>static</i> it is expected that the result set does not change.
+ *
+ * @author Sergii Karpenko (sergiy.karpenko@gmail.com)
+ */
+
+public class MultiSlicesQuery {
+
+    private final List<SliceQuery> queries;
```
martin-g, line 33: Hi!
I am having trouble implementing `org.janusgraph.diskstorage.keycolumnvalue.KeyColumnValueStore#getKeys(org.janusgraph.diskstorage.keycolumnvalue.MultiSlicesQuery, org.janusgraph.diskstorage.keycolumnvalue.StoreTransaction)` 
`MultiSlicesQuery` encapsulates completely its `queries` without any way to access them.
How is this supposed to be used ?

kptfh: Definitely need to add getter ))

martin-g: Thanks, @kptfh !

kptfh: What storage do you want to use?

martin-g: I wanted to update Janus from 0.5.3 to 0.6.0 in Apache Atlas and this class showed the problem: https://github.com/apache/atlas/blob/926fb297c13574e761db4798763780715af2477e/graphdb/janus-hbase2/src/main/java/org/janusgraph/diskstorage/hbase2/HBaseKeyColumnValueStore.java

kptfh: Hmm...
It shouldn't affect you. 
What the problem do you have?

martin-g: The only possible way to implement the new method is:
```
 @Override
    public KeySlicesIterator getKeys(MultiSlicesQuery multiSlicesQuery, StoreTransaction storeTransaction) throws BackendException {
        throw new UnsupportedOperationException("No way to use multiSlicesQuery");
    }
```

PR #2413, opened 2021-01-31
Release memory when releasing transaction

### LINKED ISSUE #2412: Transaction index cache memory leak
StandardJanusGraphTx::releaseTransaction method only closes `vertexCache` but not `indexCache`:
https://github.com/JanusGraph/janusgraph/blob/0aaf48c34b7854be964155caefd4cfcda5ba3b1e/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java#L1480-L1485 

Theoretically, this memory leak might not be observable because usually after transactions are done, the transaction handles will be garbage-collected sooner or later. However, if transaction handles are not available for GC for whatever reason, a memory leak could happen. Not sure if this should be called a bug or not. Anyway, considering `indexCache` could take a good amount of memory, it would be good to make it available for GC as early as possible.

Using the below code snippet and VisualVM, we could observe there is `indexCache` leak after transactions are committed/rollbacked, provided that transactions are not available for GC yet.

```groovy
graph = JanusGraphFactory.open("conf/janusgraph-cql.properties")

mgmt = graph.openManagement()
nameKey = mgmt.makePropertyKey('name').dataType(String.class).make()
mgmt.buildIndex("name", Vertex.class).addKey(nameKey).buildCompositeIndex()
mgmt.commit()

g = graph.traversal()
for (k = 0; k < 10; k++) {
for (i = 0; i < 10000; i++) {
g.addV().property("name", "value").next()
}
graph.tx().commit()   
}

tx1 = graph.newTransaction()
tx1.traversal().V().has("name","value").profile()
```

// total heap usage size (after manually triggering GC): 79224KB, GuavaSubqueryCache takes 0.9KB retained size
image

```groovy
tx1.rollback()
```

// total: 60710KB, GuavaSubqueryCache takes 3257KB
image

```groovy
tx2 = graph.newTransaction()
tx2.traversal().V().has("name","value").profile()
tx2.commit()
```

// total: 60808KB, 2 GuavaSubqueryCache instances
image

```groovy
txs = []
for (i = 0; i < 10000; i++) {
tx = graph.newTransaction()
tx.traversal().V().has("name","value").count()
tx.rollback()
txs.add(tx)
}
```

// total: 105373KB, 10k instances of GuavaSubqueryCache
image

```groovy
txs = [] // clear references to transactions
```

// total: 60663KB. GuavaSubqueryCache instances drop to 2.

li-boxuan:
Fixes #2412

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -1478,10 +1478,19 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
     }
 
     private void releaseTransaction() {
-        //TODO: release non crucial data structures to preserve memory?
         isOpen = false;
         graph.closeTransaction(this);
         vertexCache.close();
+        indexCache.close();
+        addedRelations.clear();
+        if (!Objects.equals(EMPTY_DELETED_RELATIONS, deletedRelations)) {
+            deletedRelations.clear();
+        }
+        if (!Objects.equals(UNINITIALIZED_LOCKS, uniqueLocks)) {
+            uniqueLocks.clear();
+        }
+        newVertexIndexEntries.close();
+        newTypeCache.clear();
     }
 
     @Override
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -2410,6 +2410,17 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
     }
 
     @Test
+    public void testTransactionScopeTransition() {
+        tx.rollback();
+
+        JanusGraphVertex newV = graph.addVertex();
+        graph.tx().commit();
+
+        newV.property("prop", "value");
+        graph.tx().commit();
+    }
+
+    @Test
     public void testNestedTransactions() {
         Vertex v1 = graph.addVertex();
         newTx();
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/AddedRelationsContainer.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/addedrelations/AddedRelationsContainer.java
@@ -41,6 +41,11 @@ public interface AddedRelationsContainer {
      */
     Collection<InternalRelation> getAll();
 
+    /**
+     * Clears the container which releases allocated memory.
+     * Calling any of the other methods after clearing the container has undetermined behavior.
+     */
+    void clear();
 
     AddedRelationsContainer EMPTY = new AddedRelationsContainer() {
         @Override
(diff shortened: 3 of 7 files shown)


li-boxuan:
Now, after rolling back a transaction, the retained size is almost 0 (rather than 3MB as in https://github.com/JanusGraph/janusgraph/issues/2412#issue-797725868):

image

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1482,6 +1482,7 @@ private void releaseTransaction() {
         isOpen = false;
         graph.closeTransaction(this);
         vertexCache.close();
+        indexCache.close();
```
mad, line 1485: Next data structures may explicitly cleared too 

```
        if (deletedRelations != EMPTY_DELETED_RELATIONS) {
            deletedRelations.clear();
        }
        if (uniqueLocks != UNINITIALIZED_LOCKS){
            uniqueLocks.clear();
        }
        addedRelations.clear();
```

mad (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1478,10 +1478,19 @@ public synchronized void rollback() {
     }
 
     private void releaseTransaction() {
-        //TODO: release non crucial data structures to preserve memory?
         isOpen = false;
         graph.closeTransaction(this);
         vertexCache.close();
+        indexCache.close();
+        addedRelations.clear();
```
porunov, line 1485: What is the complexity of 'clear' method?
If it is O(n) then I would rather don't call 'clear' at all because it may slow down transactions.
Do we reuse addedRelations after we clear it? If no, there is definitely no point for additional computation.
'addedRelations = null' will be much faster in such case and will release even more resources (no need to hold an empty  array in-memory).

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
porunov: @li-boxuan @mad 
See my comment here: https://github.com/JanusGraph/janusgraph/pull/2413/files#r571375536

porunov: Can't we just change that to?:
addedRelations = null;
deletedRelations = null;
uniqueLocks = null;

Also, I didn't check why we need to clear those data structures, it's just quick observations. My point is that some implementations of 'clear' method are really inefficient. Again, didn't check the complexity of the above clear methods

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
li-boxuan: Yeah, calling close/clear does have some overhead, although relatively very small compared to other computations in the transaction. Simply assigning null sounds like a good plan! Let me open another PR to fix this.

li-boxuan: Created https://github.com/JanusGraph/janusgraph/pull/2457

PR #2446, opened 2021-02-08
Issue 2445: Expose ES rest client connect and socket timeout config options

twilmes:
Addresses: Issue 2445

Signed-off-by: Ted Wilmes <ted.wilmes@experoinc.com>

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -21,6 +21,7 @@ import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.IN
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.*;
+import org.elasticsearch.client.RestClientBuilder;
 import org.janusgraph.diskstorage.es.compat.ESCompatUtils;
 import org.janusgraph.diskstorage.es.mapping.IndexMapping;
 import org.janusgraph.diskstorage.es.rest.util.HttpAuthTypes;
@@ -266,6 +267,16 @@ public class ElasticSearchIndex implements IndexProvider {
             "Enables cache for generated index store names. " +
                 "It is recommended to always enable index store names cache unless you have more then 50000 indexes " +
                 "per index store.", ConfigOption.Type.MASKABLE, true);
+
+    public static final ConfigOption<Integer> CONNECT_TIMEOUT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "connect-timeout",
+            "Sets the maximum connection timeout (in milliseconds).", ConfigOption.Type.MASKABLE,
+            Integer.class, RestClientBuilder.DEFAULT_CONNECT_TIMEOUT_MILLIS);
+
+    public static final ConfigOption<Integer> SOCKET_TIMEOUT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "socket-timeout",
+            "Sets the maximum socket timeout (in milliseconds).", ConfigOption.Type.MASKABLE,
+            Integer.class, RestClientBuilder.DEFAULT_SOCKET_TIMEOUT_MILLIS);
 
     public static final int HOST_PORT_DEFAULT = 9200;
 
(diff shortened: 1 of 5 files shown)


janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -267,6 +267,16 @@
                 "It is recommended to always enable index store names cache unless you have more then 50000 indexes " +
                 "per index store.", ConfigOption.Type.MASKABLE, true);
 
+    public static final ConfigOption<Integer> CONNECT_TIMEOUT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "connect-timeout",
+            "Sets the maximum connection timeout (in milliseconds).", ConfigOption.Type.MASKABLE,
+            Integer.class, 1000);
```
li-boxuan, line 273: Why 1000? The default value for Apache HTTP client seems to be -1.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -267,6 +267,16 @@
...
+
+    public static final ConfigOption<Integer> SOCKET_TIMEOUT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "socket-timeout",
+            "Sets the maximum socket timeout (in milliseconds).", ConfigOption.Type.MASKABLE,
+            Integer.class, 30000);
```
li-boxuan, line 278: Same here. The default value for the Apache HTTP client seems to be -1.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
```diff
@@ -122,7 +123,26 @@ protected RestElasticSearchClient getElasticSearchClient(RestClient rc, int scro
...
+
+        final List<RequestConfigCallback> callbackList = new LinkedList<>();
+
+        final Integer connectTimeout = config.get(ElasticSearchIndex.CONNECT_TIMEOUT);
+        final Integer socketTimeout = config.get(ElasticSearchIndex.SOCKET_TIMEOUT);
+
+        callbackList.add((requestConfigBuilder) ->
+            requestConfigBuilder.setConnectTimeout(connectTimeout).setSocketTimeout(socketTimeout));
+
+        if (callbackList.isEmpty()) {
```
li-boxuan, line 135: Seems this condition always evaluates to false.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
twilmes: @li-boxuan thanks for the review. I'm not as familiar with the more recent ES updates but isn't the ES rest client being used as opposed to Apache HTTP? If so, I think these defaults line up with what is [documented](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_timeouts.html).

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestClientSetup.java
twilmes: Yes, good catch, bad copy-paste...

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
li-boxuan: Thanks, you are right. ES rest-client internally uses Apache HTTP client (we are using 4.x HTTP client) which [uses -1 as the default value ](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/config/RequestConfig.java#L280-L292)(which is interpreted as an infinite timeout), but [ES overrides these values](https://github.com/elastic/elasticsearch/blob/master/client/rest/src/main/java/org/elasticsearch/client/RestClientBuilder.java#L50-L51). So the default values indeed are 1000, but shall we use `RestClientBuilder.DEFAULT_CONNECT_TIMEOUT_MILLIS`?

twilmes: Yes, agreed, changed to the constant values.

li-boxuan (approved):
LGTM 👍

PR #2468, opened 2021-02-20
Increase codecov after_n_builds and fix notify config

li-boxuan:
1. Notify config should be under codecov group, not coverage group
2. Remove codecov.notify.after_n_builds limit so that code coverage
is always available as a GitHub check, no matter how many codecov
reports are uploaded
3. Reduce comment.after_n_builds from 30 to 26 to reflect exact
number of codecov reports
4. Name codecov reports

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [ ] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

PR #2469, opened 2021-02-20
Fix codecov configs

li-boxuan:
1. Notify config should be under codecov group, not coverage group
2. Remove codecov.notify.after_n_builds limit so that code coverage
is always available as a GitHub check, no matter how many codecov
reports are uploaded
3. Disable pull request comment feature since it is annoying and
unstable
4. Name codecov reports

-----

### For all changes:

### For code changes:

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -12,18 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-coverage:
+codecov:
   notify:
-    after_n_builds: 30
+    wait_for_ci: yes
+
+coverage:
   status:
-    patch: off
+    patch:
+      default:
+        threshold: 5
     project:
       default:
         # allow test coverage to drop by 2%, assume that it's typically due to CI problems
         threshold: 2
 
-comment:
-  after_n_builds: 30
+comment: false
 
 # Ignore test files
 ignore:
--- a/.github/workflows/ci-backend-cql.yml
+++ b/.github/workflows/ci-backend-cql.yml
@@ -62,24 +62,35 @@ jobs:
         include:
           - module: cql
             args: "-Pcassandra3-byteordered -Dtest=\"**/diskstorage/cql/*\""
+            name: byteordered-diskstorage
           - module: cql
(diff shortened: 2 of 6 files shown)


li-boxuan:
I am pretty sure the old yaml file was invalid, because running `cat .codecov.yml | curl --data-binary @- https://codecov.io/validate` returns following result:

```
Path: coverage->notify
Wrong key 'after_n_builds' in {'after_n_builds': 30}
```

With the new changes, the API now returns "Valid!" response.

However, 
1) I don't know why codecov does not show up as a GitHub check. I suspect this needs some config changes for GitHub Actions.
2) the number of reports seems unstable. Sometimes it generates more than 30, sometimes fewer than 26. If any GitHub check is triggered again, more report(s) will be uploaded to codecov. As a consequence, sometimes we don't see codecov comment initially, but after a force-push it appears, e.g. my previous PR, https://github.com/JanusGraph/janusgraph/pull/2468

farodin91:
@li-boxuan I would deactivate the comment feature.

li-boxuan:
> I would deactivate the comment feature.

Agree, that one is quite annoying. Done.

.codecov.yml
```diff
@@ -12,18 +12,19 @@
...
 
-coverage:
+codecov:
   notify:
-    after_n_builds: 30
+    wait_for_ci: yes
+
+coverage:
   status:
     patch: off
```
farodin91, line 19: Do it have to patch the project?


It didn't updated the GitHub check.

li-boxuan: Can you elaborate? I didn't get it.

farodin91: It seems that codecov updates the github check only once.

li-boxuan: I think that has nothing to do with the patch option... but anyway let me try turning it on and observe.

farodin91: Do you have any idea? Same behavior, but patch is green

li-boxuan: I searched around but didn't find anyone else reporting this problem. I am trying out a few settings at the moment. In the end, at least we can set the after_n_builds option back.

farodin91 (approved):
The result looks good now.

Issue #1278, opened 2018-10-02
Memory Leak in JanusGraph Framework


shubham190892:
I am using janusgraph for 6 months. Now getting OOM after 4-5 days and after restarting the service It runs fine for 4-5 days. I created the heap dump and generated the analysis chart. I am not able to figure out what is the cause of OOM. Please find the screen shot and help.

<img width="1280" alt="gh-1" src="https://user-images.githubusercontent.com/6499319/46345262-8c228500-c661-11e8-8dd7-ad50b7f9fe2c.png">
<img width="1280" alt="gh-2" src="https://user-images.githubusercontent.com/6499319/46345271-93e22980-c661-11e8-8d66-5c09e95a1783.png">

doofin:
can you provide more info ? like which backend,etc

shubham190892:
I am using java 8 , including janusgraph 0.2.0 as dependecy and 3 node cassandra cluster for backend storage. Please let me know If anything specific you want to know.

doofin:
My setting is janusgraph 0.3.1 with cassandra ,have you tried the new version?

makpcheng:
I am also getting a memory leak. When I analyze the heap dump, I see an abundance of CacheVertex objects not being cleaned up.

This can easily be reproduced by exporting a large database (exporting graphml). Memory heap continues to grow without being cleaned up after export is completed.

marco-scoppetta:
Hey @makpcheng, we are having the same issue. When opening a graph and trying to export data from it, we can see the memory heap growing - eventually leading to OutOfMemory caused by GC not being able to free up any more memory. 

Have you managed to find a workaround for this? Any hints?

porunov:
I am not sure but this could be connected to #1401 which is fixed in `0.3.2` version of JanusGraph. Are you able to reproduce this issue with the `0.3.2` version?

marco-scoppetta:
@porunov I just tried, it looks a bit better, but still leaking. I will try to dig a bit deeper, please let me know if anyone has any more advice.

marco-scoppetta:
I find this comment to be potentially interesting:

`//TODO: release non crucial data structures to preserve memory?`

at this line when closing a Transaction:

https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java#L1421

I will keep searching, but as always if anyone has any idea please let me know

[...]

PR #2472, opened 2021-02-21
Fix potential ThreadLocal transaction leak [tp-tests]

li-boxuan:
This is a follow-up of #1653, which fixes some ThreadLocal transaction leak
but relies on users explicitly calling graph.tx().close(). This commit fixes
this flaw and now graph.tx().commit() or graph.tx().rollback() should close
all ThreadLocal objects used in JanusGraphBlueprintsGraph and AbstractThreadLocalTransaction.

However, note that we still have ThreadLocal leakage in case the user closes
the graph without committing/rolling back/closing ThreadLocal transactions.
Unfortunately, I don't have an appropriate solution for this at the moment.

Partial fix for #1278

-----

### For all changes:

### For code changes:


li-boxuan:
Using the sample test https://gist.github.com/mad/4230d56d73587c11dcf48eb9463ce751 provided by @mad in https://github.com/JanusGraph/janusgraph/pull/1653#issuecomment-508139294, but with a small change (remove L54 "graph.tx().close()"), we can see ThreadLocal leaks using the master version:

image


After this PR, no tx leakage:
image

However, note that we still have ThreadLocal leakage in case the user closes the graph without committing/rolling back/closing ThreadLocal transactions. Unfortunately, I don't have an appropriate solution for this.

janusgraph-core/src/main/java/org/janusgraph/core/JanusGraph.java
```diff
@@ -57,6 +57,11 @@
         reason = "JanusGraph test graph computer (FulgoraGraphComputer) " +
             "currently does not support graph filters but does not throw proper exception because doing so breaks numerous " +
             "tests in gremlin-test ProcessComputerSuite.")
+@Graph.OptOut(
```
mad, line 60: New behavior may fail users code/tests, who reuse transaction listeners within auto tx, so this is breaking change, may be add some notes to upgrade/release notes?

li-boxuan: That is a good point. Now I am wondering if we should introduce this breaking change, or leave this memory leak as it is (provided there is no other way to fix the leak caused by transactionListeners).

Alternatively, we can let users know that after a transaction is done, they will have to call `graph.tx().addTransactionListener(listener);` explicitly if they want to reuse the listener in the new transaction.

analytically (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphBlueprintsGraph.java
```diff
@@ -287,6 +289,18 @@ public GraphTransaction() {
             super(JanusGraphBlueprintsGraph.this);
         }
 
+        @Override
+        public void commit() {
+            super.commit();
+            doClose();
```
li-boxuan, line 295: This should be placed in a try - finally block to ensure threadlocals are cleaned even if commit/rollback fails.

li-boxuan (commented):
I just discovered that the same fix was applied by my coworker 1.5 years ago to fix OOM and it has been running in our production system since then. Therefore, this PR should be safe to merge. Disclaimer: we never use transactionListeners.

janusgraph-core/src/main/java/org/janusgraph/core/JanusGraph.java
li-boxuan: @mad Do you want to review again?

porunov (commented):
I never used `transactionListeners` but the changes makes sense to me.
LGTM.

PR #2502, opened 2021-03-06
Add CacheVertex::refresh method to allow clearing vertex cache

li-boxuan:
In some use cases, users might want to read from the underlying storage
backend again in the mid of a transaction, bypassing the vertex cache.
This commit adds a refresh method to CacheVertex which user could call
to clear the cache.

Related to https://lists.lfaidata.foundation/g/janusgraph-users/topic/81050716

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/operations/cache.md
+++ b/docs/operations/cache.md
@@ -47,6 +47,29 @@ which means they cannot be evicted since that would entail loosing their
 changes. Therefore, transaction which contain a lot of modifications may
 end up with a larger than configured vertex cache.
 
+Assuming your vertex is not evicted from cache, or it is evicted from
+cache but your program context still holds the reference to the vertex,
+then its properties are cached together with the vertex. This means once
+a property is queried, any subsequent reads will hit the cache. In case
+you want to force JanusGraph to read from the data storage again, or you
+simply want to save memory, you could clear the cache of that vertex manually.
+Note this is not gremlin compliant, so you need to cast your vertex into
+CacheVertex type to do the refresh:
+
+```groovy
+// first read automatically caches the property together with v
+v.property("prop").value();
+// force refresh to clear the cache
+((CacheVertex) v).refresh();
+// now a subsequent read will lead to a data storage read
+v.property("prop").value();
+```
+
+Note that refresh operation cannot guarantee your current transaction reads
+the latest data from an eventual consistent backend. You should not attempt
+to achieve Compare-And-Set (CAS) via refresh operation, even though it might
+be helpful to detect conflicts among transactions in some cases.
+
 ### Index Cache
 
 The index cache contains the results of index queries executed in the
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/CacheVertex.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/CacheVertex.java
@@ -38,6 +38,12 @@ public class CacheVertex extends StandardVertex {
         queryCache = new HashMap<>(4);
     }
 
+    public void refresh() {
+        synchronized (queryCache) {
+            queryCache.clear();
+        }
+    }
+
     protected void addToQueryCache(final SliceQuery query, final EntryList entries) {
         synchronized (queryCache) {
             //TODO: become smarter about what to cache and when (e.g. memory pressure)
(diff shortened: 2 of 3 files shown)


docs/operations/cache.md
```diff
@@ -47,6 +47,23 @@ which means they cannot be evicted since that would entail loosing their
...
+could clear the cache of that vertex manually. Note this is not gremlin
+compliant, so you need to cast your vertex into CacheVertex type to do
+the refresh:
+```groovy
+// first read automatically caches the property together with v
+v.property("prop").value();
+// force refresh to clear the cache
+((CacheVertex) v).refresh();
+// now a subsequent read will lead to a data storage read
+v.property("prop").value();
```
li-boxuan, line 64: Just to add more thoughts on this:
For storage backends like BerkeleyDB that supports locking, this operation is unnecessary because the value in data storage will not be changed simultaneously anyway; for Cassandra (and a few others) that supports eventual consistency and does not guarantee ACID, a read from data storage might still lead to an old value.

A valuable scenario seems to be:
1) The storage backend does not support (property or vertex level) locking. That is, when tx1 updates the vertex, another transaction tx2 can modify it at the same time.
2) The storage backend guarantees strong consistency. That is, once tx2 updates the vertex and commits, tx1 can read it from the storage backend immediately.

I am not even sure if there is any storage backend used by JanusGraph users that supports strong consistency (via RAFT, PAXOS, etc.) and does not support locking in any granularity. If not, then seems this API is unnecessary. Maybe it's still valuable when users use Cassandra with no replication?

janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/CacheVertex.java
```diff
@@ -38,6 +38,10 @@ public CacheVertex(StandardJanusGraphTx tx, long id, byte lifecycle) {
         queryCache = new HashMap<>(4);
     }
 
+    public void refresh() {
+        queryCache.clear();
+    }
```
porunov, line 43: I see that all operations with queryCache are synchronized except of this method. I assume this method should be thread safe as well. Thus, I guess it should be:
```
synchronized (queryCache) {
  queryCache.clear();
}
```

docs/operations/cache.md
rngcntr: Adding to your point 1.)
Although I didn't look into how JG handles locking, I would like to add that a sufficient scenario would be that only one of the transactions writes the vertex.

That said a feature like this should be used with great caution because it destroys the I in ACID on purpose.

li-boxuan: > a feature like this should be used with great caution because it destroys the I in ACID on purpose

Well said, let me add a side note to the doc. In fact, if the underlying backend storage supports ACID, then this refresh likely does not make any difference. If the underlying backend storage does not support ACID in the first place, then there is no harm to "break" the Isolation in ACID again.

li-boxuan: Another use case might actually be useful: If the vertex is a supernode, the "queryCache" stored along with the vertex can grow large. To avoid OOM or simply to save memory, the user can opt to clear the cache manually if needed.

rngcntr: Isolation and serial scheduling are guarantees that can probably also be built on top of a non ACID backend. But as long as we don't provide these guarantees, checking whether a vertex was written by another transaction can never exceed "best effort". Even if a user reads through the cache right before committing their transaction, there is still a chance that another transaction writes this vertex in the short time frame that still exists.
Therefore users should be aware that this practice is capable of lowering the chance of conflicts but can not guarantee to avoid them for sure.

li-boxuan: > Isolation and serial scheduling are guarantees that can probably also be built on top of a non ACID backend

This might be a bit off-topic, but I am interested if you have any idea/thoughts on how this could be made possible. From my understanding, this is theoretically impossible to do in JanusGraph because you would have to have a way to let different graph instances communicate/coordinate, but graph instances can only communicate via underlying storage backends.

rngcntr: You are absolutely right, that would need some form of communication like a locking approach which works across instances. But that could of course be implemented independently from the storage backend, using another technology.

li-boxuan: @rngcntr Do you want to review the doc again?

rngcntr (approved):

porunov (approved):

farodin91:
Is there a plan to get it into 0.6 or not?

porunov:
I think it makes sense to get it into 0.6.0

PR #2519, opened 2021-03-16
Increase default cql max connection configuration options

porunov:
I don't have strong opinion about this PR. I'm good merging it or closing it.
I just noticed that some novice users which start using JanusGraph with default configurations might bump to an issue with cql connections where they use only one connection and try to execute some basic benchmarks. In some situations a CQL connection might not be immediately available and thus we might see hangs between cql writes. With this simple change, users who don't configure JanusGraph at all will bump to this issue much rarely (hopefully).
That said, this is a breaking change which might be not welcoming, thus I don't have strong opinion about this PR.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
@@ -163,14 +163,14 @@ public interface CQLConfigOptions {
             "local-max-connections-per-host",
             "The maximum number of connections that can be created per host for local datacenter",
             ConfigOption.Type.FIXED,
-            1);
+            20);
 
     ConfigOption<Integer> REMOTE_MAX_CONNECTIONS_PER_HOST = new ConfigOption<>(
             CQL_NS,
             "remote-max-connections-per-host",
             "The maximum number of connections that can be created per host for remote datacenter",
             ConfigOption.Type.FIXED,
-            1);
+            5);
 
     ConfigOption<Integer> MAX_REQUESTS_PER_CONNECTION = new ConfigOption<>(
             CQL_NS,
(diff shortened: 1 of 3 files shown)


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -163,14 +163,14 @@
             "local-max-connections-per-host",
             "The maximum number of connections that can be created per host for local datacenter",
             ConfigOption.Type.FIXED,
-            1);
+            20);
```
li-boxuan, line 166: I noticed that 1 is indeed the default value used by Datastax driver. I was not aware of this config at all and we have been using the default value for a long time. I am just curious if you have any experiment/doc that demonstrates the performance difference under different settings. If we can also include that in the upgrade instruction I think it would be very helpful for users to make a call and tune their performances better.

porunov: Generally DataStax recomend to keep this setting unchanged (i.e. `1`) but sometimes this might improve performance as described in their docs: https://docs.datastax.com/en/developer/java-driver/4.10/manual/core/pooling/#tuning
I don't have a good benchmark test right now (maybe could create it later) but generally I tried it with ScyllaDB and sometimes it is improving performance when there is a single ScyllaDB node and you try to benchmark JanusGraph queries against it.
That said, I have doubts about these changes because:
1) It's a breaking change;
2) We will not use a default recommended DataStax configuration;

On another hand this PR might help new users to not bump into these problems when they start benchmarking it. 
Thus, I'm in doubt about this PR overall.
Probably, it's better to add documentation about performance tuning or something like that instead of changing default settings?

li-boxuan: Thanks for the link! If I am understanding correctly, the Datastax driver uses NIO to handle the network requests so generally, one connection is sufficient. It can only become the bottleneck when there are so many requests that the CPU becomes the bottleneck.

Since they recommend the default setting, I presume a higher number of connections brings non-negligible overhead. Thus, I would be inclined to adding performance tuning documentation.

Issue #735, opened 2017-11-22
Search count bug


robertdale:
# Setup
```
graph = JanusGraphFactory.open('conf/janusgraph-cassandra-es.properties')
mgmt = graph.openManagement()

name = mgmt.makePropertyKey('name').dataType(String.class).cardinality(SINGLE).make()
fermtype = mgmt.makePropertyKey('ferma_type').dataType(String.class).cardinality(SINGLE).make()

mgmt.buildIndex("byNameMixed", Vertex.class).addKey(name, Mapping.TEXTSTRING.asParameter()).buildMixedIndex("search"); 
mgmt.buildIndex("byFermaType", Vertex.class).addKey(fermtype).buildCompositeIndex();

mgmt.commit()

g = graph.traversal()

for (i = 0; i < 15000; i++) {
   g.addV().property("ferma_type", "com.acme.model.Person").iterate()
}
g.tx().commit()
```

# Actual:
```
gremlin> g.V().has("ferma_type", "com.acme.model.Person").count()
==>15000
gremlin> g.V().has("ferma_type", "com.acme.model.Person").count().profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
=============================================================================================================
JanusGraphStep([],[ferma_type.eq(com.acme.model...                 15000       15000          14.430    85.52
    \_condition=(ferma_type = com.acme.model.Person)
    \_isFitted=true
    \_query=multiKSQ[1]@2147483647
    \_index=byFermaType
    \_orders=[]
    \_isOrdered=true
  optimization                                                                                 0.094
  backend-query                                                                                0.000
    \_query=byFermaType:multiKSQ[1]@2147483647
  backend-query                                                                                0.000
    \_query=byFermaType:multiKSQ[1]@2147483647
  backend-query                                                                                0.000
    \_query=byFermaType:multiKSQ[1]@2147483647
  backend-query                                                    15000                       7.347
    \_query=byFermaType:multiKSQ[1]@2147483647
CountGlobalStep                                                        1           1           2.443    14.48
                                            >TOTAL                     -           -          16.874        -
gremlin> g.V().has("ferma_type", "com.acme.model.Person").count()
==>15000
gremlin> g.V().has("ferma_type", "com.acme.model.Person").has("name", "John Doe")
gremlin> g.V().has("ferma_type", "com.acme.model.Person").has("name", "John Doe").profile()
==>Traversal Metrics
[...]

PR #2520, opened 2021-03-19
Disable smart-limit by default

li-boxuan:
Prior to 0.6.0, `smart-limit` is enabled by default. It tries to guess a small limit
for each query internally, and if more results are required by user, it queries
backend again with a larger limit, and repeats until either results are exhausted or
user stops the query. However, this is not the same as paging mechanism. All interim
results will be fetched again in next round, making the whole query costly. Until
janusgraph can fully utilize the paging capacity provided by backends (e.g. Elasticsearch
scroll), I am inclined to make this option disabled by default.

Related issue: https://github.com/JanusGraph/janusgraph/issues/735

Any thoughts? Do you enable this option in your production environment?

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -259,7 +259,7 @@
     public static final ConfigOption<Boolean> ADJUST_LIMIT = new ConfigOption<>(QUERY_NS,"smart-limit",
             "Whether the query optimizer should try to guess a smart limit for the query to ensure responsiveness in " +
                     "light of possibly large result sets. Those will be loaded incrementally if this option is enabled.",
-            ConfigOption.Type.MASKABLE, true);
+            ConfigOption.Type.MASKABLE, false);
```
porunov, line 262: After updating configuration options we need to run `mvn --quiet clean install -DskipTests=true -pl janusgraph-doc -am` to regenerate configuration references documentation.
Probably we need to make it more clear

porunov:
I actually didn't know that smart limit queries the same data over and over again. Does it doubles `limit` each time or increases `limit` somehow differently? If it doubles then in some cases, the amount of data returned will be n + n\2 + n\4 + n\8... I.e. 2x overhead which isn't nice.
That said, it's interesting why this smart limit was used by default. I.e. in what cases it actually improves performance. I think it has to have some logic behind such behavior. That said, I didn't check the logic yet

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
li-boxuan: My bad, forgot about that. Thanks for reminder.

li-boxuan:
> Does it doubles limit each time or increases limit somehow differently?

It doubles the limit each time. See the code and doc in LimitAdjustingIterator:

https://github.com/JanusGraph/janusgraph/blob/3787fd4d78d2aa5e849ea836852901b19a469ec5/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java#L23-L35

and (one of its) implementation class:

https://github.com/JanusGraph/janusgraph/blob/3787fd4d78d2aa5e849ea836852901b19a469ec5/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/QueryProcessor.java#L177-L184

Ideally, this could act like a paging mechanism, so that the next query does not have to fetch every result from the very beginning, but from the cursor. However, it does not work like this. To properly utilize this smart-limit option and LimitAdjustingIterator, we need to:
1) Support paging in this iterator so that it does not fetch from beginning
1) Storage adapter must support paging. Elasticsearch supports paging, but I am not sure about other storage we use (especially primary storage backends).

Now, to my best knowledge, the only benefit of this smart-limit thing is when you are using native java API and you are iterating the results like this:

```java
Iterator iter = g.V();
int count = 0;
while (iter.hasNext()) {
    if (iter.next() matches some condition) break;
}
```
However, even the above use case can be implemented manually without the "smart-limit" support since this "smart-limit" mechanism is very naive.

li-boxuan:
Also it is worth noting that even if smart-limit is disabled, this "LimitAdjustingIterator" curse is still there because of 

https://github.com/JanusGraph/janusgraph/blob/3787fd4d78d2aa5e849ea836852901b19a469ec5/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java#L364-L368

However, this HARD_MAX_LIMIT is 100000 (compared to DEFAULT_NO_LIMIT == 1000 when smart-limit is enabled), which makes the situation slightly better compared to the case where smart-limit is enabled.

porunov (approved):
LGTM. Thank you for the explanation @li-boxuan !

Issue #1853, opened 2019-10-15
Refactor schema and index management to have unique interface


farodin91:
**Describe the feature:**
Currently, our schema and management has no clean interface. It would be good clean up the interface. We see a lot of users that having problems with our schema management or index management.

**Describe a specific use case for the feature:**
* unique interface
* index status (enable of a index)
* manage from a driver

CrazyBeeline:
Cheer! Janusgraph provides official management API

PR #2527, opened 2021-03-24
Add Service for schema management

farodin91:
Part of #1853

* Support get vertexLabel by name
* Extend client driver functions
* Move all grpc test into client round trip tests
* Move all grpc client classes into janusgraph-grpc

-----

### For all changes:

### For code changes:


janusgraph-grpc/src/main/proto/janusgraph/schema/v1/schema_manager.proto
```diff
@@ -0,0 +1,37 @@
...
+    janusgraph.types.v1.JanusGraphContext context = 1;
+    string name = 2;
+}
+
+message GetVertexLabelByNameResponse {
+    janusgraph.types.v1.VertexLabel vertex_label = 2;
+}
+
+service SchemaManagerService {
+    rpc GetVertexLabelByName(GetVertexLabelByNameRequest) returns (GetVertexLabelByNameResponse);
```
li-boxuan, line 36: In order to get vertex label by name, one has to:
1) Create a JanusGraphManagerClient to fetch JanusGraphContext
2) Create a SchemaManagerClient and use the JanusGraphContext fetched in step 1) to get vertex label

I am wondering if there is such a need. Can we simply have one service and one client?

farodin91: At the moment, i think that should be okay to have just one service. But it would become bigger over time. 

My idea of split it into two client would allow to set the context just once.

farodin91: JanusGraphManagerClient could be later used for CFG components

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public void getVertexLabelByName(
+        GetVertexLabelByNameRequest request,
+        StreamObserver<GetVertexLabelByNameResponse> responseObserver
+    ) {
+        if (request == null) {
+            responseObserver.onError(Status.INTERNAL
+                .withDescription("request is required").asRuntimeException());
+            return;
```
li-boxuan, line 37: I see sometimes you are using "asException" and sometimes "asRuntimeException". Is there a specific reason?

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/JanusGraphContextHandler.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        return JanusGraphContext.newBuilder().setGraphName(graphName).build();
+    }
+
+    public SchemaManagerProvider getSchemaManagerProviderByContext(JanusGraphContext context) {
+        Graph graph = graphManager.getGraph(context.getGraphName());
+        if (!(graph instanceof JanusGraph)) {
+            return null;
+        }
+        return new SchemaManagerProvider((JanusGraph) graph);
```
li-boxuan, line 57: Do we need to create a new SchemaManagerProvider each time the RPC server receives a request?

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import org.janusgraph.core.JanusGraph;
+import org.janusgraph.core.schema.JanusGraphManagement;
+import org.janusgraph.core.schema.VertexLabelMaker;
+import org.janusgraph.graphdb.grpc.types.VertexLabel;
+
+public class SchemaManagerProvider {
+    private final JanusGraphManagement management;
+
+    public SchemaManagerProvider(JanusGraph graph) {
+        this.management = graph.openManagement();
```
li-boxuan, line 26: We need to rollback management to prevent leakage

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerProvider.java
```diff
@@ -0,0 +1,61 @@
...
+        VertexLabelMaker vertexLabelMaker = management.makeVertexLabel(vertexLabel.getName());
+        if (vertexLabel.getReadOnly()) {
+            vertexLabelMaker.setStatic();
+        }
+        if (vertexLabel.getPartitioned()) {
+            vertexLabelMaker.partition();
+        }
+        vertexLabelMaker.make();
+
+        management.commit();
```
li-boxuan, line 40: It makes sense to commit after creating a new vertex label, however, committing at the end of this method means this SchemaManagerProvider object can only be used once, i.e. one cannot "getVertexLabelByName" after "createVertexLabel" because the management handler is already closed (committed).

I am not sure what the best approach is. Do we want to give SchemaManagerProvider the control of JanusGraphManagement's lifecycle, or give the control to the RPC service?

janusgraph-grpc/src/main/proto/janusgraph/schema/v1/schema_manager.proto
li-boxuan: Sounds fair.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/JanusGraphContextHandler.java
farodin91: I don't think we need to create SchemaManagerProvider on each call. 
Keep in mind if we are doing it differently, we have to handling the GraphManager correctly. They can be dynamic register new graphs and remove them.

My idea was to keep the logic as simple as possible.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerProvider.java
farodin91: I think we have find a solution to manage the lifecycle. 

I justed add this method for test reason (I will check to move this method into the test package). My plan is first add explore features for the schema and than start with adding and removing.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerProvider.java
farodin91: https://github.com/JanusGraph/janusgraph/pull/2527#discussion_r602708547

Isn't this question link to https://github.com/JanusGraph/janusgraph/pull/2527#discussion_r602708547 and https://github.com/JanusGraph/janusgraph/pull/2527#discussion_r602658853

farodin91:
@li-boxuan would you think of moving the grpc client stuff into the janusgraph-grpc project?

This way wouldn't requiring the grpc component in the driver or core, but you don't server project to access the grpc stuff.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerProvider.java
li-boxuan: Fair enough.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/JanusGraphContextHandler.java
li-boxuan: You think or you don't think? If I am understanding it correctly, you are creating a new SchemaManagerProvider on each call.

li-boxuan:
> would you think of moving the grpc client stuff into the janusgraph-grpc project

Make sense. janusgraph-grpc dependency is anyway needed when user wants to use grpc-client. Moving it to janusgraph-grpc module means user does not have to rely on janusgraph-driver module, right? Sounds good to me.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerImpl.java
farodin91: I will check what is best way.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/JanusGraphContextHandler.java
farodin91: At moment I would keep as simple as possible in the future would like to cache the provider some how.

farodin91:
@li-boxuan I moved all grpc client components to janusgraph-grpc and i moved the vertex creation method into the tests.

li-boxuan (approved):
LGTM 👍

farodin91:
I will merge it on wednesday, if their is no new review.

PR #2137, opened 2020-06-16
Graph Reindexing Issue Fix

shkrishn:
- Fix to avoid index corruption after reindexing RelationTypeIndex with direction IN or BOTH
- While reindexing RelationTypeIndex, edges are created according to the position of relation arity

Steps to reproduce to the self-link issue:
1) Create an Edge label (Assume name is 'link')
2) Create  an property key (Assume name is 'assockind
3) Build an 'Edge Index' by label 'link' and property key 'assockind' with Direction 'IN' or 'BOTH'
4) Create two vertices(let us call it A, B) and add a edge between A and B with label 'link' and some random value (Example : 1) for the property key 'assockind'
5) Query for all the IN edges of vertex A. Make sure the query uses the edge index. This will give no result as there is no IN edge for vertex A
	g.V().has('name' , 'A').inE().hasLabel('link').has('assocKind',1)
6) Reindex the edge index.
	mgmt.updateIndex(index, SchemaAction.REINDEX).get()
7) Execute the Step 3 once again after the reindexing. We can see a self link created for vertex A.

Complete demonstration is provided in below forum post.
https://groups.google.com/forum/#!topic/janusgraph-dev/rZPsk2CDwJ0

Signed-off-by: Shivaa Krishnan <shkrishnan@informatica.com>

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [ x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [ ] Is your initial contribution a single, squashed commit? - Single

### For code changes:
- [ x] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?
- [ ] If this PR is a documentation-only change, have you added a `[doc only]`
  tag to the first line of your commit message to avoid spending CPU cycles in
[...]

PR #2531, opened 2021-03-27
Rewrite vertex-centric reindexing test

li-boxuan:
PR #2137 Fixes a bug with VCI reindexing, but the tests are
not robust and fail manual mutation testing. This commit
rewrites the test and verifies the query results before
and after reindexing.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
@@ -136,21 +136,14 @@ public class IndexRepairJob extends IndexUpdateJob implements VertexScanJob {
 
                         Entry entry = edgeSerializer.writeRelation(janusgraphRelation, wrappedType, pos, writeTx);
                         
-                        //The below condition is check to avoid self-links which is getting created after re-indexing
-                        if(pos==0) {
-                            //Create OUT edge index entry. 
-                            //Here source will the current vertex and target vertex the other side of the relation 
-                            outAdditions.add(entry); 
-                        }else if(pos==1) {
-                            //Create IN edge index entry.
-                            //Here the source vertex is the other side of the current vertex and target will be the current vertex
+                        if (pos == 0) {
+                            outAdditions.add(entry);
+                        } else {
+                            assert pos == 1;
                             InternalVertex otherVertex = janusgraphRelation.getVertex(1);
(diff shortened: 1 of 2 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
```diff
@@ -136,21 +136,14 @@ public void process(JanusGraphVertex vertex, ScanMetrics metrics) {
...
+                        if (pos == 0) {
+                            outAdditions.add(entry);
+                        } else {
+                            Preconditions.checkArgument(pos == 1, "Invalid pos %d of relation", pos);
                             InternalVertex otherVertex = janusgraphRelation.getVertex(1);
                             StaticBuffer otherVertexKey = writeTx.getIdInspector().getKey(otherVertex.longId());
                             inAdditionsMap.computeIfAbsent(otherVertexKey, k -> new ArrayList<Entry>()).add(entry);
-                        }else {
-                            throw new IllegalStateException("Invalid position:" + pos);
                         }
```
porunov, line 146: Looks like the nature of the exception is changed from IllegalStateException to IllegalArgumentException. Any reason for that?
Also, I would prefer to use direct Java code instead of Guava code where possible due to long term task #2033 . 
That said, #2033 most likely won't be resolved anytime soon because there are many places where Guava is very handy (like cache for example), where we don't have good substitutions as for now. Thus, it's OK to use Guava but if it's easy to omit it, I would prefer omitting it.
I guess we could refactor this part to use `switch` which is compiled better or same (depends on situation) than if / else chains.
Something like:
```
switch(pos){
  case 0: outAdditions.add(entry); break;
  case 1: ... break;
  default: throw new IllegalStateException("Invalid pos " + pos + " of relation");
}
```

li-boxuan: Removing this else branch because it is lengthy, unnecessary, and cannot be covered by the test. pos can only be 0, 1, or 2, because InternalRelation::getArity either returns 1 (for property) or 2 (for edge).

I could replace Preconditions.checkArgument with assert if that sounds better.

porunov: `assert` sounds good to me

porunov (approved):
I didn't look too deep into the tests but the code LGTM overall. Thank you @li-boxuan !

PR #2532, opened 2021-03-27
Remove unused java classes to improve test coverage

li-boxuan:
I have checked that JanusGraph is not using the classes removed in this PR by using Intellij to find usage and searching the class name in the codebase. There are still some unused classes that I am not 100% sure whether some users are using, e.g. AllEdgesIterable, thus not included in this PR.

-----

### For all changes:

### For code changes:


farodin91:
We could remove also janusgraph-core/src/main/java/org/janusgraph/graphdb/schema

li-boxuan:
> We could remove also janusgraph-core/src/main/java/org/janusgraph/graphdb/schema

Thanks, I found out they are only used by tests. Thus I moved them to janusgraph-backend-testutils module.

farodin91:
If we want to remove SchemaProvider, than we should remove also test for it.

https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java#L1169-L1184

I could create also different PR for and mark it as deprecated in 0.6 and we can removed after completely.

li-boxuan:
> I could create also different PR for and mark it as deprecated in 0.6 and we can removed after completely.

Sure, I have reverted my commit to the previous one.

farodin91 (approved):
LGTM! Thank you for looking for these cases.

PR #2533, opened 2021-03-27
Deprecated org.janusgraph.graphdb.schema

farodin91:
Part of #1853

-----

### For all changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/schema/EdgeLabelDefinition.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/schema/EdgeLabelDefinition.java
@@ -20,7 +20,10 @@ import org.apache.tinkerpop.gremlin.structure.Direction;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
+ *
+ * @deprecated part of the management revamp in JG, see https://github.com/JanusGraph/janusgraph/projects/3.
  */
+@Deprecated
 public class EdgeLabelDefinition extends RelationTypeDefinition {
 
     private final boolean unidirected;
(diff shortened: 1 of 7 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/schema/EdgeLabelDefinition.java
```diff
@@ -20,7 +20,10 @@
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
+ *
+ * @deprecated part of the management revamp in JG, see https://github.com/JanusGraph/janusgraph/projects/3.
```
li-boxuan, line 24: I am just trying to understand: how is this related to the management revamp? I can see even in v0.5 branch, these classes are only used in tests.

farodin91: You could you use the explore schema. I don't think anyone is using it.

li-boxuan: Can you elaborate on "explore schema"? Never heard of it.

farodin91: You could you use the schema provider to list all vertices, edges or properties.

li-boxuan: Got it.

li-boxuan (approved):
LGTM 👍

PR #2548, opened 2021-03-31
Add GetEdgeLabelByName

farodin91:
Part of #1853

* Support get edgeLabel by name
* Add readme with a current state
* Move grpc util into a extra class
* Small breaking in protos
-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
+++ b/janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
@@ -23,11 +23,43 @@ option java_package = "org.janusgraph.graphdb.grpc.types";
 
 message VertexLabel {
     // The Internal id.
-    string id = 1;
+    google.protobuf.Int64Value id = 1;
     // The VertexLabel Name.
     string name = 2;
     // Marking as readOnly, default not readOnly.
     bool read_only = 3;
     // Marking as partitioned, default not partitioned.
     bool partitioned = 4;
+}
+
+message EdgeLabel {
+    // The Internal id.
+    google.protobuf.Int64Value id = 1;
+    // The EdgeLabel Name.
+    string name = 2;
+    enum Direction {
+        // Edge can be queried from both vertices.
+        BOTH = 0;
+        // Edge can be only queried from the outgoing vertex.
+        // Direction OUT should be only used by experts, see https://docs.janusgraph.org/schema/advschema/#unidirected-edges.
+        OUT = 1;
+    }
+    // Defines queryable direction.
+    Direction direction = 3;
+    enum Multiplicity {
+        // The given edge label specifies a multi-graph, meaning that the multiplicity is not constrained and that
+        // there may be multiple edges of this label between any given pair of vertices.
+        MULTI = 0;
+        // The given edge label specifies a simple graph, meaning that the multiplicity is not constrained but that there
+        // can only be at most a single edge of this label between a given pair of vertices.
+        SIMPLE = 1;
+        // There can only be a single in-edge of this label for a given vertex but multiple out-edges (i.e. in-unique).
+        ONE2MANY = 2;
+        // There can only be a single out-edge of this label for a given vertex but multiple in-edges (i.e. out-unique).
+        MANY2ONE = 3;
+        // There can be only a single in and out-edge of this label for a given vertex (i.e. unique in both directions).
+        ONE2ONE = 4;
+    }
+    // Defines Multiplicity of a edge.
+    Multiplicity multiplicity = 4;
 }
--- a/janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerImpl.java
+++ b/janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerImpl.java
@@ -14,9 +14,12 @@
 
 package org.janusgraph.graphdb.grpc.schema;
 
+import com.google.protobuf.GeneratedMessageV3;
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import org.janusgraph.graphdb.grpc.JanusGraphContextHandler;
+import org.janusgraph.graphdb.grpc.types.EdgeLabel;
+import org.janusgraph.graphdb.grpc.types.JanusGraphContext;
 import org.janusgraph.graphdb.grpc.types.VertexLabel;
 
 public class SchemaManagerImpl extends SchemaManagerServiceGrpc.SchemaManagerServiceImplBase {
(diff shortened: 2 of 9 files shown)


janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -23,11 +23,44 @@ option java_package = "org.janusgraph.graphdb.grpc.types";
...
+message EdgeLabel {
+    // The Internal id.
+    google.protobuf.Int64Value id = 1;
+    // The EdgeLabel Name.
+    string name = 2;
+    enum Direction {
+        // Edge can be queried from both vertices.
+        BOTH = 0;
+        // Edge can be only queried from the outgoing vertex.
+        //
```
li-boxuan, line 44: (nitpick) redundant line?

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -23,11 +23,44 @@ option java_package = "org.janusgraph.graphdb.grpc.types";
 
 message VertexLabel {
     // The Internal id.
-    string id = 1;
+    google.protobuf.Int64Value id = 1;
```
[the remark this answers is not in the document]
farodin91: I would say the ID has two use cases.
1. debugging purpose 
2. in case of name change it is helpful to have this value.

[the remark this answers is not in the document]
li-boxuan: Fair enough

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerImpl.java
```diff
@@ -56,4 +57,36 @@ public void getVertexLabelByName(
...
+            return;
+        }
+        final String edgeLabelName = request.getName();
+        if (edgeLabelName.isEmpty()) {
+            responseObserver.onError(Status.INVALID_ARGUMENT
+                .withDescription("name is required").asException());
+            return;
+        }
+        SchemaManagerProvider provider = contextHandler.getSchemaManagerProviderByContext(request.getContext());
+        EdgeLabel edgeLabel = provider.getEdgeLabelByName(edgeLabelName);
```
li-boxuan, line 83: Null check? `provider` might be null here.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerImpl.java
```diff
@@ -56,4 +57,36 @@ public void getVertexLabelByName(
...
+        if (edgeLabel == null) {
+            responseObserver.onError(Status.NOT_FOUND.asException());
```
li-boxuan, line 85: (nitpick) I think we could provide a description here. NOT_FOUND can be due to the fact that graph does not exist, or edge label does not exist.

janusgraph-server/src/main/java/org/janusgraph/graphdb/grpc/schema/util/GrpcUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.protobuf.Int64Value;
+import org.janusgraph.core.Multiplicity;
+import org.janusgraph.graphdb.grpc.types.EdgeLabel;
+import org.janusgraph.graphdb.grpc.types.VertexLabel;
+
+public class GrpcUtils {
+
+    public static VertexLabel createVertexLabelProto(org.janusgraph.core.VertexLabel vertexLabel) {
+        return VertexLabel.newBuilder()
+            .setId(Int64Value.of(vertexLabel.longId()))
```
li-boxuan, line 26: We are not coming into this part (create label gRPC API) yet, but just leaving a message here to remind us:

One thing we need to check is whether `longId()` returns a permanent id. For user-created vertices, calling `longId()` or `id()` before committing the transaction could lead to an intermediate result, if `ids.flush=false`. Not sure about vertex labels and edge labels though.

janusgraph-server/src/test/java/org/janusgraph/graphdb/grpc/JanusGraphGrpcServerBaseTest.java
```diff
@@ -51,7 +56,7 @@ protected static GraphManager getGraphManager() {
         return new DefaultGraphManager(settings);
     }
 
-    public void createVertexLabel(String graph, VertexLabel vertexLabel) {
+    public void createVertexLabel(String graph, VertexLabelOrBuilder vertexLabel) {
```
li-boxuan, line 59: Just curious: how does this "...OrBuilder" work?

farodin91: vertex can be VertexLabel or VertexLabel.Builder, you don´t need to call build to access read functions.

farodin91:
@li-boxuan I tried to fix all and tried to reduce the duplications in the SchemaManagerImpl.

li-boxuan (approved):
LGTM 👍  It would be even better if you could improve the test coverage, but I am fine without it.

janusgraph-server/src/test/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerClientTest.java
```diff
@@ -62,51 +65,127 @@ public void testGetVertexLabelByNameInvalidArgumentNullContext() {
         assertThrows(NullPointerException.class, () -> schemaManagerClient.getVertexLabelByName("test"));
     }
 
-    @Test
-    public void testGetVertexLabelByNameVertexLabelExists() {
-        final String vertexLabelName = "test";
+    @ParameterizedTest
```
li-boxuan, line 68: I suspect you need to add the following dependencies into pom.xml so that maven can execute it:
```xml
        <dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter-params</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter-engine</artifactId>
            <scope>test</scope>
        </dependency>
```

I had been wondering why my code coverage was so low in #2556 until I added these dependencies to janusgraph-cql/pom.xml. It seems otherwise, maven would not execute these parameterized tests.

farodin91: I just checked all tests are executed.

farodin91:
@li-boxuan I will try to add in the next PR some more coverage to it. My plan is to add support to query all vertices and edges for the next PR.

janusgraph-server/src/test/java/org/janusgraph/graphdb/grpc/schema/SchemaManagerClientTest.java
li-boxuan: You are right... now I am wondering why I needed them in janusgraph-cql/pom.xml...

farodin91:
@li-boxuan I will fix codacy in directly follow up PR.

PR #2551, opened 2021-04-03
Fix profiling when limit adjusting is enabled

li-boxuan:
When LimitAdjustingIterator is in effect, multiple backend queries with
different limits (e.g. 1000, 2000, 4000, ...) are fired progressively.
They are, however, not all profiled. Currently, only the last backend
query is profiled because of a bug. This commit fixes it and now users
can see timing of each backend query to better understand whether they
need smart-limit or not.

This commit also fixes a minor thing with profiler output, such that when
QUERY.NO_LIMIT is used or no limit is given, we don't display Integer.MAX_VALUE
in the profile output.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -5843,11 +5843,11 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
         Metrics deeplyNested = (Metrics) nested.getNested().toArray()[0];
         assertEquals("AND-query", deeplyNested.getName());
         // FIXME: assertTrue(deeplyNested.getDuration(TimeUnit.MICROSECONDS) > 0);
-        assertEquals("multiKSQ[1]@2147483647", deeplyNested.getAnnotation("query"));
+        assertEquals("multiKSQ[1]", deeplyNested.getAnnotation("query"));
         deeplyNested = (Metrics) nested.getNested().toArray()[1];
         assertEquals("AND-query", deeplyNested.getName());
         // FIXME: assertTrue(deeplyNested.getDuration(TimeUnit.MICROSECONDS) > 0);
-        assertEquals("multiKSQ[1]@2147483647", deeplyNested.getAnnotation("query"));
+        assertEquals("multiKSQ[1]", deeplyNested.getAnnotation("query"));
 
         // satisfied by one graph-centric query, which satisfied by in-memory filtering after one composite index query
         newTx();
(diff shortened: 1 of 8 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -5908,6 +5908,88 @@ public void testGraphCentricQueryProfiling() {
...
+                tx.addVertex("name", "bob");
+            }
+            tx.commit();
+        };
+
+        clopen(option(ADJUST_LIMIT), false);
+        dataLoader.run();
+
+        newTx();
+        Metrics mCompSingle = tx.traversal().V().has("name", "bob").profile().next().getMetrics(0);
```
li-boxuan, line 5929: It looks like this:
```
Traversal Metrics
JanusGraphStep([],[name.eq(bob)])                                   3000        3000          23.487   100.00
  constructGraphCentricQuery                                                                   0.035
  constructGraphCentricQuery                                                                   1.311
  GraphCentricQuery                                                                           23.842
    \_condition=(name = bob)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@4000
    \_index=nameIdx
    backend-query                                                   1000                       7.681
    \_query=nameIdx:multiKSQ[1]@1000
    \_limit=1000
    backend-query                                                   2000                       8.124
    \_query=nameIdx:multiKSQ[1]@2000
    \_limit=2000
    backend-query                                                   3000                       7.708
    \_query=nameIdx:multiKSQ[1]@4000
    \_limit=4000
                                            >TOTAL                     -           -          23.487        -
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
```diff
@@ -71,6 +71,8 @@ public boolean hasNext() {
             return iterator.hasNext();
         if (currentLimit>=maxLimit) return false;
 
+        //Close old iterator
+        CloseableIterator.closeIterator(iterator);
```
li-boxuan, line 75: This is needed otherwise the old iterator will not be timed properly.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/QueryProcessor.java
```diff
@@ -196,7 +196,7 @@ private LimitAdjustingIterator(BackendQueryHolder<B> backendQueryHolder) {
 
         @Override
         public Iterator<R> getNewIterator(int newLimit) {
-            if (!backendQuery.hasLimit() || newLimit>backendQuery.getLimit())
+            if (newLimit>backendQuery.getLimit())
```
li-boxuan, line 199: hasLimit() returns false iff limit == Integer.MAX_VALUE. If there is no limit, it is not only redundant but also incorrect from profiling standpoint, to update the limit (which will effectively return a new query).

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/JointIndexQuery.java
```diff
@@ -116,6 +116,7 @@ public JointIndexQuery updateLimit(int newLimit) {
             subqueries = new ArrayList<>(queries);
         }
         JointIndexQuery jointIndexQuery = new JointIndexQuery(subqueries);
+        jointIndexQuery.observeWith(this.profiler, false);
```
li-boxuan, line 119: When updating the limit and returning a new "cloned" query, we must also pass the profiler otherwise previous profiling results get lost.

janusgraph-test/src/test/java/org/janusgraph/graphdb/query/QueryTest.java
```diff
@@ -86,7 +86,7 @@ public void testQueryLimitAdjustment() {
 
         for (int i = 0; i < 20; i++) {
             tx.addVertex().property("prop1", "prop1val").element().property("prop2", "prop2val")
-                .element().property("prop3", "prop3val" + i % 2).element().property("prop4", "prop4val" + i % 2);
+                .element().property("prop3", "prop3val").element().property("prop4", "prop4val");
```
li-boxuan, line 89: This change is necessary otherwise tests fail occasionally. The reason is, the in-memory backend (seemingly) does not return results in a consistent order. Therefore, for a query that has two conditions and requires in-memory filtering, the query sometimes needs to progressively fetch more results to satisfy the number of results required. As a matter of consequence, profiling results might not necessarily be consistent.

This was not a problem (never failed before) because the profiling results were incomplete (only covered the first backend query).

li-boxuan (commented):
Adding some comments to help reviewers understand, as this commit contains some subtle points that require some context.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
rngcntr: But only in the case below, with `ADJUST_LIMIT=true`, right?

rngcntr (approved):

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
li-boxuan: Yes, thanks for the reminder. The comment above was posted at the wrong place. It belongs to the case below, i.e. with `ADJUST_LIMIT = true`.

Here the profiling output should be:
```
Traversal Metrics
JanusGraphStep([],[name.eq(bob)])                                   3000        3000         123.323   100.00
  constructGraphCentricQuery                                                                   0.086
  constructGraphCentricQuery                                                                   1.073
  GraphCentricQuery                                                                          121.164
    \_condition=(name = bob)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=multiKSQ[1]@100000
    \_index=nameIdx
    backend-query                                                   3000                     120.870
    \_query=nameIdx:multiKSQ[1]@100000
    \_limit=100000
                                            >TOTAL                     -           -         123.323        -
```

janusgraph-test/src/test/java/org/janusgraph/graphdb/query/QueryTest.java
li-boxuan: 1. I gave a wrong statement about the inconsistent order. After some experiments, it seems that the in-memory backend indeed returns in a consistent order. What actually happens is, the order of results is not consistent with the order of insertions. To be concrete, the following test would fail:

```
        JanusGraphManagement mgmt = graph.openManagement();
        PropertyKey prop1Key = mgmt.makePropertyKey("prop1").dataType(String.class).make();
        mgmt.buildIndex("prop1_idx", Vertex.class).addKey(prop1Key).buildCompositeIndex();
        mgmt.commit();

        List<Object> vertexIds = new ArrayList<>();
            Vertex v = tx.addVertex();
            v.property("prop1", "val");
            vertexIds.add(v.id());
        }
        tx.commit();

        List<Object> actualIds = graph.traversal().V().has("prop1", "val").id().toList();
        assertEquals(vertexIds, actualIds);
```

This is (at least partially) because, when JanusGraph commits, it uses an unordered HashSet to store all the mutations. Thus, in the query `graph.traversal().V().has("prop1", "prop1val").has("prop3", "prop3val0").limit(5)`, there is no guarantee whether first 10 results of `V().has("prop1", "prop1val")` has at least 5 results that satisfy `has("prop3", "prop3val0")` condition or not.

2. To my best memory, I didn't use "i % 2" for any particular use. I simply didn't want to use a constant value, to make the vertices look slightly more normal.

rngcntr: Thanks for working that out in such a detailed way!

rngcntr (approved):

li-boxuan:
If there is any further comment, I'll address in the PR https://github.com/JanusGraph/janusgraph/pull/2552

PR #2552, opened 2021-04-03
Make hard max limit configurable and use Integer.MAX_VALUE by default

li-boxuan:
As a follow-up of #2520, this commit makes HARD_MAX_LIMIT used in
GraphCentricQueryBuilder configurable. It also uses Integer.MAX_VALUE
as default value rather than the old value 100,000. This commit also
includes a benchmark program that demonstrates the performance improvement.

-----

### For all changes:

### For code changes:


li-boxuan:
Benchmark result:

```
Benchmark                               (hardMaxLimit)  (size)  (useSmartLimit)  Mode  Cnt    Score    Error  Units
GraphCentricQueryBenchmark.getVertices          100000   10000             true  avgt   10    8.835 ±  0.042  ms/op
GraphCentricQueryBenchmark.getVertices          100000   10000            false  avgt   10    6.445 ±  0.126  ms/op
GraphCentricQueryBenchmark.getVertices          100000  250000             true  avgt   10  454.994 ±  3.346  ms/op
GraphCentricQueryBenchmark.getVertices          100000  250000            false  avgt   10  509.808 ± 50.164  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647   10000             true  avgt   10   10.877 ±  2.401  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647   10000            false  avgt   10    7.852 ±  1.399  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647  250000             true  avgt   10  526.961 ± 30.369  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647  250000            false  avgt   10  320.730 ± 11.158  ms/op
```

li-boxuan:
Blocked by https://github.com/JanusGraph/janusgraph/pull/2551

rngcntr:
Have you figured out why the hard limit is set to 100,000 in the first place?

li-boxuan:
> Have you figured out why the hard limit is set to 100,000 in the first place?

My best guess is, when the titan authors started to write that code, they were hoping that the backend queries could be paged, which unfortunately does not happen. See the thread here: https://github.com/JanusGraph/janusgraph/pull/2520#issuecomment-802980616

Another reason for doing so is assuming that users would generally only need a few results, but often forget to add the limit() step. By enforcing a HARD_MAX_LIMIT, we can prevent reckless users from consuming too much CPU & memory. This is a benefit but not considerable to me, thus I prefer making the default HARD_MAX_LIMIT use max int, which would be the best in performance.

rngcntr (approved):
LGTM!
Thank you very much @li-boxuan

porunov:
> Benchmark result:
> 
> ```
> Benchmark                               (hardMaxLimit)  (size)  (useSmartLimit)  Mode  Cnt    Score    Error  Units
> GraphCentricQueryBenchmark.getVertices          100000   10000             true  avgt   10    8.835 ±  0.042  ms/op
> GraphCentricQueryBenchmark.getVertices          100000   10000            false  avgt   10    6.445 ±  0.126  ms/op
> GraphCentricQueryBenchmark.getVertices          100000  250000             true  avgt   10  454.994 ±  3.346  ms/op
> GraphCentricQueryBenchmark.getVertices          100000  250000            false  avgt   10  509.808 ± 50.164  ms/op
> GraphCentricQueryBenchmark.getVertices      2147483647   10000             true  avgt   10   10.877 ±  2.401  ms/op
> GraphCentricQueryBenchmark.getVertices      2147483647   10000            false  avgt   10    7.852 ±  1.399  ms/op
> GraphCentricQueryBenchmark.getVertices      2147483647  250000             true  avgt   10  526.961 ± 30.369  ms/op
> GraphCentricQueryBenchmark.getVertices      2147483647  250000            false  avgt   10  320.730 ± 11.158  ms/op
> ```

I'm not sure I understand this results correctly. It looks like with the hard limit `2147483647` we have improvement with the last test case only (2147483647, 250000, false). Do you know why other test cases perform slightly better with the `hardMaxLimit = 100000`?
I didn't dig deep into this issue, so they might perform the same.

li-boxuan:
@porunov Good finding, I didn't even notice that. I think it's just statistics error, but let me increase warmup & execution iterations and run the benchmark again to be sure.

li-boxuan:
I increased warmup iterations from 10 to 20, and measurement iterations from 10 to 30 locally, with the hope that statistics error can be reduced. I am not committing this change since it makes the benchmark program run much slower. The results are very similar to the previous one, but with smaller std err.
```
Benchmark                               (hardMaxLimit)  (size)  (useSmartLimit)  Mode  Cnt    Score   Error  Units
GraphCentricQueryBenchmark.getVertices          100000   10000             true  avgt   30    9.012 ± 0.028  ms/op
GraphCentricQueryBenchmark.getVertices          100000   10000            false  avgt   30    6.694 ± 0.036  ms/op
GraphCentricQueryBenchmark.getVertices          100000  250000             true  avgt   30  449.749 ± 7.606  ms/op
GraphCentricQueryBenchmark.getVertices          100000  250000            false  avgt   30  481.977 ± 7.398  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647   10000             true  avgt   30    8.862 ± 0.025  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647   10000            false  avgt   30    6.719 ± 0.251  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647  250000             true  avgt   30  471.065 ± 6.119  ms/op
GraphCentricQueryBenchmark.getVertices      2147483647  250000            false  avgt   30  296.641 ± 6.819  ms/op
```
Let's compare hardMaxLimit (2,147,483,647 v.s. 100,000):

1) we can see that there is almost no difference when the "size" is relatively small (10,000), since both hardMaxLimits are larger than 10,000.
2) when the "size" is 250,000, and smart-limit is enabled, JanusGraph will start the query from DEFAULT_NO_LIMIT which is 2,000, and doubles it iteratively. In this case, hardMaxLimit does not take effect at all. There is some difference (449.749 vs 471.065) but that is most likely due to external reasons (different CPU load when running the program, etc.).
3) when the "size" is 250,000, and smart-limit is disabled, we can clearly see the performance difference. Clearly, using 2,147,483,647 as hardMaxLimit has much better result.

porunov (approved):
LGTM. Thank you @li-boxuan !

Issue #2555, opened 2021-04-03
Make GhostVertexRemover easy to use


li-boxuan:
JanusGraph codebase contains a `GhostVertexRemover` that implements `ScanJob` interface, which could be used via:

```java
JanusGraphManagement.IndexJobFuture ghostRemover =
    graph.getBackend().buildEdgeScanJob()
         .setJob(new GhostVertexRemover(graph))
         .execute();
```

The above way, however, runs on a single instance, and thus its capacity is limited despite its use of multi-threading. It is unrealistic to use it to remove all ghost vertices on a massively large graph. We should make it easy to use in both 1) single machine 2) Hadoop system. We should, of course, also add the usage to the documentation

A good example is`IndexRemoveJob`, which also implements the `ScanJob` interface. It can be used via `ManagementSystem` (single machine) and `MapReduceIndexManagement` (map-reduce).

PR #2557, opened 2021-04-04
Refactor GhostVertexRemover and VertexJobConverter

li-boxuan:
This commit extracts common code from VertexJobConverter into the new class AbstractScanJob,
and makes both GhostVertexRemover and VertexJobConverter inherit it. Previously,
GhostVertexRemover inherits VertexJobConverter which does not make sense.

Part of #2555 

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/PartitionedVertexProgramExecutor.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/PartitionedVertexProgramExecutor.java
@@ -20,7 +20,6 @@ import org.janusgraph.diskstorage.EntryList;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.ScanMetrics;
 import org.janusgraph.graphdb.database.StandardJanusGraph;
 import org.janusgraph.graphdb.idmanagement.IDManager;
-import org.janusgraph.graphdb.olap.VertexJobConverter;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.graphdb.util.WorkerPool;
 import org.janusgraph.graphdb.vertices.PreloadedVertex;
@@ -64,7 +63,7 @@ public class PartitionedVertexProgramExecutor<M> {
         if (pVertexAggregates.isEmpty()) return; //Nothing to do here
 
         try (WorkerPool workers = new WorkerPool(numThreads)) {
-            tx = VertexJobConverter.startTransaction(graph);
+            tx = startTransaction(graph);
             for (Map.Entry<Long,EntryList> partitionedVertices : pVertexAggregates.entrySet()) {
                 if (partitionedVertices.getValue()==null) {
                     metrics.incrementCustom(GHOST_PARTITION_VERTEX);
@@ -78,6 +77,10 @@ public class PartitionedVertexProgramExecutor<M> {
         } finally {
             if (tx!=null && tx.isOpen()) tx.rollback();
         }
+    }
+
+    private StandardJanusGraphTx startTransaction(StandardJanusGraph graph) {
+        return (StandardJanusGraphTx) graph.buildTransaction().readOnlyOLAP().start();
     }
 
     private class PartitionedVertexProcessor implements Runnable {
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
@@ -25,9 +25,8 @@ import org.janusgraph.diskstorage.keycolumnvalue.SliceQuery;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.ScanMetrics;
 import org.janusgraph.diskstorage.util.BufferUtil;
 import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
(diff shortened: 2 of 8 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/PartitionedVertexProgramExecutor.java
```diff
@@ -80,6 +80,15 @@ public void run(int numThreads, ScanMetrics metrics) {
         }
     }
 
+    private StandardJanusGraphTx startTransaction(StandardJanusGraph graph) {
```
farodin91, line 83: nitpick: Would it a good idea to move startTransaction into the AbstractScanJob.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRemoveJob.java
```diff
@@ -53,7 +54,7 @@
  */
 public class IndexRemoveJob extends IndexUpdateJob implements ScanJob {
```
farodin91, line 55: nitpick: Should we use here also the AbstractScanJob?

farodin91 (commented):
I just have two questions for now.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/PartitionedVertexProgramExecutor.java
li-boxuan: But this class PartitionedVertexProgramExecutor.java is not and cannot be a subclass of AbstractScanJob.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRemoveJob.java
li-boxuan: I cannot do that since IndexRemoveJob already extends IndexUpdateJob. I cannot make IndexUpdateJob extend AbstractScanJob either, because one of its subclasses, IndexRepairJob, implements VertexScanJob interface rather than ScanJob interface.

farodin91:
@li-boxuan thanks for answering my question.

farodin91 (approved):
LGTM! Thank you @li-boxuan.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
```diff
@@ -55,52 +54,56 @@
...
-
+    public GhostVertexRemover clone() {
+        return new GhostVertexRemover(this);
+    }
 
     @Override
     public void workerIterationStart(Configuration jobConfig, Configuration graphConfig, ScanMetrics metrics) {
-        super.workerIterationStart(jobConfig, graphConfig, metrics);
-        Preconditions.checkArgument(jobConfig.has(GraphDatabaseConfiguration.JOB_START_TIME),"Invalid configuration for this job. Start time is required.");
-        this.jobStartTime = Instant.ofEpochMilli(jobConfig.get(GraphDatabaseConfiguration.JOB_START_TIME));
```
porunov, line 75: Looks like previously this line was executed after graph is initialized but before the transaction is started. Right now it is executed after the transaction is started. 
I.e. the first time `workerIterationStart` is called the `jobStartTime` will be `null`. Thus, during the execution of `startTransaction`, we will set `null` here `txb.commitTime(jobStartTime);` instead of current time. Any next execution of `workerIterationStart` will set `commitTime` of the previous transaction instead of current transaction.
It looks like a bug to me in this refactoring, but I didn't check all the refactored code yet, so I could miss something

porunov (changes requested):

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/GhostVertexRemover.java
li-boxuan: Great catch! 

> Any next execution of workerIterationStart will set commitTime of the previous transaction instead of current transaction.

This won't happen because this method is only called once.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/PartitionedVertexProgramExecutor.java
porunov: @li-boxuan What if we just move it to `StandardJanusGraphTx` itself? I'm just thinking about removing the duplication code.
Something like (not sure about method name):
```
    public static StandardJanusGraphTx startNoCacheNoInternalCheckPreloadedDataTransaction(StandardJanusGraph graph) {
        StandardTransactionBuilder txb = graph.buildTransaction().readOnly();
        txb.setPreloadedData(true);
        txb.checkInternalVertexExistence(false);
        txb.dirtyVertexSize(0);
        txb.vertexCacheSize(0);
        return (StandardJanusGraphTx)txb.start();
    }
```
If so, we can use this static method in both PartitionedVertexProgramExecutor and VertexJobConverter. 
I'm also good moving this method to a separate utility class to not mess `StandardJanusGraphTx` with such methods.

li-boxuan: I moved it to StandardTransactionBuilder class.

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2561, opened 2021-04-05
Proof of concept for properties in the future of schema management

farodin91:


janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -21,6 +21,49 @@ import "google/protobuf/wrappers.proto";
...
+
+enum VertexPropertyCardinality {
+    Single = 0;
+    List = 1;
+    Set = 2;
+}
+
+enum PropertyBinding {
+    Globally = 0;
+    Local = 1;
```
li-boxuan, line 48: What does local mean?

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -21,6 +21,49 @@ import "google/protobuf/wrappers.proto";
...
+// future possibilities
+// * per vertex/edge property
+// * enforce schema constraints but special properties that can be bound to any vertex/edge
+
+message VertexProperty {
+    google.protobuf.Int64Value id = 1;
+    string name = 2;
+    PropertyDataType dataType = 3;
+    PropertyBinding binding = 4;
+    VertexPropertyCardinality cardinality = 5;
```
li-boxuan, line 64: A vertex property can have its own meta properties.

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
farodin91: Bound to a vertex/edge.

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
farodin91: Can we define a schema for meta properties?

rngcntr: I think we can embed `repeated EdgeProperty` here, and probably rename the message type accordingly.

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
li-boxuan: Got it, but this seems to contradict to the comments below:

> // * schema constraints: property is globally defined but bind to vertex or edge

From which I inferred whether a property is bound to a vertexLabel/edgeLabel has nothing to do with whether its global or loca.

li-boxuan: I've never used the binding feature myself before, so I might be missing something here. The "two more cases" you mentioned,

> allow different datatypes for different vertexLabel with the same property name.

and 

> Global properties (can be used on all vertexLabels) with activated schema constraint

are they legitimate cases in JanusGraph now? Also, I don't get the latter case. Can you elaborate or point me to doc/code?

farodin91: I thought about possible future ideas for Schema Management.

farodin91: I thought of possible future use cases how we can define properties to be more flexible.

farodin91:
https://medium.com/@experoinc/schema-support-in-three-property-graph-databases-1beff569855e I found an article by expero.

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
farodin91: I tried to update the schema. 
@li-boxuan @rngcntr  what do you think about my updated version?

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -21,6 +21,71 @@ import "google/protobuf/wrappers.proto";
...
+    SET = 2;
+}
+
+enum PropertyConstraints {
+    // Property can be used on all vertices and properties (can only be used with locality: global)
+    OPEN = 0;
+    // Property can be used on all defined vertices or properties
+    DEFINED = 1;
+    // Property is required on all defined vertices or properties
+    REQUIRED = 2;
```
farodin91, line 52: future possibility

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -21,6 +21,71 @@ import "google/protobuf/wrappers.proto";
...
+}
+
+enum PropertyLocality {
+    // Property is defined globally for all vertices and all edges/meta properties (if cardinality is single)
+    GLOBAL = 0;
+    // Property is defined a side of a vertex or edge.
+    LOCAL = 1;
```
farodin91, line 59: future possibility

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
li-boxuan: To be honest I still don't understand the difference between

[this]( https://github.com/JanusGraph/janusgraph/pull/2561/files#diff-9c648aed04a9a8f61584cfb812ecc2727a893f6fd2a486502e75e8b3beab7f16R58-R59):
> Property is defined a side of a vertex or edge.

and 

[this:](https://github.com/JanusGraph/janusgraph/pull/2561/files#diff-9c648aed04a9a8f61584cfb812ecc2727a893f6fd2a486502e75e8b3beab7f16R63)
> schema constraints: property is globally defined but bind to vertex or edge

rngcntr: I talked to @farodin91 about the intention of this feature earlier today. To wrap it up, the idea is that although a property key (say, `direction` for example) can have different constraints for different vertex labels. For instance, the `direction` property on a vertex labeled `speaker` could be constrained to values `left` and `right`. However, the same property `direction` on a vertex labeled `entrance` could be constrained to values `north`, `east`, `south` and `west`.
That would be enabled by a local binding.

dk-github:
Wouldn't it be much cleaner to just force people to name properties
differently if their value domain is different?

Making property "bindings" local is a bit like abandoning the original
LDAP-like schema logic (where attributes are defined globally and
irrespectively of node types they can appear in), and trying to morph it
into OOP-style class/type logic, where different classes can have
properties with same name but totally different value type/semantics...
It may have been great if it was so (or maybe not), but it isn't, and if
you want to change that - that is a big "paradigm shift" I'd say - some
people may possibly be not happy about it.. Just a thought - I don't have
any strong opinion on this myself...

rngcntr:
@dk-github I don't have a strong opinion, either. But from what I understood, this is mostly about verifying the values of certain properties. Maybe the case of `left`/`right` vs `north`/`east`/... is a little bit too extreme. There are many more cases, where different verification schemes would make sense without changing the semantics. An example could be a property `email`, which should follow a format `xxxxxx@yyyyyy.zzz` for every vertex label, but for vetices labeled `student`, the format should be restricted to `<firstname>.<lastname>@yyyyyy.edu`.
Of course, all of this would be also possible if one just creates two different properties with different keys. But maybe the case of `email` makes it more clear that cases exist where it would be more convenient to use the same to use the same key.

farodin91:
@dk-github I would like to give users the option.

farodin91:
Closed by #2995  and #2940

PR #2568, opened 2021-04-11
CQLKeyColumnValueStore: Avoid fetching WRITETIME and/or TTL if no need

li-boxuan:
When meta.*.timestamps and/or meta.*.ttl are disabled (which are by default),
write timestamps and TTL info will not be retrieved. CQLKeyColumnValueStore,
however, retrieves them anyways, and simply discard them. This commit makes
CQL driver fetches them only when needed.

It also enhances relevant documentation so that users know how to make use of
these metadata if needed.

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/configs/janusgraph-cfg.md
+++ b/docs/configs/janusgraph-cfg.md
@@ -512,9 +512,9 @@ Meta data to include in storage backend retrievals
 
 | Name | Description | Datatype | Default Value | Mutability |
 | ---- | ---- | ---- | ---- | ---- |
-| storage.meta.[X].timestamps | Whether to include timestamps in retrieved entries for storage backends that automatically annotated entries with timestamps | Boolean | false | GLOBAL |
-| storage.meta.[X].ttl | Whether to include ttl in retrieved entries for storage backends that support storage and retrieval of cell level TTL | Boolean | false | GLOBAL |
-| storage.meta.[X].visibility | Whether to include visibility in retrieved entries for storage backends that support cell level visibility | Boolean | true | GLOBAL |
+| storage.meta.[X].timestamps | Whether to include timestamps in retrieved entries for storage backends that automatically annotated entries with timestamps. If enabled, timestamp can be retrieved by `element.value(ImplicitKey.TIMESTAMP.name())` or equivalently, `element.value("~timestamp")`. | Boolean | false | GLOBAL |
+| storage.meta.[X].ttl | Whether to include ttl in retrieved entries for storage backends that support storage and retrieval of cell level TTL. If enabled, ttl can be retrieved by `element.value(ImplicitKey.TTL.name())` or equivalently, `element.value("~ttl")`. | Boolean | false | GLOBAL |
+| storage.meta.[X].visibility | Whether to include visibility in retrieved entries for storage backends that support cell level visibility. If enabled, visibility can be retrieved by `element.value(ImplicitKey.VISIBILITY.name())` or equivalently, `element.value("~visibility")`. | Boolean | true | GLOBAL |
 
 ### tx
 Configuration options for transaction handling
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
@@ -28,6 +28,7 @@ import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.querybuilder.relation.Relation;
 import com.datastax.oss.driver.api.querybuilder.schema.CreateTableWithOptions;
 import com.datastax.oss.driver.api.querybuilder.schema.compaction.CompactionStrategy;
+import com.datastax.oss.driver.api.querybuilder.select.Select;
 import com.datastax.oss.driver.internal.core.cql.ResultSets;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
 import io.vavr.Tuple;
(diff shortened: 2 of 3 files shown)


li-boxuan:
This change should theoretically improve the performance because we retrieve less information from Cassandra. My simple benchmark test on my laptop shows no observable difference though.

At least, we could inspect the returned value from Cassandra by setting a debug point in StaticArrayEntryList::of method:

The below screenshot shows the master (old) behavior:
master

The below screenshot shows the new behavior:
without_fetching_metadata

We could see that less data is fetched from Cassandra now.

farodin91 (approved):
LGTM! Thank you @li-boxuan.

docs/configs/janusgraph-cfg.md
```diff
@@ -512,9 +512,9 @@ Meta data to include in storage backend retrievals
 
 | Name | Description | Datatype | Default Value | Mutability |
 | ---- | ---- | ---- | ---- | ---- |
-| storage.meta.[X].timestamps | Whether to include timestamps in retrieved entries for storage backends that automatically annotated entries with timestamps | Boolean | false | GLOBAL |
-| storage.meta.[X].ttl | Whether to include ttl in retrieved entries for storage backends that support storage and retrieval of cell level TTL | Boolean | false | GLOBAL |
-| storage.meta.[X].visibility | Whether to include visibility in retrieved entries for storage backends that support cell level visibility | Boolean | true | GLOBAL |
+| storage.meta.[X].timestamps | Whether to include timestamps in retrieved entries for storage backends that automatically annotated entries with timestamps. If enabled, timestamp can be retrieved by `element.value("~timestamp")`. | Boolean | false | GLOBAL |
```
rngcntr, line 515: Although I see no reason why e.g. `"~timestamp"` could be renamed in the future, I would still prefer the docs to refer to `ImplicitKey.TIMESTAMP.name()` instead. Otherwise, people would start writing queries which use the hard coded value as proposed here. I think we would be more flexible if that's not the case.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -150,43 +153,38 @@ public CQLKeyColumnValueStore(final CQLStoreManager storeManager, final String t
...
+        final Select getSliceSelect = selectFrom(this.storeManager.getKeyspaceName(), this.tableName)
+            .column(COLUMN_COLUMN_NAME)
+            .column(VALUE_COLUMN_NAME)
+            .where(
+                Relation.column(KEY_COLUMN_NAME).isEqualTo(bindMarker(KEY_BINDING)),
+                Relation.column(COLUMN_COLUMN_NAME).isGreaterThanOrEqualTo(bindMarker(SLICE_START_BINDING)),
+                Relation.column(COLUMN_COLUMN_NAME).isLessThan(bindMarker(SLICE_END_BINDING))
+            )
+            .limit(bindMarker(LIMIT_BINDING));
+        this.getSlice = this.session.prepare(addTTLFunction(addTimestampFunction(getSliceSelect)).build());
```
rngcntr, line 165: Nit: `getSliceSelect.addTimestampFunction().addTTLFunction()` would be more readable and also reduce nesting. It doesn't get more nitpicky than this of course ;)

rngcntr (approved):
Nice PR, thanks!

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
li-boxuan: Unfortunately, I cannot do that since the Select class is from DataStax driver package :(

li-boxuan: unless I make a wrapper over Select to enable chaining, but I doubt that might make this even more complex.

docs/configs/janusgraph-cfg.md
li-boxuan: Good idea, I'll add it and keep both in the doc. Some might prefer using string literal while others not.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
rngcntr: Oh you're right. Just leave it as it is, then.

docs/configs/janusgraph-cfg.md
rngcntr: The string literal will be updated in the docs, but it will not be updated in the user's queries once they are written.
If a user calls `ImplicitKey.TIMESTAMP.name()` in their code and it doesn't exist anymore, they will get a compiler error. But if they use the string literal, queries could fail silently, which makes it way harder to detedt.

li-boxuan: Understood. Using ImplicitKey, however, means they need to import this constant. Not sure if this class is auto-imported in the gremlin console but I would imagine in some cases users might encounter this trouble. Now we provide both choices in the doc so that they can choose whichever benefits them.

Anyways, I don't think there is any chance to rename this value, unless we decide to deprecate it in the future (e.g. visibility, which is currently not available in CQL/Hbase/Berkeley/InMemory storage). Renaming could cause backward incompatibility with minimal or no benefit. In case that happens, we will announce it as a breaking change in the changelog so that users would be notified.

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2575, opened 2021-04-17
General partial code cleanup

porunov:
Mostly stylistic code cleanups produced by IntelijIDEA. 
Applied only some cleanups which I checked.

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
```diff
@@ -113,7 +113,7 @@ protected void validateIndexStatus() {
         }
         Preconditions.checkArgument(isValidIndex, "The index %s is in an invalid state and cannot be indexed. %s", indexName, invalidIndexHint);
         // TODO consider retrieving the current Job object and calling killJob() if !isValidIndex -- would be more efficient than throwing an exception on the first pair processed by each mapper
-        log.debug("Index {} is valid for re-indexing");
+        log.debug("Index {} is valid for re-indexing", index.name());
```
li-boxuan, line 116: index.name() will lead to a backend call.

porunov: Thanks for noticing it!
I can wrap into `if`. Something like:
```
if(log.isDebugEnabled()){
  log.debug("Index "+index.name()+" is valid for re-indexing");
}
```
Do you think that's OK or do you think we should simply remove / change the log message?

porunov: @li-boxuan I wrapped it into `if` block as commented above

janusgraph-core/src/main/java/org/janusgraph/util/system/BackgroundThread.java
```diff
@@ -138,7 +138,7 @@ public void close(Duration duration) {
         try {
             join(maxWaitMs);
         } catch (InterruptedException e) {
-            log.error("Interrupted while waiting for thread {} to join",e);
+            log.error("Interrupted while waiting for thread {} to join", Thread.currentThread().getName(), e);
```
li-boxuan, line 141: What about just "getName()"?

li-boxuan (approved):
Thank you! I just left one nitpick, else looks good to me.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
```diff
@@ -549,8 +549,7 @@ public void testCacheConcurrency() throws InterruptedException {
         System.out.println("Hits: " + (getEdgeCacheRetrievals()-getEdgeCacheMisses()));
         System.out.println("Misses: " + getEdgeCacheMisses());
         assertEquals(numReads, lookups.get());
-        final int numLookupPropertyConstraints = numV;
-        assertEquals(2 * numReads + numV + numLookupPropertyConstraints, getEdgeCacheRetrievals());
+        assertEquals(2 * numReads + numV + numV, getEdgeCacheRetrievals());
```
rngcntr, line 552: This is a test so I think the additional variable is only there as an explanation of _why_ `numV` is added twice. Removing that would make me wonder why `numV` appears twice here.

janusgraph-backend-testutils/src/main/java/org/janusgraph/olap/OLAPTest.java
```diff
@@ -424,7 +424,7 @@ public boolean requiresVertexPropertyAddition() {
     public static class DegreeCounter extends StaticVertexProgram<Integer> {
 
         public static final String DEGREE = "degree";
-        public static final MessageCombiner<Integer> ADDITION = (a,b) -> a+b;
+        public static final MessageCombiner<Integer> ADDITION = Integer::sum;
```
rngcntr, line 427: I think we can also drop this and just use `Integer::sum` directly.

rngcntr (commented):
LGTM!

rngcntr (approved):

porunov:
Thank you @li-boxuan and @rngcntr ! Fixed those comments

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java
```diff
@@ -157,9 +157,9 @@ public static Mode parse(String mode) {
      * TODO Rename ZOOKEEPER_URL and "zookeeper-url" to ZOOKEEPER_URLS and
      * "zookeeper-urls" in future major releases.
      */
-    public static final ConfigOption<String[]> ZOOKEEPER_URL = new ConfigOption<String[]>(SOLR_NS,"zookeeper-url",
-            "URL of the Zookeeper instance coordinating the SolrCloud cluster",
-            ConfigOption.Type.MASKABLE, new String[] { "localhost:2181" });
+    public static final ConfigOption<String[]> ZOOKEEPER_URL = new ConfigOption<>(SOLR_NS, "zookeeper-url",
+        "URL of the Zookeeper instance coordinating the SolrCloud cluster",
```
li-boxuan, line 161: Can you revert this and the below indentation changes? It seems this indentation was by intention everywhere for CofigOptions. I would suggest fixing all or none.

porunov: Good catch! Fixed

PR #2589, opened 2021-04-25
Add checkstyle support

farodin91:
I used jenkins checkstyle config as base set and deactivated some rules to reduce the amount of changes. 

-----

### For all changes:

### For code changes:


porunov (approved):
LGTM. Thank you @farodin91 !

pom.xml
```diff
@@ -280,6 +282,66 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>com.puppycrawl.tools</groupId>
+                        <artifactId>checkstyle</artifactId>
+                        <version>8.41.1</version>
+                    </dependency>
+                </dependencies>
+                <!-- Version specified in parent POM -->
```
li-boxuan, line 295: What does this comment mean?

pom.xml
```diff
@@ -280,6 +282,66 @@
...
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>com.puppycrawl.tools</groupId>
+                        <artifactId>checkstyle</artifactId>
+                        <version>8.41.1</version>
```
li-boxuan, line 292: (nitpick) Do you want to move the version number to the properties section?

pom.xml
```diff
@@ -280,6 +282,66 @@
...
+                    </dependency>
+                </dependencies>
+                <!-- Version specified in parent POM -->
+                <configuration>
```
rngcntr, line 296: Nit: We can configure checkstyle in an external XML, by which we can avoid flooding the `pom.xml`.

rngcntr (approved):
LGTM, just one suggestion

pom.xml
farodin91: I don't see any reason why we will ever use it twice.

pom.xml
farodin91: i can try

pom.xml
li-boxuan: Never mind... I somehow thought dependabot would only be able to bump version if it's defined in properties section.

pom.xml
farodin91: You can put into a xml, i don't really like because makes maven a bit more complexer than just inline. The Checker is executed in each sub project and from their you have to find the way to a central config.

rngcntr: If that causes too much config overhead, you can also leave it in `pom.xml`

li-boxuan (approved):

PR #2591, opened 2021-04-26
Make sure timestamp_provider is set before anything else

### LINKED ISSUE #2590: Storage layer preferred TIMESTAMP_PROVIDER not used for initial setup from local file
## Description

- Version: tested with 0.3.1 and 0.4.0 but the code is still percent in master
- Storage Backend: tested with HBase but affects all

When initializing a new graph where the storage layer has a preferred `TIMESTAMP_PROVIDER` value different than the default, the initial settings are stored using the wrong timestamp resolution.

I found this bug trying to delete some GLOBAL_OFFLINE settings in a deployment using HBase as the storage layer. HBase was set to use _milliseconds_ timestamps but the graph settings specified in the `.properties` file were being saved using _microseconds_, which is the default value for `graph.timestamps`.

### Expected Behavior
All configuration values should be set using the preferred storage layer timestamp resolution when available.

### Current Behavior
Initial configuration values always use the default timestamp provider resolution.

### Steps to Reproduce
With HBase running locally using milliseconds timestamp resolution (the default).
1. Create a new graph with: `JanusGraphFactory.build().set("storage. backend","hbase").open()`
2. List the content of the `janusgraph` table with `scan "janusgraph"`

_Expected:_ timestamps are milliseconds values.
_Actual:_ timestamps are microseconds values.

BrunoBerisso:
Fix #2590 

-----

### For all changes:

### For code changes:


janusgraph-test/src/test/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilderTest.java
```diff
@@ -164,6 +165,20 @@ public void shouldSetupBackendPreferenceTimestampProvider() {
         verify(globalWrite).set(TIMESTAMP_PROVIDER, timestampProviders);
     }
 
+    @Test
+    public void shouldSetupBackendPreferenceTimestampProviderBeforeAnythingElse() {
```
rngcntr, line 169: Maybe an admin can disable the Codacy check here. We should probably even disable it globally, because there are lots of cases where a test is perfectly fine without `assert()` or `fail()`.

rngcntr (commented):
Nice and quick fix, thank you @BrunoBerisso!
Are you aware whether or not this change can affect running installations for customers? I didn't dive into it deep enough to fully understand when and how the timestamps are (de-)serialized, but I can imagine that it causes trouble if a user updates JanusGraph and it suddenly expects a different timestamp format.

BrunoBerisso:
Hi @rngcntr
Thank you for the review.

> Are you aware whether or not this change can affect running installations for customers? (...)

I think existing graphs won't be affected because this logic only applies when the settings are not frozen (`!globalWrite.isFrozen()`). After the value of `TIMESTAMP_PROVIDER` is set, the configuration is freeze (`globalWrite.freezeConfiguration()`) so this should only happen once in the lifetime of a graph.

twilmes (approved):

janusgraph-test/src/test/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilderTest.java
porunov: Disabled this check

porunov: This check is disabled now in Codacy but it still complains about this issue. I guess, it will be resolved later. Nevertheless, Codacy never blocks a PR from being merged.

porunov (approved):
LGTM. Thank you @BrunoBerisso !

janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilder.java
```diff
@@ -67,12 +67,12 @@ public ReadConfiguration buildGlobalConfiguration(ReadConfiguration localConfig,
             ModifiableConfiguration globalWrite = modifiableConfigurationBuilder.buildGlobalWrite(keyColumnValueStoreConfiguration);
 
             if (!globalWrite.isFrozen()) {
+                setupTimestampProvider(globalWrite, localBasicConfiguration, storeManager);
```
li-boxuan, line 70: Can you help me understand why this line must be put before `globalWrite.setAll(getGlobalSubset(localBasicConfiguration.getAll()));`?

By looking at the code, it seems to me that the order of these two lines does not matter.

rngcntr (dismissed):
@BrunoBerisso thanks for clarification and for the hint to `isFrozen()`!

janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/builder/ReadConfigurationBuilder.java
porunov: Yesterday I thought I was seeing the difference. For some reason I thought that the value from `localBasicConfiguration` was overwritten by `TIMESTAMP_PROVIDER.getDefaultValue()` but now checking again I'm trying to understand the difference and I don't see it anymore. Probably I missed something.
@BrunoBerisso could you please explain the difference in the behavior when we moved this line before `globalWrite.setAll(getGlobalSubset(localBasicConfiguration.getAll()));`?

BrunoBerisso: Sorry, this change doesn't fix #2590 I didn't test them properly before opening the PR.

I'll update the fix to use the correct`TIMESTAMP_PROVIDER` when creating the `KCVSConfiguration` instance used to write these settings.

li-boxuan: Sure, I've dismissed approvals in case of an accidental merge.

porunov:
@BrunoBerisso I was thinking to start preparation for 0.6.0 release soon. Do you think you need more time to finalize this PR? If you think you don't need much time to finalize this PR I'm happy to wait to include it into 0.6.0 release. Otherwise I'm thinking to retarget this PR for 0.6.1 release if you are OK with it.

porunov:
I think to remove this PR from milestone on Monday unless there are any objections

PR #2600, opened 2021-05-01
Manual cleanup

farodin91:
* Remove comment out stuff (mostly over 6 years comment out)
* Remove unused classes

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/QueryContainer.java
```diff
@@ -118,14 +112,7 @@ private QueryBuilder() {
...
 
-        private Query relations(RelationCategory returnType) {
-//            if (name==null) {
-//                if (hasSingleType()) name = getSingleType().name();
-//                else if (!requiresName) name = QUERY_NAME_PREFIX + queries.size();
-//                else throw new IllegalStateException("Need to specify an explicit name for this query");
-//            }
-
-            BaseVertexCentricQuery vq = super.constructQuery(returnType);
+        private Query relations(RelationCategory returnType) {            BaseVertexCentricQuery vq = super.constructQuery(returnType);
```
rngcntr, line 115: Split this line, please

rngcntr (changes requested):

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/QueryContainer.java
farodin91: good catch

rngcntr: +1 -1920 lines
The +1 made me curious :D

janusgraph-core/src/main/java/org/janusgraph/graphdb/util/AllEdgesIterable.java
```diff
@@ -1,92 +0,0 @@
...
-
-/**
- * Constructs {@link Iterable}s over all {@link org.janusgraph.core.JanusGraphEdge}s connecting a provided set of vertices.
- * <p>
- * Given a set of vertices, one may be interested in all edges that are contained in the subgraph spanned
- * by those vertices.
- *
- * @author Matthias Br&ouml;cheler (me@matthiasb.com);
- */
-public class AllEdgesIterable {
```
li-boxuan, line 31: I think the AllEdgesIterable and other classes under this package might be used by users directly.

li-boxuan: Should we mark them as deprecated in this release and maybe remove them in the next one?

farodin91: I think it is a good idea to deprecated them. They are not maintained over years, not tested and not documented.

farodin91: classes are deprecated

rngcntr (approved):

PR #2601, opened 2021-05-01
[Checkstyle] Optimize imports

farodin91:
* AvoidStarImport
* CustomImportOrder
  1. all other imports
  3. java imports
  4. static imports

Backed by intellij
-----

### For all changes:

### For code changes:


porunov:
@farodin91 I am just curious, do you know why do we need janusgraph imports after third party imports? Why don't we simply use alphabetical ordering for all imports? I've seen such technique in multiple projects but don't know the reason behind such structure.
Most IDE use alphabetical ordering by default. If so, by enforcing such rules we will see a lot of commits from different contributors which break this rule and then checkstyle will fail probably forcing them to fix imports ordering.
I guess, that's fine if we have a reason for such ordering but otherwise it's probably additional complexity which don't give any benefits.

farodin91:
> why do we need janusgraph imports after third party imports?

Just personal preference, i think it is really cool to see all janusgraph imports in one group.


> Why don't we simply use alphabetical ordering for all imports?
> Most IDE use alphabetical ordering by default. 

Default ordering of intellij: (This grouping seems to the most common, sometimes differently ordered.)

- thrid party here including jg
- java and javax
- static imports

> If so, by enforcing such rules we will see a lot of commits from different contributors which break this rule and then checkstyle will fail probably forcing them to fix imports ordering.

Currently is the ordering all over the place.  
I think having automate checkstyle which automatically executed each maven run, shouldn't lead to a lot commits that fails.

porunov:
@farodin91 I would prefer to not enforcing JanusGraph imports to be in a separate group if IDE will break it. If IDE are configured in this PR as well then I'm good with it.
By enforcing such a rule, we also need to make sure that all or most common Java IDE (like Intelij IDEA, Eclipse) are configured to automatically place those imports in correct groups. Otherwise I'm almost 100% sure it will be just disturbing and will cause just more problems.
I would suggest any of the following:
- Make sure that IDEs are configured for such behavior.
- Do not enforce ordering.
- Enforce default ordering.

I didn't check this PR yet. Maybe it configures import ordering. Let me know what you think about ordering

farodin91:
I used merge org.janusgraph back into thrid parties. I also added editorconfig support for import layout, this will lead to enforcement in intellij

rngcntr (approved):

porunov (approved):
I checked changes in `.editorconfig`, ` pom.xml` and several `.java` files. I didn't check all `.java` files as I believe all those changes was produces automatically.
The changes LGTM. Thank you @farodin91 !

farodin91:
@porunov I will create CTR commit to fix build issue on master.

farodin91:
https://github.com/JanusGraph/janusgraph/commit/fd4ce9a72095e8621ae5b8b5fa4c308ea43fffb0

PR #2633, opened 2021-05-29
Cleanup JanusGraphSettings

farodin91:
Follow up of upgrading to tp 3.5

-----

### For all changes:

### DIFF
--- a/janusgraph-server/src/main/java/org/janusgraph/graphdb/server/JanusGraphSettings.java
+++ b/janusgraph-server/src/main/java/org/janusgraph/graphdb/server/JanusGraphSettings.java
@@ -15,54 +15,32 @@
 package org.janusgraph.graphdb.server;
 
 import org.apache.tinkerpop.gremlin.server.Settings;
+import org.janusgraph.graphdb.server.util.JanusGraphSettingsUtils;
 import org.yaml.snakeyaml.TypeDescription;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
-import org.yaml.snakeyaml.representer.Representer;
 
-import java.io.ByteArrayInputStream;
 import java.io.FileInputStream;
-import java.io.IOException;
 import java.io.InputStream;
-import java.util.Map;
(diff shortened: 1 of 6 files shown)


janusgraph-server/src/main/java/org/janusgraph/graphdb/server/JanusGraphSettings.java
```diff
@@ -15,54 +15,32 @@
...
 
-    // TODO: Merge JanusGraphSettings with GremlinSettings which will be released in TinkerPop 3.4.11 and ongoing
-    public static JanusGraphSettings read(final InputStream stream) throws IOException {
+    public static JanusGraphSettings read(final InputStream stream) {
         Objects.requireNonNull(stream);
         Constructor constructor = new Constructor(JanusGraphSettings.class);
         TypeDescription grpcServerSettings = new TypeDescription(JanusGraphSettings.GrpcServerSettings.class);
         constructor.addTypeDescription(grpcServerSettings);
-        Representer representer = new Representer();
-        representer.getPropertyUtils().setSkipMissingProperties(true);
```
porunov, line 46: Do you know why are we removing this? Won't it lead to possible problems during Yaml parsing?

porunov (commented):
I have just a small question but otherwise LGTM

janusgraph-server/src/main/java/org/janusgraph/graphdb/server/JanusGraphSettings.java
farodin91: The code should behave as before. 

Old behavior: parse yaml for janusgraph Settings, than remove grpc setting and let the gremlin parse it again which breaks if doesn't know a property.

New behavior: single parser for JanusGraph server settings and gremlin server settings. It will also breaks if doesn't know a property.

porunov: Thanks for the clarification @farodin91 !

porunov (approved):
LGTM

PR #2640, opened 2021-06-02
Break vertices selection loop for properties prefetching when cache size is reached

porunov:
Break vertices selection loop for properties prefetching when cache size is reached

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphVertexStep.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphVertexStep.java
@@ -194,13 +194,14 @@ public class JanusGraphVertexStep<E extends Element> extends VertexStep<E> imple
             result = (Vertex.class.isAssignableFrom(getReturnClass())) ? query.vertices() : query.edges();
         }
 
-        if (batchPropertyPrefetching) {
+        if (batchPropertyPrefetching && txVertexCacheSize > 1) {
             Set<Vertex> vertices = new HashSet<>();
-            result.forEach(v -> {
-                if (vertices.size() < txVertexCacheSize ) {
-                    vertices.add((Vertex) v);
+            for(JanusGraphElement v : result){
+                vertices.add((Vertex) v);
+                if (vertices.size() >= txVertexCacheSize) {
+                    break;
                 }
-            });
+            }
 
             // If there are multiple vertices then fetch the properties for all of them in a single multiQuery to
             // populate the vertex cache so subsequent queries of properties don't have to go to the storage back end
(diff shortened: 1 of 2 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphVertexStep.java
```diff
@@ -195,25 +193,53 @@ private void setParentMultiQueryStep() {
...
+        ((BasicVertexCentricQueryBuilder) propertyMultiQuery).profiler(queryProfiler);
+
+        propertyMultiQuery.addVertex((Vertex) firstVertex);
+
+        int addedVertices = 1;
+
+        do {
+            propertyMultiQuery.addVertex((Vertex) resultIt.next());
+            ++addedVertices;
+        } while (addedVertices < txVertexCacheSize && resultIt.hasNext());
```
li-boxuan, line 238: Let's say `txVertexCacheSize == 10`, and there are 15 results. If the first 10 results are identical, `addedVertices` will reach 10 and then break the loop, even though only one unique vertex is included for batch prefetching.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphVertexStep.java
```diff
@@ -195,25 +193,53 @@ private void setParentMultiQueryStep() {
         }
 
         if (batchPropertyPrefetching) {
-            Set<Vertex> vertices = new HashSet<>();
-            result.forEach(v -> {
-                if (vertices.size() < txVertexCacheSize ) {
-                    vertices.add((Vertex) v);
-                }
-            });
```
li-boxuan, line 203: If I am understanding correctly, you are trying to optimize this for loop. Shall we do this instead?

```java
for (Vertex v : result) {
      if (vertices.size() < txVertexCacheSize ) {
            vertices.add((Vertex) v);
      } else {
            break;
      }
}
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphVertexStep.java
porunov: Thanks for catching this mistake! You are totally right

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphVertexStep.java
porunov: Updated the PR

li-boxuan (approved):
LGTM 👍

porunov:
Added a related test to have a better test coverage.

PR #1436, opened 2019-02-21
Support CQL for OLAP

FlorianHockmann:
This adds a `CqlInputFormat` to enable Spark with CQL without relying on Thrift any more.

@farodin91 and I worked together on this one.

We already tested this in production with our graph with ~250 M vertices and could successfully execute some traversals on our Scylla backend with Spark.

Fixes #985.

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there an issue associated with this PR? Is it referenced in the commit message?
- [x] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] Have you written and/or updated unit tests to verify your changes?
- [-] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [-] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [-] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [-] Have you ensured that format looks appropriate for the output in which it is rendered?
- [-] If this PR is a documentation-only change, have you added a `[skip ci]`
  tag to the first line of your commit message to avoid spending CPU cycles in
  Travis CI when no code, tests, or build configuration are modified?

### Note:
Please ensure that once the PR is submitted, you check Travis CI for build issues and submit an update to your PR as soon as possible.

PR #2649, opened 2021-06-09
Use all hosts available in CqlBinaryInputFormat config

li-boxuan:
When setting up connections to Cassandra cluster for OLAP traversal,
it is recommended to provide all contact points, such that even if
some of them are unavailable at that time, the connections can still
be established.

This has been running for some time in our production cluster. So far, so good.
I also verified that even if one Cassandra node is not available when submitting
the MapReduce job, everything still works smoothly.

This is a follow-up of https://github.com/JanusGraph/janusgraph/pull/1436#discussion_r629018494

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/hadoop/formats/cql/CqlBinaryInputFormat.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/hadoop/formats/cql/CqlBinaryInputFormat.java
@@ -64,7 +64,7 @@ public class CqlBinaryInputFormat extends AbstractBinaryInputFormat {
         super.setConf(config);
 
         // Copy some JanusGraph configuration keys to the Hadoop Configuration keys used by Cassandra's ColumnFamilyInputFormat
-        ConfigHelper.setInputInitialAddress(config, janusgraphConf.get(GraphDatabaseConfiguration.STORAGE_HOSTS)[0]);
+        ConfigHelper.setInputInitialAddress(config, String.join(",", janusgraphConf.get(GraphDatabaseConfiguration.STORAGE_HOSTS)));
         if (janusgraphConf.has(GraphDatabaseConfiguration.STORAGE_PORT))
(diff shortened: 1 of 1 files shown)


janusgraph-cql/src/main/java/org/janusgraph/hadoop/formats/cql/CqlBinaryInputFormat.java
```diff
@@ -64,7 +65,7 @@ public void setConf(final Configuration config) {
         super.setConf(config);
 
         // Copy some JanusGraph configuration keys to the Hadoop Configuration keys used by Cassandra's ColumnFamilyInputFormat
-        ConfigHelper.setInputInitialAddress(config, janusgraphConf.get(GraphDatabaseConfiguration.STORAGE_HOSTS)[0]);
+        ConfigHelper.setInputInitialAddress(config, Joiner.on(",").join(janusgraphConf.get(GraphDatabaseConfiguration.STORAGE_HOSTS)));
```
FlorianHockmann, line 68: (nitpick) Can't we simply use:

 ```java
String.join(",", janusgraphConf.get(GraphDatabaseConfiguration.STORAGE_HOSTS))
``` 

Just so we don't need a third party library for this?

FlorianHockmann (approved):
LGTM, I only have one small question / suggestion.

janusgraph-cql/src/main/java/org/janusgraph/hadoop/formats/cql/CqlBinaryInputFormat.java
li-boxuan: Thanks, we use guava in our codebase a lot so I used it without any particular reason. Plus, I wasn't aware of this API 😄

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2660, opened 2021-06-13
Optimize getHasContainers usage

porunov:
Calls to `getHasContainers` of `JanusGraphStep` create new `ArrayList` and also spends time to populate that list. So, this commit tries to reduce calls to this method when not needed (i.e. reuse already computed values). Also, this PR reduces unnecessary creation of some collections (no need to create lists when they are used only once as Iterable. It would make sense to create lists if they are used more than once).

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/HasStepFolder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/HasStepFolder.java
@@ -49,7 +49,6 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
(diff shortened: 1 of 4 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexFilterOptimizerStrategy.java
```diff
@@ -111,12 +111,13 @@ private OptimizableQueryType analyzeSubSteps(List<Step> steps) {
         } else if (steps.get(1) instanceof HasStep) {
             // Check if this filter traversal matches the pattern: _.inV/outV/otherV.hasId(x)
             HasStep hasStep = (HasStep) steps.get(1);
-            if (hasStep.getHasContainers().size() != 1) {
+            List hasContainers = hasStep.getHasContainers();
```
li-boxuan, line 114: nitpick:
```suggestion
            List<HasContainer> hasContainers = hasStep.getHasContainers();
```
so that you don't need to do casting on line 120, i.e. `(HasContainer) hasContainers.get(0);`

li-boxuan (approved):
LGTM, except one nitpick

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AdjacentVertexFilterOptimizerStrategy.java
porunov: Thank you for suggestion. Resolved

li-boxuan (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/HasStepFolder.java
```diff
@@ -168,13 +167,13 @@ static void foldInHasContainer(final HasStepFolder janusgraphStep, final Travers
...
-                final Iterable<HasContainer> containers = ((HasContainerHolder) currentStep).getHasContainers().stream().map(JanusGraphPredicateUtils::convert).collect(Collectors.toList());
-                if  (validFoldInHasContainer(currentStep, true)) {
-                    janusgraphStep.addAll(containers);
-                    currentStep.getLabels().forEach(janusgraphStep::addLabel);
-                    traversal.removeStep(currentStep);
-                }
+            } else if (currentStep instanceof HasContainerHolder && validFoldInHasContainer(currentStep, true)){
+                final HasContainerHolder currentHasContainerHolder = (HasContainerHolder) currentStep;
+                final Iterable<HasContainer> containers = () -> currentHasContainerHolder.getHasContainers()
+                    .stream().map(JanusGraphPredicateUtils::convert).iterator();
```
rngcntr, line 173: (somewhat) nitpick: I remember you once discovered that unnecessary usage of `stream()` can cause computation overhead. You could easily avoid streaming here and replace it by `currentHasContainerHolder.getHasContainers().forEach(...)`.
`JanusGraphStep` only has an `addAll()` method, but no `add()`. Maybe it would make sense to implement one.

rngcntr (approved):
I'm fine with merging this, but I have one comment.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/HasStepFolder.java
porunov: I didn't want to put `add()` implementation into this PR but it makes sense to implement `add()` method and change the logic to `currentHasContainerHolder.getHasContainers().forEach(...)` but I would prefer to do that in a separate PR. We already used `stream` here

rngcntr: Alright, thanks for clarification!

PR #2681, opened 2021-06-24
Fix deserializing multiple self-loop edges

### LINKED ISSUE #2669: Missing self-loop edges when reading graph data using Apache Spark
# Summary

 - Version: 0.5.3
 - Storage backend: cql
 - Link to discussed bug: https://lists.lfaidata.foundation/g/janusgraph-users/topic/83394832
 - Expected behavior: using `vertex.edges(Direction.OUT)` on vertices read in Apache Spark should return all outbound edges of that vertex
 - Current behavior: using `vertex.edges(Direction.OUT)` returns all non-looping edges and only one self-loop edge, ignoring other self-loop edges if there are multiple

# Description

While debugging some Apache Spark jobs that process data from a Janusgraph graph. i noticed some issues with self-loop edges (edges that connect a vertex to itself). The data is read using:

`javaSparkContext.newAPIHadoopRDD(hadoopConfiguration(), CqlInputFormat.class, NullWritable.class, VertexWritable.class)`

When I try to process all outbound edges of a single vertex in this RDD using:

`vertex.edges(Direction.OUT)`

and that vertex has multiple self-loop edges with the same edge label, the iterator always returns only one such edge. Edges that are not self-loop are all returned as expected.

To give a specific example, if I have a vertex V0 with edges that E1, E2, E3, E4, E5 that lead to vertices V1, V2, V3, V4, V5, the call `vertex.edges(Direction.OUT)` will return an iterator that iterates over all five edges. However, if I have a vertex V0 with edges E1, E2, E3 that lead to V1, V2, V3, and self-loop edges EL1, EL2, EL3, the iterator will iterate over E1, E2, E3, and only one of (EL1, EL2, EL3), giving a total of four edges instead of the expected six.

After further analysis, I came upon this commit:

https://github.com/JanusGraph/janusgraph/commit/d3006dc939c1b640bb263806abd3fd6bee630d12

which explicitly added code that skips deserializing multiple self-loop edges. The code from the linked commit is still present in 

https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java

and seems to be the cause of this unexpected behavior.

mmarovic:
Fixes #2669

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
+++ b/janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
@@ -36,6 +36,7 @@ import org.janusgraph.graphdb.relations.RelationCache;
 import org.janusgraph.graphdb.types.TypeInspector;
 import org.janusgraph.hadoop.formats.util.input.JanusGraphHadoopSetup;
 import org.janusgraph.hadoop.formats.util.input.SystemTypeInspector;
+import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -59,14 +60,13 @@ public class JanusGraphVertexDeserializer implements AutoCloseable {
         this.idManager = setup.getIDManager();
     }
 
+    private boolean edgeExists(Vertex vertex, RelationType type, RelationCache possibleDuplicate) {
+        Iterator<Edge> it = vertex.edges(possibleDuplicate.direction, type.name());
 
-    private static Boolean isLoopAdded(Vertex vertex, String label) {
-        Iterator<Vertex> adjacentVertices = vertex.vertices(Direction.BOTH, label);
+        while (it.hasNext()) {
+            Edge edge = it.next();
 
-        while (adjacentVertices.hasNext()) {
-            Vertex adjacentVertex = adjacentVertices.next();
-
-            if(adjacentVertex.equals(vertex)){
+            if (edge.id().equals(possibleDuplicate.relationId)) {
                 return true;
             }
         }
@@ -158,8 +158,8 @@ public class JanusGraphVertexDeserializer implements AutoCloseable {
                     // We don't know the label of the other vertex, but one must be provided
                     TinkerVertex adjacentVertex = getOrCreateVertex(relation.getOtherVertexId(), null, tg);
 
-                    // handle self-loop edges
-                    if (tv.equals(adjacentVertex) && isLoopAdded(tv, type.name())) {
+                    // skip self-loop edges that were already processed, but from a different direction
+                    if (tv.equals(adjacentVertex) && edgeExists(tv, type, relation)) {
                         continue;
                     }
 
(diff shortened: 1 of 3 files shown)


porunov:
@mmarovic Thank you for the fix! Could you please resolve conflicts and add a test to cover the bug?

mmarovic:
@porunov I added a test to cover the specific bug. Also, I changed the PR base to v0.5 because that's the original branch I based my branch on. However, this seems to cause issues with the codacity code analysis.

Should I change the base back to master and resolve conflicts or leave it as it is now?

porunov:
@mmarovic Thank you! `v0.5` uses Travis CI still but it's disabled now, so I'm not sure if `v0.5` will be supported. I would recommend to change back to `master` branch as we can run CI only on `master` branch right now (using GitHub Actions).

porunov:
@mmarovic I think you probably need to add a copy of your test `testReadMultipleSelfEdges` to `HBaseSnapshotInputFormatIT` and overwrite it. I guess you you will need to add a line:
```
        // Take a snapshot of the graph table
        hBaseContainer.createSnapshot(snapshotName, table);
```

similarly as it is in `testReadSelfEdge` (didn't check that).

P.S. That's definitely against DRY principle and we should refactor that but no need to do such refactoring in your PR.

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
```diff
@@ -59,14 +60,14 @@ public JanusGraphVertexDeserializer(final JanusGraphHadoopSetup setup) {
         this.idManager = setup.getIDManager();
     }
 
+    private Boolean edgeExists(Vertex vertex, RelationCache possibleDuplicate) {
```
porunov, line 63: Not need to return wrapper type. I would simply return `boolean` instead of `Boolean` here.

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
```diff
@@ -59,14 +60,14 @@ public JanusGraphVertexDeserializer(final JanusGraphHadoopSetup setup) {
...
+        final RelationType type = typeManager.getExistingRelationType(possibleDuplicate.typeId);
```
porunov, line 64: We already computed this `type` before, so no need to compute it again. Instead you could pass this `type` into this method. I.e. could be something like `private boolean edgeExists(Vertex vertex, RelationType type, RelationCache possibleDuplicate)`

porunov (approved):
LGTM. Thank you @mmarovic ! The fix makes sense to me I just added two optional nitpicks

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
mmarovic: I originally left it this way because the previous implemnetation also returned the wrapper type. However, I agree with you and will change it.

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
mmarovic: True, I'll change this as well.

porunov: Thank you! I see that the latest force push didn't update the commit, so I unresolved this commits

porunov (approved):

li-boxuan (approved):
LGTM 👍  Thank you for the fix!

PR #2684, opened 2021-06-27
Reduce redundant global centric query build

li-boxuan:
-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
@@ -266,7 +266,7 @@ public class GraphCentricQueryBuilder implements JanusGraphQuery<GraphCentricQue
         //Prepare constraints
         final MultiCondition<JanusGraphElement> conditions;
         if (this.globalConstraints.size() == 1) {
-            conditions = QueryUtil.constraints2QNF(tx, constraints);
+            conditions = QueryUtil.constraints2QNF(tx, globalConstraints.get(0));
             if (conditions == null) return GraphCentricQuery.emptyQuery(resultType);
         } else {
             conditions = new Or<>();
@@ -301,7 +301,7 @@ public class GraphCentricQueryBuilder implements JanusGraphQuery<GraphCentricQue
             indexLimit = Math.min(hardMaxLimit,
                 QueryUtil.adjustLimitForTxModifications(tx, conditions.numChildren() - coveredClauses.size(), indexLimit));
             query = new BackendQueryHolder<>(selectedIndex.getQuery().updateLimit(indexLimit),
-                    coveredClauses.size() == conditions.numChildren(), selectedIndex.isSorted());
+                    coveredClauses.size() == conditions.numChildren() || coveredClauses.contains(conditions), selectedIndex.isSorted());
         } else {
             query = new BackendQueryHolder<>(new JointIndexQuery(), false, selectedIndex.isSorted());
         }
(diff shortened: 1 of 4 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -301,7 +301,7 @@ public GraphCentricQuery constructQueryWithoutProfile(final ElementCategory resu
             indexLimit = Math.min(hardMaxLimit,
                 QueryUtil.adjustLimitForTxModifications(tx, conditions.numChildren() - coveredClauses.size(), indexLimit));
             query = new BackendQueryHolder<>(selectedIndex.getQuery().updateLimit(indexLimit),
-                    coveredClauses.size() == conditions.numChildren(), selectedIndex.isSorted());
+                    coveredClauses.containsAll(conditions) || coveredClauses.contains(conditions), selectedIndex.isSorted());
```
li-boxuan, line 304: when `conditions` is an `Or` object, and that `OR` condition is covered, then `coveredClauses` will only contain one `OR` object. In this case, `coveredClauses.size() == 1` while `conditions.numChildren() >= 1` (depending on how many children are nested within the `OR` condition. Thus, we need `coveredClauses.contains(conditions)`.

On the other hand, when `conditions` is an `And` object, then `coveredClauses` will contain the nested conditions that it cover. Under this scenario, `coveredClauses.size() == conditions.numChildren` is equivalent to `coveredClauses.containsAll(conditions)`.

This bug does not affect correctness but does harm to performance (extra in-memory filtering step is done in `QueryProcessor` even if unnecessary).

li-boxuan (commented):
This PR contains 3 fixes. None of the bugs impairs correctness, but all of them do harm to performance. Adding comments below to explain each of them.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
```diff
@@ -266,7 +266,7 @@ public GraphCentricQuery constructQueryWithoutProfile(final ElementCategory resu
         //Prepare constraints
         final MultiCondition<JanusGraphElement> conditions;
         if (this.globalConstraints.size() == 1) {
-            conditions = QueryUtil.constraints2QNF(tx, constraints);
+            conditions = QueryUtil.constraints2QNF(tx, globalConstraints.get(0));
```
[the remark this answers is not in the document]
rngcntr: Thanks for your explanation, I wouldn't have been able to understand these changes properly without your comments :+1:

rngcntr (approved):
Looks good, thank you very much!

rngcntr (changes requested):
It is only now that I see there are still test errors. Maybe I was too fast to approve this...

Solr issues look very minor (renamed `name` for `name_s` somewhere). What concerns me more is the failing `CQLOperationCountingTest`. Would you look into that, please?

li-boxuan:
@rngcntr Thanks for the reminder. Have fixed the Solr test failure. The CQLOperationCountingTest failure disappeared after a rerun.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/GraphCentricQueryBuilder.java
porunov: >  Under this scenario, coveredClauses.size() == conditions.numChildren is equivalent to coveredClauses.containsAll(conditions).

Just wanted to double check this statement. Does it mean that this condition can be replaced by the next condition?
```
coveredClauses.size() == conditions.numChildren() || coveredClauses.contains(conditions)
```
Probably not but I wanted to double check because if the result is the same than `equality` operation will be faster to compute than `containsAll` operation. 
Also, `containsAll` will return `true` if `conditions.size() == 0` or if simply the amount of `conditions` elements is smaller than the amount of `coveredClauses` elements but they all equal. Thus, wanted to double check that it is the expected behavior

li-boxuan: Case `conditions.size() == 0` and case `conditions.numChildren() < coveredClauses.size()` are never gonna happen in theory. I will change it back to `coveredClauses.size() == conditions.numChildren()` for performance reason.

rngcntr (approved):

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2688, opened 2021-06-29
Fix NPE when updating mixed index

### LINKED ISSUE #1876: Index fields of an used mixed index are not updated with new fields after adding a new index key
After modifing a **used** mixed index, the newly added key is not added into the index for current JanusGraph instance. New instances of JanusGraph will have the indexed field but current instance will not have this field. This behaviour isn't reproduced if we didn't use an index in traversal before modifing the index.

- Version: 0.5.0-SNAPSHOT
- Storage Backend: ScyllaDB 3.0
- Mixed Index Backend: ElasticSearch 6.8.0

### Unit test:

```
@Test
public void shouldUpdateIndexFieldsAfterIndexModification() throws InterruptedException, ExecutionException {

    String uniqueSuffix = "_s"+System.currentTimeMillis();
    String key1 = "testKey1"+uniqueSuffix;
    String key2 = "testKey2"+uniqueSuffix;
    String key3 = "testKey3"+uniqueSuffix;
    String vertexL = "testVertexLabel"+uniqueSuffix;
    String indexName = "testIndex"+uniqueSuffix;

    JanusGraphManagement mgmt = janusGraph.openManagement();

    mgmt.makePropertyKey(key1).dataType(Long.class).make();
    mgmt.makePropertyKey(key2).dataType(Long.class).make();
    mgmt.makeVertexLabel(vertexL).make();
    mgmt.commit();

    mgmt = janusGraph.openManagement();
    mgmt.buildIndex(indexName, Vertex.class).indexOnly(mgmt.getVertexLabel(vertexL)).addKey(mgmt.getPropertyKey(key1)).addKey(mgmt.getPropertyKey(key2)).buildMixedIndex("search");
    mgmt.commit();

    ManagementSystem.awaitGraphIndexStatus(janusGraph, indexName).call();
    mgmt = janusGraph.openManagement();
    mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.ENABLE_INDEX).get();
    mgmt.commit();

    mgmt = janusGraph.openManagement();
    PropertyKey testKey3 = mgmt.makePropertyKey(key3).dataType(Long.class).make();
    mgmt.addIndexKey(mgmt.getGraphIndex(indexName), testKey3);
    mgmt.commit();

    ManagementSystem.awaitGraphIndexStatus(janusGraph, indexName).status(SchemaStatus.REGISTERED, SchemaStatus.ENABLED).call();

    mgmt = janusGraph.openManagement();
    mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.REINDEX).get();
    mgmt.commit();

    JanusGraphVertex vertex = janusGraph.addVertex(vertexL);
    vertex.property(key1, 1L);
    vertex.property(key2, 2L);
    vertex.property(key3, 3L);

    janusGraph.tx().commit();
}
```

### Configs:
```
query.batch=true
query.fast-property=false
query.smart-limit=false
storage.batch-loading=false
storage.buffer-size=10240
storage.cql.batch-statement-size=200
schema.default = none
storage.connection-timeout=2400000
storage.read-time=2400000
storage.write-time=2400000
storage.cassandra.frame-size-mb=50
cassandra.thrift.framed.size_mb=50
storage.backend=cql
storage.hostname=cassandra.endpoint
storage.username=cassandra
storage.password=cassandra
storage.cassandra.astyanax.local-datacenter=LOCAL_DC
index.search.backend=elasticsearch
index.search.hostname=elasticsearch.endpoint
index.search.index-name=janusgraph
index.search.elasticsearch.create.sleep=10
index.search.elasticsearch.bulk-refresh=false
ids.block-size=1000000
ids.flush=false
```

### Stack Trace

```
""2019-11-01 13:27:07 [main] ERROR o.j.g.database.StandardJanusGraph - Could not commit transaction [8] due to exception
"java.lang.NullPointerException: null
	at org.janusgraph.graphdb.database.IndexSerializer.getIndexUpdates(IndexSerializer.java:332)
	at org.janusgraph.graphdb.database.StandardJanusGraph.prepareCommit(StandardJanusGraph.java:573)
	at org.janusgraph.graphdb.database.StandardJanusGraph.commit(StandardJanusGraph.java:727)
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.commit(StandardJanusGraphTx.java:1380)
	at org.janusgraph.graphdb.tinkerpop.JanusGraphBlueprintsGraph$GraphTransaction.doCommit(JanusGraphBlueprintsGraph.java:297)
	at org.apache.tinkerpop.gremlin.structure.util.AbstractTransaction.commit(AbstractTransaction.java:104)
	at com.janusgraphtest.JBCIndexTest.shouldUpdateIndexFieldsAfterIndexModification(JBCIndexTest.java:208)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:628)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:117)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:184)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:180)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:127)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:68)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:135)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:32)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:51)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:229)
	at org.junit.platform.launcher.core.DefaultLauncher.lambda$execute$6(DefaultLauncher.java:197)
	at org.junit.platform.launcher.core.DefaultLauncher.withInterceptedStreams(DefaultLauncher.java:211)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:191)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:128)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:69)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMainV2.main(AppMainV2.java:131)


org.janusgraph.core.JanusGraphException: Could not commit transaction due to exception during persistence

	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.commit(StandardJanusGraphTx.java:1391)
	at org.janusgraph.graphdb.tinkerpop.JanusGraphBlueprintsGraph$GraphTransaction.doCommit(JanusGraphBlueprintsGraph.java:297)
	at org.apache.tinkerpop.gremlin.structure.util.AbstractTransaction.commit(AbstractTransaction.java:104)
	at com.janusgraphtest.JBCIndexTest.shouldUpdateIndexFieldsAfterIndexModification(JBCIndexTest.java:208)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:628)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:117)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:184)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:180)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:127)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:68)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:135)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:125)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:135)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:122)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:80)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:32)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:51)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:229)
	at org.junit.platform.launcher.core.DefaultLauncher.lambda$execute$6(DefaultLauncher.java:197)
	at org.junit.platform.launcher.core.DefaultLauncher.withInterceptedStreams(DefaultLauncher.java:211)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:191)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:128)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:69)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMainV2.main(AppMainV2.java:131)
Caused by: java.lang.NullPointerException
	at org.janusgraph.graphdb.database.IndexSerializer.getIndexUpdates(IndexSerializer.java:332)
	at org.janusgraph.graphdb.database.StandardJanusGraph.prepareCommit(StandardJanusGraph.java:573)
	at org.janusgraph.graphdb.database.StandardJanusGraph.commit(StandardJanusGraph.java:727)
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.commit(StandardJanusGraphTx.java:1380)
	... 59 more
```

rayokota:
Fix NPE when updating mixed index by ensuring the cache is not stale.  Fixes #1876.

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -2793,4 +2793,60 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
 
     }
 
+    @Test
+    public void shouldUpdateIndexFieldsAfterIndexModification() throws InterruptedException, ExecutionException {
+        String key1 = "testKey1";
+        String key2 = "testKey2";
+        String key3 = "testKey3";
+        String vertexL = "testVertexLabel";
+        String indexName = "mixed";
+
+        PropertyKey p1 = mgmt.makePropertyKey(key1).dataType(Long.class).make();
+        PropertyKey p2 = mgmt.makePropertyKey(key2).dataType(Long.class).make();
+        mgmt.makeVertexLabel(vertexL).make();
+
+        JanusGraphIndex index = mgmt.buildIndex(indexName, Vertex.class)
+            .indexOnly(mgmt.getVertexLabel(vertexL))
+            .addKey(mgmt.getPropertyKey(key1))
+            .addKey(mgmt.getPropertyKey(key2))
+            .buildMixedIndex(INDEX);
+        if (index.getIndexStatus(p1) == SchemaStatus.INSTALLED) {
+            mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.REGISTER_INDEX).get();
+            mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.ENABLE_INDEX).get();
+        } else if (index.getIndexStatus(p1) == SchemaStatus.REGISTERED) {
+            mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.ENABLE_INDEX).get();
+        }
+        mgmt.commit();
+
+        JanusGraphVertex vertex = graph.addVertex(vertexL);
+        vertex.property(key1, 111L);
+        vertex.property(key2, 222L);
+
+        graph.tx().commit();
+        //By default ES indexes documents each second. By sleeping here we guarantee that documents are indexed.
+        // It is just for testing. A better approach is to use Refresh API.
+        Thread.sleep(1500L);
+
+        List<Vertex> result = graph.traversal().V().hasLabel(vertexL).has(key1, 111L).toList();
+
+        assertEquals(1, result.size());
+
+        mgmt = graph.openManagement();
+        PropertyKey testKey3 = mgmt.makePropertyKey(key3).dataType(Long.class).make();
+        mgmt.addIndexKey(mgmt.getGraphIndex(indexName), testKey3);
+        mgmt.commit();
+
+        ManagementSystem.awaitGraphIndexStatus(graph, indexName).status(SchemaStatus.REGISTERED, SchemaStatus.ENABLED).call();
+
+        mgmt = graph.openManagement();
+        mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.REINDEX).get();
+        mgmt.commit();
+
+        vertex = graph.addVertex(vertexL);
+        vertex.property(key1, 1L);
+        vertex.property(key2, 2L);
+        vertex.property(key3, 3L);
+
+        graph.tx().commit();
+    }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
@@ -28,6 +28,7 @@ import org.janusgraph.core.JanusGraph;
 import org.janusgraph.core.JanusGraphEdge;
 import org.janusgraph.core.JanusGraphElement;
 import org.janusgraph.core.JanusGraphException;
+import org.janusgraph.core.JanusGraphTransaction;
 import org.janusgraph.core.JanusGraphVertex;
 import org.janusgraph.core.JanusGraphVertexProperty;
 import org.janusgraph.core.Multiplicity;
@@ -247,6 +248,9 @@ public class ManagementSystem implements JanusGraphManagement {
             managementLogger.sendCacheEviction(updatedTypes, evictGraphFromCache, updatedTypeTriggers, getOpenInstancesInternal());
             for (JanusGraphSchemaVertex schemaVertex : updatedTypes) {
                 schemaCache.expireSchemaElement(schemaVertex.longId());
+                for (JanusGraphTransaction tx : graph.getOpenTransactions()) {
+                    tx.expireSchemaElement(schemaVertex.longId());
+                }
             }
         }
 
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -1576,4 +1576,11 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
         return !addedRelations.isEmpty() || !deletedRelations.isEmpty();
     }
 
+    @Override
+    public void expireSchemaElement(final long id) {
+        final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
+        if (v instanceof JanusGraphSchemaVertex) {
+            ((JanusGraphSchemaVertex) v).resetCache();
+        }
+    }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -1578,9 +1578,11 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
 
     @Override
     public void expireSchemaElement(final long id) {
-        final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
-        if (v instanceof JanusGraphSchemaVertex) {
-            ((JanusGraphSchemaVertex) v).resetCache();
+        if (vertexCache.contains(id)) {
+            final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
+            if (v instanceof JanusGraphSchemaVertex) {
+                ((JanusGraphSchemaVertex) v).resetCache();
+            }
         }
     }
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphTransaction.java
@@ -125,4 +125,6 @@ public interface JanusGraphTransaction extends Transaction {
      */
     boolean hasModifications();
 
+
+    void expireSchemaElement(final long id);
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphTransaction.java
@@ -125,6 +125,5 @@ public interface JanusGraphTransaction extends Transaction {
      */
     boolean hasModifications();
 
-
     void expireSchemaElement(final long id);
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementLogger.java
@@ -93,6 +93,9 @@ public class ManagementLogger implements MessageReader {
                 for (int i = 0; i < numEvictions; i++) {
                     long typeId = VariableLong.readPositive(in);
                     schemaCache.expireSchemaElement(typeId);
+                    for (JanusGraphTransaction tx : graph.getOpenTransactions()) {
+                        tx.expireSchemaElement(typeId);
+                    }
                 }
                 final GraphCacheEvictionAction action = serializer.readObjectNotNull(in, GraphCacheEvictionAction.class);
                 Preconditions.checkNotNull(action);


mbrukman:
@rayokota — thanks for the PR! Great to see you in this repo. :-)
And thanks for adressing the CLA and DCO checks quickly.

Would you mind adding the unit test provided by @porunov in issue https://github.com/JanusGraph/janusgraph/issues/1876#issue-516337652 to show that this fixes the problem identified there?

rayokota:
Hi @mbrukman!  Thanks for the quick response.  The fix above solved my issue.  I'll try to add the unit test, but still not that familiar with this repo.  Any suggestions on where to add the unit test?

mbrukman:
@rayokota — I would suggest https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-test/src/test/java/org/janusgraph/graphdb/database/IndexSerializerTest.java as that's the parallel test file for the file you're modifying here, but @porunov may have some other thoughts, since he wrote the test in question in the original issue he filed.

The other question is whether this is actually a unit test or if it actually needs to have storage + indexing backends to be running to manifest the underlying issue, which makes it more of a system test and not a unit test. @porunov, what do you think?

rayokota:
Alas, the NPE still occurs with this fix, possibly less often (but not clear).  @porunov, do you have any suggestions on how I might put together a real fix for the NPE?  I'm still making my way around the code base, so any pointers would be appreciated.  Thanks!

porunov:
> Alas, the NPE still occurs with this fix, possibly less often (but not clear). @porunov, do you have any suggestions on how I might put together a real fix for the NPE? I'm still making my way around the code base, so any pointers would be appreciated. Thanks!

Hi @rayokota , I didn't dig deep into the issue yet. I may check it later to see why are we getting NPE there.

rayokota:
@porunov , it looks like this fix addresses the NPE after all (I was running into some additional issues around trying to remove a property key from an index).  I added a copy of your test - it fails without the calls to `resetCache`, and succeeds with the calls to `resetCache`.  I also marked some fields that are affected by `resetCache` as `volatile`.  `IndexTypeWrapper.fieldMap` was already volatile, this just makes the rest of the fields affected by `resetCache` volatile as well.

porunov:
@rayokota looks like DCO failed. Would you be able to squash your commits please?

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java
```diff
@@ -334,6 +334,7 @@ private static boolean indexAppliesTo(IndexType index, JanusGraphElement element
                     update = new IndexUpdate<>(iIndex, updateType, getIndexKey(iIndex, record), getIndexEntry(iIndex, record, relation), relation);
                 } else {
                     assert relation.valueOrNull(type)!=null;
+                    index.resetCache();
```
porunov, line 337: I noticed that this place is called every time we added new properties which are covered by some index.
So I was thinking, maybe we should reset cache every time we add a new key to the index (i.e. update index itself) instead of every time we add a new vertex property which is covered by some mixed index. After checking it, I see that we already do it here:
https://github.com/JanusGraph/janusgraph/blob/c1b11d4b848303ae0a2df102123fe018abad18d2/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java#L707
Thus, it confuses me. After checking it more, I noticed that we often create a new instance of `MixedIndexTypeWrapper`. Even so, it uses the same `base` the actual instance changes. 
Checking it more, I see that each call to this method produces a new instance even if we don't need a new instance:
https://github.com/JanusGraph/janusgraph/blob/c1b11d4b848303ae0a2df102123fe018abad18d2/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/vertices/JanusGraphSchemaVertex.java#L177
I tried to change the logic to actually cache the result instead of producing a new instance and all tests passed: https://github.com/porunov/janusgraph/commit/ce9191ef302093c9ae9b13ca5d13dab1cbe8e4ad (see changes in `JanusGraphSchemaVertex.java`). I can't confirm that caching is OK here because it actually changes the logic a little bit (a new instance don't have cached `fieldMap`. Thus, it will repopulate that map on the first interaction. The only question is: do we really need to repopulate that map each time or is it OK to cache it in some / all situations?).
What I see is that `asIndexType()` is called many times in `RelationTypeVertex.getKeyIndexes()` here:
https://github.com/JanusGraph/janusgraph/blob/c1b11d4b848303ae0a2df102123fe018abad18d2/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/vertices/RelationTypeVertex.java#L113

The above is just unfinished research and it could potentially be somehow connected to this PR. I can't yet verify if this fix is correct or not. I mean, it definitely fixes the problem which is great but I don't know if there are any bad consequences which we get if we apply this fix (i.e. is there any performance degradation or GC pressure increase due to this fix?). This fix could probably be right I just isn't too familiar with this part of the code. Thus, couldn't verify properly. Hope someone could help verifying it.

cc @li-boxuan @rngcntr @mad

janusgraph-core/src/main/java/org/janusgraph/graphdb/types/indextype/CompositeIndexTypeWrapper.java
```diff
@@ -58,7 +58,7 @@ public SchemaStatus getStatus() {
         return base.getStatus();
     }
 
-    IndexField[] fields = null;
+    volatile IndexField[] fields = null;
```
porunov, line 61: I can't say if `volatile` is OK here or not. Again, I'm not much familiar with this part of the code but my thoughts are the next:
If we check `getFieldKeys()` method implementation we could see that at the beginning the author explicitly copies the reference from class variable to method level variable to not lose the reference. I.e. the author expects that the code could be executed concurrently but instead of synchronizing the code, the author prefers to explicitly compute this code multiple times if such situation acquires. I.e. looks like the author thinks that race condition will happen rarely here but if it ever happen it is OK to compute the same code several times.
Now about `volatile`. It makes sense of course to compute this code only once and then to reuse the variable by other threads but `volatile` will force us to always read and write this variable directly to RAM and never to CPU cache which could potentially affect performance.
I believe the author didn't put `volatile` here on purpose because most likely it will rarely happen that all threads which go here read stale variable from CPU cache. It can only happen if there are already multiple existing threads and all of them cached `null` in CPU and then they execute this code sequentially and CPU don't clear that variable from cache. Sounds very unlikely but even if that happens, we are back to synchronization problem. As explained above, it looks like the author didn't want to apply order guarantee to this part of the code because it's okay to compute this code multiple times and the cashed result can be either `null` or the same as in other threads (i.e. could be different instance of the array but the result in that array is the same as in other threads CPU cache).
Again, the above are just my thoughts. I can't say for sure if we need `volatile` here (and below) or not.

rayokota:
@porunov , I looked at this a bit more and think I have a better solution.

The problem is that in the test, line 2830 starts a new transaction and caches some schema vertexes in the transaction cache.  If you remove line 2830, the caching does not occur and the test passes. 

I think a better solution is that when `mgmt.commit()` is called, in addition to expiring the schema vertexes from the schema cache, it should also expire the schema vertexes from the caches of any open transactions.   In this case, expiring the schema vertex from the transaction cache just means to call `resetCache` on it, since we can't remove the schema vertex since it may possibly have a nonempty `addedRelations`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -247,6 +248,9 @@ public synchronized void commit() {
             managementLogger.sendCacheEviction(updatedTypes, evictGraphFromCache, updatedTypeTriggers, getOpenInstancesInternal());
             for (JanusGraphSchemaVertex schemaVertex : updatedTypes) {
                 schemaCache.expireSchemaElement(schemaVertex.longId());
+                for (JanusGraphTransaction tx : graph.getOpenTransactions()) {
+                    tx.expireSchemaElement(schemaVertex.longId());
```
li-boxuan, line 252: I didn't verify but I think this only solves the problem partially. The same NPE could occur if the transaction runs on a different JanusGraph instance (typically on another machine).

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1576,4 +1576,11 @@ public boolean hasModifications() {
         return !addedRelations.isEmpty() || !deletedRelations.isEmpty();
     }
 
+    @Override
+    public void expireSchemaElement(final long id) {
+        final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
```
li-boxuan, line 1581: This call will fetch this vertex from the storage backend if it does not exist in the cache, which is unnecessary. I think a better way is to create a new method, e.g. `getIfPresent` in `GuavaVertexCache.java`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java
li-boxuan: Are you saying https://github.com/porunov/janusgraph/commit/ce9191ef302093c9ae9b13ca5d13dab1cbe8e4ad fix #1876? I don't see a test in that commit thus want to confirm.

rayokota:
Thanks for the review @li-boxuan !  I made the changes that you requested.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java
porunov: No, it doesn't fix the problem. That's just a different thing which I noticed

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
rayokota: I added the code to ManagementLogger::read as well

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
rayokota: Thanks, I used the existing method `contains`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java
li-boxuan: Got it, thanks for the clarification!

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
li-boxuan: Thanks! I am just thinking if we could test this. Probably try opening two graphs in a single test? I am not sure if we have such tests in the code base.

rayokota: I looked for tests to copy but didn't find any.  Let me know if you know of one.

@li-boxuan , otherwise, are you happy with this PR?  Thanks

li-boxuan: No problem, I can add a test case for that afterward. Thank you!

li-boxuan (approved):

porunov (approved):
LGTM. Thank you @rayokota !

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -2793,4 +2793,60 @@ public void testOrderByWithRange() {
...
+        mgmt = graph.openManagement();
+        mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.REINDEX).get();
+        mgmt.commit();
+
+        vertex = graph.addVertex(vertexL);
+        vertex.property(key1, 1L);
+        vertex.property(key2, 2L);
+        vertex.property(key3, 3L);
+
+        graph.tx().commit();
```
li-boxuan, line 2850: Having thought this for a while, I now wonder if we actually should reset the cache of other transactions. Just to recap this bug: an update to the mixed index would cause open transactions to fail if they happen to touch the same index. The fix in this PR tries to make schema changes transparent to all transactions.

Now my question is, what isolation level should we provide? I know JanusGraph barely has any isolation level (some stuff is transparent while others are not), but if we want to "imitate" stronger isolation, say, snapshot isolation, then shouldn't we let those transactions abort? In that case, the fix to the bug would be to provide a better error message.

In fact, while writing test cases, I found out that the fix in this PR is incomplete. If you add 

```java
assertTrue(graph.traversal().V().hasLabel(vertexL).has(key3, 3L).hasNext());
```

after the commit, the test would fail, because the mixed index commit fails (JanusGraph by design does not throw exceptions for mixed index commit). To fix this, we likely need to refresh some other cache as well.

li-boxuan: > To fix this, we likely need to refresh some other cache as well.

This seems viable but non-trivial. Likely we need to update the information stored in IndexSerializer.IndexInfoRetriever::indexes, if we want to continue in this way.

I am not sure which way we should choose and I am happy to discuss. If we choose to let transactions abort, then we should revert this PR. If we choose to continue on the path suggested by this PR, we need to add a fix (as explained in the previous comment). For the latter case, I still propose to revert this PR before that fix is applied.

porunov: Good observations! I personally don't have strong opinion on the way we should treat it. I'm OK with both solutions (i.e. abort transactions or provide the fix in IndexSerializer). Nevertheless, I think we should note the behavior in the documentation so that it wouldn't surprise anyone.

li-boxuan: I don't have a strong opinion either. I have created https://github.com/JanusGraph/janusgraph/pull/2698 which I think is needed anyways. I (temporarily) reverted this PR in #2698 since the fix in this PR is not complete.

PR #2689, opened 2021-07-03
Refactor JanusGraphSchemaVertex to return List instead of Iterable

porunov:
getRelated is always computed as List so there is no sense to return Iterable instead of List.
In case the implementation of getRelated is changed to return Iterable then we can change the
dependant logic back to use Iterable instead of List.
Right now using List instead of Iterable allows us to create ArrayList in some places with correct size
instead of LinkedList which generally takes more space.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/vertices/JanusGraphSchemaVertex.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/types/vertices/JanusGraphSchemaVertex.java
@@ -40,6 +40,8 @@ import org.janusgraph.graphdb.types.system.BaseKey;
 import org.janusgraph.graphdb.types.system.BaseLabel;
 import org.janusgraph.graphdb.vertices.CacheVertex;
 
+import java.util.List;
+
 public class JanusGraphSchemaVertex extends CacheVertex implements SchemaSource {
 
     public JanusGraphSchemaVertex(StandardJanusGraphTx tx, long id, byte lifecycle) {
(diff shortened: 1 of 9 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/types/vertices/JanusGraphSchemaVertex.java
```diff
@@ -176,11 +175,9 @@ public SchemaStatus getStatus() {
...
     public IndexType asIndexType() {
         Preconditions.checkArgument(getDefinition().containsKey(TypeDefinitionCategory.INTERNAL_INDEX),"Schema vertex is not a type vertex: [%s,%s]", longId(), name());
-        if (getDefinition().<Boolean>getValue(TypeDefinitionCategory.INTERNAL_INDEX)) {
-            return new CompositeIndexTypeWrapper(this);
-        } else {
-            return new MixedIndexTypeWrapper(this);
-        }
+        return getDefinition().<Boolean>getValue(TypeDefinitionCategory.INTERNAL_INDEX) ?
+            new CompositeIndexTypeWrapper(this) :
+            new MixedIndexTypeWrapper(this);
```
porunov, line 180: As explained here https://github.com/JanusGraph/janusgraph/pull/2688#discussion_r663368800 it looks like we can actually cache the returning `IndexType`. The example implementation is here: https://github.com/porunov/janusgraph/commit/ce9191ef302093c9ae9b13ca5d13dab1cbe8e4ad (all tests passed).
That said, as I'm not sure 100% here that it's OK to cache the resulting object. Thus, I didn't do it in this PR. That said, I think we should check it and possibly change the implementation of this method to actually cache the result instead of returning a new instance every time (probably in the next PR if we can prove that it's OK to cache the result).

li-boxuan (approved):
LGTM 👍

PR #2698, opened 2021-07-08
Abort transaction if mixed index schema gets updated concurrently

li-boxuan:
An update to the mixed index would cause open transactions to fail and throw
NullPointerException if they happen to touch the same index. #2688 fixes this
by making schema changes transparent to all transactions. The fix is, however,
incomplete in the sense that transaction now still fails but fails silently.

Generally, there are two solutions:

1) Do not attempt to fix this issue, but provide a better error message.
2) Continue the fix in #2688 and make sure schema caches are cleared.

This commit reverts #2688 and provides a better error message when aborting
the transaction. Note that even if we choose 2) later, this commit still
remains valuable, because schema updates are propagated asynchronously
among JanusGraph instances.

This commit does not judge whether we should do 2) or not. It reverts #2688
simply because the fix is incomplete and can possibly cause more trouble than it
attempts to fix.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -44,6 +44,7 @@ import org.janusgraph.core.JanusGraphTransaction;
 import org.janusgraph.core.JanusGraphVertex;
 import org.janusgraph.core.JanusGraphVertexProperty;
 import org.janusgraph.core.PropertyKey;
+import org.janusgraph.core.SchemaViolationException;
 import org.janusgraph.core.VertexLabel;
 import org.janusgraph.core.attribute.Cmp;
 import org.janusgraph.core.attribute.Geo;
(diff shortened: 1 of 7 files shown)


porunov (approved):
LGTM. Thank you @li-boxuan !
I would also add some information about such transactions here: https://docs.janusgraph.org/master/schema/#changing-schema-elements but we can do it in a separate PR

li-boxuan:
> I would also add some information about such transactions here: https://docs.janusgraph.org/master/schema/#changing-schema-elements but we can do it in a separate PR

You are right. I guess we could do that once we have more ideas or consents on https://github.com/JanusGraph/janusgraph/pull/2688#discussion_r666046369

rayokota:
@li-boxuan , @porunov if you don't have strong objections, I would prefer to try to pursue option #2.  This is mainly because this issue is affecting our production use of Apache Atlas with JanusGraph.

If it turns out that a solution is not possible or unwieldy, then I will understand if you no longer wish to pursue option #2.  At that point we would have to decide whether to maintain a private version of JanusGraph with these patches, which is less than ideal.

In any case, I've proposed a potential solution to the latest issue here:  https://github.com/JanusGraph/janusgraph/pull/2699.  It is only a few more lines of code.  Please let me know what you think and if there are any obvious problems with the new fix.  I really appreciate your feedback.  Thanks in advance.

li-boxuan:
@rayokota I am fine with the continuation. I am merging this PR first (just in case your PR cannot be merged timely, before 0.6.0 release). Can you then do the changes from #2688 again in #2699? I will review #2699 soon. Thanks and sorry for the inconvenience.

PR #2699, opened 2021-07-08
Properly invalidate stale cached index data

rayokota:
This is a continuation PR to https://github.com/JanusGraph/janusgraph/pull/2688, that properly invalidates stale cached data from index transactions.

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -52,6 +52,7 @@ import org.janusgraph.core.schema.VertexLabelMaker;
 import org.janusgraph.diskstorage.BackendException;
 import org.janusgraph.diskstorage.BackendTransaction;
 import org.janusgraph.diskstorage.EntryList;
+import org.janusgraph.diskstorage.indexing.IndexTransaction;
 import org.janusgraph.diskstorage.keycolumnvalue.SliceQuery;
 import org.janusgraph.graphdb.query.graph.MixedIndexCountQueryBuilder;
 import org.janusgraph.diskstorage.util.time.TimestampProvider;
@@ -109,6 +110,7 @@ import org.janusgraph.graphdb.transaction.subquerycache.SubqueryCache;
 import org.janusgraph.graphdb.transaction.vertexcache.GuavaVertexCache;
 import org.janusgraph.graphdb.transaction.vertexcache.VertexCache;
 import org.janusgraph.graphdb.types.CompositeIndexType;
+import org.janusgraph.graphdb.types.IndexType;
 import org.janusgraph.graphdb.types.StandardEdgeLabelMaker;
 import org.janusgraph.graphdb.types.StandardPropertyKeyMaker;
 import org.janusgraph.graphdb.types.StandardVertexLabelMaker;
@@ -1581,5 +1583,27 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
     @Override
     public boolean hasModifications() {
         return !addedRelations.isEmpty() || !deletedRelations.isEmpty();
+    }
+
+    @Override
+    public void expireSchemaElement(final long id) {
+        if (vertexCache.contains(id)) {
+            final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
+            if (v instanceof JanusGraphSchemaVertex) {
+                JanusGraphSchemaVertex sv = (JanusGraphSchemaVertex) v;
+                sv.resetCache();
+                if (sv.getDefinition().containsKey(TypeDefinitionCategory.INTERNAL_INDEX)) {
+                    IndexType indexType = sv.asIndexType();
+                    if (indexType.isMixedIndex()) {
+                        // Invalidate mixed index
+                        String store = indexType.getBackingIndexName();
+                        if (txHandle.hasIndexTransaction(store)) {
+                            IndexTransaction indexTx = txHandle.getIndexTransaction(store);
+                            indexTx.invalidate(sv.name());
+                        }
+                    }
+                }
+            }
+        }
     }
 }
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/diskstorage/indexing/IndexProviderTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/diskstorage/indexing/IndexProviderTest.java
@@ -109,6 +109,11 @@ public abstract class IndexProviderTest {
             public KeyInformation.StoreRetriever get(String store) {
                 return mappings::get;
             }
+
+            @Override
+            public void invalidate(String store) {
+                mappings.remove(store);
+            }
         };
     }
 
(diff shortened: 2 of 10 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1588,7 +1589,13 @@ public void expireSchemaElement(final long id) {
         if (vertexCache.contains(id)) {
             final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
             if (v instanceof JanusGraphSchemaVertex) {
-                ((JanusGraphSchemaVertex) v).resetCache();
+                JanusGraphSchemaVertex sv = (JanusGraphSchemaVertex) v;
+                sv.resetCache();
+                if (sv.getDefinition().containsKey(TypeDefinitionCategory.INTERNAL_INDEX)) {
+                    String store = sv.getDefinition().getValue(TypeDefinitionCategory.BACKING_INDEX,String.class);
+                    IndexTransaction idxTx = txHandle.getIndexTransaction(store);
+                    idxTx.invalidate(sv.name());
```
li-boxuan, line 1597: Thanks, I believe this works but not ideal. If I understand it correctly, it clears cache for the whole mixed index store, which includes caches for many other indices.

li-boxuan (commented):
This looks good, thank you! Can you please rebase your PR? You would have to do your old changes again here, unfortunately.

Btw, it's actually easy to test propagations between multiple graph instances. You could just call `JanusGraphFactory.open(config)` to open another graph instance in your test. See https://github.com/li-boxuan/janusgraph/commit/15dea4a for an example.

rayokota:
Thanks @li-boxuan !  I really appreciate your help with this.  I've rebased the PR as you requested, and I've re-added my changes.

porunov:
CQL tests are constantly failing for this PR

rayokota:
@porunov , @li-boxuan , I'll take a look.  It seems the error is `org.janusgraph.core.JanusGraphException: Index did not stabilize within the given amount of time. For sufficiently long wait periods this is most likely caused by a failed/incorrectly shut down JanusGraph instance or a lingering transaction.`.  It seems now that the index is not in the cache, somehow the test is not able to get it directly (due to a lingering transaction)?  I'm also having trouble running the CQL tests locally (due to `org.testcontainers.containers.ContainerLaunchException: Container startup failed`), so any help would be appreciated!

porunov:
@rayokota You need docker to be running on your system where you run tests. Also, default CQL ports should not be already in use by any processes

rayokota:
Thanks @porunov , I had to increase the memory settings for Docker.  I am able to reproduce the issue now.  I'll take a look.

rayokota:
@porunov , @li-boxuan I've fixed the issue with the CQL tests.  Apparently the name of a composite index was being passed to `getIndexTransaction`, which is only for mixed indexes.  This caused an NPE.  I've fixed the code to only call `getIndexTransaction` for mixed indexes.

rayokota:
@porunov , @li-boxuan this PR is ready for re-review, thanks!

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1582,4 +1583,27 @@ public final boolean isClosed() {
...
+                JanusGraphSchemaVertex sv = (JanusGraphSchemaVertex) v;
+                sv.resetCache();
+                if (sv.getDefinition().containsKey(TypeDefinitionCategory.INTERNAL_INDEX)
+                    && !sv.getDefinition().<Boolean>getValue(TypeDefinitionCategory.INTERNAL_INDEX)) {
+                    try {
+                        // Invalidate mixed index
+                        String store = sv.getDefinition()
+                            .getValue(TypeDefinitionCategory.BACKING_INDEX, String.class);
+                        IndexTransaction indexTx = txHandle.getIndexTransaction(store);
+                        indexTx.invalidate(sv.name());
```
porunov, line 1601: I would generally prefer using if else conditions when possible but if that's complicated I think it's fine to use try catch

porunov (approved):
LGTM. Thank you @rayokota ! I have one comment but that's just my thoughts and not requesting changes.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1582,4 +1583,27 @@ public final boolean isClosed() {
...
     }
+
+    @Override
+    public void expireSchemaElement(final long id) {
+        if (vertexCache.contains(id)) {
+            final InternalVertex v = vertexCache.get(id, externalVertexRetriever);
+            if (v instanceof JanusGraphSchemaVertex) {
+                JanusGraphSchemaVertex sv = (JanusGraphSchemaVertex) v;
+                sv.resetCache();
+                if (sv.getDefinition().containsKey(TypeDefinitionCategory.INTERNAL_INDEX)
```
li-boxuan, line 1594: I would prefer calling `sv.asIndexType().isMixedIndex()` which is shorter and does the same thing.

li-boxuan (commented):
I think it's in a good shape! I just have some minor doubts & comments.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
rayokota: Thanks for the review again @li-boxuan!  I've changed the code to use `sv.asIndexType().isMixedIndex()` as you requested.

I've also removed the try-catch as you and @porunov requested.  It was added to handle a possible NPE thrown by `getIndexTransaction`.  Instead, I've added a method `hasIndexTransaction` to avoid the NPE.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
rayokota: It was added to handle a possible NPE thrown by getIndexTransaction. Instead, I've added a method hasIndexTransaction to avoid the NPE.

porunov:
@rayokota Codecov says that there are many places in this PR which are not covered by tests. Will you be able to cover it? That said, sometimes it may produce a wrong report, if so, just ignore my comment

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
rayokota: Actually, I think I still need a check that the `JanusGraphSchemaVertex` is indeed an index type.

rayokota:
@porunov , I think that was a preliminary Codecov report before all the tests had run.

@li-boxuan , the PR is ready for re-review again, thanks!

porunov:
> @porunov , I think that was a preliminary Codecov report before all the tests had run.

Oh, you are right. Sorry for the noise 😅

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
li-boxuan: > I think I still need a check that the JanusGraphSchemaVertex is indeed an index type.

I don't think that is necessary but there is no harm to do so.

li-boxuan (approved):
Thank you! Can you please squash your commits?

rayokota:
@li-boxuan , I've squashed the commits as you've requested, thanks!

PR #2700, opened 2021-07-09
Add configuration options for ExecutorService

porunov:

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/advanced-topics/executor-service.md
+++ b/docs/advanced-topics/executor-service.md
@@ -1,0 +1,36 @@
+# Backend ExecutorService
+
+By default JanusGraph uses `ExecutorService` to process some queries in parallel 
+(see `storage.parallel-backend-ops` configuration option).  
+JanusGraph allows to configure the executor service which is used via configuration options 
+provided in `storage.parallel-backend-executor-service` configuration section.
+
+Currently JanusGraph has the next ExecutorService implementations which can be used (controlled via 
+`storage.parallel-backend-executor-service.class`):
+
+* `fixed` - fixed thread pool size;
+* `cached` - cached thread pool size;
+* Custom ExecutorService;
+
+## Custom ExecutorService
+
+To use custom `ExecutorService` the configuration option `storage.parallel-backend-executor-service.class` must be 
+provided which must be the full class name of the `ExecutorService` implementation class.
+
+The provided class which implements `ExecutorService` must have either a public constructor with
+[ExecutorServiceConfiguration](https://javadoc.io/doc/org.janusgraph/janusgraph-core/latest/org/janusgraph/diskstorage/configuration/ExecutorServiceConfiguration.html)
+argument (preferred constructor) or a public parameterless constructor. 
+When both constructors are available the constructor with `ExecutorServiceConfiguration` argument will be used.
+
+When custom `ExecutorService` is provided with `ExecutorServiceConfiguration` it isn't required to 
+use any of the configuration options provided in `ExecutorServiceConfiguration` but sometimes it is
+convenient to use the provided configurations. The configuration options provided in `ExecutorServiceConfiguration` 
+are typically those configuration options which are provided via configurations from `storage.parallel-backend-executor-service`.
+
+# Cassandra backend ExecutorService
+
+Apart from the general executor service mentioned in the previous section, Cassandra backend uses an additional 
+`ExecutorService` to process CQL queries. The rules by which the Cassandra backend `ExecutorService` is built are the 
+same as the rules which are used to build parallel backend queries `ExecutorService` (described above). 
+The only difference is that the configuration for Cassandra backend `ExecutorService` are provided via configuration 
+options under `storage.cql.executor-service`.
--- a/docs/configs/janusgraph-cfg.md
+++ b/docs/configs/janusgraph-cfg.md
@@ -348,7 +348,6 @@ Configuration options for the storage backend.  Some options are applicable only
 | storage.hostname | The hostname or comma-separated list of hostnames of storage backend servers.  This is only applicable to some storage backends, such as cassandra and hbase. | String[] | 127.0.0.1 | LOCAL |
 | storage.page-size | JanusGraph break requests that may return many results from distributed storage backends into a series of requests for small chunks/pages of results, where each chunk contains up to this many elements. | Integer | 100 | MASKABLE |
 | storage.parallel-backend-ops | Whether JanusGraph should attempt to parallelize storage operations | Boolean | true | MASKABLE |
-| storage.parallel-backend-ops-thread-pool-size | Thread pool size which is used for parallel requests when storage.parallel-backend-ops is enabled. If not set the pool size will be equal to number of processors multiplied by 2. | Integer | (no default value) | MASKABLE |
 | storage.password | Password to authenticate against backend | String | (no default value) | LOCAL |
 | storage.port | The port on which to connect to storage backend servers. For HBase, it is the Zookeeper port. | Integer | (no default value) | LOCAL |
 | storage.read-only | Read-only database | Boolean | false | LOCAL |
@@ -405,6 +404,17 @@ CQL storage backend options
 | storage.cql.speculative-retry | The speculative retry policy. One of: NONE, ALWAYS, <X>percentile, <N>ms. | String | (no default value) | FIXED |
 | storage.cql.use-external-locking | True to prevent JanusGraph from using its own locking mechanism. Setting this to true eliminates redundant checks when using an external locking mechanism outside of JanusGraph. Be aware that when use-external-locking is set to true, that failure to employ a locking algorithm which locks all columns that participate in a transaction upfront and unlocks them when the transaction ends, will result in a 'read uncommitted' transaction isolation level guarantee. If set to true without an appropriate external locking mechanism in place side effects such as dirty/non-repeatable/phantom reads should be expected. | Boolean | false | MASKABLE |
 | storage.cql.write-consistency-level | The consistency level of write operations against Cassandra | String | QUORUM | MASKABLE |
+
+### storage.cql.executor-service
+Configuration options for CQL executor service which is used to process CQL queries.
+
+
+| Name | Description | Datatype | Default Value | Mutability |
+| ---- | ---- | ---- | ---- | ---- |
+| storage.cql.executor-service.class | The implementation of `ExecutorService` to use. The full name of the class which extends `ExecutorService` which has either a public constructor with `ExecutorServiceConfiguration` argument (preferred constructor) or a public parameterless constructor. Other accepted options are: `fixed` - fixed thread pool size of `max-pool-size` size; `cached` - cached thread pool size; | String | fixed | LOCAL |
+| storage.cql.executor-service.core-pool-size | Core pool size for executor service. May be ignored if custom executor service is used (depending on the implementation of the executor service). | Integer | 10 | LOCAL |
+| storage.cql.executor-service.keep-alive-time | Keep alive time in milliseconds for executor service. When the number of threads is greater than the `core-pool-size`, this is the maximum time that excess idle threads will wait for new tasks before terminating. Ignored for `fixed` executor service and may be ignored if custom executor service is used (depending on the implementation of the executor service). | Long | 60000 | LOCAL |
+| storage.cql.executor-service.max-pool-size | Maximum pool size for executor service. Ignored for `fixed` and `cached` executor services. May be ignored if custom executor service is used (depending on the implementation of the executor service). | Integer | 2147483647 | LOCAL |
 
 ### storage.cql.metrics
 Configuration options for CQL metrics
@@ -520,6 +530,17 @@ Meta data to include in storage backend retrievals
 | storage.meta.[X].timestamps | Whether to include timestamps in retrieved entries for storage backends that automatically annotated entries with timestamps. If enabled, timestamp can be retrieved by `element.value(ImplicitKey.TIMESTAMP.name())` or equivalently, `element.value("~timestamp")`. | Boolean | false | GLOBAL |
 | storage.meta.[X].ttl | Whether to include ttl in retrieved entries for storage backends that support storage and retrieval of cell level TTL. If enabled, ttl can be retrieved by `element.value(ImplicitKey.TTL.name())` or equivalently, `element.value("~ttl")`. | Boolean | false | GLOBAL |
 | storage.meta.[X].visibility | Whether to include visibility in retrieved entries for storage backends that support cell level visibility. If enabled, visibility can be retrieved by `element.value(ImplicitKey.VISIBILITY.name())` or equivalently, `element.value("~visibility")`. | Boolean | true | GLOBAL |
+
+### storage.parallel-backend-executor-service
+Configuration options for executor service which is used for parallel requests when `storage.parallel-backend-ops` is enabled.
+
+
+| Name | Description | Datatype | Default Value | Mutability |
+| ---- | ---- | ---- | ---- | ---- |
+| storage.parallel-backend-executor-service.class | The implementation of `ExecutorService` to use. The full name of the class which extends `ExecutorService` which has either a public constructor with `ExecutorServiceConfiguration` argument (preferred constructor) or a public parameterless constructor. Other accepted options are: `fixed` - fixed thread pool size of `max-pool-size` size; `cached` - cached thread pool size; | String | fixed | LOCAL |
+| storage.parallel-backend-executor-service.core-pool-size | Core pool size for executor service. May be ignored if custom executor service is used (depending on the implementation of the executor service).If not set the core pool size will be equal to number of processors multiplied by 2. | Integer | (no default value) | LOCAL |
+| storage.parallel-backend-executor-service.keep-alive-time | Keep alive time in milliseconds for executor service. When the number of threads is greater than the `core-pool-size`, this is the maximum time that excess idle threads will wait for new tasks before terminating. Ignored for `fixed` executor service and may be ignored if custom executor service is used (depending on the implementation of the executor service). | Long | 60000 | LOCAL |
+| storage.parallel-backend-executor-service.max-pool-size | Maximum pool size for executor service. Ignored for `fixed` and `cached` executor services. May be ignored if custom executor service is used (depending on the implementation of the executor service). | Integer | 2147483647 | LOCAL |
 
 ### tx
 Configuration options for transaction handling
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
@@ -29,13 +29,10 @@ import java.net.Inet4Address;
 import java.net.UnknownHostException;
 import java.util.Map;
 
-import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.PARALLEL_BACKEND_OPS;
-import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.PARALLEL_BACKEND_OPS_THREAD_POOL_SIZE;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.REPLACE_INSTANCE_IF_EXISTS;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.UNIQUE_INSTANCE_ID;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.UNIQUE_INSTANCE_ID_HOSTNAME;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.UNIQUE_INSTANCE_ID_SUFFIX;
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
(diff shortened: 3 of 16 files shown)


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -583,4 +583,33 @@
...
+        "max-pool-size",
+        "Maximum pool size for CQL executor service.",
+        ConfigOption.Type.LOCAL,
+        Integer.class,
+        100);
+
+    ConfigOption<Long> EXECUTOR_SERVICE_KEEP_ALIVE_TIME = new ConfigOption<>(
+        EXECUTOR_SERVICE,
+        "keep-alive-time",
+        "Keep alive time in milliseconds for CQL executor service.",
```
li-boxuan, line 611: (nitpick) I think we might want to either paste the explanation from ThreadPoolExecutor here:

```
when the number of threads is greater than the core, this is the maximum time that excess idle threads will wait for new tasks before terminating
```
or give a pointer to the ThreadPoolExecutor javadoc. Otherwise by simply looking at the documentation itself, one might not know what it is.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java
```diff
@@ -583,4 +583,33 @@
             "Can be used when `" + REQUEST_TRACKER_CLASS + "` is set to `RequestLogger`.",
         ConfigOption.Type.LOCAL,
         Boolean.class);
+
+    ConfigNamespace EXECUTOR_SERVICE = new ConfigNamespace(
+        CQL_NS,
+        "executor-service",
+        "Configuration options for CQL executor service which is used to process CQL queries.");
```
li-boxuan, line 590: I guess we can say that it uses ThreadPoolExecutor under the hood.

li-boxuan (commented):
As discussed here https://github.com/JanusGraph/janusgraph/discussions/2545#discussioncomment-551453, I am not sure why we need to use vavr at the first place, but this PR itself looks good to me as it does not bring any breaking change and makes existing fixed parameters configurable. I just left two nitpicks on the documentation part.

porunov:
@li-boxuan Thanks for the review! Reading docs and https://github.com/JanusGraph/janusgraph/discussions/2545 I now think that we should give users to chose executor services they want to use instead of forcing them using this executor service. Also, as @cdegroc already noted, there is no any sense providing `maximumPoolSize` to the current implementation because it is simply ignored. I will try to make more customizable configurations for executor services and will update this PR

porunov:
@li-boxuan , @cdegroc I added documentation regarding new functionality and covered it by tests. Would be great to hear your thoughts regarding this PR

cdegroc:
> @li-boxuan , @cdegroc I added documentation regarding new functionality and covered it by tests. Would be great to hear your thoughts regarding this PR

Thanks. Today is a holiday here, but I'll have a look tomorrow and will test with a custom ExecutorService.

docs/advanced-topics/executor-service.md
```diff
@@ -0,0 +1,34 @@
...
+When both constructors are available the constructor with `ExecutorServiceConfiguration` argument will be used.
+
+When custom `ExecutorService` is provided with `ExecutorServiceConfiguration` it isn't required to 
+use any of the configuration options provided in `ExecutorServiceConfiguration` but sometimes it is
+convenient to use the provided configurations. The configuration options provided in `ExecutorServiceConfiguration` 
+are typically those configuration options which are provided via configurations from `storage.parallel-backend-executor-service`.
+
+## CQL ExecutorService
+
+CQL also uses `ExecutorService` to process CQL queries. The rules by which the CQL
```
li-boxuan, line 31: (nitpick) Non Cassandra users might not be able to understand this section until they realize CQL is a protocol for Cassandra. We could probably use "Cassandra backend" interchangeably to avoid confusion.

docs/configs/janusgraph-cfg.md
```diff
@@ -406,6 +405,17 @@ CQL storage backend options
...
 | storage.cql.write-consistency-level | The consistency level of write operations against Cassandra | String | QUORUM | MASKABLE |
 
+### storage.cql.executor-service
+Configuration options for CQL executor service which is used to process CQL queries.
+
+
+| Name | Description | Datatype | Default Value | Mutability |
+| ---- | ---- | ---- | ---- | ---- |
+| storage.cql.executor-service.class | The implementation of `ExecutorService` to use. The class full name of the class which extends `ExecutorService` which has either a public constructor with `ExecutorServiceConfiguration` argument (preferred constructor) or a public parameterless constructor. Other accepted options are: `fixed` - fixed thread pool size of `max-pool-size` size; `cached` - cached thread pool size; | String | fixed | LOCAL |
```
li-boxuan, line 414: [DISCUSS] I am thinking if we should allow not using executor service at all. Maybe I will try what happens when we don't use a thread pool and vavr at all, and then compare the performance.

li-boxuan (commented):
I am really impressed by your efficiency 😄  Haven't taken a detailed look yet, but looks great in general!

docs/configs/janusgraph-cfg.md
porunov: I actually was thinking about it as well. I'm not sure why we actually need CQL thread pool at all. It looks like with current implementation in `master` branch we can execute up to 10 parallel CQL queries and never more because this thread pool is limiting us which is crazy unless I'm missing something. Maybe it was introduced to limit the amount of parallel interaction with the same CQL storage backend but I'm not sure. Nevertheless, with this PR users can either increase this thread pool size or just simply switch to cached thread pool size.
In general, I think we need to allow users to optionally disable / enable CQL thread pool usage as we do with parallel backend queries thread pool. That said, I think this functionality should go to a separate PR. Would be great if you could check performance difference between using and not using thread pool there.

porunov: Checking `CQLStoreManager.mutateManyUnlogged` closer now I see that we may execute a sequence of requests in parallel. Thus, having the executor service make sense there. At the end we would still wait until all the requests are finished but with executor service we gain 2 features here:
- executing a sequence of requests in parallel
- limiting of the amount of parallel executions of such sequences

I assume executing CQL requests sequentially without the executor service will make queries always take only 1 current thread which might be slower but won't block other queries.
With limited executor service we can process such queries in parallel but a single query can potentially take all the available threads which essentially will delay the execution of other queries.
With unlimited executor service (i.e. `cached` executor service OR custom unlimited executor service) we will execute all sequences of all queries in parallel but we may potentially acquire too many threads which may lead to OOM or decrease the overall performance due to threads context switching overhead.

I think we should use CQL executor service by default (as we do right now) but make it configurable (as this PR does). Executing queries without executor service is a separate feature which in some situations I think could be beneficial (resources or performance optimization) but we should double check if we gain any improvements when executing queries not via executor service.

li-boxuan: Agree that "Executing queries without executor service is a separate feature" that should be done in a separate PR.

I suspect the following statement is not accurate:
> executing CQL requests sequentially without the executor service will make queries always take only 1 current thread

The reason is `this.session.executeAsync` is asynchronous and returns a `CompletionStage`. When we have multiple batches, they are not executing in sequence. We only need to wait for their completeness in sequence.

porunov: Oh, you are right. I was thinking if we change executeAsync to execute then we would have sequential queries execution. If we leave executeAsync then all queries will be executed in parallel but then we won't be able to limit the amount of parallel queries if we don't have executorService.

li-boxuan: > we won't be able to limit the amount of parallel queries if we don't have executorService

That's my guess as well: https://github.com/JanusGraph/janusgraph/discussions/2545#discussioncomment-620084

cdegroc:
> @li-boxuan , @cdegroc I added documentation regarding new functionality and covered it by tests. Would be great to hear your thoughts regarding this PR

I confirm this is generic enough to plug a custom (e.g. scaling or instrumented) thread pool 👍

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -526,10 +529,55 @@ public boolean apply(@Nullable String s) {
...
+        ConfigOption.Type.LOCAL,
+        Integer.class);
+
+    public static final ConfigOption<Integer> PARALLEL_BACKEND_EXECUTOR_SERVICE_MAX_POOL_SIZE = new ConfigOption<>(
+        PARALLEL_BACKEND_EXECUTOR_SERVICE,
+        "max-pool-size",
+        "Maximum pool size for executor service. Ignored for `fixed` executor service and may be ignored " +
+            "if custom executor service is used (depending on the implementation of the executor service)."+
+            "If not set the core pool size will be equal to number of processors multiplied by "
+            +ExecutorServiceBuilder.THREAD_POOL_SIZE_SCALE_FACTOR+".",
```
cdegroc, line 553: I think it should be "If not set, the _max_ pool size will be Integer.MAX_VALUE."

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java
```diff
@@ -674,4 +668,17 @@ private static Locker openManagedLocker(String classname, String lockerName) {
...
+
+    private ExecutorService buildExecutorService(Configuration configuration){
+
+        Integer corePoolSize = configuration.getOrDefault(PARALLEL_BACKEND_EXECUTOR_SERVICE_CORE_POOL_SIZE);
+        Integer maxPoolSize = configuration.getOrDefault(PARALLEL_BACKEND_EXECUTOR_SERVICE_MAX_POOL_SIZE);
+        Long keepAliveTime = configuration.getOrDefault(PARALLEL_BACKEND_EXECUTOR_SERVICE_KEEP_ALIVE_TIME);
+        String executorServiceClass = configuration.getOrDefault(PARALLEL_BACKEND_EXECUTOR_SERVICE_CLASS);
+
+        ExecutorServiceConfiguration executorServiceConfiguration =
+            new ExecutorServiceConfiguration(executorServiceClass, corePoolSize, maxPoolSize, keepAliveTime, null);
```
cdegroc, line 680: Can we use the occasion to provide a ThreadFactory that would name those threads? Or maybe in a follow-up PR?
I.e. pass in `new ThreadFactoryBuilder().setNameFormat("backend[%02d]").build()` instead of `null`.

docs/configs/janusgraph-cfg.md
```diff
@@ -521,6 +531,17 @@ Meta data to include in storage backend retrievals
...
+| storage.parallel-backend-executor-service.core-pool-size | Core pool size for executor service. May be ignored if custom executor service is used (depending on the implementation of the executor service).If not set the core pool size will be equal to number of processors multiplied by 2. | Integer | (no default value) | LOCAL |
+| storage.parallel-backend-executor-service.keep-alive-time | Keep alive time in milliseconds for CQL executor service. When the number of threads is greater than the `core-pool-size`, this is the maximum time that excess idle threads will wait for new tasks before terminating. Ignored for `fixed` executor service and may be ignored if custom executor service is used (depending on the implementation of the executor service). | Long | 60000 | LOCAL |
+| storage.parallel-backend-executor-service.max-pool-size | Maximum pool size for executor service. Ignored for `fixed` executor service and may be ignored if custom executor service is used (depending on the implementation of the executor service).If not set the core pool size will be equal to number of processors multiplied by 2. | Integer | 2147483647 | LOCAL |
+
```
cdegroc, line 544: Same as https://github.com/JanusGraph/janusgraph/pull/2700/files#r670582778

docs/configs/janusgraph-cfg.md
cdegroc: Sorry if I'm missing something but if the "CQL executor service" was removed, wouldn't the Backend threadpool size limit the number of concurrent requests anyway?

li-boxuan: > wouldn't the Backend threadpool size limit the number of concurrent requests anyway

That's a good point, but it is not always the case.

https://github.com/JanusGraph/janusgraph/blob/a28e314dcf09d1ae2f6b2ab341fe3f7f44fb5cc4/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java#L301-L304

From above we can see that backend thread pool is not used when 1) PARALLEL_BACKEND_OPS is disabled, OR 2) the number of queries is one.

I would say the behavior of the combination usage of the backend thread pool and CQL thread pool is very confusing and it would be great if we can document it.

li-boxuan: Limiting the number of concurrent CQL queries seems not very useful to me anyways, because users can tune the CQL driver configurations to limit using configurations like `MAX_REQUESTS_PER_CONNECTION`.

porunov: I think @li-boxuan is right here. I added a separate ticket #2707 to track the feature request to add possibility to disable ExecutorService for CQL requests.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java
porunov: I opened a separate ticket to track it: #2708 
I would prefer doing it in a follow-up PR

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
porunov: I think we probably don't even need to have this statement because the default value is present in the documentation. So, probably it's just unnecessary to duplicate that information. I removed this statement

porunov:
@cdegroc I updated the PR to resolve your comments. Let me know what you think about it

cdegroc:
> @cdegroc I updated the PR to resolve your comments. Let me know what you think about it
Yeah you are right about the default values. Thanks!

li-boxuan (commented):
Thanks for this feature! Most of my comments are nitpicks, except one regarding the behavior of the cached pool.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ExecutorServiceBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+        if(keepAliveTime == null){
+            throw new IllegalStateException("To use "+CACHED_THREAD_POOL_CLASS+
+                " executor service keepAliveTime must be provided.");
+        }
+
+        log.info("Initiated cached thread pool of core size {}, max size {} and keep alive time {} ms",
+            corePoolSize, maxPoolSize, keepAliveTime);
+
+        if(threadFactory == null){
+            return new ThreadPoolExecutor(corePoolSize, maxPoolSize, keepAliveTime, TimeUnit.MILLISECONDS, new SynchronousQueue<>());
```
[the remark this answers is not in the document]
porunov: Thanks for noticing that. Totally makes sense but what about `keepAliveTime`? If we just do `Executors.newCachedThreadPool()` then the time will always be 1 minute with no possibility to change that.
What if we change it to `new ThreadPoolExecutor(0, Integer.MAX_VALUE, keepAliveTime, TimeUnit.MILLISECONDS, new SynchronousQueue<>());`? Or do you prefer `Executors.newCachedThreadPool()`?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ExecutorServiceBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+                if(parameterlessConstructor == null){
+                    throw new IllegalStateException(configurationClass + " has neither public constructor which accepts "
+                        +ExecutorServiceConfiguration.class.getName() + " nor parameterless public constructor.");
+                }
+                executorService = (ExecutorService) parameterlessConstructor.newInstance();
+            }
+
+        } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+            throw new IllegalStateException("Couldn't create a new instance of "+configurationClass +
+                ". Please, check that it has a public constructor without arguments.", e);
```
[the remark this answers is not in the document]
porunov: It can be reached if there is a single private parameterless constructor or single private constructor with `ExecutorService` .
See test `shouldNotCreateBackendExecutorServiceIfClassHasNoPublicConstructor`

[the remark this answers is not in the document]
porunov: The above code returns up to 2 constructors. Without and with parameter. The above code doesn't check if those constructors are actually available to be accessed. Thus, the code if this `try` can actually fail with `private` constructors only.

docs/advanced-topics/executor-service.md
```diff
@@ -0,0 +1,34 @@
...
+* `cached` - cached thread pool size;
+* Custom ExecutorService;
+
+## Custom ExecutorService
+
+To use custom `ExecutorService` the configuration option `storage.parallel-backend-executor-service.class` must be 
+provided which must be the full class name of the `ExecutorService` implementation class.
+
+The provided class which implements `ExecutorService` must have either a public constructor with 
+`ExecutorServiceConfiguration` argument (preferred constructor) or a public parameterless constructor.
```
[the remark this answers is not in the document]
porunov: The javadoc isn't published with `ExecutorServiceConfiguration` yet but I could just add a future link which will fail to be resolved until 0.6.0 is released. Will it be OK? Or do you mean something else?

janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ExecutorServiceBuilder.java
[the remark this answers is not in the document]
li-boxuan: Got it, so the exception message here emphasizes "public" rather than "has a constructor". That makes sense.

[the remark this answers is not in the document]
porunov: Yes, that's right. The exception is more about `public` here.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/configuration/ExecutorServiceBuilder.java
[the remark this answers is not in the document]
li-boxuan: Makes sense. In that case I guess both `corePoolSize` and `keepAliveTime` are useful, and we could do `new ThreadPoolExecutor(corePoolSize, Integer.MAX_VALUE, keepAliveTime, TimeUnit.MILLISECONDS, new SynchronousQueue<>());`. That being said, I didn't verify by myself whether this works as expected or not.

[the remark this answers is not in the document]
porunov: Makes sense. Will do.
After the executor creation I can just invoke `prestartAllCoreThreads` as you suggested to properly support `corePoolSize`. I checked via debugger that after `prestartAllCoreThreads` is executed, the `corePoolSize` of workers is created internally.

porunov:
@li-boxuan updated this PR with changes to resolve your comments.
I also added a new test which tests that configuration is correctly provided to the custom ExecutorService implementation. Codacy doesn't like it because it's unsafe (assigning `static` variable in public constructor) but that's just for testing, so I think that's fine.
I also tried to improve the error message we talked here https://github.com/JanusGraph/janusgraph/pull/2700#discussion_r672916777 to make it slightly more clear.

porunov:
Ignored `CustomTestExecutorService.java` in Codacy.

docs/advanced-topics/executor-service.md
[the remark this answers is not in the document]
li-boxuan: Yes, I think that's fine.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/GraphDatabaseConfigurationInstanceIdTest.java
```diff
@@ -108,11 +117,11 @@ public void instanceIdShouldEqualHostnamePlusSuffix() throws UnknownHostExceptio
     }
 
     @Test
-    public void shouldCreateCustomBackendThreadPoolSize() throws UnknownHostException {
+    public void shouldCreateCustomBackendFixedThreadPoolSize() {
```
li-boxuan, line 120: Nitpick and not necessary to resolve: It seems these tests are placed in the wrong place. Looks like this test class was intended for instance id tests.

docs/advanced-topics/executor-service.md
```diff
@@ -0,0 +1,35 @@
...
+When both constructors are available the constructor with `ExecutorServiceConfiguration` argument will be used.
+
+When custom `ExecutorService` is provided with `ExecutorServiceConfiguration` it isn't required to 
+use any of the configuration options provided in `ExecutorServiceConfiguration` but sometimes it is
+convenient to use the provided configurations. The configuration options provided in `ExecutorServiceConfiguration` 
+are typically those configuration options which are provided via configurations from `storage.parallel-backend-executor-service`.
+
+# Cassandra backend ExecutorService
+
+Cassandra backend also uses `ExecutorService` to process CQL queries. The rules by which the Cassandra backend
```
li-boxuan, line 32: (nitpick) I think we can make it clearer by saying sth like "Apart from the general executor service mentioned in the previous section, Cassandra backend uses an additional ExecutorService..."

porunov:
Thank you @li-boxuan . Added fixes to this PR for your latest comments

li-boxuan (approved):
Thank you for this great feature! LGTM 👍

PR #2702, opened 2021-07-10
Optimize profile result for values() step if query.batch=false

### LINKED ISSUE #2185: values() step is much slower than valueMap() step
```groovy
graph = JanusGraphFactory.open("inmemory")
g = graph.traversal()
for (i = 0; i < 100; i++) {
g.addV().property("prop", "value").iterate();
}
g.V().valueMap().profile()
g.V().values().profile()
```
Tested on 0.6.0 SNAPSHOT version.
image
image

li-boxuan:
When query.batch=false, we don't need to replace PropertiesStep with JanusGraphPropertiesStep which
aims to cache the traversers and fire multiple queries in batches.

Notice that JanusGraphPropertiesStep profiles every single query, which makes values().profile()
output extremely lengthy and slow. It also gives users an illusion that values() is much slower
than valueMap() step (provided query.batch=false) which is not true. By using the ordinary
TinkerPop propertiesStep here, the values().profile() becomes faster and produces more concise output.

Fixes #2185

This PR also contains a benchmark program. We can see that without this PR, getValuesProfile is significantly slower than getValueMapProfile, while getValues is as fast as getValueMap.

Before this PR,
```
Benchmark                                       (size)  Mode  Cnt     Score     Error  Units
VertexCentricQueryBenchmark.getValueMap          10000  avgt    5    43.872 ±   8.219  ms/op
VertexCentricQueryBenchmark.getValueMap         100000  avgt    5   523.589 ±  53.983  ms/op
VertexCentricQueryBenchmark.getValueMapProfile   10000  avgt    5    45.226 ±   0.501  ms/op
VertexCentricQueryBenchmark.getValueMapProfile  100000  avgt    5   516.469 ±   8.443  ms/op
VertexCentricQueryBenchmark.getValues            10000  avgt    5    36.449 ±   0.417  ms/op
VertexCentricQueryBenchmark.getValues           100000  avgt    5   499.776 ±   5.431  ms/op
VertexCentricQueryBenchmark.getValuesProfile     10000  avgt    5   122.361 ±   9.163  ms/op
VertexCentricQueryBenchmark.getValuesProfile    100000  avgt    5  1388.586 ± 119.154  ms/op
```

With this PR, getValuesProfile is as fast as getValueMapProfile.
```
Benchmark                                       (size)  Mode  Cnt    Score     Error  Units
VertexCentricQueryBenchmark.getValueMap          10000  avgt    5   40.883 ±   0.764  ms/op
VertexCentricQueryBenchmark.getValueMap         100000  avgt    5  490.148 ±  10.919  ms/op
VertexCentricQueryBenchmark.getValueMapProfile   10000  avgt    5   41.499 ±   0.357  ms/op
VertexCentricQueryBenchmark.getValueMapProfile  100000  avgt    5  574.121 ±  89.151  ms/op
VertexCentricQueryBenchmark.getValues            10000  avgt    5   36.851 ±   1.218  ms/op
VertexCentricQueryBenchmark.getValues           100000  avgt    5  516.501 ±  58.846  ms/op
VertexCentricQueryBenchmark.getValuesProfile     10000  avgt    5   47.177 ±   5.043  ms/op
VertexCentricQueryBenchmark.getValuesProfile    100000  avgt    5  572.094 ± 146.230  ms/op
```

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphLocalQueryOptimizerStrategy.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphLocalQueryOptimizerStrategy.java
@@ -60,10 +60,11 @@ public class JanusGraphLocalQueryOptimizerStrategy extends AbstractTraversalStra
         }
 
         boolean batchPropertyPrefetching = janusGraph.getConfiguration().batchPropertyPrefetching();
+        boolean useMultiQuery = janusGraph.getConfiguration().useMultiQuery();
         int txVertexCacheSize = janusGraph.getConfiguration().getTxVertexCacheSize();
 
         applyJanusGraphVertexSteps(traversal, batchPropertyPrefetching, txVertexCacheSize);
-        applyJanusGraphPropertiesSteps(traversal);
+        applyJanusGraphPropertiesSteps(traversal, useMultiQuery);
         inspectLocalTraversals(traversal);
     }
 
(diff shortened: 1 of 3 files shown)


li-boxuan:
This does not seem to be a good solution. Just noticed there is some logic in JanusGraphPropertiesStep::makeQuery that optimizes a query like `graph.traversal().V().properties().hasKey("weight").hasValue(P.lt(3)).value()` and generates a backend query `[(~key = weight AND ~value < 3 AND PROPERTY AND visibility:normal)]`, while `PropertiesStep` would lead to a backend query `[(PROPERTY AND visibility:normal)]`. In other words, `JanusGraphPropertiesStep` can push down the predicates while `PropertiesStep` cannot.

Therefore I am closing this PR. I don't have a good solution for the profiling issue. Probably we could stop profiling if there are too many similar items.

PR #2705, opened 2021-07-15
Test graph interactions in shouldUpdateIndexFieldsAfterIndexModification

li-boxuan:
This tests that once a new property is added to an existing mixed index
in one graph instance, other graph instances can receive the broadcast message
and expire the stale schema cache.

Continuation of #2699

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -18,6 +18,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Sets;
+import io.github.artsok.RepeatedIfExceptionsTest;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
@@ -3029,9 +3030,9 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
 
     }
 
-    @Test
+    @RepeatedIfExceptionsTest(repeats = 4, minSuccess = 2)
     public void shouldUpdateIndexFieldsAfterIndexModification() throws InterruptedException, ExecutionException {
-        clopen(option(FORCE_INDEX_USAGE), true);
+        clopen(option(FORCE_INDEX_USAGE), true, option(LOG_READ_INTERVAL, MANAGEMENT_LOG), Duration.ofMillis(5000));
         String key1 = "testKey1";
         String key2 = "testKey2";
         String key3 = "testKey3";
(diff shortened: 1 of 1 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -3080,9 +3084,18 @@ public void shouldUpdateIndexFieldsAfterIndexModification() throws InterruptedEx
         mgmt.updateIndex(mgmt.getGraphIndex(indexName), SchemaAction.REINDEX).get();
         mgmt.commit();
 
-        vertex = graph.addVertex(T.label, vertexL, key1, 1L, key2, 2L, key3, 3L);
+        graph2.addVertex(T.label, vertexL, key1, 1L, key2, 2L, key3, 3L);
+        JanusGraphException ex = assertThrows(JanusGraphException.class, () -> graph2.tx().commit());
```
rngcntr, line 3088: This exception only occurs because `graph2` did not have enough time to adopt the schema change yet, right? If so, can we be sure that this has not happened yet?
I think although it is a good idea to have a test for this behavior, this one could easily fail randomly, if the schema change is adopted early.

li-boxuan: You are right, but I cannot come up with a non-intrusive way to make sure `graph2` hasn't be able to read from `ManagementLogger` and adopt the schema changes. FWIW, the interval for ManagementLogger::read call is defined in GraphDatabaseConfiguration::LOG_READ_INTERVAL, which is by default 5 seconds. That's also why we need to wait for 5 seconds below.

I ran this test in GitHub CI for 3 times and many times locally. So far, so good. That being said, let me know if you have a better way to improve the resilience of this test.

rngcntr: The only way I can imagine is by introducing some very ugly `synchronized` blocks and locks within the `ManagementLogger::read` method. But what about marking this test as flaky?

li-boxuan: I marked this test as `RepeatedIfExceptionsTest`.

rngcntr: That would cover the case where 5 seconds are not enough time for the synchronization process. My concerns were about the first test. But as I have no idea how to set up a test environment for the first case either, I am fine with leaving it as is.

rngcntr (approved):

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
li-boxuan: `RepeatedIfExceptionsTest` is intended to cover the first test assertion. If L3088 does not throw an exception, then JUnit will throw an `AssertionFailedError`, which leads to a retry of the whole test. The scenario for L3088 to fail the test is that schema changes might get adopted too soon. Note that schema changes are read every 5 seconds. Thus, the chance for the first case to fail is existent but small. By retrying 5 times, we reduce the possibility to a very minimal level.

rngcntr: Alright then. I did not know that `@RepeatedIfExceptionsText` also repeats the test for assertion errors.

Issue #455, opened 2017-08-09
Error acquiring write lock while committing management transaction with bigtable backend


amyth:
Hey, 
I am using janusgraph with bigtable backend and am trying to create a property key. As soon as I commit this transaction using `management.commit()`, I get the following error:

```
org.janusgraph.core.JanusGraphException: Could not commit transaction due to exception during persistence
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.commit(StandardJanusGraphTx.java:1374)
	at org.janusgraph.graphdb.database.management.ManagementSystem.commit(ManagementSystem.java:235)
	at org.janusgraph.core.schema.JanusGraphManagement$commit$0.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:117)
	at groovysh_evaluate.run(groovysh_evaluate:3)
	at org.codehaus.groovy.vmplugin.v7.IndyInterface.selectMethod(IndyInterface.java:232)
	at org.codehaus.groovy.tools.shell.Interpreter.evaluate(Interpreter.groovy:70)
	at org.codehaus.groovy.tools.shell.Groovysh.execute(Groovysh.groovy:190)
	at org.apache.tinkerpop.gremlin.console.GremlinGroovysh.super$3$execute(GremlinGroovysh.groovy)
	at sun.reflect.GeneratedMethodAccessor14.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1215)
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.invokeMethodOnSuperN(ScriptBytecodeAdapter.java:132)
	at org.apache.tinkerpop.gremlin.console.GremlinGroovysh.execute(GremlinGroovysh.groovy:72)
[...]

PR #2716, opened 2021-07-22
Change LOCK_WAIT configuration from GLOBAL_OFFLINE to MASKABLE

li-boxuan:
LOCK_WAIT is used in JanusGraph locking protocol to determine how long the system should wait
for a lock application to be acked by the storage backend (see ConsistentKeyLocker::writeSingleLock).
It is also the time waited at the end of all lock applications before verifying that
lock applications were successful (see ConsistentKeyLocker::checkSingleLock).

The recommended value is a small multiple of the average consistent write time, thus typically a
small number, e.g. 100 milliseconds. However, in some use cases, the consistent write time can
be much larger. For example, there can be a significant latency for developers to launch a JanusGraph
instance locally, connecting to the storage backend located in another country. In this case, a short LOCK_WAIT
would lead to timeouts and thus transactions would fail. Thus, it would be useful if this LOCK_WAIT
is maskable. Despite this legitimate usecase, this commit adds a notice to remind users of potential
risks when using different LOCK_WAIT values across JanusGraph instances.

Related to https://github.com/JanusGraph/janusgraph/issues/455

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -648,16 +648,12 @@ public boolean apply(@Nullable String s) {
             "Number of times the system attempts to acquire a lock before giving up and throwing an exception",
             ConfigOption.Type.MASKABLE, 3);
 
-    /**
-     * The number of milliseconds the system waits for a lock application to be acknowledged by the storage backend.
-     * Also, the time waited at the end of all lock applications before verifying that the applications were successful.
-     * This value should be a small multiple of the average consistent write time.
-     */
```
porunov, line 655: Why this comment was removed?

porunov (commented):
I have only one question. Otherwise looks good!

Also, this PR might be related to this issue / comment: https://github.com/JanusGraph/janusgraph/issues/455#issuecomment-646446405

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
li-boxuan: I removed it because it simply duplicates the code without any extra information.

porunov: Oh, you are right. I somehow checked added lines only without checking unchanged lines. Now I see that it's just a duplication

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #2718, opened 2021-07-23
Use latest KeyInformation in LuceneIndex

rayokota:
The KeyInformation is cached in the LuceneCustomAnalyzer.
This data may be stale if additional keys have been added
to the index.  Therefore we always use the latest
KeyInformation that is passed in. 

In our production use case, we found that the missing KeyInformation
was causing the wrong analyzer to be used.

It appears that SolrIndex and ElasticIndex don't cache KeyInformation,
so don't have this problem.

Some get/put calls were also replaced with computeIfAbsent.

(It may be possible to remove the `writeLock` by using a `ConcurrentHashMap`,
but I'll leave that for a future investigation.)

Signed-off-by: Robert Yokota <rayokota@gmail.com>

-----

### For all changes:

### For code changes:


rayokota:
@porunov @li-boxuan , I also found some stale index key data is cached in the LuceneIndex.  Please take a look.  Thanks!

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -742,10 +736,10 @@ private Query combineTerms(String key, List<String> terms, Function<Term, Query>
     }
 
     private LuceneCustomAnalyzer delegatingAnalyzerFor(String store, KeyInformation.IndexRetriever information2) {
-        if (!delegatingAnalyzers.containsKey(store)) {
-            delegatingAnalyzers.put(store, new LuceneCustomAnalyzer(store, information2, Analyzer.PER_FIELD_REUSE_STRATEGY));
-        }
-        return delegatingAnalyzers.get(store);
+        LuceneCustomAnalyzer analyzer = delegatingAnalyzers.computeIfAbsent(store, key ->
+            new LuceneCustomAnalyzer(store, information2, Analyzer.PER_FIELD_REUSE_STRATEGY));
+        analyzer.setInformations(information2);
```
rayokota, line 741: Update the KeyInformation here.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -207,44 +207,42 @@ private Directory getStoreDirectory(String store) throws BackendException {
...
+                return new IndexWriter(getStoreDirectory(store), iwc);
+            } catch (final IOException | BackendException e) {
+                log.error("Could not create writer", e);
+                return null;
             }
+        });
+        if (writer == null) {
+            throw new PermanentBackendException("Could not create writer");
         }
+        ((LuceneCustomAnalyzer) writer.getAnalyzer()).setInformations(informations);
```
rayokota, line 224: Update the KeyInformation here.

li-boxuan:
Thank you for your contribution! Is it possible to write a test case to reproduce the problem you said?

li-boxuan:
Apologize - I closed this PR by mistake. Reopened.

rayokota:
Thanks for taking a look @li-boxuan .  I've added a unit test that fails without this fix.

li-boxuan:
>  I've added a unit test that fails without this fix.

@rayokota I tried your unit test but it succeeded (on master branch) even without this fix.

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -207,44 +207,42 @@ private Directory getStoreDirectory(String store) throws BackendException {
...
             iwc.setOpenMode(IndexWriterConfig.OpenMode.CREATE_OR_APPEND);
             try {
-                writer = new IndexWriter(getStoreDirectory(store), iwc);
-                writers.put(store, writer);
-            } catch (final IOException e) {
-                throw new PermanentBackendException("Could not create writer", e);
+                return new IndexWriter(getStoreDirectory(store), iwc);
+            } catch (final IOException | BackendException e) {
+                log.error("Could not create writer", e);
+                return null;
```
li-boxuan, line 218: I personally prefer the old style, i.e. without using `computeIfAbsent`. Now if the user has a higher log level, this root cause will not be available to users.

li-boxuan:
Hi @rayokota  do you have a chance to address https://github.com/JanusGraph/janusgraph/pull/2718#issuecomment-887953075?

Issue #2722, opened 2021-07-27
NoSuchMethodError when enabling CQL driver metrics


cdegroc:
### Summary
While testing JanusGraph's master branch (0.6.0), I've stumbled on the following issue and would appreciate if anyone could confirm it.

- Version: master (a8925e90cdd5eb872ee8f937c9b56d9bf0f3caab)
- Storage Backend: CQL
- Expected Behavior: JanusGraph would start with CQL driver metrics enabled
- Current Behavior: `java.lang.NoSuchMethodError: com.codahale.metrics.MetricRegistry.gauge` is thrown
- Steps to Reproduce:
  - Using JanusGraph's master branch
  - Build and start a docker image running `janusgraph-server.sh`
  - Create a graph from a configuration including properties
```
storage.cql.metrics.node-enabled = pool.open-connections,pool.available-streams,pool.in-flight,bytes-sent,bytes-received,cql-messages
storage.cql.metrics.session-enabled = bytes-sent,bytes-received,cql-requests,cql-client-timeouts,throttling.delay,throttling.queue-size,throttling.errors
```
  - Graph creation fails with a `NoSuchMethodError` (stack trace below)

### Stack Trace

```
jce-janusgraph   | 2021-07-27 13:42:34,895 [gremlin-server-session-1] WARN  org.apache.tinkerpop.gremlin.server.op.AbstractEvalOpProcessor - 0 0 - Exception processing a script on request [RequestMessage{, requestId=774b3c82-18a5-4a92-b01e-65acc0812a22, op='eval', processor='session', args={gremlin=graph = createGraph(graphName), session=5a48bd92-4457-4cd0-9224-8f93fbdfd484, manageTransaction=false, evaluationTimeout=600000, userAgent=Gremlin Console/1.21.0, batchSize=64, maintainStateAfterException=false}}].
jce-janusgraph   | java.lang.NoSuchMethodError: com.codahale.metrics.MetricRegistry.gauge(Ljava/lang/String;Lcom/codahale/metrics/MetricRegistry$MetricSupplier;)Lcom/codahale/metrics/Gauge;
jce-janusgraph   | 	at com.datastax.oss.driver.internal.core.metrics.DropwizardMetricUpdater.lambda$initializeGauge$1(DropwizardMetricUpdater.java:110) ~[java-driver-core-4.12.0.jar:na]
[...]

PR #2723, opened 2021-07-27
Exclude metrics-core 3.x from gremlin-server

cdegroc:
- Exclude `metrics-core` 3.x from gremlin-server.
- Import JmxReporter from `metrics-jmx` as it moved there in 4.x.

Fixes issue #2722

Signed-off-by: Clement de Groc <clement.degroc@datadoghq.com>

-----

### For all changes:

### For code changes:


li-boxuan:
@cdegroc Thank you for finding and fixing the problem! Is it possible to include a test that reproduces what you said in #2722?

cdegroc:
> @cdegroc Thank you for finding and fixing the problem! Is it possible to include a test that reproduces what you said in #2722?

Hey @li-boxuan 👋  I completely understand your point and having a single test we could keep would be great.

It's a bit harder here, because the issue only appears when JanusGraph's classpath has jars in a specific order.
The order can be different when JanusGraph is started with `janusgraph-server.sh` and in unit tests (and even there, it can depend on the content of Maven's pom files).

I've tried my best nonetheless, and I've added a test demonstrating the issue here: https://github.com/cdegroc/janusgraph/commit/fbeba221e46aa44b19115eac78b9dce6a7027479
You can see the test [failing in the CI](https://github.com/cdegroc/janusgraph/runs/3182468766) and should be able to reproduce the issue locally.

Of course, for such a change, we could also just wait for more users to test 0.6/master to see if they hit this issue.

porunov (approved):
LGTM. Thank you @cdegroc !

li-boxuan:
@cdegroc Thanks for the explanation! I think the `testCqlMetricsInitialization` test in your commit fbeba221e46aa44b19115eac78b9dce6a7027479 is worth including in this PR. What do you think?

cdegroc:
> @cdegroc Thanks for the explanation! I think the `testCqlMetricsInitialization` test in your commit [fbeba22](https://github.com/JanusGraph/janusgraph/commit/fbeba221e46aa44b19115eac78b9dce6a7027479) is worth including in this PR. What do you think? This is just to ensure (to some extent) the same problem does not occur again.

@li-boxuan I've added the test to the PR.

porunov:
@cdegroc could you please rebase you PR against the latest commit in `master`? It has a fix for `dist-tests`.

li-boxuan (approved):
LGTM 👍

PR #2741, opened 2021-08-03
Make CQL executor service usage optional [cql-tests]

### LINKED ISSUE #2707: Make CQL ExecutorService optional
As discussed here: https://github.com/JanusGraph/janusgraph/discussions/2545
And noted by @li-boxuan here: https://github.com/JanusGraph/janusgraph/pull/2700#discussion_r670907692

We can limit the number of concurrent CQL requests using `MAX_REQUESTS_PER_CONNECTION`. Looks like using a separate ExecutorService just adds additional overhead. I think adding possibility to use CQL storage backend without additional ExecutorService make sense. I would propose making ExecutorService optional and then remove it completely in the later version if we found that it's not useful at all.

porunov:
This PR adds possibility to work with CQL without ExecutorService at all. I.e. relying on `MAX_REQUESTS_PER_CONNECTION` and internal DataStax pools instead of relying on the default or custom pool.
Potentially working with disabled CQL ExecutorService should be the same or better in performance and have less overhead (i.e. less unnecessary created threads).
If we prove that working with disabled ExecutorService is better in terms of memory management and / or performance, we can disable CQL ExecutorService by default in JanusGraph and potentially deprecate / remove the usage of unnecessary CQL ExecutorService.

Performance tests conducted:
https://github.com/JanusGraph/janusgraph/pull/2741#issuecomment-894331436
https://github.com/JanusGraph/janusgraph/pull/2741#issuecomment-894635159
https://github.com/JanusGraph/janusgraph/pull/2741#issuecomment-894656248
https://github.com/JanusGraph/janusgraph/pull/2741#issuecomment-894663680

Before opening this PR I also checked that `CQLGraphTest` tests passes with disabled ExecutorService by default (i.e. tests in `JanusGraphTest`) but if we test all `CQLGraphTest` tests with enabled and disabled `ExecutorService` it will double the execution time for CQL tests. Thus, I simply extended `testConsistencyEnforcement` and `testConcurrentConsistencyEnforcement` to run their parametrized tests with both enabled and disabled `ExecutorService`. I don't know, but maybe we need to add a flow to trigger all CQL tests with disabled `ExecutorService` when we create a commit with `[cql-tests]`. 

Related discussions with @li-boxuan and @cdegroc:
https://github.com/JanusGraph/janusgraph/pull/2700#discussion_r669480986
https://github.com/JanusGraph/janusgraph/discussions/2545#discussioncomment-551453

Fixes #2707

-----

### For all changes:

### For code changes:


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -347,20 +348,41 @@ public String getName() {
...
+        CompletableFuture<AsyncResultSet> asyncResult = this.session.executeAsync(this.getSlice.boundStatementBuilder()
+            .setByteBuffer(KEY_BINDING, query.getKey().asByteBuffer())
+            .setByteBuffer(SLICE_START_BINDING, query.getSliceStart().asByteBuffer())
+            .setByteBuffer(SLICE_END_BINDING, query.getSliceEnd().asByteBuffer())
+            .setInt(LIMIT_BINDING, query.getLimit())
+            .setConsistencyLevel(getTransaction(txh).getReadConsistencyLevel()).build())
+            .toCompletableFuture();
+        AsyncResultSet asyncResultSet = interruptibleWait(asyncResult);
+        return fromResultSet(asyncResultSet, this.getter);
+    }
```
cdegroc, line 384: NIT: Most of the code is identical. I would put this code in a shared method and move the part that's different in `getSlice`. What do you think?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/function/CQLExecutionServiceMutateManyLoggedFunction.java
```diff
@@ -0,0 +1,55 @@
...
+import org.janusgraph.diskstorage.BackendException;
+import org.janusgraph.diskstorage.cql.CQLKeyColumnValueStore;
+import org.janusgraph.diskstorage.util.time.TimestampProvider;
+
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+
+import static org.janusgraph.diskstorage.cql.CQLKeyColumnValueStore.EXCEPTION_MAPPER;
+
+public class CQLExecutionServiceMutateManyLoggedFunction extends AbstractCQLMutateManyLoggedFunction{
```
cdegroc, line 30: NIT: ExecutorService instead of ExecutionService?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/function/CQLSimpleMutateManyUnloggedFunction.java
```diff
@@ -0,0 +1,99 @@
+// Copyright 2021 JanusGraph Authors
```
cdegroc, line 1: I've ran a diff between this file and CQLExecutionServiceMutateManyUnloggedFunction and most of the code is identical. Is there a way we could isolate the Future/CompletableFuture diffs so the rest of the code remains shared between classes?

cdegroc:
Thanks for doing this @porunov 👍  I'm really curious to see the performance differences!
If the goal is to merge this as-is and not just run performance tests, my main concern would be about code duplication and maintainability.

porunov:
@cdegroc Thanks for the review! I applied your suggestions and tried to reduce code duplication more.

porunov:
> Thanks for doing this @porunov +1 I'm really curious to see the performance differences!

The main purpose of this PR is to be able to run the same CQL queries without using potentially unnecessary threads. I don't think there will be much performance difference between enabled and disabled ExecutorService.
The only difference I could think of is if you have high throughput ability on your server but you configured high `storage.cql.max-requests-per-connection` and low `storage.cql.executor-service.max-pool-size`. If so, yes, there should theoretically be performance difference. Other than that, performance improvement could also come from less context switches between different threads when we are not using additional ExecutorService (i.e. we simply have less unnecessary threads on a server) but I wouldn't give that much expectations either. 

That said, I didn't really conduct and performance testing.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/function/mutate/CQLSimpleMutateManyUnloggedFunction.java
```diff
@@ -0,0 +1,67 @@
...
+                    resultList.add(completableFuture);
+                })
+            );
+        });
+
+        for(CompletableFuture<AsyncResultSet> resultPart : resultList){
+            try {
+                resultPart.get();
+            } catch (InterruptedException | ExecutionException e) {
+                return Optional.of(e);
```
li-boxuan, line 61: Question: if we don't call `.get()` for the rest of futures, will it cause any leak?

porunov: I don't think so. Future isn't necessarily to be executed. We will simply loose links to all futures here and they should be picked up by GC unless we messed anywhere.

porunov: Here is a good answer regarding it: https://stackoverflow.com/a/30707238/3846192

porunov:
Made very simple benchmark test which simply adds 100,000 empty vertices, 2000 vertices with a property and 1000 edges with a property.
After that it traverses all vertices and gets properties for these vertices as well as edges for the vertices and properties of the edges.
The test with enabled by default CQL executor service is located here: https://gist.github.com/porunov/ce3422a8266d9feb4c89452ff705b9ce
The test with disabled CQL executor service is located here (the same test but with added `configuration.set(EXECUTOR_SERVICE_ENABLED.toStringWithoutRoot(), false);`): https://gist.github.com/porunov/9b0c10ec0c06c1ba32695897db9b2412

The results are below:
```
VM version: JDK 11.0.12, OpenJDK 64-Bit Server VM, 11.0.12+7
```

**On master branch with enabled CQL executor service**:
```
Benchmark                             (assignTimestamp)  (atomicBatch)  (batchSize)  Mode  Cnt      Score   Error  Units
TraverseBenchmark.testTraversingTime               true           true           20  avgt       49474.682          ms/op
TraverseBenchmark.testTraversingTime               true           true            1  avgt       49918.248          ms/op
TraverseBenchmark.testTraversingTime               true          false           20  avgt       42622.825          ms/op
TraverseBenchmark.testTraversingTime               true          false            1  avgt       44450.675          ms/op
TraverseBenchmark.testTraversingTime              false           true           20  avgt       44478.228          ms/op
TraverseBenchmark.testTraversingTime              false           true            1  avgt       40274.519          ms/op
TraverseBenchmark.testTraversingTime              false          false           20  avgt       41803.870          ms/op
TraverseBenchmark.testTraversingTime              false          false            1  avgt       41076.673          ms/op
```

**On this branch with enabled by default CQL executor service**:
```
Benchmark                             (assignTimestamp)  (atomicBatch)  (batchSize)  Mode  Cnt      Score   Error  Units
TraverseBenchmark.testTraversingTime               true           true           20  avgt       40880.514          ms/op
TraverseBenchmark.testTraversingTime               true           true            1  avgt       41069.039          ms/op
TraverseBenchmark.testTraversingTime               true          false           20  avgt       41636.535          ms/op
TraverseBenchmark.testTraversingTime               true          false            1  avgt       41224.890          ms/op
TraverseBenchmark.testTraversingTime              false           true           20  avgt       40837.060          ms/op
TraverseBenchmark.testTraversingTime              false           true            1  avgt       41950.444          ms/op
TraverseBenchmark.testTraversingTime              false          false           20  avgt       40490.052          ms/op
TraverseBenchmark.testTraversingTime              false          false            1  avgt       41422.679          ms/op
```

**On this branch with disabled CQL executor service**:
```
Benchmark                             (assignTimestamp)  (atomicBatch)  (batchSize)  Mode  Cnt      Score   Error  Units
TraverseBenchmark.testTraversingTime               true           true           20  avgt       36145.111          ms/op
TraverseBenchmark.testTraversingTime               true           true            1  avgt       35734.069          ms/op
TraverseBenchmark.testTraversingTime               true          false           20  avgt       36026.026          ms/op
TraverseBenchmark.testTraversingTime               true          false            1  avgt       36430.664          ms/op
TraverseBenchmark.testTraversingTime              false           true           20  avgt       35902.646          ms/op
TraverseBenchmark.testTraversingTime              false           true            1  avgt       35687.994          ms/op
TraverseBenchmark.testTraversingTime              false          false           20  avgt       36476.968          ms/op
TraverseBenchmark.testTraversingTime              false          false            1  avgt       35778.240          ms/op
```

As seen above, the execution time for tests on `master` branch and this branch with enabled by default CQL executor service is similar in performance. The execution time for tests with disabled CQL executor service on this branch is lower for all the cases tested.
Thus, I can confirm that working with disabled executor service may be better in performance in some situations. I can't say in all situations yet because I simply didn't test all the cases but probably most of the cases will benefit from disabled CQL executor service (at least we can see about 10% increase in performance for this case on my machine). 
Notice, this test is executing benchmarks in one thread, thus it doesn't test the multi threaded access by multiple clients to the same `CQLStoreManager`. Probably, to test that case we could remove `RUN_COUNT` (to always return the same `keyspace` configuration) and add additional check to `setup` method which will not create schema & data if schema is already initialized.

P.S. I didn't expect to see 10% improvement for the test. I think it could be just a bad test. More various tests are needed to confirm the performance improvement.

porunov:
Simple write test: https://gist.github.com/porunov/6a6f1b0a5dd654871714f663d4a83c51
To test disabled executor service you need to add `configuration.set(EXECUTOR_SERVICE_ENABLED.toStringWithoutRoot(), false);` to `setup` method.

Results are below.

**Master branch**:
```
Benchmark                     (assignTimestamp)  (atomicBatch)  (batchSize)  Mode  Cnt    Score      Error  Units
WriteBenchmark.testWriteTime               true           true           20  avgt    3  820.602 ± 1239.062  ms/op
WriteBenchmark.testWriteTime               true           true            1  avgt    3  742.753 ± 2447.104  ms/op
WriteBenchmark.testWriteTime               true          false           20  avgt    3  730.592 ± 2318.229  ms/op
WriteBenchmark.testWriteTime               true          false            1  avgt    3  826.049 ± 4052.735  ms/op
WriteBenchmark.testWriteTime              false           true           20  avgt    3  777.851 ± 2107.037  ms/op
WriteBenchmark.testWriteTime              false           true            1  avgt    3  776.049 ± 2353.500  ms/op
WriteBenchmark.testWriteTime              false          false           20  avgt    3  714.974 ± 2048.363  ms/op
WriteBenchmark.testWriteTime              false          false            1  avgt    3  838.614 ± 5144.424  ms/op
```

**This branch with enabled CQL executor service by default**:
```
Benchmark                     (assignTimestamp)  (atomicBatch)  (batchSize)  Mode  Cnt    Score      Error  Units
WriteBenchmark.testWriteTime               true           true           20  avgt    3  750.470 ± 1792.536  ms/op
WriteBenchmark.testWriteTime               true           true            1  avgt    3  902.898 ± 6193.790  ms/op
WriteBenchmark.testWriteTime               true          false           20  avgt    3  702.975 ± 2318.588  ms/op
WriteBenchmark.testWriteTime               true          false            1  avgt    3  840.229 ± 5032.483  ms/op
WriteBenchmark.testWriteTime              false           true           20  avgt    3  835.900 ± 4783.879  ms/op
WriteBenchmark.testWriteTime              false           true            1  avgt    3  899.230 ± 6222.718  ms/op
WriteBenchmark.testWriteTime              false          false           20  avgt    3  736.021 ± 2091.344  ms/op
WriteBenchmark.testWriteTime              false          false            1  avgt    3  819.863 ± 4776.885  ms/op
```

**This branch with disabled CQL executor service**:
```
Benchmark                     (assignTimestamp)  (atomicBatch)  (batchSize)  Mode  Cnt    Score      Error  Units
WriteBenchmark.testWriteTime               true           true           20  avgt    3  883.683 ± 6222.681  ms/op
WriteBenchmark.testWriteTime               true           true            1  avgt    3  761.552 ± 1324.556  ms/op
WriteBenchmark.testWriteTime               true          false           20  avgt    3  687.512 ± 1920.683  ms/op
WriteBenchmark.testWriteTime               true          false            1  avgt    3  733.218 ± 3038.014  ms/op
WriteBenchmark.testWriteTime              false           true           20  avgt    3  766.956 ± 2653.458  ms/op
WriteBenchmark.testWriteTime              false           true            1  avgt    3  878.886 ± 6276.718  ms/op
WriteBenchmark.testWriteTime              false          false           20  avgt    3  683.059 ± 2509.469  ms/op
WriteBenchmark.testWriteTime              false          false            1  avgt    3  682.410 ± 2408.270  ms/op
```

Conclusion about the "write" test: Write tests didn't show performance differences between `master` branch with enabled CQL executor service, this branch with enabled executor service and this branch with disabled executor service.

porunov:
Multi threaded test: https://gist.github.com/porunov/0b01f62da2440f9567c0a47e82c0785d

I was comparing the result of the test to this one: https://github.com/JanusGraph/janusgraph/pull/2741#issuecomment-894331436

Both tests do the same (i.e. same data is created and same data is traversed).

Notice, I tested only the first case when `assignTimestamp = true, atomicBatch = true, batchSize = 20`.  To test with all cases, change the test parameters to include all values (with all parameters the amount of tests will be `48`).

As always, to test disable executor service you need to add `configuration.set(EXECUTOR_SERVICE_ENABLED.toStringWithoutRoot(), false);` to `setup` method.

The results are the next:

**Master branch (enabled CQL executor service)**:
```
Benchmark                                          (assignTimestamp)  (atomicBatch)  (batchSize)  (threadsCount)  Mode  Cnt       Score   Error  Units
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               1  avgt        40961.136          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               3  avgt        37205.040          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               5  avgt        44785.057          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              10  avgt        51406.648          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              20  avgt        80802.159          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              50  avgt       120320.723          ms/op
```

**This branch with enabled CQL executor service by default**:

```
Benchmark                                          (assignTimestamp)  (atomicBatch)  (batchSize)  (threadsCount)  Mode  Cnt       Score   Error  Units
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               1  avgt        41784.622          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               3  avgt        37371.193          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               5  avgt        45240.449          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              10  avgt        51302.146          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              20  avgt        77971.718          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              50  avgt       120231.977          ms/op
```

**This branch with disabled CQL executor service**:
```
Benchmark                                          (assignTimestamp)  (atomicBatch)  (batchSize)  (threadsCount)  Mode  Cnt      Score   Error  Units
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               1  avgt       36841.976          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               3  avgt       42757.243          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               5  avgt       43562.073          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              10  avgt       45961.538          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              20  avgt       54717.432          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              50  avgt       90342.676          ms/op
```

As seen above, again, `master` branch and this branch are the same in performance when executor service is enabled (default case).
When it comes to disabled CQL executor service, we see performance improvement.
There was a statistical anomaly in the second case when the amount of threads were `3`. The execution time had to be similar or bigger than the execution time with `1` thread but was faster. Thus, I think we can disregard that case.

porunov:
Also, there is a possibility that we reach some limits in some of the fixed executor pools, so to be sure we don't bump into the limit I rerun above tests with `cached` thread pools. I.e. I added the next two configurations:
```
configuration.set(PARALLEL_BACKEND_EXECUTOR_SERVICE_CLASS.toStringWithoutRoot(), "cached");
configuration.set(EXECUTOR_SERVICE_CLASS.toStringWithoutRoot(), "cached");
```

The results with enabled CQL executor pool got a little bit better but still not as good as with disabled CQL executor pool.

**This branch with enabled CQL executor service**:
```
Benchmark                                          (assignTimestamp)  (atomicBatch)  (batchSize)  (threadsCount)  Mode  Cnt       Score   Error  Units
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               1  avgt        38876.357          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               3  avgt        36788.180          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               5  avgt        44827.199          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              10  avgt        51498.062          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              20  avgt        64224.704          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              50  avgt       111967.714          ms/op
```

**This branch with disabled CQL executor service**:
```
Benchmark                                          (assignTimestamp)  (atomicBatch)  (batchSize)  (threadsCount)  Mode  Cnt      Score   Error  Units
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               1  avgt       36634.309          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               3  avgt       41743.145          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20               5  avgt       44011.099          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              10  avgt       45750.356          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              20  avgt       56144.066          ms/op
TraverseMultithreadedBenchmark.testTraversingTime               true           true           20              50  avgt       93831.007          ms/op
```

li-boxuan (commented):
The code looks good to me, I just have a nitpick about the doc.

docs/configs/janusgraph-cfg.md
```diff
@@ -413,6 +413,7 @@ Configuration options for CQL executor service which is used to process CQL quer
 | ---- | ---- | ---- | ---- | ---- |
 | storage.cql.executor-service.class | The implementation of `ExecutorService` to use. The full name of the class which extends `ExecutorService` which has either a public constructor with `ExecutorServiceConfiguration` argument (preferred constructor) or a public parameterless constructor. Other accepted options are: `fixed` - fixed thread pool size of `max-pool-size` size; `cached` - cached thread pool size; | String | fixed | LOCAL |
 | storage.cql.executor-service.core-pool-size | Core pool size for executor service. May be ignored if custom executor service is used (depending on the implementation of the executor service). | Integer | 10 | LOCAL |
+| storage.cql.executor-service.enabled | Whether to use CQL executor service to process queries or not. If not used, the parallelism will be controlled internally by DSE driver via `storage.cql.max-requests-per-connection` parameter only which may be preferable in some situations. | Boolean | true | LOCAL |
```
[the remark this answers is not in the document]
porunov: Updated configuration parameter description and added more information to the documentation about CQL executor service. Here is the diff: https://github.com/JanusGraph/janusgraph/compare/246ee51168c6159bf054f81ea05f68ae65f34208..ffe4150b938f6483f56922a9032a637a3d702e0b

li-boxuan (approved):
LGTM 👍  Thank you for the huge contribution!

cdegroc:
Thanks @porunov! I was out and couldn't review it again. But I'm looking forward to testing it! 👏

Issue #2746, opened 2021-08-06
config skip-schema-check=true is not honored for HBase


jigs1993:
- Version: latest (tested lowest was 0.3.3)
- Storage Backend: Hbase
- Link to discussed bug: https://lists.lfaidata.foundation/g/janusgraph-users/message/6065
- Expected Behavior: when `skip-schema-check` is set in the janusgraph config to `true` the schema check(which requires CREATE perm in Hbase) should be ignore at all places 
- Current Behavior: while calling `getDeployment()` in [HbaseStorageManager](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-hbase/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java) it tries to access the [code](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-hbase/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java#L732) in hbase which requires create permission without checking `skip-schema-check` config value
- Steps to Reproduce: run it on hbase storage backend without create permission to the user, mostly for the range queries requiring key partitions

### Stack Trace 

```
2021-08-05 22:21:15,580 WARN [uber-SubtaskRunner] org.janusgraph.diskstorage.hbase.HBaseStoreManager: Unexpected exception during getDeployment()
java.lang.RuntimeException: org.janusgraph.diskstorage.TemporaryBackendException: Temporary failure in storage backend
	at org.janusgraph.diskstorage.hbase.HBaseStoreManager.getDeployment(HBaseStoreManager.java:379)
	at org.janusgraph.diskstorage.hbase.HBaseStoreManager.getFeatures(HBaseStoreManager.java:418)
	at org.janusgraph.graphdb.configuration.builder.GraphDatabaseConfigurationBuilder.build(GraphDatabaseConfigurationBuilder.java:51)
	at org.janusgraph.core.JanusGraphFactory.open(JanusGraphFactory.java:161)
	at org.janusgraph.core.JanusGraphFactory.open(JanusGraphFactory.java:132)
	at org.janusgraph.core.JanusGraphFactory.open(JanusGraphFactory.java:79)
	...
Caused by: org.janusgraph.diskstorage.TemporaryBackendException: Temporary failure in storage backend
	at org.janusgraph.diskstorage.hbase.HBaseStoreManager.ensureTableExists(HBaseStoreManager.java:748)
[...]

PR #2748, opened 2021-08-06
honoring skip-schema-check for hbase

jigs1993:
Signed-off-by: jigs1993 <jigar.9408266552@gmail.com>

discussion thread: https://lists.lfaidata.foundation/g/janusgraph-users/message/6065

ISSUE: https://github.com/JanusGraph/janusgraph/issues/2746
-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-hbase/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
+++ b/janusgraph-hbase/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
@@ -540,8 +540,14 @@ public class HBaseStoreManager extends DistributedStoreManager implements KeyCol
     public List<KeyRange> getLocalKeyPartition() throws BackendException {
         List<KeyRange> result = new LinkedList<>();
         try {
-            ensureTableExists(
-                tableName, getCfNameForStoreName(GraphDatabaseConfiguration.SYSTEM_PROPERTIES_STORE_NAME), 0);
+            if (skipSchemaCheck) {
+                logger.debug("Skipping schema check");
+                if (!exists()) throw new PermanentBackendException("Table " + tableName + " doesn't exist in HBase!");
+            } else {
+                logger.debug("Performing schema check");
+                ensureTableExists(
+                    tableName, getCfNameForStoreName(GraphDatabaseConfiguration.SYSTEM_PROPERTIES_STORE_NAME), 0);
+            }
             Map<KeyRange, ServerName> normed = normalizeKeyBounds(cnx.getRegionLocations(tableName));
 
             for (Map.Entry<KeyRange, ServerName> e : normed.entrySet()) {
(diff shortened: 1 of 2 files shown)


porunov:
@jigs1993 thank you for the bug fix. Let us know if you have problems with signing CLA via EasyCLA.
Also, would you be able to add a test case for this bug, so that we don't bump into it in the future? Normally a test should fail on the master branch but succeed on a bugfix branch.

porunov (approved):
LGTM. Thank you @jigs1993 !

jigs1993:
Thanks @porunov!

janusgraph-hbase/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -540,8 +540,14 @@ public boolean exists() throws BackendException {
     public List<KeyRange> getLocalKeyPartition() throws BackendException {
         List<KeyRange> result = new LinkedList<>();
         try {
-            ensureTableExists(
-                tableName, getCfNameForStoreName(GraphDatabaseConfiguration.SYSTEM_PROPERTIES_STORE_NAME), 0);
+            if (skipSchemaCheck) {
+                logger.debug("Skipping schema check");
+                exists();
```
li-boxuan, line 545: Question: when we skip schema check, what is the point of calling `exists()` method?

jigs1993: I guess the purpose of this parameter is to skip all DDL on hbase when `skip-schema-check=true` assuming user do not have create/admin privilege. In this can we assume that the tables are already present and all column families are created beforehand. So this check is just to verify the existence of the table and fail early if it is not present

li-boxuan: Got it. In that case should we throw an exception if `exists()` returns false? Otherwise, calling this method seemingly does nothing.

porunov: @jigs1993 will you be able to change the logic of this case to throw an exception as suggested by @li-boxuan ? Unless you have other thoughts about it

jigs1993: @li-boxuan @porunov thanks for pointing this out! I will update it to throw the exception if table doesn't exist

janusgraph-hbase/src/main/java/org/janusgraph/diskstorage/hbase/HBaseStoreManager.java
```diff
@@ -540,8 +540,14 @@ public boolean exists() throws BackendException {
     public List<KeyRange> getLocalKeyPartition() throws BackendException {
         List<KeyRange> result = new LinkedList<>();
         try {
-            ensureTableExists(
-                tableName, getCfNameForStoreName(GraphDatabaseConfiguration.SYSTEM_PROPERTIES_STORE_NAME), 0);
+            if (skipSchemaCheck) {
+                logger.debug("Skipping schema check");
+                if (!exists()) throw new TemporaryBackendException("Table " + tableName + " doesn't exist in HBase!");
```
li-boxuan, line 545: This should be a `PermanentBackendException`. `TemporaryBackendException` is used when the unavailability is temporary.

li-boxuan (commented):
Hi @jigs1993 I only have one suggestion, and the rest looks good to me! Can you please also squash all your commits into a single commit?

li-boxuan (approved):
LGTM 👍  I amended your commit message and changed your test by a little bit (using `assertThrows` to make sure the expected exception is thrown).

PR #2756, opened 2021-08-14
Fix scala dependency conflict in spark traversal [cql-tests]

li-boxuan:
This aims to fix Spark traversal caused by scala version mismatch. Tested locally using the following steps:

```
:plugin use tinkerpop.hadoop
:plugin use tinkerpop.spark
graph = GraphFactory.open('conf/hadoop-graph/read-cql.properties')
g = graph.traversal().withComputer(SparkGraphComputer)
g.V().count()
```

TODOs:

- [x] Check if Spark traversal works for HBase
- [x] Check if Spark traversal works on a Spark standalone cluster
- [ ] Check if Spark traversal works on a Spark yarn cluster

### DIFF
--- a/janusgraph-cql/pom.xml
+++ b/janusgraph-cql/pom.xml
@@ -71,7 +71,6 @@
         <dependency>
             <groupId>org.apache.cassandra</groupId>
             <artifactId>cassandra-all</artifactId>
-            <optional>true</optional>
             <exclusions>
                 <!-- Use more recent version of jbcrypt from gremlin-groovy -->
                 <exclusion>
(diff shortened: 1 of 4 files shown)


janusgraph-cql/pom.xml
```diff
@@ -71,7 +71,6 @@
         <dependency>
             <groupId>org.apache.cassandra</groupId>
             <artifactId>cassandra-all</artifactId>
-            <optional>true</optional>
```
li-boxuan, line 74: This is needed otherwise Spark traversal does not work for CQL. I checked 0.5.3's distribution and saw cassandra-all.jar under lib folder. My guess is it was included by some other libraries as a transitive dependency.

porunov (approved):
LGTM. Thank you @li-boxuan . Checked that the provided test fails on `master` branch and works with Cassandra + ElasticSearch on your branch.

janusgraph-cql/pom.xml
li-boxuan: In v0.5, `cassandra-all` was introduced by `janusgraph-cassandra` module, which is removed from v0.6. Therefore we need to make this non-optional to make Spark traversal work.

li-boxuan:
I am having trouble connecting it to a Spark standalone cluster (running on my laptop). I never set up a Spark cluster before so I am not sure if this is my issue.

li-boxuan:
I am having trouble running Spark traversal on a pseudo yarn cluster. Most likely it is due to my misconfiguration since I had no experience setting up yarn cluster or Spark...

Since JanusGraph official documentation does not mention how to run Spark traversal on yarn, I am inclined to merge this despite not testing it properly.

porunov:
I'm good with it. Thank you @li-boxuan !

cdegroc:
Looks good to me too. Thanks @li-boxuan!

li-boxuan:
I found out a janusgraph dependency issue which might be related to my failure to run spark traversal on yarn cluster... still investigating and will update here.

PR #2757, opened 2021-08-14
Ensure CQLSession is closed properly even on CQLStoreManager construction exception

porunov:
This PR simply wraps the code after `CQLSession` is created in `try catch` and then trigger `close` method in case the exception was thrown later in the constructor.

From https://docs.datastax.com/en/developer/java-driver/4.13/manual/core/ it is noted that `CqlSession` must always be properly closed.

I found that CQLSession isn't properly closed if the exception is thrown during CQLStoreManager construction after CQLSession is initialized.
The next property helped to find the leak `advanced.session-leak.threshold` (you can read more about it here: https://docs.datastax.com/en/developer/java-driver/4.13/manual/core/configuration/reference/).

The test which will fail on master branch (similar one is included in this PR):
```
@Test
public void shouldProperlyCloseSessionOnException() throws NoSuchFieldException, IllegalAccessException {

    ModifiableConfiguration configuration = Mockito.spy(getBaseStorageConfiguration());

    configuration.set(METRICS_JMX_ENABLED, true);
    configuration.set(BASIC_METRICS, true);
    configuration.set(METRICS_SESSION_ENABLED, new String[]{"bytes-sent", "bytes-received", "connected-nodes"});

    MetricRegistry metricRegistrySpy = spyOnMetricsRegistry();

    verify(metricRegistrySpy, never()).remove(any());

    Mockito.doThrow(RuntimeException.class).when(configuration).get(BATCH_STATEMENT_SIZE);

    RuntimeException runtimeException = Assertions.assertThrows(RuntimeException.class, () ->
        new CQLStoreManager(configuration));
    
    verify(metricRegistrySpy, atLeastOnce()).remove(any());
}

private MetricRegistry spyOnMetricsRegistry() throws NoSuchFieldException, IllegalAccessException {
    MetricRegistry metricRegistrySpy = Mockito.spy(MetricManager.INSTANCE.getRegistry());
    Field registryField = MetricManager.INSTANCE.getClass().getDeclaredField("registry");
    registryField.setAccessible(true);
    registryField.set(MetricManager.INSTANCE, metricRegistrySpy);
    return metricRegistrySpy;
}
```

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
@@ -26,7 +26,6 @@ import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata;
 import com.datastax.oss.driver.internal.core.auth.PlainTextAuthProvider;
 import com.datastax.oss.driver.internal.core.ssl.DefaultSslEngineFactory;
 import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.vavr.Tuple;
 import io.vavr.collection.Array;
 import io.vavr.collection.HashMap;
@@ -40,24 +39,19 @@ import org.janusgraph.diskstorage.StaticBuffer;
 import org.janusgraph.diskstorage.StoreMetaData.Container;
 import org.janusgraph.diskstorage.common.DistributedStoreManager;
 import org.janusgraph.diskstorage.configuration.Configuration;
-import org.janusgraph.diskstorage.configuration.ExecutorServiceBuilder;
-import org.janusgraph.diskstorage.configuration.ExecutorServiceConfiguration;
-import org.janusgraph.diskstorage.cql.function.mutate.CQLExecutorServiceMutateManyLoggedFunction;
-import org.janusgraph.diskstorage.cql.function.mutate.CQLExecutorServiceMutateManyUnloggedFunction;
+import org.janusgraph.diskstorage.cql.builder.CQLMutateManyFunctionBuilder;
+import org.janusgraph.diskstorage.cql.builder.CQLMutateManyFunctionWrapper;
+import org.janusgraph.diskstorage.cql.builder.CQLStoreFeaturesBuilder;
+import org.janusgraph.diskstorage.cql.builder.CQLStoreFeaturesWrapper;
 import org.janusgraph.diskstorage.cql.function.mutate.CQLMutateManyFunction;
-import org.janusgraph.diskstorage.cql.function.mutate.CQLSimpleMutateManyLoggedFunction;
-import org.janusgraph.diskstorage.cql.function.mutate.CQLSimpleMutateManyUnloggedFunction;
-import org.janusgraph.diskstorage.configuration.ExecutorServiceInstrumentation;
 import org.janusgraph.diskstorage.keycolumnvalue.KCVMutation;
 import org.janusgraph.diskstorage.keycolumnvalue.KeyColumnValueStore;
 import org.janusgraph.diskstorage.keycolumnvalue.KeyColumnValueStoreManager;
 import org.janusgraph.diskstorage.keycolumnvalue.KeyRange;
-import org.janusgraph.diskstorage.keycolumnvalue.StandardStoreFeatures;
 import org.janusgraph.diskstorage.keycolumnvalue.StoreFeatures;
 import org.janusgraph.diskstorage.keycolumnvalue.StoreTransaction;
 import org.janusgraph.hadoop.CqlHadoopStoreManager;
 import org.janusgraph.util.stats.MetricManager;
-import org.janusgraph.util.system.NetworkUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -69,7 +63,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ThreadFactory;
 
 import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.truncate;
 import static com.datastax.oss.driver.api.querybuilder.SchemaBuilder.createKeyspace;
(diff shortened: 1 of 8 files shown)


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -179,93 +179,101 @@ public CQLStoreManager(final Configuration configuration) throws BackendExceptio
...
-            } else {
-                executeManyFunction = new CQLExecutorServiceMutateManyUnloggedFunction(batchSize,
-                    session, openStores, times, executorService, assignTimestamp, this::sleepAfterWrite);
-            }
-        } else {
-            this.executorService = null;
-            if(atomicBatch){
-                executeManyFunction = new CQLSimpleMutateManyLoggedFunction(times,
-                    assignTimestamp, openStores, session, this::sleepAfterWrite);
+            initializeJmxMetrics();
```
li-boxuan, line 185: This try block seems quite long. Do you consider moving the logic within try block to a separate method?

porunov: There are some `final` fields to which we assign value in constructor: `executorService`, `executeManyFunction`, `deployment`, `storeFeatures`. We can assign values to these fields in constructor only, thus we can't simply move this logic to a separate method. I can just refactor the logic to pass maybe builders to the `CQLStoreManager` constructor and than trigger those builders and assign the result values to these final fields. What do you think about it?

li-boxuan: That sounds great to me! Otherwise it's fine to keep it as it is.

li-boxuan:
Question: now that `advanced.session-leak.threshold` is available, can we make use of this option to test that sessions are not leaked?

porunov:
> Question: now that advanced.session-leak.threshold is available, can we make use of this option to test that sessions are not leaked?

Potentially yes. Probably, we could just check for logs and if we see any log with session leak warning - we fail the test. I will check how to intercept logs in unit tests later

porunov:
@li-boxuan updated the PR with your suggestions.

Added test `shouldLogSessionLeakWarning` which checks that `SESSION_LEAK_THRESHOLD` parameter works (should pass on master).

Added test `shouldProperlyCloseSessionOnExceptionAndNotLogSessionLeakWarnings` which checks that `CQLSession` is closed correctly even on exception (should fail on master).

Made very small refactoring. More refactoring will be needed but decided to do it in stages rather then everything in one PR.

li-boxuan (commented):
Hi @porunov I only have one small comment below, otherwise looks good to me!

janusgraph-cql/pom.xml
```diff
@@ -33,6 +33,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>uk.org.lidalia</groupId>
+            <artifactId>slf4j-test</artifactId>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
porunov: Do you mean `version` tag or some other tag. If `version` than this dependency is added to the root `pom.xml`  with the version set. I.e.
```
<dependency>
    <groupId>uk.org.lidalia</groupId>
    <artifactId>slf4j-test</artifactId>
    <version>1.2.0</version>
    <scope>test</scope>
</dependency>
```

You can check that there are other loggers added below without version set as well (the version is set in the root `pom`):
```
<dependency>
    <groupId>uk.org.lidalia</groupId>
    <artifactId>slf4j-test</artifactId>
    <scope>test</scope>
</dependency>

<!-- Logging backends. -->
<!-- Enforce a classpath ordering constraint to ensure log4j is used for logging. -->
<dependency>
    <groupId>org.slf4j</groupId>
    <artifactId>slf4j-log4j12</artifactId>
    <optional>true</optional>
</dependency>
<dependency>
    <groupId>ch.qos.logback</groupId>
</dependency>
``` 

Overall we shouldn't enforce usage of any loggers when deploying to Sonatype but I guess it was added because we also provide janusgraph distributions. I didn't think about it yet but probably later we will need to think of excluding loggers implementations from deployed jars to maven and include logger dependencies when building janusgraph distributions. 
In this situation I just added a `test` logger dependency which won't be available in distributions (only for tests).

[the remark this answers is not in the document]
li-boxuan: Yes, I meant to say version tag. And you are right, I somehow missed the change in the root pom.xml (I didn't scroll down to the bottom).

li-boxuan (approved):
LGTM 👍 verified one of the new tests fails on master

porunov:
I restarted CQL tests multiple times and they always run for very long time and the one test fails.
I have assumption, that by including `uk.org.lidalia` test logger we store all logs in memory without clearing them. So, probably, it causes some huge overhead and such problems.
Unfortunately, I could use appeasers  in Junit because it requires us to use logback logger but we use something else in our tests.
Probable solution could be resolving https://github.com/JanusGraph/janusgraph/discussions/2765 but I'm not sure yet how to do that.

li-boxuan:
@porunov Did you try the `clear()` method? I see it on https://projects.lidalia.org.uk/slf4j-test/ example

porunov:
> @porunov Did you try the clear() method? I see it on https://projects.lidalia.org.uk/slf4j-test/ example

I didn't know about these methods. That actually makes a lot of sense. I added the below clear methods. Hope it will help. 
```
@AfterEach
public void clearLoggers(){
    TestLoggerFactory.clear();
}

@AfterAll
public static void clearAllLoggers(){
    TestLoggerFactory.clearAll();
}
```

porunov:
@li-boxuan unfortunately neither `TestLoggerFactory.clear();` nor `TestLoggerFactory.clear();`  worke as expected. CQL tests still take a lot of time to be executed.
Thus, I updated the code with that reflection code to see if it works.
I checked that default loggers implementations didn't change compared to `master` branch. Thus, I hope it should work now unless I missed something.

porunov:
I was able to replace that logic with reflection only, so I don't see the reason of using additional dependency anymore. I removed that dependency because it causes some troubles and `clear()` doesn't work for some reason. Instead of that, I added `TestLoggerUtils` which should be sufficient to check logs. That class uses reflection to replace the logger of a specific class and then replace it back at the end of the test.
We may use this class in several other scenarios to check for expected logs.

porunov:
@li-boxuan I'm merging this PR but if you are not OK with the new way I check logs I'm totally fine reverting it or fixing your post merge review

PR #2764, opened 2021-08-18
Fix KCVSLog not correctly shutdowns after closure if storage is shutdowned shortly after closure

porunov:
In case we shutdown storage backend shortly after closure message pullers will not be closed and will be retried even when the thread is interrupted. It will log 'Could not read messages for timestamp [2021-08-18T20:29:59.057Z] (this read will be retried)'

Not sure yet how to test it properly

Signed-off-by: Oleksandr Porunov <alexandr.porunov@gmail.com>

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

PR #2769, opened 2021-08-19
Graceful shutdown

porunov:
This is a replacement for #2764 .
The main problem I noticed is that when JanusGraph looses connection to storage or index backend than the `close` method will not be graceful (not all underlying threads will be finished sometimes).

This PR checks Graceful shutdown for CQL connections. I didn't test it with other storage backends, so maybe additional implementation is required for other storage backends.

The added test `shouldGracefullyCloseGraphWhichLostAConnection` clearly shows there is threads leak in `master` branch during `close()` with lost connection.
To check this test in master branch you should remove the usage of two new added parameters at the beginning `EXECUTOR_SERVICE_MAX_SHUTDOWN_WAIT_TIME` and `PARALLEL_BACKEND_EXECUTOR_SERVICE_MAX_SHUTDOWN_WAIT_TIME`.

Also, you may add `sleep` after closing graph in hope that those leaked threads will be finished but it won't help. I tested it with on master branch with `Thread.sleep(130000L);` and those leaked threads didn't finish.

After this PR all threads are finished normally on shutdown even when connection is lost.

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.TimeUnit;
+
+public class ExecuteUtil {
+
+    public static void executeWithCatching(ExecuteExceptionallyFunction function, BackendExceptionWrapper backendExceptionWrapper){
+        try {
+            function.execute();
+        } catch (BackendException backendException){
+            if(backendExceptionWrapper.getBackendException() == null){
+                backendExceptionWrapper.setBackendException(backendException);
```
li-boxuan, line 31: So this backendExceptionWrapper only catches the first exception, right? Is this by intention? I thought you would collect all exception stack traces and print them out together.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/log/kcvs/KCVSLog.java
```diff
@@ -441,7 +455,7 @@ private KCVSMessage parseMessage(Entry msg) {
      * @param persistor
      * @return
      */
-    private Future<Message> add(StaticBuffer content, int partitionId, ExternalPersistor persistor) {
+    private synchronized Future<Message> add(StaticBuffer content, int partitionId, ExternalPersistor persistor) {
```
li-boxuan, line 458: Why synchronized? I suspect this might bring great overhead when WAL is enabled (i.e. SYSTEM_LOG_TRANSACTIONS == True).

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
porunov: I made it just for simplicity but I see that it may make sense to collect all the exceptions history. I will change that.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/log/kcvs/KCVSLog.java
porunov: Great observation! I noticed that all the other methods which access volatile variable `isOpen` are synchronized. This methods checks that as well, so, I thought it makes sense to make it synchronized. I need to double check if we need to access isOpen here or if it's not critical to execute this method after closure

li-boxuan: If `isOpen` needs to be synchronized, we probably should synchronize `verifyOpen` method only. This `add` method can be called by multiple threads concurrently so synchronizing the whole method could bring great overhead.

porunov: I don't think it makes much sense synchronizing verifyOpen method. The main idea of synchronizing is to acquire the lock, check condition, process logic, release the lock.
If we are synchronizing verifyOpen method the flow will be:
acquire the lock, check condition, release lock, process logic. Which means logic processing isn't protected.
Nevertheless, if this method can be executed in parallel then we definitely don't need to synchronize is with simple lock. I think I have an idea for this issue. I will change the logic here

porunov: It looks like it's not critical to have synchronization here. We will simply throw `JanusGraphException` instead of `ResourceUnavailableException` in case of concurrency happens here.
I.e. in case we check that `isOpen == true` we enter `add` method, after that another thread triggers `close()` method and there is a chance of storage being closed during that time (the same may happen even when we don't trigger `close` method. For example, storage backend connection is lost). In such a situation there is a chance of trying to send a message when we don't have a backend connection anymore which will result in `JanusGraphException`. Even if this threads retries to executor `add` method right after receiving `JanusGraphException` (for any reason) we will send `ResourceUnavailableException` on the second execution because `isOpen` variable is `volatile`. Thus, I think it's OK to not having any synchronization mechanism here (i.e. leave it as it is).
I will remove `synchronize` from this method but if anyone thinks we need some synchronization here, I'm happy to discuss another better synchronization solution for this use case (i.e. we could simple have atomic integer which counts amount of current enters to this method. If so, in `close()` method after changing `isOpen` to `false` we could simple await until that variable is equal to `0`).

li-boxuan: > I don't think it makes much sense synchronizing verifyOpen method

You are right, I somehow thought you wanted to control concurrent access to `if (!isOpen)` block in `verifyOpen` method. Now it's clear to me. I also agree that it might be a better idea to leave it as it is, considering the tradeoff between performance and consistency.

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public static void gracefulExecutorServiceShutdown(ExecutorService executorService, long maxWaitTime){
+        if(executorService == null){
+            return;
+        }
+        executorService.shutdown();
+        try{
+            executorService.awaitTermination(maxWaitTime, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException throwable){
+            // ignored
```
li-boxuan, line 59: I guess at least we can log it?

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
```diff
@@ -0,0 +1,65 @@
...
+            } else if(throwable instanceof RuntimeException){
+                throw (RuntimeException) throwable;
+            } else {
+                throw new UnhandledException(throwable);
+            }
+        }
+    }
+
+    public static void gracefulExecutorServiceShutdown(ExecutorService executorService, long maxWaitTime){
+        if(executorService == null){
```
li-boxuan, line 52: We can do
```suggestion
        if(executorService == null || !executorService.isShutdown()){
```
to make this function idempotent.

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
```diff
@@ -0,0 +1,65 @@
...
+            return;
+        }
+        executorService.shutdown();
+        try{
+            executorService.awaitTermination(maxWaitTime, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException throwable){
+            // ignored
+        }
+        if(!executorService.isTerminated()){
+            executorService.shutdownNow();
```
li-boxuan, line 62: nitpick: can we also log here?

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
```diff
@@ -230,12 +238,17 @@ TableMetadata getTableMetadata(final String name) throws BackendException {
...
-            this.session.close();
-        } finally {
-            if(this.executorService != null && !executorService.isShutdown()){
-                this.executorService.shutdownNow();
+            ExceptionWrapper exceptionWrapper = new ExceptionWrapper();
+            executeWithCatching(this::clearJmxMetrics, exceptionWrapper);
+            executeWithCatching(session::close, exceptionWrapper);
+            if(exceptionWrapper.getThrowable() != null){
+                if(exceptionWrapper.getThrowable() instanceof RuntimeException){
+                    throw (RuntimeException) exceptionWrapper.getThrowable();
```
li-boxuan, line 246: Why not using `throwIfException`?

li-boxuan (commented):
Looks roughly good to me! Just a few small suggestions.

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
porunov: added logging

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
porunov: added logging

janusgraph-core/src/main/java/org/janusgraph/util/system/ExecuteUtil.java
porunov: I think it should be `if(executorService == null || !executorService.isTerminated()())` because if it isn't terminated in specified amount of time, the underlying threads of executor service should be interrupted. Changed it

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java
porunov: Changed it

porunov: Diff: https://github.com/JanusGraph/janusgraph/compare/fa1ec0d02376b353e2755e51894adff269299540..c659d549f83389162ed7b9cfb82e845a085513ea

li-boxuan (approved):
LGTM 👍

PR #2783, opened 2021-09-02
Cleanup duplicate configurations and add datacenter options to cql configuration for better visibility

porunov:
Add CQL datacenter configuration just for better visibility.
Remove duplicate configurations from `conf/gremlin-server/*.properties` which present in `conf/*.properties`.
Use automatic configuration generation instead of custom comments for better consistency in `janusgraph-cql-configurationgraph.properties`.

-----

### For all changes:

### For code changes:


janusgraph-dist/src/assembly/static/conf/gremlin-server/janusgraph-cql-es-server.properties
```diff
@@ -1,161 +0,0 @@
...
-# storage backend.  After starting the database for the first time, this
-# file's copy of this setting is ignored.  Use JanusGraph's Management System
-# to read or modify this value after bootstrapping.
-index.search.elasticsearch.client-only=true
-
-# Or start ES inside the JanusGraph JVM
-#index.search.backend=elasticsearch
-#index.search.directory=db/es
-#index.search.elasticsearch.client-only=false
-#index.search.elasticsearch.local-mode=true
```
porunov, line 161: These properties are invalid and `client-only` mode isn't supported for ElasticSearch for quite some time. These properties will be just ignored. After removing them, this file will be identical to `conf/janusgraph-cql-es.properties`.
Thus, removing this file.

janusgraph-dist/src/assembly/static/conf/gremlin-server/janusgraph-berkeleyje-es-server.properties
```diff
@@ -1,26 +0,0 @@
-# Copyright 2019 JanusGraph Authors
```
porunov, line 1: Identical to `janusgraph-berkeleyje.properties`. Thus, removing this file.

janusgraph-dist/src/assembly/static/conf/gremlin-server/janusgraph-berkeleyje-server.properties
```diff
@@ -1,17 +0,0 @@
-# Copyright 2019 JanusGraph Authors
```
porunov, line 1: Identical to `janusgraph-berkeleyje-es.properties`. Thus, removing this file.

janusgraph-dist/src/assembly/static/conf/gremlin-server/janusgraph-berkeleyje-es-server.properties
FlorianHockmann: I think you mean `conf/janusgraph-berkeleyje-es.properties`. One small difference I see between the two files is that the other doesn't contain the `index.search.hostname`. But I'm not sure if that is really helpful.
In general, I'm wondering why we had these `.properties` files under `gremlin-server` any way. Aren't we always using `.properties` files directly from the conf dir?

FlorianHockmann (commented):
LGTM in general, I only added one comment where I'm not 100% sure.

janusgraph-dist/src/assembly/static/conf/gremlin-server/janusgraph-berkeleyje-es-server.properties
porunov: You are right, I confused `conf/janusgraph-berkeleyje-es.properties` with `conf/janusgraph-berkeleyje.properties`. 

> One small difference I see between the two files is that the other doesn't contain the index.search.hostname 

That's was fixed in the previous commit. If you check these files in `master` branch, they are using identical properties but for `index.search.hostname` we use `127.0.0.1` instead of `elasticsearch` because in many cases `elasticsearch` isn't a default hostname for elasticsearch but for `127.0.0.1` it's clear that it expects local ES installation:
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-dist/src/assembly/cfilter/conf/janusgraph-berkeleyje-es.properties
https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-dist/src/assembly/static/conf/gremlin-server/janusgraph-berkeleyje-es-server.properties
`index.search.hostname` isn't necessary to be used because it already default to `127.0.0.1` but I was thinking about it more like about documentation. I noticed that this property was used in `conf/gremlin-server/janusgraph-berkeleyje-es-server.properties` and thus decided to add it in the previous commit.

>  In general, I'm wondering why we had these .properties files under gremlin-server any way. Aren't we always using .properties files directly from the conf dir?

It's a good question but I wonder it myself. I can only guess that these files were initially used for `conf/gremlin-server/*.yaml` configurations but than become obsolete after some time and nobody removed them but I could be wrong. I couldn't find other usages of these files.

FlorianHockmann (approved):

farodin91:
@porunov You delete the server yaml which could lead to issue, if people don't want to use https://docs.janusgraph.org/operations/configured-graph-factory/. The normal server yamls used JanusGraphFactory instead fo ConfiguredGraphFactory which works differently.

porunov:
I didn't delete yaml files. I only deleted duplicate. properties files.

porunov:
Only janusgraph-dist/src/assembly/cfilter/conf/janusgraph-cql-configurationgraph.properties is using ConfiguredGraphFactory (as it should). All other configurations are using JanusGraphFactory.  
You can download the latest JanusGraph distribution 0.6.0 to check the generated configurations. All configurations are there. I removed duplicates only.

farodin91:
What do you think of renaming `#JANUSGRAPHCFG` to `#JANUSGRAPHHANDLE`? Otherwise it is a bit confusing for me.

porunov:
> What do you think of renaming #JANUSGRAPHCFG to #JANUSGRAPHHANDLE? Otherwise it is a bit confusing for me.

As I understand `CFG` means `configuration`. 
By `HANDLE` do you mean that this configuration property will be generated from real JanusGraph configuration?
I personally don't have preference between `CFG` or `HANDLE`. I'm OK with any.

farodin91:
I did read CGF which stand for ConfiguredGraphFactory.

PR #2795, opened 2021-09-12
Cleanup main branch

farodin91:
I removed a lot deprecated stuff.

-----

### For all changes:

### For code changes:


janusgraph-hadoop/src/test/resources/org/janusgraph/hadoop/formats/graphson/incremental-custom-cerberus-load.groovy
```diff
@@ -21,7 +24,7 @@ JanusGraphVertex getOrCreateVertex(faunusVertex, graph, context, log) {
     if (null == uniqueValue) {
         throw new RuntimeException("The provided Faunus vertex does not have a property for the unique key: " + faunusVertex)
     }
-    Iterator<Vertex> itty = graph.query().has(uniqueKey, uniqueValue).vertices().iterator()
+    Iterator<Vertex> itty = graph.query().has(uniqueKey, uniqueValue).vertiexStream().iterator()
```
rngcntr, line 27: Did you test this? I don't know the method call, but it looks to me like this should be `vertexStream()` not `vertiexStream()`.

farodin91:
@rngcntr Would you like to review?

docs/changelog.md
```diff
@@ -83,6 +83,39 @@ For more information on features and bug fixes in 1.0.0, see the GitHub mileston
 
 #### Upgrade Instructions
 
+##### Removal of deprecated classes/methods/functionalities
```
rngcntr, line 86: As these are all internal classes/methods, do you think it's worth mentioning all of them explicitly in the upgrade instructions?

janusgraph-dist/pom.xml
```diff
@@ -633,6 +520,7 @@
                             <configuration>
                                 <forkCount>1</forkCount>
                                 <reuseForks>false</reuseForks>
+                                <argLine>-Xmx1024m</argLine>
```
rngcntr, line 523: Why did you add this line and why didn't we need it previously?

farodin91:
>  * Why did you remove coverity?
 >    -  Was it already removed earlier and these are just leftovers?
 >    -  Is it not functional anymore?
 >    -  Or did you decide on your own that we don't want it anymore?

The last coverity run is 4 years ago. 
The CI code is removed in 2019.
I don't like that we link in our readme stuff that's not updated over 4 years. 
If we want to re-add this, we can find the code in the history.

> Was ist announced earlier that we discontinue the binary builds? I didn't even know that we had these.

* As far, i can remember, we never released binary builds, deb's or dnf's. 
* If we want to re-add this, we can find the code in the history. I don't like code that is not supported by anyone us or either tested. There's a really old ticket about packages with not much attention.

> I do not understand at all why you replaced vertices(), edges(), properties() and query() in some places, but not in others. I can still find over a thousand occurrences of these calls all across the repository.

I just replaced deprecated methods. They were deprecated multiple years ago.

docs/changelog.md
farodin91: It could be interesting for people that are using internal methods. :see_no_evil:

janusgraph-dist/pom.xml
farodin91: I will check

rngcntr:
> I just replaced deprecated methods. They were deprecated multiple years ago.
I think it's quite inconvenient to use `edgeStream()` for IndexTransactions, but `edges()` for JanusGraphTransactions. In addition, there are still references to the now removed methods in some of our markdown docs.

farodin91:
> I think it's quite inconvenient to use edgeStream() for IndexTransactions, but edges() for JanusGraphTransactions. 

Before, it was also inconvenient different return types: JanusGraphIndexQuery.edges() did return a Iterable<Result<JanusGraphEdge>> and Graph.edges() do return a Iterator<Edge>.

farodin91:
I will split up the PR into multiple small ones.

pom.xml
```diff
@@ -435,7 +435,7 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-failsafe-plugin</artifactId>
-                    <version>2.22.2</version>
+                    <version>3.0.0-M5</version>
```
FlorianHockmann, line 438: Why is this change part of this PR? Seems unrelated to me.

FlorianHockmann (approved):

PR #2812, opened 2021-09-22
CQL default request-timeout should be JanusGraph default

### LINKED ISSUE #2811: CQL request-timeout does not default to 10s
### Setup
- Version: 0.6.0
- Storage Backend: cql
- Mixed Index Backend: none

### Description
We run JanusGraph 0.5.x and are testing JanusGraph 0.6.0.
With JanusGraph 0.6.0, we've observed an increase in `INFO` logs such as
```
org.janusgraph.diskstorage.TemporaryBackendException: Temporary failure in storage backend
[...]
Caused by: com.datastax.oss.driver.api.core.DriverTimeoutException: Query timed out after PT2S
```

We don't override request-timeout anywhere so we were expecting JanusGraph 0.6.0 to use a 10s request-timeout by default (#2770, #2776).

After checking the code, I think the issue could be that `if`: https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/builder/CQLProgrammaticConfigurationLoaderBuilder.java#L89

If `REQUEST_TIMEOUT` is not set explicitly, then it won't be forwarded to CQL DriverConfigLoader.

### Workaround
- Set `REQUEST_TIMEOUT` explicitly

cdegroc:
CQLConfigOption storage.cql.request-timeout defaults to 10s.
But this default value is not forwarded to CQL DriverConfigLoader.

Fixes #2811

Signed-off-by: Clement de Groc <clement.degroc@datadoghq.com>

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLConfigTest.java
+++ b/janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLConfigTest.java
@@ -13,6 +13,8 @@
 
 package org.janusgraph.diskstorage.cql;
 
+import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
+import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
 import com.datastax.oss.driver.internal.core.tracker.RequestLogger;
 import org.apache.commons.io.FileUtils;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
@@ -21,9 +23,13 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.util.WithOptions;
 import org.janusgraph.JanusGraphCassandraContainer;
 import org.janusgraph.core.JanusGraphFactory;
 import org.janusgraph.core.schema.JanusGraphManagement;
+import org.janusgraph.diskstorage.configuration.BasicConfiguration;
 import org.janusgraph.diskstorage.configuration.ConfigElement;
+import org.janusgraph.diskstorage.configuration.Configuration;
 import org.janusgraph.diskstorage.configuration.ExecutorServiceBuilder;
+import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
 import org.janusgraph.diskstorage.configuration.WriteConfiguration;
+import org.janusgraph.diskstorage.cql.builder.CQLProgrammaticConfigurationLoaderBuilder;
 import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
 import org.janusgraph.graphdb.configuration.JanusGraphConstants;
 import org.janusgraph.graphdb.database.StandardJanusGraph;
@@ -44,6 +50,8 @@ import java.io.IOException;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.Collections;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -80,6 +88,7 @@ import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.CO
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.IDS_RENEW_TIMEOUT;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.INITIAL_STORAGE_VERSION;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.PARALLEL_BACKEND_EXECUTOR_SERVICE_MAX_SHUTDOWN_WAIT_TIME;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.ROOT_NS;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_HOSTS;
 import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.STORAGE_PORT;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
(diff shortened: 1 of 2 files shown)


janusgraph-cql/src/test/java/org/janusgraph/diskstorage/cql/CQLConfigTest.java
```diff
@@ -392,6 +401,18 @@ public void shouldBeAbleToSetupCustomRequestTimeout() {
...
+    public void defaultProgrammaticConfigurationShouldUseJanusGraphDefaultRequestTimeout() {
+        // Create a CQLProgrammaticConfigurationLoaderBuilder from default values
+        Configuration config = new ModifiableConfiguration(ROOT_NS, getConfiguration(), BasicConfiguration.Restriction.NONE);
+        CQLProgrammaticConfigurationLoaderBuilder builder = new CQLProgrammaticConfigurationLoaderBuilder();
+
+        // DriverConfigLoader should use CQLConfigOptions.REQUEST_TIMEOUT default value and not DataStax default value of 2s
+        DriverConfigLoader loader = builder.build(config, Collections.emptyList(), Duration.of(1, ChronoUnit.SECONDS));
+        Duration requestTimeout = loader.getInitialConfig().getDefaultProfile().getDuration(DefaultDriverOption.REQUEST_TIMEOUT);
+        assertEquals(Duration.of(REQUEST_TIMEOUT.getDefaultValue(), ChronoUnit.MILLIS), requestTimeout);
+    }
```
cdegroc, line 414: Fails without the change in CQLProgrammaticConfigurationLoaderBuilder.java

porunov (approved):
LGTM. Thank you @cdegroc !

li-boxuan (approved):
LGTM because I think we should use the default value if it is provided. Just one question for both of you: what was the default REQUEST_TIMEOUT we used in JanusGraph v0.5? I didn't find any place that we set up this configuration. If I am not missing anything, it means in 0.5.3 we were using the DataStax default value. Does this mean the latest CQL driver changed its default request-timeout from 10s to 2s?

porunov:
> LGTM because I think we should use the default value if it is provided. Just one question for both of you: what was the default REQUEST_TIMEOUT we used in JanusGraph v0.5? I didn't find any place that we set up this configuration. If I am not missing anything, it means in 0.5.3 we were using the DataStax default value. Does this mean the latest CQL driver changed its default request-timeout from 10s to 2s?

We were using 10 seconds in 0.5. That said, we were using a wrong parameter for that (connection_timeout instead of request_timeout). It was changed here: https://github.com/JanusGraph/janusgraph/commit/5a7c0ab3842f1a433060dd46a61288ac4d135198

li-boxuan:
@porunov I can it was introduced by this commit:
https://github.com/JanusGraph/janusgraph/blob/65cad01828f1a455cc51435f8efaf309111c0768/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLStoreManager.java#L250
which was not included in any v0.5 release. Thus I doubt what value was the default one in v0.5 version.

cdegroc:
Thanks for double checking @li-boxuan.

I agree: in v0.5.3, the `CONNECTION_TIMEOUT` is stored in `DistributedStoreManager.connectionTimeoutMS`: https://github.com/JanusGraph/janusgraph/blob/c2b49734b0a0917784286bfdc6e20dd7ba2aa254/janusgraph-core/src/main/java/org/janusgraph/diskstorage/common/DistributedStoreManager.java#L78

But then it doesn't seem used by the `CQLStoreManager`. And I found not connection/read timeout set at the session-level, nor at the statement-level.

So it also seems to me that we were using DataStax's driver default timeout in v0.5.

While surprising, it looks like the default value changed when the driver's configuration was reworked in 4.x:
- Cassandra 3.9.x default driver-side timeout was set to 12s in [SocketOptions.java](https://github.com/datastax/java-driver/blob/3.9.0/driver-core/src/main/java/com/datastax/driver/core/SocketOptions.java#L31-L40) (see this section in [DataStax's 3.9.x java-driver documentation](https://docs.datastax.com/en/developer/java-driver/3.9/manual/socket_options/#driver-read-timeout)).
- Cassandra 4.13.x default read timeout is set to 2s in the [reference configuration](https://docs.datastax.com/en/developer/java-driver/4.13/manual/core/configuration/reference/#reference-configuration) that is [loaded by default](https://docs.datastax.com/en/developer/java-driver/4.13/manual/core/configuration/reference/#reference-configuration).

If I can make some time, I'll try to confirm those values by running both versions locally.

cdegroc:
I've written a unit test (with `testcontainers` and [`toxiproxy`](https://www.testcontainers.org/modules/toxiproxy/)) to compare the behaviors of JanusGraph 0.5 and 0.6 when CQL requests time out.

The code (identical for 0.5 and 0.6) is here: [v0.5](https://github.com/cdegroc/janusgraph/commit/467b527215f15e6a848db56d87ae56950946ec43), [v0.6](https://github.com/cdegroc/janusgraph/commit/17948e2d78f2e73f9d40e2728803c3fd0a10fee9)

Setting `CONNECTION_TIMEOUT` to `5s` does not seem to change the behaviors of both versions:
- JanusGraph 0.5 times out after 12s by default.
- JanusGraph 0.6 times out after 2s by default.

Based on that and previous comments, I think we should:
- Merge this PR that ensures that the default request timeout is used if no value is provided by the user.
- Update `storage.connection-timeout` default value to `12s` (instead of 10s currently) so it matches JanusGraph 0.5.x.

If you think of additional tests to make, just let me know!

li-boxuan:
@cdegroc Do you want to update the default value in the same PR, or open up a new PR?
Btw is it possible to include your test in the codebase too? I think that would be helpful to catch regression issues.

cdegroc:
Should I target branch `v0.6` for this commit?

Regarding the test, I'm not 100% sold: it's testing a default timeout value in a dependency (DataStax CQL driver) and will consume time as it brings up a test container and wait for Xs until the requests time out. It sounded more like a 1-time test to make to me.
But either way is fine. If you really think it's worth having this test, I'll add it.

li-boxuan:
> Should I target branch v0.6 for this commit?

Yeah, that would be great! 

>  it's testing a default timeout value in a dependency (DataStax CQL driver) and will consume time as it brings up a test container and waits for Xs until the requests time out.

I was thinking about tagging it as a release test and only letting GitHub Actions execute it when required, but I think you are right, this particular one-time test was needed only because we had a bug that is solved by this PR.

li-boxuan (approved):

li-boxuan:
@porunov Do you want to review again?

PR #2838, opened 2021-10-22
docs.Dockerfile: Use python official image

li-boxuan:
The doc build has been failing in the past few days. See https://github.com/JanusGraph/janusgraph/runs/3921615242?check_suite_focus=true. This commit uses the official python docker image so that we don't need to worry about gcc dependencies.

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs.Dockerfile
+++ b/docs.Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.12
+FROM python:3.8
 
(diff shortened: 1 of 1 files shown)


li-boxuan:
I have no idea why https://github.com/JanusGraph/janusgraph/runs/3985377771?check_suite_focus=true is failing. From [here](https://github.com/JanusGraph/janusgraph/runs/3985377771?check_suite_focus=true#step:5:228) it seems the deploy-doc is still using the old script, but I don't see where any cache is used in the deploy-doc step.

PR #2860, opened 2021-11-21
Add config printer to janusgraph-server.sh

### LINKED ISSUE #2857: The jvm.options feature of janusgraph-server.sh is broken
Since janusgraph-0.6.0 the starting script, renamed to `bin/janusgraph-server.sh`, reads the `JAVA_OPTIONS` from the file `conf/jvm.options`. As [indicated by user m.leinweber@..,](https://lists.lfaidata.foundation/g/janusgraph-users/topic/jvm_options_broken/86896083) the reading of these `jvm.options` file is broken. Instead of reading all options, only the very last option is applied.

# Steps to reproduce
Copy `bin/janusgraph-server.sh` to `bin/janusgraph-server-tmp.sh` and add the following debug lines starting from line 110:
```
echo $JAVA_OPTIONS_FILE
echo $JAVA_OPTIONS
exit 1
```

# Workaround
Edit the `bin/janusgraph-server.sh` file manually and replace line 103 reading:  
```
  for opt in `grep "^-" $JAVA_OPTIONS_FILE`
```
with  
```
  for opt in "`grep '^-' $JAVA_OPTIONS_FILE`"
```
As often in failing bash scripts, quoting was done in an improper way.

farodin91:
Fixes #2857

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-dist/src/assembly/static/bin/janusgraph-server.sh
+++ b/janusgraph-dist/src/assembly/static/bin/janusgraph-server.sh
@@ -108,11 +108,17 @@ COLLECTED_JAVA_OPTIONS_FILE=""
 
 # Read user-defined JVM options from jvm.options file
 if [[ -z "$JAVA_OPTIONS_FILE" ]]; then
-  JAVA_OPTIONS_FILE="$JANUSGRAPH_CONF/jvm.options"
+  jver=$(java -version 2>&1 | grep 'version' 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]"."a[2]}')
+  if [[ $jver == "1.8" ]]; then                
+    JAVA_OPTIONS_FILE="$JANUSGRAPH_CONF/jvm-8.options"
+  else
+    JAVA_OPTIONS_FILE="$JANUSGRAPH_CONF/jvm-11.options"
+  fi
 fi
 if [[ -f "$JAVA_OPTIONS_FILE" ]]; then
-  for opt in `grep "^-" $JAVA_OPTIONS_FILE`
+  for opt in "$(grep '^-' $JAVA_OPTIONS_FILE)"
   do
+    opt=$(echo "$opt" | xargs)
     COLLECTED_JAVA_OPTIONS_FILE="$JAVA_OPTIONS_FILE_CURATED $opt"
   done
 fi
@@ -254,8 +260,17 @@ startForeground() {
   fi
 }
 
+printConfig() {
+    echo "JANUSGRAPH_HOME: ${JANUSGRAPH_HOME}"
+    echo "JANUSGRAPH_YAML: ${JANUSGRAPH_YAML}"
+    echo "JANUSGRAPH_SERVER_CMD: ${JANUSGRAPH_SERVER_CMD}"
+    echo "LOG_DIR: ${LOG_DIR}"
+    echo "PID_FILE: ${PID_FILE}"
+    echo "JAVA_OPTIONS: ${JAVA_OPTIONS}"
+}
+
 printUsage() {
-  echo "Usage: $0 {start [conf file]|stop|restart [conf file]|status|console|usage <group> <artifact> <version>|<conf file>}"
+  echo "Usage: $0 {start [conf file]|stop|restart [conf file]|status|console|config [conf file]|usage <group> <artifact> <version>|<conf file>}"
   echo
   echo "    start           Start the server in the background. Configuration file can be specified as a second argument
                     or as JANUSGRAPH_YAML environment variable. If configuration file is not specified
@@ -267,6 +282,7 @@ printUsage() {
   echo "    status          Check if the server is running"
   echo "    console         Start the server in the foreground. Same rules are applied for configurations as described
                     in \"start\" command"
+  echo "    config          Print out internal script variable to debug config"
   echo "    usage           Print out this help message"
   echo
   echo "In case command is not specified and the configuration is specified as the first argument, JanusGraph Server will
@@ -286,6 +302,10 @@ case "$1" in
   start)
     useGremlinServerConfiguration "$2"
     start
+    ;;
+  config)
+    useGremlinServerConfiguration "$2"
+    printConfig
     ;;
   stop)
     stop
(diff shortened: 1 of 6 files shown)


janusgraph-dist/src/assembly/static/bin/janusgraph-server.sh
```diff
@@ -254,8 +260,18 @@ startForeground() {
   fi
 }
 
+printConfig() {
+    echo "JANUSGRAPH_HOME: ${JANUSGRAPH_HOME}"
+    echo "JANUSGRAPH_YAML: ${JANUSGRAPH_YAML}"
+    echo "JANUSGRAPH_SERVER_CMD: ${JANUSGRAPH_SERVER_CMD}"
+    echo "LOG_DIR: ${LOG_DIR}"
+    echo "PID_FILE: ${PID_FILE}"
+    echo "JAVA_OPTIONS: ${JAVA_OPTIONS}"
```
FlorianHockmann, line 269: This should print the full `JAVA_OPTIONS` including everything from `jvm-11.options` / ``jvm-8.options`, right? I just build the distribution archive on my machine from this branch and then executed the script to get the output produced here, but the `JAVA_OPTIONS` don't look very complete:

```bash
$ bin/janusgraph-server.sh config
JANUSGRAPH_HOME: /mnt/c/Workspace/GitHub/janusgraph/janusgraph-dist/target/janusgraph-full-0.6.1-SNAPSHOT
JANUSGRAPH_YAML: /mnt/c/Workspace/GitHub/janusgraph/janusgraph-dist/target/janusgraph-full-0.6.1-SNAPSHOT/conf/gremlin-server/gremlin-server.yaml
JANUSGRAPH_SERVER_CMD: org.janusgraph.graphdb.server.JanusGraphServer
LOG_DIR: /mnt/c/Workspace/GitHub/janusgraph/janusgraph-dist/target/janusgraph-full-0.6.1-SNAPSHOT/logs
PID_FILE: /mnt/c/Workspace/GitHub/janusgraph/janusgraph-dist/target/janusgraph-full-0.6.1-SNAPSHOT/run/janusgraph.pid
JAVA_OPTIONS:  -XX:+CMSClassUnloadingEnabled  -javaagent:/mnt/c/Workspace/GitHub/janusgraph/janusgraph-dist/target/janusgraph-full-0.6.1-SNAPSHOT/lib/jamm-0.3.0.jar
COLLECTED_JAVA_OPTIONS_FILE:  -XX:+CMSClassUnloadingEnabled
```

This is still just the last line from the `OPTIONS` file `XX:+CMSClassUnloadingEnabled` together with the `javaagent`, isn't it?

janusgraph-dist/src/assembly/static/bin/janusgraph-server.sh
```diff
@@ -254,8 +260,18 @@ startForeground() {
...
+    echo "COLLECTED_JAVA_OPTIONS_FILE: ${COLLECTED_JAVA_OPTIONS_FILE}"
```
FlorianHockmann, line 270: Here, I would expect to get a file path as this ends on `FILE` but I'm actually getting its content:

```
COLLECTED_JAVA_OPTIONS_FILE:  -XX:+CMSClassUnloadingEnabled
```

is that supposed to be like that? If so, then I think that the name is confusing. I'm also not sure if we really want to include this output here. Isn't that only useful to debug the problem this is supposed to fix?

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/JanusGraphAssemblyBaseIT.java
```diff
@@ -128,6 +128,9 @@ protected static void unzip(String dir, String zipFile) throws IOException, Inte
 
     protected static void command(File dir, String... command) throws IOException, InterruptedException {
         ProcessBuilder pb = new ProcessBuilder(command);
+
+        Map<String, String> env = pb.environment();
```
FlorianHockmann, line 132: Why was this change necessary? Is that related to the changes in `janusgraph-server.sh`?

janusgraph-dist/src/assembly/static/conf/jvm-11.options
```diff
@@ -0,0 +1,77 @@
...
+#  GC SETTINGS  #
+#################
+
+### CMS Settings
+
+-XX:+UseG1GC
+-XX:+CMSParallelRemarkEnabled
+-XX:SurvivorRatio=8
+-XX:MaxTenuringThreshold=1
+-XX:CMSInitiatingOccupancyFraction=75
```
FlorianHockmann, line 72: Aren't all these `CMS[...]` values config options of the `ConcMarkSweep` garbage collector that we had configured for Java 8, but not that we use G1 instead they shouldn't have any effect, right?

FlorianHockmann (changes requested):

janusgraph-dist/src/test/java/org/janusgraph/pkgtest/JanusGraphAssemblyBaseIT.java
farodin91: Yeah, it now reads the value from options. It would set the memory to 4G and github actions doesn't have enough memory/

janusgraph-dist/src/assembly/static/bin/janusgraph-server.sh
farodin91: I will check again. Could be my last commit to fix codacy issues.

janusgraph-dist/src/assembly/static/bin/janusgraph-server.sh
farodin91: Will remove this log as they are already include in JAVA_OPTIONS

janusgraph-dist/src/assembly/static/conf/jvm-11.options
farodin91: lines are removed

farodin91:
@FlorianHockmann Would you to review it again?

FlorianHockmann (approved):
Just checked again and it all looks good now.

farodin91:
@FlorianHockmann I will change the base branch to 0.6.

PR #2876, opened 2021-12-01
Implement our own GraphBinary Geoshape serialization

### LINKED ISSUE #2856: Add proper GraphBinary serialization for Geoshapes
**Describe the feature:**
JanusGraph [got basic support for GraphBinary in version 0.6.0](https://docs.janusgraph.org/changelog/#graphbinary-is-now-supported). Unfortunately, it still uses a custom Java based binary serialization format for Geoshapes that is implemented in the Java library `Spatial4j`. This makes it very cumbersome to add support for these GraphBinary types to other languages, like .NET. This serialization is also very bloated. It produces for example a GraphBinary value consisting of 73 bytes for a single `Geoshape.Point` instance that only contains two double values (so the serialized value should only consist of 16 bytes).
This defies the purpose of GraphBinary which is supposed to be language agnostic and result in minimal payloads.

I suggest that we implement GraphBinary serialization of Geoshapes completely by ourselves to stay close to the GraphBinary standard and to make it easy to implement support for GraphBinary in other languages.
While we're doing that, we can also improve the serialization by adding a dedicated GraphBinary type for each Geoshape type (Point, Circle, ...) instead of using just a single Geoshape type for all of them. This would also simplify the serialization.

This will of course be a breaking change but we can probably at least keep support for deserialization of Geoshapes coming from 0.6 drivers in the server.

Being able to support GraphBinary for JanusGraph in other languages.

FlorianHockmann:
This makes it much easier to support GraphBinary outside of Java. It also reduces the payload size by a factor of 3 to 4 (taken from the examples used in the tests):

| Geoshape type      | Payload size (B) with previous serialization | Payload size (in B) with new serialization | Relation (new / old) |
|--------------------|----------------------------------------------|--------------------------------------------|----------------------|
| Point              | 73                                           | 22                                         | 30.1%                |
| Circle             | 105                                          | 30                                         | 28.6%                |
| Polygon            | 373                                          | 90                                         | 24.1%                |
| GeometryCollection | 633                                          | 164                                        | 25.9%                |

It is unfortunately not possible to add the Geoshape types also directly as GraphBinary types which would simplify the serialization and also reduce the payload size even more as we wouldn't have to serialize the Geoshape GraphBinary type together with the Geoshape type code. This however doesn't work with how Geoshapes are implemented right now in
Java as they are all represented by a single class `Geoshape`. We would need to split this class into different classes for the Geoshape types which seems out of scope for now as we just want to improve the GraphBinary serialization here.

Fixes #2856

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -86,6 +86,21 @@ For more information on features and bug fixes in 1.0.0, see the GitHub mileston
 
 We are dropping support for HBase 1.
 
+##### Breaking change for Geoshape GraphBinary serialization
+
+Support for the [GraphBinary](http://tinkerpop.apache.org/docs/3.5.1/dev/io/#graphbinary) serialization format was
+added in JanusGraph 0.6.0. This also included support to serialize Geoshapes via GraphBinary. The implementation of the
+Geoshape serializer was unfortunately closely tied to the Java library `Spatial4j` that we are using to implement
+Geoshapes in Java. This made it very complicated to add support for GraphBinary in other languages than Java. To make
+it easier to support GraphBinary in non-Java environments like .NET, we have completely reimplemented the GraphBinary
+serialization of Geoshapes in this version.
+
+This is a breaking change for users who have already adopted GraphBinary and who are using Geoshapes. It is necessary
+to update JanusGraph Server and all (Java) clients that use GraphBinary at the same time since JanusGraph Server with
+an older version will not be able to read a Geoshape created by a client that is already on version 1.0.0 and vice
+versa.
+Users who do not use GraphBinary yet or who are not using Geoshapes are not affected by this change.
+
 ##### Removal of deprecated classes/methods/functionalities
 
 ###### Classes/Interfaces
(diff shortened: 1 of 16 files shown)


farodin91 (approved):
LGTM! Thank you @FlorianHockmann. 

I discuss with @FlorianHockmann offline about the breaking change. As we just started to support GraphBinary, it should a big issue to introduce a breaking change. 

It could be a good idea to add a hint for 0.6 in the docs that we will introduce a breaking change with 1.0. So people are aware that GraphBinary isn't stable yet.

FlorianHockmann:
> As we just started to support GraphBinary, it should a big issue to introduce a breaking change.

I assume you mean "_not_ a big issue" 😅 

> It could be a good idea to add a hint for 0.6 in the docs that we will introduce a breaking change with 1.0. So people are aware that GraphBinary isn't stable yet.

Good idea, I added a small note there.

farodin91:
Not a big issue is correct. 🙊

farodin91:
We have to cherry pick the note, if this got merged.

FlorianHockmann:
> We have to cherry pick the note, if this got merged.

The problem is that we can only do that after 1.0.0 is released as we would otherwise refer and link to a release that isn't out yet. So, I'm not sure how we should approach this. Maybe we should just not merge it to the 0.6 branch at all and just keep it on master?

Issue #1236, opened 2018-08-30
Exclude log4j.properties in pom.xml


huynhtk:
The janusgraph-core-0.3.0.jar includes log4j.properties which is overriding the existing application's log4j.properties file.

pluradj:
It appears that the `log4j.properties` has been packaged in janusgraph-core.jar going all the way back to titan-core.jar. It is probably safe to remove from the jar file.

@huynhtk According to the log4j docs on the [default initialization procedure](https://logging.apache.org/log4j/1.2/manual.html#Default_Initialization_Procedure), the preferred way to specify the default initialization file is through the `log4j.configuration` system property. For example, the Gremlin Console uses this flag `-Dlog4j.configuration=conf/log4j-console.properties` to pull in the properties from the local config file rather than from the jar file.

TotalForgot:
Dear,

It also appears in janusgraph-core-0.3.1.jar
I totally agree that it would be better to remove from the jar file.

Best
TotalForgot

DatPhanTien:
Dear 

It is also included in janusgraph-0.4.0.jar

And it is crazy that our java projects after declaring dependency on JanusGraph cannot logout properly without overriding another log4j.properties file.

We believe that this file should be removed from the jar, and one will need to provide in extra the proper log4j.properties as his/her expectation when starting the application.

Including this file in jar brings abnormal behaviors, and this is an anti-pattern.

Best
Tien Dat

PR #2890, opened 2021-12-13
Upgrade log4j to 2.17

farodin91:
-----

### For all changes:

### For code changes:


farodin91:
I want to see failing builds.

janusgraph-dist/src/assembly/static/conf/logback.xml
```diff
@@ -1,12 +0,0 @@
-<configuration>
```
li-boxuan, line 1: Seems `* src/assembly/static/cassandra/conf/logback.xml` from `janusgraph-dist/README.md` needs to be removed.

In addition, I wonder why we remove this file.

janusgraph-core/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
li-boxuan, line 1: Does this file still cause the same problem as reported in https://github.com/JanusGraph/janusgraph/issues/1236? It seems to me that instead of distributing `log4j.properties`, we are now distributing `log4j2.xml`.

To be honest, I don't understand why we have `log4j` configuration in the first place. I thought libraries should only use `slf4j`, and let applications choose their logging backends and provide configurations. I might be totally wrong though.

li-boxuan (commented):
I did a quick pass and had a few questions. Fundamentally, I don't quite understand why we are using `log4j` rather than using `slf4j` only and giving users the flexibility to choose their logging backend.

janusgraph-dist/src/assembly/static/conf/logback.xml
farodin91: No this is required for Cassandra distribution.

janusgraph-core/src/main/resources/log4j2.xml
farodin91: I will remove this. 

I will check where i can remove log4j and where not. 

If you are using janusgraph as library, it would good to let help choose their own logging solution. If you are using janusgraph standalone, it would inconvenient. You would have to setup your own logging solution. 

For tests, it is helpful to be able to get logs. 

I know cql is depend on Cassandra-all which includes logback. Here I'm unable to remove the dependency from the pom.

farodin91:
We are using slf4j for logging in the default case. Just the _printing_ is done by log4j.

janusgraph-core/src/main/resources/log4j2.xml
li-boxuan: > If you are using janusgraph standalone, it would inconvenient. You would have to setup your own logging solution.

I am not familiar with the logging libraries and their mechanisms, so my question might be dumb. If users use JanusGraph server (so-called janusgraph standalone), is it sufficient to provide `log4j` configurations and dependencies only in `janusgraph-server/pom.xml`?

janusgraph-dist/src/assembly/static/conf/logback.xml
li-boxuan: Ya, my bad. Btw do we remove this file because we are not using `logback` anyways?

farodin91: We are replacing it.

janusgraph-core/src/main/resources/log4j2.xml
farodin91: I have to try.

farodin91:
@li-boxuan I think I just have to remove the log4j.XML in the core pom. Log4j and Logback are runtime scoped which prevent a direct dependency to log4j. I will go though all pom and check that logback and log4j are scoped runtime. We should prevent log4j that log4j is required dependency.

farodin91:
@li-boxuan Would you like to check again?

li-boxuan (commented):
Looks good to me except for a few minor doubts as follows. In addition, do you want to include https://github.com/JanusGraph/janusgraph/issues/1236 in your commit message?

li-boxuan:
Can you rebase this PR given that https://github.com/JanusGraph/janusgraph/pull/2901 is merged (which uses old log4j config)?

farodin91:
@li-boxuan I updated the log4j, fixed your comments and rebased.

janusgraph-dist/src/assembly/static/conf/log4j2-console.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="warn">
+    <Appenders>
+        <Console name="STDOUT" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss} %-5level %class.%method{36} - %msg%n"/>
+        </Console>
+    </Appenders>
+    <Loggers>
+        <Root level="info">
```
li-boxuan, line 9: The old file `janusgraph-dist/src/assembly/static/conf/log4j-console.properties` has more configurations (e.g. `log4j.logger.org.apache.hadoop.conf.Configuration.deprecation=OFF`). Do we not need those anymore?

janusgraph-benchmark/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="warn">
+    <Appenders>
+        <Console name="STDOUT" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss} %-5level %class.%method{36} - %msg%n"/>
+        </Console>
+    </Appenders>
+    <Loggers>
+        <Root level="info">
```
li-boxuan, line 9: The logging level needs to be at least the warning level, otherwise, there will be [too many logs](https://github.com/JanusGraph/janusgraph/runs/4703815873?check_suite_focus=true#step:6:933).

janusgraph-berkeleyje/src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,15 @@
...
+<Configuration status="warn">
+    <Appenders>
+        <File name="TestFile" fileName="target/test.log">
+            <PatternLayout>
+                <Pattern>%d{HH:mm:ss} %-5level %class.%method{36} - %msg%n</Pattern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Root level="info">
```
li-boxuan, line 11: The old logging level was `error`.

janusgraph-dist/src/assembly/static/conf/log4j2-console.xml
farodin91: log4j.logger.org.apache.hadoop.conf.Configuration.deprecation=OFF is any longer needed.

farodin91: For the other log loglevel changes, we have to see behavior in production what levels would be great. I don't really love to just copy values without knowing why. JanusGraph has a lot of stuff that is not really maintained at moment.

li-boxuan: My only concern is, this might change the behavior in the production, in which case we should make it clear in the changelog. However, I personally don't understand everything in the old log configuration, so I am not confident to review these configuration changes. Hope someone else could help with the review.

farodin91: I will add a note to changelog.

farodin91: Done

farodin91:
@li-boxuan  Would you like to review it again?

docs/changelog.md
```diff
@@ -136,6 +136,13 @@ Users who do not use GraphBinary yet or who are not using Geoshapes are not affe
 
 We are dropping support for old serialization format of JanusGraph predicates. The old predicates serialization format is only used by client older than 0.6. 
 The change only affects Gryo and GraphSON.
+##### Upgrade of log4j to version 2
+
+This change requires a new log4j configuration. You can find an example configuration in `conf/log4j2-server.xml`. As fact of the changed configuration format, 
+we cleanup all configurations. This could lead to unexpect new log lines. Please open an issue, if you have see any unwanted log line.
```
li-boxuan, line 142:
```suggestion
we clean up all configurations. This could lead to unexpected new log lines. Please open an issue, if you see any unwanted log line.
```

docs/changelog.md
```diff
@@ -136,6 +136,13 @@ Users who do not use GraphBinary yet or who are not using Geoshapes are not affe
 
 We are dropping support for old serialization format of JanusGraph predicates. The old predicates serialization format is only used by client older than 0.6. 
 The change only affects Gryo and GraphSON.
+##### Upgrade of log4j to version 2
+
+This change requires a new log4j configuration. You can find an example configuration in `conf/log4j2-server.xml`. As fact of the changed configuration format,
```
li-boxuan, line 141:
```suggestion
This change requires a new log4j configuration. You can find an example configuration in `conf/log4j2-server.xml`. As a result of the changed configuration format, 
```

docs/changelog.md
```diff
@@ -136,6 +136,13 @@ Users who do not use GraphBinary yet or who are not using Geoshapes are not affe
 
 We are dropping support for old serialization format of JanusGraph predicates. The old predicates serialization format is only used by client older than 0.6. 
 The change only affects Gryo and GraphSON.
+##### Upgrade of log4j to version 2
+
+This change requires a new log4j configuration. You can find an example configuration in `conf/log4j2-server.xml`. As fact of the changed configuration format, 
+we cleanup all configurations. This could lead to unexpect new log lines. Please open an issue, if you have see any unwanted log line.
+
+!!! note 
+    Log4j is just used for standalone server deployments and JanusGraph testing.
```
li-boxuan, line 145:
```suggestion
    Log4j is only used for standalone server deployments and JanusGraph testing.
```

li-boxuan (approved):
LGTM, thank you for the upgrade and cleanup! I just have a few nitpicks on the wording in the changelog:

farodin91:
@li-boxuan  changelog is changed

docs/changelog.md
```diff
@@ -137,6 +137,14 @@ Users who do not use GraphBinary yet or who are not using Geoshapes are not affe
 We are dropping support for old serialization format of JanusGraph predicates. The old predicates serialization format is only used by client older than 0.6. 
 The change only affects Gryo and GraphSON.
 
+##### Upgrade of log4j to version 2
+
+This change requires a new log4j configuration. You can find an example configuration in `conf/log4j2-server.xml`. As a result of the changed configuration format, 
+we clean up all configurations. This could lead to unexpected new log lines. Please open an issue, if you have see any unwanted log line.
```
li-boxuan, line 143: nitpick: if you have see -> if you see

li-boxuan (approved):
LGTM! Just a small typo, not necessary to fix.

PR #2899, opened 2021-12-24
 Fix NullPointerException caused by transaction release

### LINKED ISSUE #2898: StandardJanusGraphTx::expireSchemaElement throws NPE due to race condition
- Version: 0.6.0, master
- Storage Backend: doesn't matter
- Mixed Index Backend: doesn't matter
- Current Behavior: No NPE should be thrown

I was running [this benchmark](https://gist.github.com/li-boxuan/98d4dc18215e28b239c7433c8c39fb3f) and encountered NPE on v0.6 and master branches (provided that I changed concurrency in https://github.com/JanusGraph/janusgraph/blob/e64facbcdefd9d63ede4bedfbf15b78dc56d4751/janusgraph-core/src/main/java/org/janusgraph/diskstorage/Backend.java#L430-L435 from 1 to 8).

The exception trace is:

```
java.lang.NullPointerException
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.expireSchemaElement(StandardJanusGraphTx.java:1593)
	at org.janusgraph.graphdb.database.management.ManagementLogger.read(ManagementLogger.java:97)
	at org.janusgraph.diskstorage.log.util.ProcessMessageJob.run(ProcessMessageJob.java:46)
```

Apparently, `vertexCache` is null because the transaction has already been closed.

This bug was uncovered by 0.6.0 (08fd019) because now the internal resources are cleared after the transaction gets closed.

The cause is a race condition. Right after `graph.getOpenTransactions()` is invoked in ManagementLogger, the transaction could be released (i.e. `releaseTransaction()` being called) concurrently. Then, when ManagementLogger calls `expireSchemaElement` method for that transaction, NPE occurs.

This is similar to the bug discussed in https://lists.lfaidata.foundation/g/janusgraph-users/topic/potential_transaction_issue/85970858, where NPEs occur because some transaction internal structures are accessed even if the transaction has already been closed. Nonetheless, it is not yet clear what is the exact reason for this NPE.

li-boxuan:
Fixes JanusGraph#2898. Due to a race condition, vertexCache could be cleared before
it is accessed. To avoid NPEs, we cannot simply assign null to vertexCache
to release the memory.

This commit also surpresses https://lists.lfaidata.foundation/g/janusgraph-users/topic/potential_transaction_issue/85970858
by using a dummy empty vertexCache after transaction release. The root cause
for this bug is unknown, but likely it is due to another race condition.
To keep the behavior consistent with v0.5 and before, this commit surpresses
this issue and writes a warning log whenever the issue happens.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -95,8 +95,10 @@ import org.janusgraph.graphdb.relations.StandardVertexProperty;
 import org.janusgraph.graphdb.tinkerpop.JanusGraphBlueprintsTransaction;
 import org.janusgraph.graphdb.transaction.addedrelations.AddedRelationsContainer;
 import org.janusgraph.graphdb.transaction.addedrelations.ConcurrentAddedRelations;
+import org.janusgraph.graphdb.transaction.addedrelations.EmptyAddedRelations;
 import org.janusgraph.graphdb.transaction.addedrelations.SimpleAddedRelations;
 import org.janusgraph.graphdb.transaction.indexcache.ConcurrentIndexCache;
+import org.janusgraph.graphdb.transaction.indexcache.EmptyIndexCache;
 import org.janusgraph.graphdb.transaction.indexcache.IndexCache;
 import org.janusgraph.graphdb.transaction.indexcache.SimpleIndexCache;
 import org.janusgraph.graphdb.transaction.lock.CombinerLock;
@@ -105,8 +107,10 @@ import org.janusgraph.graphdb.transaction.lock.IndexLockTuple;
 import org.janusgraph.graphdb.transaction.lock.LockTuple;
 import org.janusgraph.graphdb.transaction.lock.ReentrantTransactionLock;
 import org.janusgraph.graphdb.transaction.lock.TransactionLock;
+import org.janusgraph.graphdb.transaction.subquerycache.EmptySubqueryCache;
 import org.janusgraph.graphdb.transaction.subquerycache.GuavaSubqueryCache;
 import org.janusgraph.graphdb.transaction.subquerycache.SubqueryCache;
+import org.janusgraph.graphdb.transaction.vertexcache.EmptyVertexCache;
 import org.janusgraph.graphdb.transaction.vertexcache.GuavaVertexCache;
 import org.janusgraph.graphdb.transaction.vertexcache.VertexCache;
 import org.janusgraph.graphdb.types.CompositeIndexType;
@@ -229,7 +233,7 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
      * Transaction-local data structure for unique lock applications so that conflicting applications can be discovered
      * at the transactional level.
      */
-    private volatile ConcurrentMap<LockTuple, TransactionLock> uniqueLocks;
+    private volatile Map<LockTuple, TransactionLock> uniqueLocks;
 
     //####### Other Data structures
     /**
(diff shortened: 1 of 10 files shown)


farodin91 (approved):
LGTM! Thank you @li-boxuan.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -504,6 +504,10 @@ private InternalVertex getExistingVertex(long vertexId) {
     }
 
     public InternalVertex getInternalVertex(long vertexId) {
+        if (vertexCache == null) {
```
rngcntr, line 507: The class `StandardJanusGraphTx` already has a `verifyOpen()` method, which checks `isClosed()` and throws an exception if the transaction is closed. I quickly skimmed the `StandardJanusGraphTx` class and it looks to me like `isClosed()` is always equal to `vertexCache == null`, `addedRelations == null` and `deletedRelations == null`, but it's way more readable.

I suggest to add a `warnIfClosed()` method and call it here and also in `getExistingVertex()`, `getInternalVertices()`, `connectRelation()`, `getSchemaVertex()` and `expireSchemaElement()`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -1582,11 +1586,12 @@ public final boolean isClosed() {
 
     @Override
     public boolean hasModifications() {
-        return !addedRelations.isEmpty() || !deletedRelations.isEmpty();
+        return addedRelations != null && deletedRelations != null && (!addedRelations.isEmpty() || !deletedRelations.isEmpty());
```
rngcntr, line 1589:
```suggestion
        return isOpen() && (!addedRelations.isEmpty() || !deletedRelations.isEmpty());
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
rngcntr: I don't know exactly how transactions are handled between threads, but if multiple threads can access the same transaction, we need to synchronize state-changing calls to the same transaction.

li-boxuan: > if multiple threads can access the same transaction, we need to synchronize state-changing calls to the same transaction

TBH I am not sure if race condition is the cause for the NPE in `getInternalVertex`.

rngcntr: Me neither and it's hard to debug as well

li-boxuan: I now assign a dummy empty cache to the variable `vertexCache` in transaction release method.

rngcntr (approved):

li-boxuan:
@rngcntr That's great! Did you also find the root cause for https://lists.lfaidata.foundation/g/janusgraph-users/topic/potential_transaction_issue/85970858? In other words, how does it happen that users accidentally use a closed transaction?

rngcntr:
@li-boxuan The thread you mention sounds very similar to the problem we were facing prior to the fix. The issue was that the TinkerPop framework called another `hasNext()` on each traversal after its transaction has been committed. In most cases, this is not a problem for us, but sometimes - especially when using `limit()` steps - we observe race conditions.

li-boxuan:
@rngcntr Thanks, what you said is very likely to be true - since both users reported that they were using the TinkerPop server.

PR #2901, opened 2021-12-24
Add concurrency level config in ManagementSystem::updateIndex

li-boxuan:
ManagementSystem uses an local executor service to run reindex and index removal
processes, available via ManagementSystem::updateIndex method. Unfortunately, it
only uses one thread which is inefficient. This commit changes the default concurrency
level from one to the number of available processors, which is fetched in runtime.
This commit also adds a new API for updateIndex which allows users to configure the
concurrency level, i.e. number of threads to run the job concurrently.

This commit comes along with a benchmark program. Benchmarking result on a 6-core
MacBook Pro (2019) is as follows:
```
Benchmark(multi-threaded)           (size)  Mode  Cnt    Score    Error  Units
MgmtOlapJobBenchmark.runReindex       10000  avgt    3  174.800 ± 74.395  ms/op
MgmtOlapJobBenchmark.runRemoveIndex   10000  avgt    3  118.072 ±  4.144  ms/op
MgmtOlapJobBenchmark.runReindex      100000  avgt    3  488.368 ± 134.655  ms/op
MgmtOlapJobBenchmark.runRemoveIndex  100000  avgt    3  148.176 ±  31.914  ms/op


Benchmark(single-threaded)           (size)  Mode  Cnt    Score     Error  Units
MgmtOlapJobBenchmark.runReindex       10000  avgt    3  341.851 ± 266.476  ms/op
MgmtOlapJobBenchmark.runRemoveIndex   10000  avgt    3  116.425 ±   5.799  ms/op
MgmtOlapJobBenchmark.runReindex      100000  avgt    3  1872.342 ± 567.239  ms/op
MgmtOlapJobBenchmark.runRemoveIndex  100000  avgt    3   143.134 ±  49.882  ms/op
```
The benchmark result shows a significant improvement with multi-threading in reindex
job. Index removal job does not see performance difference, likely because the test
scope is not large enough.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-benchmark/src/main/java/org/janusgraph/MgmtOlapJobBenchmark.java
+++ b/janusgraph-benchmark/src/main/java/org/janusgraph/MgmtOlapJobBenchmark.java
@@ -1,0 +1,127 @@
+// Copyright 2021 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph;
+
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.janusgraph.core.Cardinality;
+import org.janusgraph.core.JanusGraph;
+import org.janusgraph.core.JanusGraphFactory;
+import org.janusgraph.core.PropertyKey;
+import org.janusgraph.core.schema.JanusGraphManagement;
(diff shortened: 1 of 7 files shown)


janusgraph-benchmark/src/main/java/org/janusgraph/MgmtOlapJobBenchmark.java
```diff
@@ -0,0 +1,127 @@
...
+        mgmt.commit();
+    }
+
+
+    @TearDown(Level.Iteration)
+    public void tearDown() {
+        graph.close();
+    }
+
+    public static void main(String[] args) throws RunnerException {
```
farodin91, line 118: Do we need the main?

farodin91 (approved):
Thank you @li-boxuan.  LGTM!

janusgraph-benchmark/src/main/java/org/janusgraph/MgmtOlapJobBenchmark.java
li-boxuan: Not needed for CI, but it would be handy to run this particular benchmark locally with the main function.

PR #2913, opened 2022-01-04
Add Support matrix to readme

farodin91:
* Remove unsupported version from hbase

-----

### For all changes:


FlorianHockmann:
Do we really want to have this matrix in two places? I think that will at some point result in us forgetting to update it in both places which will be very confusing then. So, if we think that it should be prominently shown in the README, then I'd remove it from the changelog.
But do we really need this in the README? I think that a short README that only contains the most important information about the project with pointers to other resources with more information is preferable over a longer README.

PR #2929, opened 2022-01-08
Second run to improve java 11

farodin91:
I disabled solr 7 for java 11. 

-----

### For all changes:

### For code changes:


farodin91:
@FlorianHockmann I've updated the PR.

.github/workflows/ci-backend-cql.yml
```diff
@@ -98,18 +98,18 @@ jobs:
             java: 8
           - module: cql
             install-args: "-Pjava-11"
```
FlorianHockmann, line 100: This file contains 12 test entries for Java 8, but only 3 for Java 11. Shouldn't all tests be also executed with Java 11?

docs/changelog.md
```diff
@@ -149,6 +149,11 @@ we clean up all configurations. This could lead to unexpected new log lines. Ple
 !!! note 
     Log4j is only used for standalone server deployments and JanusGraph testing.
 
+##### Add support for Java 11
+
+All tests are no running on Java 11. Java 11 won't yet anymore the full build distribution.
```
FlorianHockmann, line 154: (nitpick) Should be _are no**w** running on Java 11_.

But I don't understand the second sentence at all. The full build distribution is removed now?

janusgraph-server/pom.xml
```diff
@@ -88,4 +88,12 @@
             </exclusions>
         </dependency>
     </dependencies>
+    <build>
+        <resources>
+            <resource>
+                <directory>${basedir}/src/main/resources</directory>
```
FlorianHockmann, line 94: I don't quite understand the purpose of this addition. I don't see any resources here under `src/main/resources`.

docs/changelog.md
```diff
@@ -149,6 +149,14 @@ we clean up all configurations. This could lead to unexpected new log lines. Ple
 !!! note 
     Log4j is only used for standalone server deployments and JanusGraph testing.
 
+##### Add support for Java 11
+
+All tests are now running on Java 11. We encourage everyone to test out Java 11.
+
+!!! note
+    Currently, we aren't able to release full distribution build with Java 11.
```
FlorianHockmann, line 157: I'm not sure what this means to be honest. We still build both distribution zip archives with Java 8, right? Does the full distribution then also require Java 8 as its runtime?
Or is this just informing users that we currently cannot build the full distribution with Java 11, although we also don't build the normal distribution with Java 8?

farodin91: I added an extra line

farodin91:
@FlorianHockmann  Would you like to check again?

docs/changelog.md
FlorianHockmann: Where? I don't see anything added in the diff, but the note is now missing completely.

FlorianHockmann:
@farodin91 My problem with this PR is mostly that it's not clear to me what the state of support for Java 11 and 8 are after we merge this.

Are we supporting Java 11 in general, but use Java 8 by default, e.g., to build the distribution archives? Or are we already using Java 11 for everything?
And what about the runtime for users? Should users stay on Java 8 for now if they want highest stability and only adopt Java 11 on their own risk? Or should everybody now update to Java 11 as soon as possible?

The answers to these questions should be clear to users after they've read the changelog. I could maybe suggest a wording for the changelog, but I would need to know the answers myself first 😅

FlorianHockmann (approved):
We talked about this offline and changed the formulation for the changelog.

PR #2933, opened 2022-01-12
Ensure Jetty versions are aligned

FlorianHockmann:
We're not directly using Jetty but a few dependencies have a dependency on it which is why it's in the pom.xml so we can prevent dependency conflicts. We currently seem to use the same Jetty version as Solr which is also the newest Jetty version used by any of our dependencies right now.

Using a variable for the Jetty version also enables Dependabot to update the version for all three Jetty dependencies together.

My motivation for this PR is simply that I looked into why Dependabot isn't able to update Jetty at the moment (#2870 + #2923). But it looks like we're not using Jetty ourselves so we just have to ensure that we don't get dependency conflicts from our dependencies that depend on Jetty.

Signed-off-by: Florian Hockmann <fh@florian-hockmann.de>

-----

### For all changes:

### For code changes:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,6 +123,8 @@
         <grpc.version>1.42.1</grpc.version>
         <protoc.version>3.15.3</protoc.version>
         <jmh.version>1.33</jmh.version>
+        <!-- align with org.apache.solr:solr-solrj -->
+        <jetty.version>9.4.44.v20210927</jetty.version>
     </properties>
     <modules>
         <module>janusgraph-grpc</module>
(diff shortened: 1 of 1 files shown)


FlorianHockmann:
@farodin91: Could you please check whether this makes sense, as you've added Jetty to the `pom.xml` [here](https://github.com/JanusGraph/janusgraph/commit/51562e36fd05dba05239f5208c12d061574602cb)?

If so, then we can decline the two Dependabot PRs which try to update Jetty and maybe let Dependabot even ignore these dependencies as we really have to wait for Solr to update Jetty, I guess.

farodin91 (approved):

PR #2937, opened 2022-01-13
Don't set CQL NETTY_{ADMIN,IO}_SIZE by default

### LINKED ISSUE #2936: storage.cql.netty.admin-size default value is 0 instead of 2
### Context:
- Version: 0.6.0
- Storage Backend: cql

### Description:
Cassandra's default size for Netty admin-group is 2 [doc](https://docs.datastax.com/en/developer/java-driver/4.13/manual/core/performance/#thread-pooling), [code](https://github.com/datastax/java-driver/blob/4.13.0/core/src/main/resources/reference.conf#L2219)

JanusGraph 0.6.0 overrides it and sets it to 0 (`Runtime.getRuntime().availableProcessors() * 2`) by default [doc](https://docs.janusgraph.org/configs/configuration-reference/#storagecqlnetty), [code 1](https://github.com/JanusGraph/janusgraph/blob/v0.6.0/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLConfigOptions.java#L351), [code 2](https://github.com/JanusGraph/janusgraph/blob/v0.6.0/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/builder/CQLProgrammaticConfigurationLoaderBuilder.java#L165).

We should probably leave `NETTY_IO_SIZE`/`NETTY_ADMIN_SIZE` unset by default and let Cassandra pick the defaults.
If that's not doable, we could set the default value to 2 in JanusGraph.

cdegroc:
Fixes #2936

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/configs/janusgraph-cfg.md
+++ b/docs/configs/janusgraph-cfg.md
@@ -457,8 +457,8 @@ Configuration options related to the Netty event loop groups used internally by 
 
 | Name | Description | Datatype | Default Value | Mutability |
 | ---- | ---- | ---- | ---- | ---- |
-| storage.cql.netty.admin-size | The number of threads for the event loop group used for admin tasks not related to request I/O (handle cluster events, refresh metadata, schedule reconnections, etc.). If this is set to 0, the driver will use `Runtime.getRuntime().availableProcessors() * 2`. | Integer | 0 | LOCAL |
-| storage.cql.netty.io-size | The number of threads for the event loop group used for I/O operations (reading and writing to Cassandra nodes). If this is set to 0, the driver will use `Runtime.getRuntime().availableProcessors() * 2`. | Integer | 0 | LOCAL |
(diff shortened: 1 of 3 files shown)


porunov (commented):
Thank you @cdegroc .
LGTM but I would probably target `master` branch here because it looks like a breaking change unless there is a reason to introduce this change in 0.6.1 release.
I would also recommend to add this configuration change in upgrade instructions here: https://github.com/JanusGraph/janusgraph/blob/master/docs/changelog.md#upgrade-instructions
I.e. what parameters were changed and what to do if you want to preserve the old default behavior

li-boxuan:
@porunov I don't think it's a breaking change because it seems to fix a regression problem caused by the 0.6.0 release. If it's true, then it makes sense to release it in v0.6.

porunov:
> @porunov I don't think it's a breaking change because it seems to fix a regression problem caused by the 0.6.0 release. If it's true, then it makes sense to release it in v0.6.

Got it. You right then

porunov (approved):
LGTM. Thank you @cdegroc !

FlorianHockmann (approved):

PR #2940, opened 2022-01-13
Add basic vertex property support for grpc

farodin91:
Part of #1853

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
+++ b/janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
@@ -21,6 +21,56 @@ import "google/protobuf/wrappers.proto";
 option java_multiple_files = true;
 option java_package = "org.janusgraph.graphdb.grpc.types";
 
+enum PropertyDataType {
+    PROPERTY_DATA_TYPE_UNSPECIFIED = 0;
+    // java class: String.class
+    PROPERTY_DATA_TYPE_STRING = 1;
+    // java class: Character.class
+    PROPERTY_DATA_TYPE_CHARACTER = 2;
+    // java class: Boolean.class
+    PROPERTY_DATA_TYPE_BOOLEAN = 3;
+    // java class: Byte.class
+    PROPERTY_DATA_TYPE_INT8 = 4;
+    // java class: Short.class
+    PROPERTY_DATA_TYPE_INT16 = 5;
+    // java class: Integer.class
+    PROPERTY_DATA_TYPE_INT32 = 6;
+    // java class: Long.class
+    PROPERTY_DATA_TYPE_INT64 = 7;
+    // java class: Float.class
+    PROPERTY_DATA_TYPE_FLOAT32 = 8;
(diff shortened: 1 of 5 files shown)


janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -21,6 +21,54 @@ import "google/protobuf/wrappers.proto";
 option java_multiple_files = true;
 option java_package = "org.janusgraph.graphdb.grpc.types";
 
+enum PropertyDataType {
+    // java class: String.class
+    STRING = 0;
```
porunov, line 26: Usually by convention it is suggested to use `UNSPECIFIED` element for `0` enum value.
https://developers.google.com/protocol-buffers/docs/style#enums

Do you think we should follow this style convention? I don't have preference but thought could be good to discuss it

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -21,6 +21,56 @@ import "google/protobuf/wrappers.proto";
...
+    google.protobuf.Int64Value id = 1;
+    // The VertexProperty Name.
+    string name = 2;
+    // The Data type.
+    PropertyDataType data_type = 3;
+    enum Cardinality {
+        CARDINALITY_UNSPECIFIED = 0;
+        // Vertex can have the property just once.
+        CARDINALITY_SINGLE = 1;
+        // Vertex can have the multiple property of the same.
```
FlorianHockmann, line 65:
```suggestion
        // Vertex can have multiple properties of the same name.
```

same below

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
```diff
@@ -30,6 +80,8 @@ message VertexLabel {
     bool read_only = 3;
     // Marking as partitioned, default not partitioned.
     bool partitioned = 4;
+    // Contains all constrainted properties at the moment.
```
FlorianHockmann, line 83:
```suggestion
    // Contains all constrained properties at the moment.
```

FlorianHockmann (approved):
The changes look good to me. It's probably a good idea that we review the API again when we add some docs for them. I think that will make it easier to inspect how the API feels from a user's point of view

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
farodin91: fixed

farodin91:
@porunov Do you want to review it again?

janusgraph-grpc/src/main/proto/janusgraph/types/v1/schema_types.proto
FlorianHockmann: Your updated commit fixed this comment here, but the comment two lines below is exactly like this.

Issue #1221, opened 2018-08-13
Support String-type custom vertex ID


czheo:
Though #147 is merged, we still need to call `graph.getIDManager().toVertexId(custom_id)` to explicitly convert a custom vertex ID to a valid Janusgraph vertex ID. I feel it's unfriendly for the end users. For example, when I use gremlinpython library with Janus, I cannot get the IDManager instance from the Python script.
Looking at https://docs.aws.amazon.com/neptune/latest/userguide/access-graph-gremlin-differences.html, I wonder if it's possible to have a similar String-type custom vertex ID as Neptune, where users can provide any string as the vertex ID, and JanusGraph will automatically hash the string and convert it to a Long type internal ID. We don't even need to expose the Long type internal ID to the end users.
Are there any blockers preventing us to do so?

carlspring:
Are there any updates on this issue? Has anybody looked into this any further? We would really like to be able to use UUID-s, for example.

siddharthTyagi:
Same here. I've been using Janusgraph for local development with deployment to aws neptune. Neptune supports only String id values.

li-boxuan:
This does not seem to be too complicated. I am planning to work on this recently.

li-boxuan:
It is not technically impossible - in fact, a working implementation is easy. However, it might not work as you expected.

JanusGraph expects the long id to be within the range [1, 36028797018963968), if you are using the default setting for `cluster.max-partitions`. 36028797018963968 looks like a very large number, but if you consider the mapping from alphanumeric to numeric, this is not that large.

Let's say we want to use digits and lower-case alphabets as custom vertex id. That is 10 + 26 = 36 characters. Now let's consider converting this alphanumeric string to a number. This has to be one-to-one mapping to avoid duplicates. One handy way is to consider the dictionary of the 36 alphanumeric characters as a new base. Then our goal is to convert a "number" from base36 to base10. 

For example, let's say, our dictionary is: "0123456789abcdefghijklmnopqrstuvwxyz". Then for a custom vertexId "vid", the converted long value would be:

`id = 31 * 36 * 36 + 18 * 36 + 13 = 40837`. 

We can see how a short 3-character string gets converted to a long value with 5 digits.

If our custom vertexId is "myvid98134", the converted long value would be `597172750407467008` which is already close to the upper bound. If our custom vertexId is "myvid981340", then we will exceed the upper bound thus get rejected.

[...]

PR #2942, opened 2022-01-14
Refactor: Use Object ID type rather than long type in APIs [tp-tests]

li-boxuan:
This commit refactors the codebase by changing all long-type IDs into Object-type in
all method signatures, method return types, etc. No functionality is affected, except that
IDs are now stored as objects rather than primitives in the memory, which increases
overhead. This is the first step to implementing #1221.

See https://github.com/JanusGraph/janusgraph/issues/1221#issuecomment-1007910018 for the full implementation.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphElement.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphElement.java
@@ -60,18 +60,20 @@ public interface JanusGraphElement extends Element, Idfiable, Removable {
      * @see #hasId
      */
     @Override
-    default Object id() {
-        return longId();
-    }
+    Object id();
 
     /**
      * Unique identifier for this entity. This id can be temporarily assigned and might change.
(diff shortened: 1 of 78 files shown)


janusgraph-core/src/main/java/org/janusgraph/core/JanusGraphElement.java
```diff
@@ -60,18 +60,7 @@ public interface JanusGraphElement extends Element, Idfiable, Removable {
...
-    }
-
-    /**
-     * Unique identifier for this entity. This id can be temporarily assigned and might change.
-     * Use {@link #id()} for the permanent id.
-     *
-     * @return Unique long id
-     */
-    @Override
-    long longId();
```
farodin91, line 74: Do we want to deprecated this function before removing?

li-boxuan: Good idea!

Issue #2325, opened 2021-01-01
Cassandra 4 support


DennisBB:
When Cassandra 4 is gonna be supported?

porunov:
There are no stable Cassandra 4 releases yet. Right now the latest release is `cassandra-4.0-beta4`.
That said, I didn't check Cassandra 4 beta compatibility with JanusGraph yet.

DennisBB:
> Apache Cassandra 3.0 is supported until 6 months after 4.0 release (date TBD)

I think it's better to start preparing till it's out

gklijs:
Release candidate is out now, just fyi, not using JanusGraph at the moment.

DennisBB:
Cassandra 4 has just been released that supports Java 11

Can JanusGraph 0.6.0 be released with Cassandra 4 support?

It would make sense since both should be able to run on Java 11

porunov:
@farodin91 @FlorianHockmann do you think it's possible to upgrade Cassandra to 4.x knowing that `Thrift` was removed from 4.x version but we use it in `CqlBinaryInputFormat` for Hadoop?
```
import org.apache.cassandra.thrift.SlicePredicate;
import org.apache.cassandra.thrift.SliceRange;
```

FlorianHockmann:
It should work without Thrift. @pluradj tried that out with our PR #1436 with Cassandra that had Thrift disabled completely. So, I think that this should also work with Cassandra 4, but we should definitely test that. I guess these types were still in the Thrift package but we don't actually use the Thrift protocol only because we use these types.

We will however need to replace these types with something else when we update our `cassandra.version` to 4 as the `thrift` package was removed completely in that version.

DennisBB:
Can Cassandra 4 also be supported in 0.6.0?

farodin91:
Closed by  #3033

PR #3033, opened 2022-05-08
activate cql tests for cassandra 4 [tp-tests]

### LINKED ISSUE #3017: Dist tests failing
It looks like `dist-tests (8)` started to fail suddenly. I assume we use some external calls to download some dependencies / artifacts / data from external sources in that test and something was changes (maybe a URL or something else). Now all PR's and a `master` branch fails because of this change.

Here is the link to the failed test: https://github.com/JanusGraph/janusgraph/runs/6128877287

As seen from #3025 (https://github.com/JanusGraph/janusgraph/runs/6194674118?check_suite_focus=true) previously this test was working but after trying it again it started to fail.

- Version: master

### Stack Trace

```
Error:  Failures: 
Error:    JanusGraphScriptIT.testGraphOfTheGodsGraphSONFull:33->JanusGraphAssemblyBaseIT.unzipAndRunExpect:113->JanusGraphAssemblyBaseIT.unzipAndRunExpect:84->JanusGraphAssemblyBaseIT.parseTemplateAndRunExpect:106->JanusGraphAssemblyBaseIT.expect:128->JanusGraphAssemblyBaseIT.command:171 expected: <0> but was: <1>
```

farodin91:
Fixes #3017

* Upgrade Cassandra testing dist version to 4.0.6
* Enable full dist builds on java 11 tests
* Fix log collection during local testing
-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,7 +27,7 @@ All currently supported verions of JanusGraph are listed below.
 | JanusGraph | Storage Version | Cassandra | HBase | Bigtable | Elasticsearch | Solr | TinkerPop | Spark | Scala |
 | ----- | ---- | ---- | ---- | ---- | ---- | ---- | --- | ---- | ---- |
 | 0.6.z | 2 | 3.0.z, 3.11.z | 1.6.z, 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 7.y, 8.y | 3.5.z | 3.0.z | 2.12.z |
-| 1.0.z | 2 | 3.0.z, 3.11.z | 1.6.z, 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 8.y | 3.5.z | 3.0.z | 2.12.z |
+| 1.0.z | 2 | 3.11.z, 4.0.z | 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 8.y | 3.5.z | 3.0.z | 2.12.z |
 
 #### End-of-Life
 The versions of JanusGraph listed below are outdated and will no longer receive bugfixes.
@@ -58,8 +58,8 @@ compile "org.janusgraph:janusgraph-core:1.0.0"
 
 **Tested Compatibility:**
 
-* Apache Cassandra 3.0.14, 3.11.10
-* Apache HBase 1.6.0, 2.2.7
+* Apache Cassandra 3.11.10, 4.0.6
+* Apache HBase 2.2.7
 * Google Bigtable 1.3.0, 1.4.0, 1.5.0, 1.6.0, 1.7.0, 1.8.0, 1.9.0, 1.10.0, 1.11.0, 1.14.0
 * Oracle BerkeleyJE 7.5.11
 * Elasticsearch 6.0.1, 6.6.0, 7.17.5
@@ -67,6 +67,11 @@ compile "org.janusgraph:janusgraph-core:1.0.0"
 * Apache Solr 8.11.1
 * Apache TinkerPop 3.5.3
 * Java 8, 11
+
+**Installed versions in the Pre-Packaged Distribution**
+
+* Cassandra 4.0.6
+* Elasticsearch 7.14.0
 
 #### Changes
 
@@ -154,7 +159,7 @@ we clean up all configurations. This could lead to unexpected new log lines. Ple
 JanusGraph now officially supports Java 11 in addition to Java 8. We encourage everyone to update to Java 11.
 
 !!! note
-    The distribution zip archives are however still built with Java 8 since the full distribution includes Cassandra which will only support Java 11 in version 4.
+    The pre-packaged distribution now requires Java 11.
 
 ##### Drop support for 
 
--- a/janusgraph-dist/pom.xml
+++ b/janusgraph-dist/pom.xml
@@ -22,7 +22,7 @@
         <assembly.static.dir>${project.basedir}/src/assembly/static</assembly.static.dir>
         <assembly.resources.dir>${project.basedir}/src/assembly/resources</assembly.resources.dir>
         <assembly.es.dir>${project.build.directory}/elasticsearch-${elasticsearch.version}</assembly.es.dir>
-        <assembly.cassandra.dir>${project.build.directory}/apache-cassandra-${cassandra.version}</assembly.cassandra.dir>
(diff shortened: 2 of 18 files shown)


janusgraph-dist/pom.xml
```diff
@@ -13,6 +13,7 @@
 
     <properties>
         <top.level.basedir>${project.basedir}/..</top.level.basedir>
+        <cassandra-dist.version>4.0.3</cassandra-dist.version>
```
FlorianHockmann, line 16: Why did you add a specific `cassandra-dist.version` instead of upgrading the `cassandra.version`? This now means that we use `cassandra.version` (which is still `3.11.10`)  for tests, but `cassandra-dist.version` (`4.0.3`) in the zip distribution.
Shouldn't we always use the same version for our tests of the CQL backend as the one we include in the distribution?

janusgraph-dist/src/assembly/static/cassandra/conf/jvm-server.options
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 JanusGraph Authors
```
FlorianHockmann, line 1: The file has been renamed, but the `README.md` in `janusgraph-dist` still uses the old name.

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
```diff
@@ -83,7 +83,7 @@ wait_for_cassandra() {
     while [ $now_s -le $stop_s ]; do
         echo -n .
         # The \r\n deletion bit is necessary for Cygwin compatibility
-        statusbinary="`$BIN/../cassandra/bin/nodetool statusbinary 2>/dev/null | tr -d '\n\r'`"
+        statusbinary="`$BIN/../cassandra/bin/nodetool -Dcom.sun.jndi.rmiURLParsing=legacy statusbinary 2>/dev/null | tr -d '\n\r'`"
```
FlorianHockmann, line 86: This disables new validations of URL parsing which seems to have been added for security reasons. Is it really necessary to disable this?

janusgraph-dist/pom.xml
FlorianHockmann, line 16: Apart from the two different versions: If we now ship with Cassandra 4 by default, then this should definitely be named in the `CHANGELOG.md`. It currently doesn't even mention that JanusGraph is compatible with Cassandra 4.

FlorianHockmann (changes requested):

porunov:
@farodin91 Thank you for the fixes! Do you think it's possible to split Cassandra version update (which I think is related to #2325) and dist tests fix (which is related to #3017) to 2 different PRs? I think we need to fix dist tests bug in `v0.6` branch. Otherwise we won't be able to release 0.6.x versions anymore.

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
farodin91: It fixes the test.

farodin91:
@FlorianHockmann  Would you like to review?

FlorianHockmann:
> Would you like to review?

I just tried to review this again, but I'm still a bit confused: Is this now the PR to update Cassandra to version 4? In that case, this also fixes #2325 as @porunov noted, but then it should definitely be clearly documented in the changelog.

farodin91:
cql hadoop can't be replaced directly. We have to switch to the spark read by datastax.

docs/changelog.md
```diff
@@ -27,7 +27,7 @@ All currently supported verions of JanusGraph are listed below.
 | JanusGraph | Storage Version | Cassandra | HBase | Bigtable | Elasticsearch | Solr | TinkerPop | Spark | Scala |
 | ----- | ---- | ---- | ---- | ---- | ---- | ---- | --- | ---- | ---- |
 | 0.6.z | 2 | 3.0.z, 3.11.z | 1.6.z, 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 7.y, 8.y | 3.5.z | 3.0.z | 2.12.z |
-| 1.0.z | 2 | 3.0.z, 3.11.z | 1.6.z, 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 8.y | 3.5.z | 3.0.z | 2.12.z |
+| 1.0.z | 2 | 3.11.z | 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 8.y | 3.5.z | 3.0.z | 2.12.z |
```
FlorianHockmann, line 30: Shouldn't 4.0.z also be added here as a Cassandra Version that 1.0.z is compatible with?

docs/changelog.md
```diff
@@ -68,6 +68,11 @@ compile "org.janusgraph:janusgraph-core:1.0.0"
 * Apache TinkerPop 3.5.3
 * Java 8, 11
 
+**Installed version in JanusGraph with Embedded Cassandra and Elasticsearch**
```
FlorianHockmann, line 71:
```suggestion
**Installed versions in the Pre-Packaged Distribution**
```

farodin91:
@FlorianHockmann Would you like to review it again?

farodin91:
@porunov Would you like to review this PR?

janusgraph-dist/src/assembly/static/cassandra/bin/cassandra
```diff
@@ -82,27 +92,6 @@ elif [ -r "$CASSANDRA_INCLUDE" ]; then
...
-    done
-else
-    JAVA=java
-fi
-
-if [ -z $JAVA ] ; then
-    echo Unable to find java executable. Check JAVA_HOME and PATH environment variables. >&2
-    exit 1;
-fi
-
```
porunov, line 105: Why was this removed? I see `JAVA` variable is still used in this script.

porunov (commented):
Thank you @farodin91 ! I have one comment below

janusgraph-dist/src/assembly/static/cassandra/bin/cassandra
farodin91: This file is mostly https://github.com/apache/cassandra/blob/trunk/bin/cassandra with some lines changed.

porunov: I see it was removed here: https://github.com/apache/cassandra/commit/6ba2fb9395226491872b41312d978a169f36fcdb#diff-ca56a2b54ac847b9ff3908aac256aaa7f160680271725927a4b1917e64206011

Not sure why, but I guess that's fine

porunov (approved):
LGTM. Thank you @farodin91 !

janusgraph-dist/pom.xml
farodin91: I will cleanup those variables, if i have removed the `cassandra-all` dependency.

farodin91:
@FlorianHockmann I've fixed both comments.

FlorianHockmann (approved):

PR #3036, opened 2022-05-11
fix dist build

### LINKED ISSUE #3017: Dist tests failing

farodin91:
Fixes #3017

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-dist/pom.xml
+++ b/janusgraph-dist/pom.xml
@@ -208,8 +208,10 @@
                         </goals>
                         <configuration>
                             <url>https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${elasticsearch.version}-linux-x86_64.tar.gz</url>
-                            <outputFileName>elasticsearch.tar.gz</outputFileName>
-                            <outputDirectory>${project.build.directory}</outputDirectory>
+                            <outputFileName>elasticsearch-${elasticsearch.version}-linux-x86_64.tar.gz</outputFileName>
+                            <outputDirectory>${project.build.directory}/</outputDirectory>
+                            <unpack>true</unpack>
+                            <sha512>${elasticsearch.version.sha512}</sha512>
                         </configuration>
                     </execution>
                     <execution>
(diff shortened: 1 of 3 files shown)


janusgraph-dist/pom.xml
```diff
@@ -208,8 +208,10 @@
                         </goals>
                         <configuration>
                             <url>https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${elasticsearch.version}-linux-x86_64.tar.gz</url>
-                            <outputFileName>elasticsearch.tar.gz</outputFileName>
-                            <outputDirectory>${project.build.directory}</outputDirectory>
+                            <outputFileName>elasticsearch-${elasticsearch.version}-linux-x86_64.tar.gz</outputFileName>
+                            <outputDirectory>${project.build.directory}/</outputDirectory>
+                            <unpack>true</unpack>
+                            <sha512>30764d5838009dc36d8f5c6c7249e65f323b5bd843027b47b37b91566d28cac95c4b5e6db1decce748f9ccd404a7ca4ba4fa9632bcd8b43fb27b0d93cc7b8f27</sha512>
```
FlorianHockmann, line 214: Is this the hash of the archive? Then we'll have to update it every time we update to another version, right? Do we want to do that for security reasons or do we maybe want to omit this check here?

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
```diff
@@ -48,7 +48,7 @@ fi
 : ${GSRV_PORT:=8182}
 
 : ${SLEEP_INTERVAL_S:=2}
-VERBOSE=
+VERBOSE=true
```
FlorianHockmann, line 51: Is this a leftover from debugging or do we really want to enable verbose logging by default?

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
```diff
@@ -83,7 +83,7 @@ wait_for_cassandra() {
     while [ $now_s -le $stop_s ]; do
         echo -n .
         # The \r\n deletion bit is necessary for Cygwin compatibility
-        statusbinary="`$BIN/../cassandra/bin/nodetool statusbinary 2>/dev/null | tr -d '\n\r'`"
+        statusbinary="`$BIN/../cassandra/bin/nodetool -Dcom.sun.jndi.rmiURLParsing=legacy statusbinary 2>/dev/null | tr -d '\n\r'`"
```
FlorianHockmann, line 86: Copying [my comment from the other PR](https://github.com/JanusGraph/janusgraph/pull/3033#discussion_r868954744):

This disables new validations of URL parsing which seems to have been added for security reasons. Is it really necessary to disable this?

janusgraph-dist/pom.xml
farodin91: We have to update each time, but otherwise maven thought the file is already cached. So you run the tests with old version.

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
farodin91: It was on of the tests.

janusgraph-dist/pom.xml
FlorianHockmann: OK, then I guess we have to update the values manually each time.

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
farodin91: I have no idea it happened in the first place.

porunov: @FlorianHockmann @farodin91 I think Java 8 was upgraded in CI because we didn't specify correct fixed version (which we probably should do) and it caused the issue described [here](https://community.datastax.com/questions/13764/java-version-for-cassandra-3113.html) (see the answer [here](https://community.datastax.com/answers/13817/view.html)).
So, I believe this is the correct way to fix that issue as for now but later we will need to upgrade Cassandra and remove this legacy URL parsing (looks like [related issue](https://github.com/apache/cassandra/pull/1586) was closed recently).

farodin91:
@porunov I activated verbose only for debugging purpose.

porunov (approved):

janusgraph-dist/src/assembly/static/bin/janusgraph.sh
FlorianHockmann: Thanks for looking into this and the detailed explanation, @porunov! Then

FlorianHockmann (approved):

PR #3080, opened 2022-06-16
Allow removal of global configuration properties

### LINKED ISSUE #555: JanusGraphManagement does not allow to remove any property once set
``JanusGraphManagement`` has only methods ``get()`` and ``set()`` for changing config values.
Set method does not accept null values, which means there is NO way to clear any property once set.

As an example if I have configured some mixed index backend using property:
``index.search.backend=solr``
there is NO way to tell Janus to actually stop using this backend.

Since there can be an arbitrary number of indexing backends and there is NO default value for ``index.[X].*`` properties, there must be a way to actually remove them without dropping the whole janus database.

cdegroc:
Fixes #555

Usage
```
mgmt = graph.openManagement()
mgmt.set(<config_key>, null)
mgmt.commit()
```

I've confirmed locally that, once removed, configuration keys can't be retrieved using `ManagementSystem#get()` and don't appear anymore in `graph.getBackend().getGlobalSystemConfig()`.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-test/src/test/java/org/janusgraph/diskstorage/configuration/UserModifiableConfigurationTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/diskstorage/configuration/UserModifiableConfigurationTest.java
@@ -1,0 +1,48 @@
+// Copyright 2022 JanusGraph Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.diskstorage.configuration;
+
+import org.apache.commons.configuration2.BaseConfiguration;
+import org.janusgraph.diskstorage.configuration.backend.CommonsConfiguration;
+import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
+import org.janusgraph.util.system.ConfigurationUtil;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class UserModifiableConfigurationTest {
+
+    private UserModifiableConfiguration newBaseConfiguration() {
+        final BaseConfiguration base = ConfigurationUtil.createBaseConfiguration();
+        final CommonsConfiguration config = new CommonsConfiguration(base);
+        return new UserModifiableConfiguration(new ModifiableConfiguration(GraphDatabaseConfiguration.ROOT_NS, config.copy(), BasicConfiguration.Restriction.NONE));
+    }
+
+    @Test
+    public void testRemove() {
+        UserModifiableConfiguration configuration = newBaseConfiguration();
+        configuration.set("storage.backend", "inmemory");
+        assertEquals("inmemory", configuration.get("storage.backend"));
+
+        configuration.set("storage.backend", null);
+        assertEquals("null", configuration.get("storage.backend"));
+
+        configuration.set("index.search.backend", "lucene");
+        assertEquals("lucene", configuration.get("index.search.backend"));
+        assertEquals("+ search\n", configuration.get("index"));
+
+        configuration.set("index.search.backend", null);
+        assertEquals("", configuration.get("index"));
+    }
+}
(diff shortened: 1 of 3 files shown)


janusgraph-test/src/test/java/org/janusgraph/diskstorage/configuration/UserModifiableConfigurationTest.java
```diff
@@ -0,0 +1,48 @@
+// Copyright 2021 JanusGraph Authors
```
FlorianHockmann, line 1: (nitpick) 2022

FlorianHockmann (commented):
The changes look good to me in general, but should we briefly mention this in the docs?

Also what about [this comment from @robertdale](https://github.com/JanusGraph/janusgraph/issues/555):

> If there becomes the ability to clear index.*.backend properties, be sure to document that indexes on that backend would first have to be DISABLE and REMOVE (which effectively disables using that backend).

How will JanusGraph behave if an index backend is configured and in use for some mixed index and the index backend is then removed from the config? Will it still try to use it for traversals that could use that index?

li-boxuan:
> How will JanusGraph behave if an index backend is configured and in use for some mixed index and the index backend is then removed from the config? Will it still try to use it for traversals that could use that index?

This definitely needs experiments (or even better, having a test case to showcase that), but AFAIK this will cause read & writes involving that indexing backend to fail. The reason is that schema information is stored in the storage backend, not in the index backend. So JanusGraph will assume the mixed indexes still exist and try connecting to the indexing backend.

I think it could be a good improvement to prevent users from removing an index backend in config if there is an active mixed index for that index backend.

li-boxuan (commented):
I am okay with this PR - What @FlorianHockmann said [here](https://github.com/JanusGraph/janusgraph/pull/3080#pullrequestreview-1010199223) made sense, but we could create a follow-up issue for this. For now, my suggestion is that we can:

Rather than this breaking change, add a `remove` API and add some Javadoc saying that removal of config might not work properly under all circumstances (e.g. index settings) so use it at your own risk.

After all, it is clear that this functionality is useful in many cases, so some advanced users could still utilize it. We don't want to mention it in the user doc until we have figured out and documented all edge cases.

cdegroc:
> Rather than this breaking change, add a `remove` API and add some Javadoc saying that removal of config might not work properly under all circumstances (e.g. index settings) so use it at your own risk.
> 
> After all, it is clear that this functionality is useful in many cases, so some advanced users could still utilize it. We don't want to mention it in the user doc until we have figured out and documented all edge cases.

Thanks @li-boxuan. Adding a `remove` API would be cleaner and the "use at your own risk" approach sounds right to me.

The disadvantage is that this API will be different from `ConfiguredGraphFactory`, but we could in turn update `ConfiguredGraphFactory` to have a similar API (for instance, add a `removeConfiguration(String graphName, Configuration config)` method.

One issue is that `Configuration` are key/value collections and so we would have to use `removeConfiguration(String graphName, Iterable<String> configKeys)`.

li-boxuan:
> The disadvantage is that this API will be different from ConfiguredGraphFactory, but we could in turn update ConfiguredGraphFactory to have a similar API (for instance, add a removeConfiguration(String graphName, Configuration config) method

That sounds good to me. @FlorianHockmann what do you think?

FlorianHockmann:
> That sounds good to me. @FlorianHockmann what do you think?

Sure, sounds good to me! :+1:

FlorianHockmann:
Just stumbled upon this PR again when looking into the open issues for the 0.6.3 milestone. It's not clear to me any more what we agreed on here 🙁 

If I understand our discussion here correctly then what's mostly missing is some addition to the Java docs that explains how this removal works but also warns that it won't work for all options.
Is that correct?
And @cdegroc, are you planning to come back to this PR any time soon? Then we can include it in the 0.6.3 release.

If you come back to this PR, then it probably also makes sense to change the target branch to `master` and add the labels _backport_  and _backport/v0.6_. We have recently changed our merging strategy: Instead of requiring contributors to target `v0.6` and then merge the PR first into `v0.6` and then merge that into `master`, we now only target `master` and let a bot create a separate backporting PR if we want a change to also land into `v0.6`.
But we can alternatively probably also merge this into `v0.6` and then cherry-pick the commit into `master`.

This seems to be still open:
> The disadvantage is that this API will be different from ConfiguredGraphFactory, but we could in turn update ConfiguredGraphFactory to have a similar API (for instance, add a removeConfiguration(String graphName, Configuration config) method.

Do we want to do this as part of this PR or in a follow-up one? Doing it in a different PR is OK to me, but could you please create an issue for that, @cdegroc? (I think you have a better understanding of this to properly explain what needs to be done.)

porunov:
To me it looks like a good feature to have but not a critical bug. Thus, I propose to reschedule this PR to v0.6.4 or 1.0.0. And start 0.6.3 release process.

li-boxuan:
I would like to have this PR in 1.0.0 and 0.6.4, so I am inclined to merge it. We could add an one-liner in release notes that it's a breaking change.

li-boxuan (approved):

cdegroc:
👋 Sorry for not answering earlier. I'll try to catch up and see if there's anything I can help with.

cdegroc:
Created #4061 to address comments regarding the interface. Let me know what you think.

PR #3101, opened 2022-06-24
Setup Cassandra benchmark env and add CQLMultiQueryBenchmark

li-boxuan:
Currently, all our benchmarks are using the in-memory backend. This PR sets up the infrastructure for Cassandra-based benchmarks and adds a multi-query benchmark using the CQL backend.

-----

### For all changes:

### For code changes:

### DIFF
--- a/.github/workflows/ci-benchmark.yml
+++ b/.github/workflows/ci-benchmark.yml
@@ -44,11 +44,18 @@ jobs:
         with:
           java-version: 1.8
       - run: mvn clean install --projects janusgraph-all -Pjanusgraph-cache -Dmaven.javadoc.skip=true ${{ env.BUILD_MAVEN_OPTS }}
-
+      - run: mvn verify --projects janusgraph-all -Pjanusgraph-cache ${{ env.VERIFY_MAVEN_OPTS }}
   benchmark:
     name: Performance regression check
     runs-on: ubuntu-latest
+    needs: build-all
     steps:
+      # install CCM for CQL benchmarks
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - run: pip install git+https://github.com/li-boxuan/ccm.git
+
       - uses: actions/checkout@v2
       - uses: actions/cache@v2
         with:
@@ -61,6 +68,9 @@ jobs:
           java-version: 1.8
       - run: mvn clean install -Pjanusgraph-benchmark ${{ env.BUILD_MAVEN_OPTS }} -Dgpg.skip=true
       - run: mvn verify --projects janusgraph-benchmark
+        env:
+          CCM_CLUSTER_START_DEFAULT_TIMEOUT: 300
+          CCM_UPDATE_PID_DEFAULT_TIMEOUT: 300
 
       # Download previous benchmark result from cache (if exists)
       - name: Download previous benchmark data
--- a/janusgraph-benchmark/src/main/java/org/janusgraph/BenchmarkRunner.java
+++ b/janusgraph-benchmark/src/main/java/org/janusgraph/BenchmarkRunner.java
@@ -14,6 +14,7 @@
 
 package org.janusgraph;
 
+import com.datastax.oss.driver.api.testinfra.ccm.CcmBridge;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
 import org.openjdk.jmh.results.Result;
 import org.openjdk.jmh.results.RunResult;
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-benchmark.yml
```diff
@@ -30,25 +30,16 @@ permissions:
   contents: write
 
 jobs:
-  build-all:
```
farodin91, line 33: Could you re-add this section?

li-boxuan: Tbh I don't understand why we need this section. Do you mind helping understand?

li-boxuan: I see, thanks for the explanation! I should also make `benchmark` job dependent on this `build-all` job, right?

farodin91: Correct

.github/workflows/ci-benchmark.yml
```diff
@@ -44,11 +44,18 @@ jobs:
...
   benchmark:
     name: Performance regression check
     runs-on: ubuntu-latest
+    needs: build-all
     steps:
+      # install CCM for CQL benchmarks
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - run: pip install git+https://github.com/li-boxuan/ccm.git
```
li-boxuan, line 57: I tried many many times and the CI only worked with my custom changes to `ccm`. I will submit a PR to the upstream repository.

farodin91 (commented):
Just interest. Why didn't you use testcontainers for the cql tests?

li-boxuan:
> Why didn't you use testcontainers for the cql tests?

That's a good question. I guess we could use test containers, but IIUC it does not support multiple Cassandra instances. `ccm` gives us more flexibility in the Cassandra cluster setup. It is also the tool used by Cassandra's official [distributed test suite](https://github.com/apache/cassandra-dtest). Also, I am not sure but I suspect using Cassandra in docker would bring additional overhead that might affect the stability of benchmark results. But yeah, there is no strong reason against testcontainers.

li-boxuan:
This is a blocker for https://github.com/JanusGraph/janusgraph/pull/2942

janusgraph-benchmark/src/main/java/org/janusgraph/CQLMultiQueryBenchmark.java
```diff
@@ -0,0 +1,103 @@
...
+    @Param({"100", "500"})
+    int fanoutFactor;
+
+    JanusGraph graph;
+
+    public WriteConfiguration getConfiguration() {
+        ModifiableConfiguration config = GraphDatabaseConfiguration.buildGraphConfiguration();
+        config.set(GraphDatabaseConfiguration.STORAGE_BACKEND,"cql");
+        config.set(CQLConfigOptions.LOCAL_DATACENTER, "dc1");
+        config.set(GraphDatabaseConfiguration.USE_MULTIQUERY, true);
```
rngcntr, line 56: How about adding the `USE_MULTIQUERY` flag as a `@Param` to provide some insights to the performance boost multi query brings?

janusgraph-benchmark/src/main/java/org/janusgraph/BenchmarkRunner.java
```diff
@@ -62,17 +49,65 @@ public static void main(String[] args) throws RunnerException, IOException {
...
+        System.out.println("Cassandra version = " + bridge.getCassandraVersion());
+
+        cqlBuilder.include("CQL.*Benchmark");
+        transformResults(new Runner(cqlBuilder.build()).run(), outputs);
+    }
+
+    public static void main(String[] args) throws RunnerException, IOException, InterruptedException {
+        final boolean runSpecifiedTests = args.length > 0;
+        final List<Map<String, Object>> outputs = new ArrayList<>();
+        final Supplier<ChainedOptionsBuilder> builderSupplier = () -> new OptionsBuilder()
```
rngcntr, line 85: Just a stylistic comment: Is there a reason to use the lambda supplier in this case instead of a private method like `getDefaultOptionsBuilder()`. This would shorten the `main` method and thus make it easier to read and easier to extend.

rngcntr (commented):
Hi @li-boxuan, thanks for your PR!

janusgraph-benchmark/src/main/java/org/janusgraph/CQLMultiQueryBenchmark.java
li-boxuan: That's a good idea, but unfortunately, it will take much much longer time for the benchmark to finish. I have tried it manually and (as we would expect) `USE_MULTIQUERY = true` makes the queries much much faster.

rngcntr: Ouch, I didn't expect the time to be on the scale of hours. In this case, I think it's fine as it is.

rngcntr (approved):

li-boxuan:
@farodin91 Do you want to review again?

farodin91 (approved):

PR #3149, opened 2022-07-20
Add gremlin script evaluation API

li-boxuan:
Currently, there is no way to evaluate a gremlin query in plain string format
if users deploy JanusGraph as an embedded Java library rather than JanusGraph
servers. This commit adds a new API that allows users to use the `JanusGraph`
Java object to directly evaluate gremlin queries.

An example usage is `graph.eval("g.V().has('_v', 1).values('_v').hasNext()")`.

-----

### For all changes:

### For code changes:


janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
```diff
@@ -230,6 +248,23 @@ public String getGraphName() {
         return this.name;
     }
 
+    @Override
+    public Object eval(String gremlinScript) {
+        Objects.requireNonNull(scriptEngine, String.format("%s.%s is not enabled", SCRIPT_EVAL.getNamespace().getName(), SCRIPT_EVAL.getName()));
+        JanusGraphTransaction tx = newTransaction();
+        try {
+            Bindings bindings = scriptEngine.createBindings();
+            bindings.put("g", tx.traversal().withStrategies(ReadOnlyStrategy.instance()));
```
porunov, line 257: I don't know but do we need to create a new 'Bindings' instance each time we evaluate a script? Can't we just create 'Bindings' instance in the constructor?

Also a separate comment about ReadOnlyStrategy. Why do we want script evaluation to be read only? As far as I know we could do that by creating a read-only default transaction but then it would force all default transactions of this graph to be read-only. So, maybe it makes sense to add another configuration which allows to enable / disable read-only mode?

porunov (commented):
Thank you @li-boxuan ! Looks good but I have several questions about 'Bindings' instance below.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
li-boxuan: > I don't know but do we need to create a new 'Bindings' instance each time we evaluate a script? Can't we just create 'Bindings' instance in the constructor

If we create an instance in constructor, say, `bindings.put("g", graph.traversal().withStrategies(ReadOnlyStrategy.instance()));`, then the transaction will be visible to uncommitted results.

Example 1:
```
graph.addVertex("dummy") // create a vertex
graph.eval("g.V().hasLabel('dummy')") // this will return true because it can see outside changes
```
This is not so great because it sees uncommitted changes. This might be acceptable to some users though. The next example demonstrates why we must not do this.

Example 2:
```
graph.addVertex("dummy") // create a vertex
graph.eval("g.tx().rollback()") // malicious/accidental rollback
graph.tx().commit() // nothing is committed because this transaction has been rolled back
```
There is no (easy) way to prevent user from submitting a `g.tx().rollback()` script. Therefore, we must create a new transaction whenever this `graph.eval` method is called.

li-boxuan: > As far as I know we could do that by creating a read-only default transaction but then it would force all default transactions of this graph to be read-only

I am not sure if I understand you correctly here. Calling `tx.traversal().withStrategies(ReadOnlyStrategy.instance())` is fine since it will not affect other transactions.

> Why do we want script evaluation to be read only

This is debatable. I did it mainly as a precaution (malicious queries like `g.V().drop().iterate()` can happen and JanusGraph user might not have control over it).
If we want to allow writes, the question is, how do we implement it? Shall we do it like `graph.eval(String script, boolean commit)`?

porunov: > This is debatable. I did it mainly as a precaution (malicious queries like g.V().drop().iterate() can happen and JanusGraph user might not have control over it).

I see. I would probably like to have a small debate about it because it gives users a false feeling of feeling that you can safely expose script evaluation and it will be guaranteed that anything they execute there will be read-only. If we talk specifically about malicious queries then this script engine should way more protected. For example, I didn't test this query myself but what would happen to the next evaluation?
```
g2 = ((JanusGraph) g.getGraph()).buildTransaction().start().traversal(); g2.V().drop().iterate(); g2.tx().commit();
```
Moreover, by exposing script evaluation you literally giving access to the flow JVM and by accessing it you can do whatever you want with the server. For example, someone could get a dump of the whole memory or get some credentials, or access the network this machine is hosted in, etc. 
By giving such access to the user you should probably limit a lot of things in the script evaluation engine (at least following some guideline from TinkerPop [here](https://tinkerpop.apache.org/docs/current/reference/#script-execution)).
Creating a secure evaluation would be really great but that requires us to configure gremlin script engine quite a bit.
So, I would probably not consider it as a protection against malicious script. Thus, I don't see any reason why this specific traversal is read-only. If use doesn't want to change anything - they should not write any scripts which change data.
If we proceed with read-only default traversal then I bet we will have a feature request soon which allows traversal to write some data, or even worse, users will start using hacks like:
```
g2 = ((JanusGraph) g.getGraph()).buildTransaction().start().traversal(); g2.addV("dummy").next(); g2.tx().commit();
```
So, I guess I would actually make `read-only` traversal creation as an optional feature. Moreover I would explain that it simply creates a read-only traversal but it doesn't protect users from creating `write` transactions inside scripts. I would also probably add some documentation which says that it isn't safe to expose this evaluation to clients as it doesn't protect from executing arbitrary code. And then I would probably point to TinkerPop documentation about the used script evaluation engine.
That said, I could be wrong but would like to her your thoughts about it. Maybe creating read-only default traversal is actually good by default. I'm just not sure right now.

porunov: > If we create an instance in constructor, say, bindings.put("g", graph.traversal().withStrategies(ReadOnlyStrategy.instance()));, then the transaction will be visible to uncommitted results.

I see. I'm just thinking that it could be desirable in some situations.
What if we create 2 different bindings? 
Let's say we have an interface like:
`eval(String script, boolean readUncommitted)`.
Let's say now the first binding is a default binding which is created in the constructor and assigned be default (i.e. allows read-uncommitted) but if `readUncommitted` is `false` - it actually creates a new binding as you do right now and not uses a default binding.
If so, those users who need to read elements from uncommitted transaction can do that and those users who don't need it - can omit that functionality.
What do you think about it?

li-boxuan: > If we talk specifically about malicious queries then this script engine should be way more protected.

That's a great point! Indeed, a query like the following would circumvent the read-only limitation:

```java
g2 = g.getGraph().traversal(); g2.V().drop().iterate(); g2.tx().commit();
```

The fix is easy: we could use `GremlinLangScriptEngine` instead of `GremlinGroovyScriptEngine`. The former one requires the query to be a single gremlin query while the latter allows you to submit a script. I guess we could make it configurable and let users know the differences between the two.

> If so, those users who need to read elements from uncommitted transaction can do that

That is fair, but I cannot think of a use case for that. Do you have any in mind?

porunov: > The fix is easy: we could use GremlinLangScriptEngine instead of GremlinGroovyScriptEngine. The former one requires the query to be a single gremlin query while the latter allows you to submit a script. I guess we could make it configurable and let users know the differences between the two.

That's a good alternative I believe. Allowing users to chose between full Gremlin + arbitrary scripts execution vs limited Gremlin only. I see that `GremlinLangScriptEngine` allows only gremlin queries without any lambdas. I didn't investigate if it has any vulnerabilities against malicious queries. From the first look I don't see any security problems with it but as I said I didn't investigate it.
So, I would vote +1 for making it configurable as you proposed.

> That is fair, but I cannot think of a use case for that. Do you have any in mind?

To be fair I don't see good use cases for read-only queries when we need access to uncommitted elements. Maybe the only benefit would be that we are saving heap memory by not creating redundant `GraphTraversalSource` and `Bindings` for each execution of `eval` method but I didn't check what is a memory foot print for these objects actually. I guess in situations when the user uses default JanusGraph transaction for scripts executions only we don't actually need to create a new `GraphTraversalSource` and a new `Bindings` because there will be no any uncommitted elements in that transaction.
That said, for a situation when `eval`  allows writes then there could actually be some use cases (not sure how good are they but still). For example:
1st execution: `.eval("g.addV().property('id', 123)")`
2nd execution: `.eval("g.addV().property('id', 321)")`
3rd execution: `.eval("g.V().has('id', 123).addE('someEdge').to(g.V().has('id', 321).next())")`
I guess this example is kind of bad because the user could write all of that in a single query but I was just trying to create a use case where user depends on some uncommitted elements in current traversal.
I was thinking, maybe we could create a configuration options like:

`graph.script-eval.enabled - If "true" then attaches Script engine evaluation to JanusGraph instance and allows to use "eval" method.`

`graph.script-eval.use-new-traversal - If "true" (default) then uses a new traversal per evaluation execution to prevent access to read-uncommitted elements. Otherwise uses a default traversal which is shared between all executions of "eval" method. Notice, when "false" it will allow read-uncommitted, thus make sure that it's a desired behavior for scripts evaluation.`.

`graph.script-eval.read-only - If "true" (default) then a binded traversal is created with read-only capabilities. Notice, when "GremlinGroovyScriptEngine" is used it won't prevent a user to send a malicious query which creates a new traversal without those restrictions (please see "query.script-eval.engine" configuration option for more details).`.

`graph.script-eval.engine - Full class path of script engine to be used. Should implement "ScriptEngine" interface. Reserved keywords are "GremlinLangScriptEngine" (default) (https://tinkerpop.apache.org/javadocs/current/full/org/apache/tinkerpop/gremlin/jsr223/GremlinLangScriptEngine.html) and "GremlinGroovyScriptEngine" (https://tinkerpop.apache.org/javadocs/current/full/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptEngine.html). Please see possible security problems with "GremlinGroovyScriptEngine" here (https://tinkerpop.apache.org/docs/current/reference/#script-execution).`.

Also, notice I change configuration options prefix from `query` to `graph` because in this situation we actually configure JanusGraph instance itself and not queries execution. From my understanding `query` prefix is more about how queries are processed by JanusGraph (any queries. Like queries via script evaluation, direct Gremlin queries, direct JanusGraph API queries) and `graph.script-eval` is less about how queries are processed but more about enabling script evaluation for this particular JanusGraph instance or not. 

What do you think about these configuration options? Do you think `read-only` or `use-new-traversal` might be redundant?

li-boxuan: The problem with `graph.script-eval.use-new-traversal` I see is, how do you control when to commit the transaction? In your previous example,
```
1st execution: .eval("g.addV().property('id', 123)")
2nd execution: .eval("g.addV().property('id', 321)")
3rd execution: .eval("g.V().has('id', 123).addE('someEdge').to(g.V().has('id', 321).next())")
```
When shall we commit this transaction? Should we let user explicitly commit it?

porunov: > When shall we commit this transaction? Should we let user explicitly commit it?

I would say that it should be on user's chine when they need to commit a transaction. I think they can execute `g.tx().commit()` when they need to do that. Again for convenience we can provide additional configuration which commits current default transaction each time we execute `eval`.

 > I feel like things can get very tricky and there are many edge cases. Maybe we could open a new issue as a follow-up?

Agree. We can always use a new traversal as for now and create a separate issue for making it configurable later.

docs/configs/janusgraph-cfg.md
```diff
@@ -520,7 +529,7 @@ HBase storage options
 | storage.hbase.short-cf-names | Whether to shorten the names of JanusGraph's column families to one-character mnemonics to conserve storage space | Boolean | true | FIXED |
 | storage.hbase.skip-schema-check | Assume that JanusGraph's HBase table and column families already exist. When this is true, JanusGraph will not check for the existence of its table/CFs, nor will it attempt to create them under any circumstances.  This is useful when running JanusGraph without HBase admin privileges. | Boolean | false | MASKABLE |
 | storage.hbase.snapshot-name | The name of an existing HBase snapshot to be used by HBaseSnapshotInputFormat | String | janusgraph-snapshot | LOCAL |
-| storage.hbase.snapshot-restore-dir | The temporary directory to be used by HBaseSnapshotInputFormat to restore a snapshot. This directory should be on the same File System as the HBase root dir. | String | /tmp | LOCAL |
+| storage.hbase.snapshot-restore-dir | The temporary directory to be used by HBaseSnapshotInputFormat to restore a snapshot. This directory should be on the same File System as the HBase root dir. | String | /var/folders/nc/5yy0j48d5z3dx_57kvd_zsjh0000gq/T/ | LOCAL |
```
porunov, line 532: Looks like an accident change

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
```diff
@@ -230,6 +248,37 @@ public String getGraphName() {
         return this.name;
     }
 
+    @Override
+    public Object eval(String gremlinScript, boolean commit) {
```
porunov, line 252: I think I would better rename `boolean commit` to `boolean readOnlyBoundTraversal` (or some similar name) so that it would be clear that the bound traversal is created with read-only mode.
In Javadoc or JanusGraph documentation I would also add information that this method commits a newly created transaction when `readOnlyBoundTraversal == false` or rollbacks a newly created transaction when `readOnlyBoundTraversal == true`.
The reason for that is because we will be able to extend implementations later by adding a new method like:
`eval(String script, boolean readOnlyBoundTraversal, boolean wrapTraversalInNewTx)`

Then an implementation of previous `eval` will look like:
```
public Object eval(String gremlinScript, boolean readOnlyBoundTraversal) {
  return eval(gremlinScript, readOnlyBoundTraversal, true);
}
```

li-boxuan: `readOnlyBoundTraversal` is ambiguous IMO coz non-readonly does not necessarily mean it will be committed.

I do see your point here, that's a valid concern. 👍 My feeling is, once we allow users to use existing threaded transaction, it would be insufficient to use a combination of binary flags to control. My proposal is, once we extend implementation later, we could use an enum to control:

- `newTxReadOnly`: open a new transaction and is read-only
- `newTxAutoCommit`: open a new transaction and do auto-commit
- `threadedTxReadOnly`: use existing threaded tx (could see uncommitted results) and is read-only
- `threadedTxAutoCommit`: use existing threaded tx (could see uncommitted results) and do auto-commit
- `threadedTxNormal`: use existing threaded tx (could see uncommitted results), read is allowed, but no auto-commit

As of now, I am inclined to using the name `commit` rather than sth like `readOnly`. I will add documentation once this PR is merged for sure.

porunov: I guess instead of developing different variants of `eval` method and try to guess all possible use-cases we could teach people in the documentation on how to create their own script evaluation engine which isn't restricted to a specific use-case.
I.e. I would imagine a simple documentation which points to TinkerPop documentation and then maybe adds some examples like:
```
JanusGraph graph = ...;
GremlinLangScriptEngine scriptEngine = new GremlinLangScriptEngine();
Bindings bindings = scriptEngine.createBindings();
bindings.put("graph", graph);
bindings.put("g", graph.traversal());
```

li-boxuan: > I guess instead of developing different variants of eval method and try to guess all possible use-cases we could teach people in the documentation on how to create their own script evaluation engine which isn't restricted to a specific use-case.

That's a good point! I'd still like to have this method within JG so that it's handy to use. My own usecase only involves read-only transactions so I am not even sure if we shall expose the auto-commit functionality or not. What do you think?

porunov: > My own usecase only involves read-only transactions so I am not even sure if we shall expose the auto-commit functionality or not. What do you think?

To be fair I don't have much preference on that. If we add good documentation I'm OK with any case (i.e. with `commit` and without `commit`).

porunov (approved):
LGTM. Thank you @li-boxuan !

PR #3162, opened 2022-07-30
Refactor: Use Object ID type rather than long type in APIs [tp-tests]

li-boxuan:
This commit refactors the codebase by changing all long-type ID into Object-type in
all method signatures, return types, etc. No functionality is affected, except that
IDs are now stored as objects rather than primitives in the memory, which increases
memory footprint by a little bit. This is the first step to implement #1221.

This PR is to replace https://github.com/JanusGraph/janusgraph/pull/2942 because the benchmark PR commenter is only available from the JanusGraph repo.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/VertexMemoryHandler.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/VertexMemoryHandler.java
@@ -26,7 +26,6 @@ import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
 import org.janusgraph.core.JanusGraphEdge;
 import org.janusgraph.core.JanusGraphException;
-import org.janusgraph.core.JanusGraphVertex;
 import org.janusgraph.core.JanusGraphVertexProperty;
 import org.janusgraph.graphdb.vertices.PreloadedVertex;
 
@@ -46,14 +45,14 @@ class VertexMemoryHandler<M> implements PreloadedVertex.PropertyMixing, Messenge
 
     protected final FulgoraVertexMemory<M> vertexMemory;
     private final PreloadedVertex vertex;
-    protected final long vertexId;
+    protected final Object vertexId;
     private boolean inExecute;
 
     VertexMemoryHandler(FulgoraVertexMemory<M> vertexMemory, PreloadedVertex vertex) {
         assert vertex!=null && vertexMemory!=null;
         this.vertexMemory = vertexMemory;
         this.vertex = vertex;
-        this.vertexId = vertexMemory.getCanonicalId(vertex.longId());
+        this.vertexId = vertexMemory.getCanonicalId(vertex.id());
         this.inExecute = false;
     }
 
@@ -131,7 +130,7 @@ class VertexMemoryHandler<M> implements PreloadedVertex.PropertyMixing, Messenge
 
             return edges.stream()
                         .flatMap(e -> {
-                            long canonicalId = vertexMemory.getCanonicalId(((JanusGraphEdge) e).otherVertex(vertex).longId());
+                            Object canonicalId = vertexMemory.getCanonicalId(((JanusGraphEdge) e).otherVertex(vertex).id());
                             return vertexMemory.getMessage(canonicalId, localMessageScope)
                                                .map(msg -> msg == null ? null : edgeFct.apply(msg, e));
                         })
@@ -154,9 +153,7 @@ class VertexMemoryHandler<M> implements PreloadedVertex.PropertyMixing, Messenge
             vertexMemory.sendMessage(vertexId, m, messageScope);
         } else {
             ((MessageScope.Global) messageScope).vertices().forEach(v -> {
-                long vertexId;
-                if (v instanceof JanusGraphVertex) vertexId=((JanusGraphVertex)v).longId();
-                else vertexId = (Long)v.id();
+                long vertexId = ((Number) v.id()).longValue();
                 vertexMemory.sendMessage(vertexMemory.getCanonicalId(vertexId), m, messageScope);
             });
         }
@@ -174,7 +171,7 @@ class VertexMemoryHandler<M> implements PreloadedVertex.PropertyMixing, Messenge
                 return super.receiveMessages(messageScope);
             } else {
                 final MessageScope.Local<M> localMessageScope = (MessageScope.Local) messageScope;
-                return vertexMemory.getAggregateMessage(vertexId,localMessageScope);
+                return vertexMemory.getAggregateMessage(((Number) vertexId).longValue(), localMessageScope);
             }
         }
     }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/ElementUtils.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/ElementUtils.java
@@ -16,7 +16,6 @@ package org.janusgraph.graphdb.tinkerpop;
 
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.janusgraph.core.JanusGraphEdge;
-import org.janusgraph.core.JanusGraphVertex;
 import org.janusgraph.graphdb.relations.RelationIdentifier;
 
 /**
@@ -24,24 +23,19 @@ import org.janusgraph.graphdb.relations.RelationIdentifier;
  */
 public class ElementUtils {
 
-    public static long getVertexId(Object id) {
-        if (null == id) return 0;
+    public static Object getVertexId(Object id) {
+        if (null == id) return null;
 
-        if (id instanceof JanusGraphVertex) //allows vertices to be "re-attached" to the current transaction
-            return ((JanusGraphVertex) id).longId();
+        if (id instanceof Vertex) //allows vertices to be "re-attached" to the current transaction
+            return ((Vertex) id).id();
         if (id instanceof Long)
             return (Long) id;
         if (id instanceof Number)
             return ((Number) id).longValue();
-
         try {
-            // handles the case of a user passing a "detached" Vertex (DetachedVertex, StarVertex, etc).
-            if (id instanceof Vertex)
-                return Long.parseLong(((Vertex) id).id().toString());
-            else
-                return Long.parseLong(id.toString());
+            return Long.parseLong(id.toString());
         } catch (NumberFormatException e) {
-            return 0;
+            return null;
         }
     }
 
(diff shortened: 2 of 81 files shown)


li-boxuan:
Just did a force-commit to force benchmark CI to run again. The result [here](https://github.com/JanusGraph/janusgraph/pull/3162#commitcomment-79831336) was too good to be true, and was likely caused by variance in GitHub Actions environment itself.

li-boxuan:
It seems the GitHub CI environment varies a lot these days - thus, the benchmark result is not very reliable. I ran benchmarks locally and seems there's no regression or improvement by this PR.

FYI the benchmark result is: [benchmark.md](https://github.com/JanusGraph/janusgraph/files/9234044/benchmark.md)

analytically:
@li-boxuan can rebase?

janusgraph-test/src/test/java/org/janusgraph/graphdb/query/vertex/VertexIdListTest.java
```diff
@@ -0,0 +1,110 @@
...
+        VertexIdList list = new VertexIdList(tx, Arrays.asList("x1", "x2", "y1"), true);
+        VertexIdList otherList = new VertexIdList(tx, Arrays.asList(2, 4), true);
+        list.addAll(otherList);
+        assertEquals(Arrays.asList(2, 4, "x1", "x2", "y1"), list.getIDs());
+        assertTrue(list.isSorted());
+    }
+
+    @Test
+    public void testMergeUnsortedHeterogeneousIdLists() {
+        VertexIdList list = new VertexIdList(tx, new ArrayList<>(Arrays.asList("x1", "x2", "y1")), true);
```
FlorianHockmann, line 81: Shouldn't this be:
```suggestion
        VertexIdList list = new VertexIdList(tx, new ArrayList<>(Arrays.asList("x1", "x2", "y1")), false);
```

FlorianHockmann (commented):
Sorry for the extremely late review, I had actually started reviewing this PR months ago and also wrote some first inline comments back then, but didn't finish the review and then never really got back to it.
Any way, great contribution. The changes look good to me and I only wrote minor comments, but I'm not sure whether other places also need to be changed. But I guess, if it builds and all tests pass, then it should be OK overall.

janusgraph-test/src/test/java/org/janusgraph/graphdb/query/vertex/VertexIdListTest.java
li-boxuan: Nope. This is to showcase that merging a sorted list with an unsorted list would result in an unsorted list.

FlorianHockmann: Understood, the naming scheme just made me expect this to be basically the same as `testMergeUnsortedLongIdLists`, just with a mixture of strings and longs. You could maybe add this as an additional test specifically for the scenario of merging a sorted with an unsorted list to reduce this confusion, but it's also not really important.

FlorianHockmann (approved):

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
```diff
@@ -55,23 +54,22 @@ public long getTypeId() {
         return typeId;
     }
 
-    public long getOutVertexId() {
+    public Object getOutVertexId() {
         return outVertexId;
     }
 
-    public long getInVertexId() {
-        Preconditions.checkState(inVertexId != 0);
```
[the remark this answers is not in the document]
li-boxuan: We cannot. A bit context here: a `RelationIdentifier` is essentially an id for each edge / property, which consists of `outVertexId`, `typeId`, `relationId` (a unique long value id for this edge/property), and an optional `inVertexId` (only existent for edges but not properties).

Previously, the `inVertexId` is initialized as 0. On [RelationIdentifier.java line 92 before this change](https://github.com/JanusGraph/janusgraph/pull/3162/files#diff-4181397444bf720764051d1d50a66b0846fe2e5c2f0279d147c5ad40c2698ab5L92), it guarantees this `getInVertexId` is only called when `inVertexId != 0`. Now we essentially move this check to the caller side (see IndexRecordUtil.java:266 in the new code, which doesn't really make much of a difference), so we have to remove the `!=  0`, or, in the new codebase, `!= null` check from this function.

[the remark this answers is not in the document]
FlorianHockmann: Thanks for the explanation! I guess it makes sense to just return the `inVertexId` as is. If it's not initialized, e.g., for a property, and therefore will be `null`, then the caller has to decide how to handle that.

li-boxuan:
I'll merge this PR at the end of this week if no one else plans to review it.

porunov (commented):
I see many places where we still use `longId` assuming a vertex has `long` id and not custom id (for example, below in the comment). Will it work normally? How will JanusGraph messaging bus work with non-long vertex ids? I assume it may be a breaking change because logs will now have non-long ids, but I could be wrong.

I also see `IndexRecordUtil.element2String(Object elementId)` which assumes that the id is either `Long` or `RelationIdentifier` and never anything else.

You could potentially also change signature of the next method: https://github.com/JanusGraph/janusgraph/blob/6549949a9c734da19527ce4808e6be51c53f6a68/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/util/StaleIndexRecordUtil.java#L164 but it's not necessary. I can change it later after this PR is merged.

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/computer/VertexMemoryHandler.java
```diff
@@ -154,9 +153,7 @@ public void sendMessage(MessageScope messageScope, M m) {
             vertexMemory.sendMessage(vertexId, m, messageScope);
         } else {
             ((MessageScope.Global) messageScope).vertices().forEach(v -> {
-                long vertexId;
-                if (v instanceof JanusGraphVertex) vertexId=((JanusGraphVertex)v).longId();
-                else vertexId = (Long)v.id();
+                long vertexId = ((Number) v.id()).longValue();
                 vertexMemory.sendMessage(vertexMemory.getCanonicalId(vertexId), m, messageScope);
```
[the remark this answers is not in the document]
li-boxuan: This PR does not intend to bring about any breaking or functional change. For now, we just assume its a long value, but in the subsequent PRs, this will be changed. The message bus will also require breaking changes.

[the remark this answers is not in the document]
li-boxuan: btw

>  You could potentially also change signature of the next method

`forceRemoveVertexFromGraphIndex` is changed in this PR. Maybe you are looking at the wrong version?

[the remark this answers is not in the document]
porunov: You right. I missed that. Thank you!

janusgraph-core/src/main/java/org/janusgraph/util/datastructures/AbstractIdListUtil.java
```diff
@@ -14,60 +14,85 @@
...
+        validateId(id2);
+        if (id1 instanceof Number && id2 instanceof String) return -1;
+        if (id1 instanceof String && id2 instanceof Number) return 1;
+        if (id1 instanceof String) {
+            assert id2 instanceof String;
+            return ((String) id1).compareTo((String) id2);
+        } else {
+            assert id1 instanceof Number;
+            assert id2 instanceof Number;
+            return Double.compare(((Number) id1).doubleValue(), ((Number) id2).doubleValue());
```
porunov, line 48: Why `Double` and not `Long`? I don't remember but aren't we loosing precision after 2^53 here?

janusgraph-driver/src/main/java/org/janusgraph/util/IDUtils.java
```diff
@@ -0,0 +1,34 @@
...
+        if (vertexId == null) {
+            throw new IllegalArgumentException("vertex id cannot be null");
+        }
+        if (vertexId instanceof String) {
+            // string-type vertex id is allowed
+            return;
+        }
+        if (vertexId instanceof Number) {
+            if (((Number) vertexId).longValue() <= 0) {
+                throw new IllegalArgumentException(String.format("vertex id %d is non-positive", ((Number) vertexId).longValue()));
```
porunov, line 28: Not related to this PR but do you know by any chance why negative vertex ids are not allowed?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/ElementUtils.java
```diff
@@ -16,32 +16,26 @@
...
             return (Long) id;
         if (id instanceof Number)
             return ((Number) id).longValue();
-
         try {
-            // handles the case of a user passing a "detached" Vertex (DetachedVertex, StarVertex, etc).
-            if (id instanceof Vertex)
-                return Long.parseLong(((Vertex) id).id().toString());
-            else
-                return Long.parseLong(id.toString());
```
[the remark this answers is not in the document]
li-boxuan: I combined `id instanceof Vertex` case with `id instanceof JanusGraphVertex`. See line 29 of new version.

janusgraph-driver/src/main/java/org/janusgraph/util/IDUtils.java
li-boxuan: Iff `ids.flush=false`, then JanusGraph will assign a temporary negative ID when an element is created and generates a unique positive ID only when the transaction commits. Thus, a negative number represents a temporary id.

porunov: Thank you, got it! I guess it would make sense to think about reducing that scope (I guess `2^31` would be enough for temporary ids and the rest could be potentially used for normal ids) but it's definitely out of the scope of this PR.

janusgraph-core/src/main/java/org/janusgraph/util/datastructures/AbstractIdListUtil.java
li-boxuan: Great catch! And I shouldn't have allowed double value in the first place.

porunov (approved):
LGTM! Thank you @li-boxuan !

m-thirumal:
This  feature will be available for release 1.0.0? Because this is not available in Release candidate `janusgraph-1.0.0-rc1`. 

Still getting the same error

```
gremlin> g.addV('user').property(T.id, 'f3695384-8442-447e-ba25-914e5b8deb08').property('name', 'Thirumal')
Cannot cast java.lang.String to java.lang.Number
```

li-boxuan:
@m-thirumal This feature is WIP. I have the code ready but need some time to clean it up and let the community review it. It may or may not be released in 1.0.0.

PR #3166, opened 2022-08-09
Fix mixed index query for neq condition

### LINKED ISSUE #3164: Incorrect result for query that filters on unset property when an index is present
- Version: 0.6.2
- Storage Backend: inmemory
- Mixed Index Backend: lucene
- Link to discussed bug: https://lists.lfaidata.foundation/g/janusgraph-users/topic/92645197#6578
- Expected Behavior:
  I would expect the result to stay the same even when an index is present.
- Current Behavior:
  The result is different when an index is present.
- Steps to Reproduce:
```java
PropertiesConfiguration conf = ConfigurationUtil.loadPropertiesConfig("conf/test.properties");
JanusGraph graph = JanusGraphFactory.open(conf);

JanusGraphManagement m = graph.openManagement();

VertexLabel l = m.makeVertexLabel("L").make();
PropertyKey p = m.makePropertyKey("p").dataType(Short.class).make();
PropertyKey q = m.makePropertyKey("q").dataType(UUID.class).make();
m.buildIndex("someName", Vertex.class).addKey(p).addKey(q).indexOnly(l).buildMixedIndex("search");
m.commit();

g.addV("L").property("p", (short) 1).next();
g.tx().commit();

System.out.println(g.V().hasLabel("L").has("q").count().next()); // 0
System.out.println(g.V().hasLabel("L").has("q", not(eq(UUID.randomUUID()))).count().next()); // 1
```
With the following configuration:
```
gremlin.graph=org.janusgraph.core.JanusGraphFactory
storage.backend=inmemory
index.search.backend=lucene
index.search.directory=data/searchindex
schema.default=none
```
According to Marc this bug also happens with Cassandra and Elasticsearch. Furthermore, the bug doesn't seem to be related to the `UUID` type of property `q`.

li-boxuan:
If a mixed index consists of multiple properties, then an element can be
indexed as long as it contains one of these properties. In this case, a
query like `has(keyA, neq(val))` could match an element without `keyA`
as long as it is indexed by this mixed index. This commit fixes this bug
by adding an existence condition, i.e. matching an element only if it
has the given key present.

Fixes #3164

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -1487,6 +1487,16 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         mgmt.addIndexKey(pindex, text, getTextMapping());
         mgmt.makeEdgeLabel("knows").signature(name).make();
         mgmt.makePropertyKey("uid").dataType(String.class).signature(text).make();
+        final PropertyKey number = makeKey("number", Integer.class);
+        final PropertyKey flag = makeKey("flag", Boolean.class);
+        final PropertyKey date = makeKey("date", Date.class);
+        final PropertyKey geo = makeKey("geo", Geoshape.class);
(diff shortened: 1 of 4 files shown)


cdegroc (approved):
LGTM

Issue #3185, opened 2022-08-27
Performance issue on vertex cache with Guava implementation


ministat:
JanusGraph's GuavaVertexCache will cause memory leak in 100% read and heavy read scenario, especially used in OLAP cases. The Guava cache should be replaced by Caffiene cache to fix this issue according to Guava's community feedback.
See https://github.com/google/guava/issues/2408 and https://github.com/google/guava/issues/2063.

In my local JanusGraph, after I replaced Guava cache with Caffeine cache, the OLAP case, for example, g.V().count() takes 18 minutes to finish the scan of a graph with 500 million vertex and 6 billion edges. Before this fix, it takes 45 minutes. So, the fix gets ~3X benefit.

For confirmed bugs, please report:

- Version: All versions before 0.6.2, including 0.6.2
- Storage Backend: hbase, inmemory, cql
- Mixed Index Backend: 
- Link to discussed bug: https://github.com/google/guava/issues/2408, https://github.com/google/guava/issues/2063
- Expected Behavior: vertex cache should not be the top 1 memory leak suspect
- Current Behavior: Guava vertex cache caused leak suspects from JVM heap dump
- Steps to Reproduce: develop a java app which initialize GuavaVertexCache with 30 element size, and launch multiple threads to call get() in a loop. The loop should be, for example, 1,000,000. Use 

### Stack Trace (if you have one)

```
paste stack trace here
```

li-boxuan:
Thanks for reporting @ministat! Seems a duplicate of https://github.com/JanusGraph/janusgraph/issues/871 (which was created by the author of Guava, interesting!). Usually, we close newer issues if they are duplicates, but your discovery and experiments are very valuable so I'd like to keep both issues open. Would you like to contribute your implementation back to the community by submitting a PR? That would be very helpful!

ministat:
Thanks for your quick response.

I don't know how to pass CLA check, and created duplicate PR by mistake.

What I have found is replacing Guava cache with Caffeine cache gives big benefit. In my OLAP test, I got 3X improvement.


发送自 Outlook<http://aka.ms/weboutlook>

________________________________
发件人: Boxuan Li ***@***.***>
发送时间: 2022年8月28日 7:54
收件人: JanusGraph/janusgraph ***@***.***>
抄送: Hongjiang Zhang ***@***.***>; Mention ***@***.***>
主题: Re: [JanusGraph/janusgraph] Performance issue on vertex cache with Guava implementation (Issue #3185)


[...]

Issue #871, opened 2018-01-06
TinyLFU cache (vs LRU)


ben-manes:
There are a number of LRU caches, such as Guava's and the custom `ConcurrentLRUCache`. Guava's amortizes the locking required by LRU, whereas `ConcurrentLRUCache` uses a dedicated thread per instance to perform a full sweep after a threshold size is crossed. Both use LRU, which is a simple algorithm but underperforms in db / search / analytical workloads due to scans and one-hit wonders polluting the cache. In most cases frequency is a better indicator, due to caches following a Zipf-like distribution, but traditionally was not O(1) and difficult to implement well.

TinyLFU is a modern policy that has [near optimal hit rates](https://github.com/ben-manes/caffeine/wiki/Efficiency) while also being O(1) and has a low memory footprint. [Caffeine](https://github.com/ben-manes/caffeine) is an implementation that has [superior performance](https://github.com/ben-manes/caffeine/wiki/Benchmarks) compared to Guava and scales linearly with the cpu count. It implements Guava's API, is feature compatibility, and is used by other databases like Cassandra, Druid, and Neo4J. It is also recommended by the Guava team when closing their issues as won't fix, probably because I coauthored Guava's cache with them.

I think the eviction policy and concurrency would be a good fit for Janus, as well be trivial to migrate to. We can easily run a trace against a production workload if you'd like to see the hit rate curve.

robertdale:
Yes, if you can provide a test plan and the before/after metrics, that would be great.

ben-manes:
Since I do not have a production deployment, you're best bet is to add tracing to the most critical cache (e.g. a node/vertex cache). This could simply be a logger that appends to its own file, capturing the key's hash code, and canary'd for an hour or so. We can then run it through a simulator to see the hit rate curve across many sizes and policies. That would have minimal impact and provide useful data to understand your access patterns.

PR #3188, opened 2022-08-28
Replace Guava cache with Caffeine cache for vertex cache

ministat:
Fix the issue #3185

Guava cache has performance bug in 100% read scenario. The solution is to use Caffeine cache which is also suggested by Guava community. See google/guava#2408 for more details. Meanwhile, Caffeine cache beats Guava cache in all scenarios.

The following result is for read scenario.

Benchmark (cacheType) Mode Cnt Score Error Units
VertexCacheBenchmark.run guava avgt 5 0.001 ± 0.001 ms/op
VertexCacheBenchmark.run caffeine avgt 5 ≈ 10⁻⁴ ms/op

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/pom.xml
+++ b/janusgraph-core/pom.xml
@@ -81,6 +81,16 @@
             <artifactId>jsr305</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.checkerframework</groupId>
+                    <artifactId>checker-qual</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
     </dependencies>
     <build>
         <directory>${basedir}/target</directory>
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -111,8 +111,8 @@ import org.janusgraph.graphdb.transaction.lock.TransactionLock;
 import org.janusgraph.graphdb.transaction.subquerycache.EmptySubqueryCache;
 import org.janusgraph.graphdb.transaction.subquerycache.GuavaSubqueryCache;
 import org.janusgraph.graphdb.transaction.subquerycache.SubqueryCache;
+import org.janusgraph.graphdb.transaction.vertexcache.CaffeineVertexCache;
 import org.janusgraph.graphdb.transaction.vertexcache.EmptyVertexCache;
-import org.janusgraph.graphdb.transaction.vertexcache.GuavaVertexCache;
 import org.janusgraph.graphdb.transaction.vertexcache.VertexCache;
 import org.janusgraph.graphdb.types.CompositeIndexType;
 import org.janusgraph.graphdb.types.IndexType;
@@ -319,7 +319,7 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
                     config.getVertexCacheSize(), effectiveVertexCacheSize, MIN_VERTEX_CACHE_SIZE);
         }
 
-        vertexCache = new GuavaVertexCache(effectiveVertexCacheSize,concurrencyLevel,config.getDirtyVertexSize());
+        vertexCache = new CaffeineVertexCache(effectiveVertexCacheSize,config.getDirtyVertexSize());
 
         indexCache = new GuavaSubqueryCache(concurrencyLevel, config.getIndexCacheWeight());
 
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
@@ -1,0 +1,137 @@
+// Copyright 2022 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.transaction.vertexcache;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import com.github.benmanes.caffeine.cache.RemovalListener;
+import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import com.google.common.base.Preconditions;
+import org.jctools.maps.NonBlockingHashMapLong;
+import org.janusgraph.graphdb.internal.InternalVertex;
+import org.janusgraph.graphdb.vertices.AbstractVertex;
+import org.janusgraph.util.datastructures.Retriever;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentMap;
+
+public class CaffeineVertexCache implements VertexCache {
+    private static final Logger log =
+        LoggerFactory.getLogger(CaffeineVertexCache.class);
+
+    private final ConcurrentMap<Long, InternalVertex> volatileVertices;
+    private final Cache<Long, InternalVertex> cache;
+
+    private long createdTime;
+
+    public CaffeineVertexCache(final long maxCacheSize, final int initialDirtySize) {
+        volatileVertices = new NonBlockingHashMapLong<>(initialDirtySize);
+        log.debug("Created dirty vertex map with initial size {}", initialDirtySize);
+
+        Caffeine<Long, InternalVertex> cacheBuilder = Caffeine.newBuilder().maximumSize(maxCacheSize)
+            .removalListener(new CaffeineRemovalListener())
+            .executor(Runnable::run); // according to the https://github.com/ben-manes/caffeine/discussions/757
+        if (log.isDebugEnabled()) {
+            cacheBuilder = cacheBuilder.recordStats();
+        }
+        cache = cacheBuilder.build();
+        log.debug("Created vertex cache with max size {}", maxCacheSize);
+        createdTime = System.currentTimeMillis();
+    }
+
+    @Override
+    public boolean contains(long id) {
+        Long vertexId = id;
+        return cache.getIfPresent(vertexId) != null || volatileVertices.containsKey(vertexId);
+    }
+
+    @Override
+    public InternalVertex get(long id, Retriever<Long, InternalVertex> retriever) {
+        final Long vertexId = id;
+
+        InternalVertex vertex = cache.getIfPresent(vertexId);
+
+        if (vertex == null) {
+            InternalVertex newVertex = volatileVertices.get(vertexId);
+
+            if (newVertex == null) {
+                newVertex = retriever.get(vertexId);
+            }
+            assert newVertex != null;
+            final InternalVertex v = newVertex;
+            try {
+                vertex = cache.get(vertexId, (k) -> v);
+            } catch (Exception e) { throw new AssertionError("Should not happen: "+e.getMessage()); }
+            assert vertex!=null;
+        }
+
+        return vertex;
+    }
+
+    @Override
+    public void add(InternalVertex vertex, long id) {
+        Preconditions.checkNotNull(vertex);
+        Preconditions.checkArgument(id != 0);
+        Long vertexId = id;
+
+        cache.put(vertexId, vertex);
+        if (vertex.isNew() || vertex.hasAddedRelations())
+            volatileVertices.put(vertexId, vertex);
+    }
+
+    @Override
+    public List<InternalVertex> getAllNew() {
+        final List<InternalVertex> vertices = new ArrayList<>(10);
+        for (InternalVertex v : volatileVertices.values()) {
+            if (v.isNew()) vertices.add(v);
+        }
+        return vertices;
+    }
+
+    @Override
+    public synchronized void close() {
+        if (log.isDebugEnabled()) {
+            long end = System.currentTimeMillis();
+            CacheStats stats = cache.stats();
+            log.debug("Caffeine cache (lifespan: {}ms) stats: {}", end - createdTime, stats);
+        }
+        volatileVertices.clear();
+        cache.invalidateAll();
+        cache.cleanUp();
+    }
+
+    class CaffeineRemovalListener implements RemovalListener<Long, InternalVertex> {
+
+        @Override
+        public void onRemoval(@Nullable Long key,
+                              @Nullable InternalVertex internalVertex,
+                              @Nonnull RemovalCause removalCause) {
+            if (removalCause == RemovalCause.EXPLICIT) {
+                assert volatileVertices.isEmpty();
+                return;
+            }
+            assert (removalCause == RemovalCause.SIZE || removalCause == RemovalCause.REPLACED) : "Cause: " + removalCause;
+            InternalVertex v = internalVertex;
+            if (((AbstractVertex) v).isTxOpen() && (v.isModified() || v.isRemoved())) {
+                volatileVertices.putIfAbsent(key, v);
+            }
+        }
+    }
+}
--- a/janusgraph-all/pom.xml
+++ b/janusgraph-all/pom.xml
@@ -30,6 +30,12 @@
             <groupId>org.janusgraph</groupId>
             <artifactId>janusgraph-grpc</artifactId>
             <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.checkerframework</groupId>
+                    <artifactId>checker-qual</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.janusgraph</groupId>
--- a/janusgraph-benchmark/pom.xml
+++ b/janusgraph-benchmark/pom.xml
@@ -76,7 +76,11 @@
             <artifactId>commons-exec</artifactId>
             <version>1.3</version>
         </dependency>
-
+        <dependency>
+            <groupId>org.easymock</groupId>
+            <artifactId>easymock</artifactId>
+            <version>${easymock.version}</version>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/janusgraph-benchmark/src/main/java/org/janusgraph/VertexCacheBenchmark.java
+++ b/janusgraph-benchmark/src/main/java/org/janusgraph/VertexCacheBenchmark.java
@@ -1,0 +1,171 @@
+// Copyright 2022 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph;
+
+import org.easymock.EasyMock;
+import org.easymock.EasyMockSupport;
+import org.janusgraph.core.PropertyKey;
+import org.janusgraph.core.RelationType;
+import org.janusgraph.core.schema.DefaultSchemaMaker;
+import org.janusgraph.core.schema.PropertyKeyMaker;
+import org.janusgraph.diskstorage.util.time.TimestampProvider;
+import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
+import org.janusgraph.graphdb.database.EdgeSerializer;
+import org.janusgraph.graphdb.database.IndexSerializer;
+import org.janusgraph.graphdb.database.StandardJanusGraph;
+import org.janusgraph.graphdb.database.serialize.Serializer;
+import org.janusgraph.graphdb.idmanagement.IDManager;
+import org.janusgraph.graphdb.internal.ElementLifeCycle;
+import org.janusgraph.graphdb.internal.InternalVertex;
+import org.janusgraph.graphdb.query.index.IndexSelectionStrategy;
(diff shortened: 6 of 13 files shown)


porunov:
Thank you for the contribution @ministat , @clovertrail !
Replacing Guava cache with Caffeine cache definitely makes sense to me.
That said, to accept your contribution you need to sign CLA to give JanusGraph permission to use the contribution you created.
Would you be able to sign EasyCLA?
Signing a CLA might be a problem if you signed CLA for one GitHub account but made a commit using information from another GitHub account.
In your case I see that the commit is made by @clovertrail but the PR is opened by @ministat .
In this case you need to use `clovertrail` as a GitHub account when signing EasyCLA.
If you intended to use `ministat` account then you will need to change the author of the commit to use name and email of `ministat` GitHub account.
Below I will show possible solutions on how to correctly sign the commit in your case.
1) Solution 1 - the case when you want to use `ministat` account only. In this case we need to update commit with the correct author information and sign the commit (by using `-s` parameter).
```
git config user.name "Hongjiang Zhang"
git config user.email "your.email@example.org"
git commit --amend -s --author="Hongjiang Zhang <your.email@example.org>"
git push -f
```
After that login into GitHub with `ministat` username. And click the next link to sign CLA:
https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/1543822/77385607/3188/#/?version=2

2) Solution 2 - the case when you want to use `clovertrail` account only.
```
git config user.name "Your Full Name"
git config user.email "email.of.clovertrail@example.org"
git commit --amend -s
git push -f
```
After that login into GitHub with `clovertrail` username. And click the next link to sign CLA:
https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/1543822/77385607/3188/#/?version=2

Please, let me know if you have any problems with signing EasyCLA, and I will help you with signing it.

ministat:
@porunov Thanks for your suggestion. I have passed the CLA checked. Appreciated!

porunov (commented):
Checking the dependency tree I see that we get Caffeine 2.3.1 from TinkerPop dependency (`org.apache.tinkerpop:gremlin-groovy:3.5.4`).
I'm just thinking if we should use the latest Caffeine 3.1.1 or should we stick to the one provided by TinkerPop. Do you know if there were any breaking changes? If no, then I would prefer trying out the latest Caffeine dependency 3.1.1.

ben-manes:
@porunov The main reason for the major version bump was that v3.x requires Java 11 or higher, whereas previously it was Java 8. This allowed the implementation code to switch from sun.misc.Unsafe to VarHandles (fully removed in 3.0.2). Otherwise the changes were minor clean ups that had little impact on users (see [release notes](https://github.com/ben-manes/caffeine/releases/tag/v3.0.0)).

porunov:
> @porunov The main reason for the major version bump was that v3.x requires Java 11 or higher, whereas previously it was Java 8. This allowed the implementation code to switch from sun.misc.Unsafe to VarHandles (fully removed in 3.0.2). Otherwise the changes were minor clean ups that had little impact on users (see [release notes](https://github.com/ben-manes/caffeine/releases/tag/v3.0.0)).

Thank you for the clarification! JanusGraph supports Java 11 but we still write everything in Java 8 to support clients on Java 8, so I think in this case it should be safe to use Caffeine 2.9.3.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    public CaffeineVertexCache(final long maxCacheSize, final int initialDirtySize) {
+        volatileVertices = new NonBlockingHashMapLong<>(initialDirtySize);
+        log.debug("Created dirty vertex map with initial size {}", initialDirtySize);
+
+        cache = Caffeine.newBuilder().maximumSize(maxCacheSize)
+            .removalListener(new CaffeineRemovalListener())
+            .recordStats()
```
porunov, line 68: I see that cache statistics is always enabled here, even so many applications won't use it. It's definitely helpful for analyses but many times it won't be needed. Moreover, I don't know if there is any memory / performance overhead with having statistics enabled. 
What I noticed is that you report statistics only when closing cache and only when debug is enabled. Thus, I would recommend here to enable `.recordStats()` only when debug log level is enabled. I.e. something like this:
```
Caffeine<Long, InternalVertex> cacheBuilder = Caffeine.newBuilder().maximumSize(maxCacheSize)
    .removalListener(new CaffeineRemovalListener())
    .executor(Runnable::run); // according to the https://github.com/ben-manes/caffeine/discussions/757
if(log.isDebugEnabled()){
    cacheBuilder = cacheBuilder.recordStats();
}
cache = cacheBuilder.build();
```
Another option (maybe even better but not necessary to do this in this PR) is to add a new configuration option to JanusGraph to enable / disable cache stats recording.

porunov (commented):
The PR looks good but I have small nitpicks below.

Could you please add explicit Caffeine dependency to root `pom.xml` like the next:
```
<dependency>
    <groupId>com.github.ben-manes.caffeine</groupId>
    <artifactId>caffeine</artifactId>
    <version>${caffeine.version}</version>
</dependency>
```
Also, specify `<caffeine.version>2.9.3</caffeine.version>`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
porunov: Also, in case you do this with debug enabled / disabled then make sure in `close` method there is no NPE thrown by any chance (I just didn't check if that is even possible). But for additional protection you can wrap that code in additional if.
I.e.:
```
if(log.isDebugEnabled()){
    long end = System.currentTimeMillis();
    CacheStats stats = cache.stats();
    log.debug("Caffeine cache (lifespan: {}ms) stats: {}", end - createdTime, stats);  
}
```

ben-manes: I haven't benchmarked this for a while, but have recently been advising a developer writing a similar cache for .NET. He recently ported Java's `LongAdder` to replace an equivalent of `AtomicLong` for statistics, since I mentioned that was a likely bottleneck. You can see how there is a degradation but the penalty is not noticeable in an application as the performance is very high. (https://github.com/bitfaster/BitFaster.Caching/pull/219)

porunov:
@ministat ignore failed Hbase and dist tests. Looks like there is a problem with getting docker image for some reason but it's unrelated to your changes

ministat:
Add explicit Caffeine dependency, and excludes the lower version of org.checkerframework:checker-qual from guava, otherwise, there are many dependency convergence errors like the following:
```
Dependency convergence error for org.checkerframework:checker-qual:jar:3.12.0:compile paths to dependency are:
+-org.janusgraph:janusgraph-dist:pom:1.0.0-SNAPSHOT
  +-org.apache.tinkerpop:hadoop-gremlin:jar:3.5.4:compile
    +-com.google.guava:guava:jar:31.1-jre:compile
      +-org.checkerframework:checker-qual:jar:3.12.0:compile
and
+-org.janusgraph:janusgraph-dist:pom:1.0.0-SNAPSHOT
  +-org.apache.tinkerpop:gremlin-groovy:jar:3.5.4:compile
    +-com.github.ben-manes.caffeine:caffeine:jar:2.9.3:compile
      +-org.checkerframework:checker-qual:jar:3.19.0:compile

[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for org.checkerframework:checker-qual:3.12.0 paths to dependency are:
+-org.janusgraph:janusgraph-dist:1.0.0-SNAPSHOT
  +-org.apache.tinkerpop:hadoop-gremlin:3.5.4
    +-com.google.guava:guava:31.1-jre (managed) <-- com.google.guava:guava:16.0.1
      +-org.checkerframework:checker-qual:3.12.0
and
+-org.janusgraph:janusgraph-dist:1.0.0-SNAPSHOT
  +-org.apache.tinkerpop:gremlin-groovy:3.5.4
    +-com.github.ben-manes.caffeine:caffeine:2.9.3 (managed) <-- com.github.ben-manes.caffeine:caffeine:2.3.1
      +-org.checkerframework:checker-qual:3.19.0
```

ministat:
@porunov Thanks for you comments, I have updated my patch.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+
+    public CaffeineVertexCache(final long maxCacheSize, final int initialDirtySize) {
+        volatileVertices = new NonBlockingHashMapLong<>(initialDirtySize);
+        log.debug("Created dirty vertex map with initial size {}", initialDirtySize);
+
+        Caffeine<Long, InternalVertex> cacheBuilder = Caffeine.newBuilder().maximumSize(maxCacheSize)
+            .removalListener(new CaffeineRemovalListener())
+            .recordStats()
```
porunov, line 68: Looks like you forgot to remove `.recordStats()` here. I.e. you are enabling it 2 times, here and below.

ministat: My fault!

ministat: My fault!

ministat: Fixed

porunov:
@ministat I fixed HBase tests issue in the latest commit in `master` and `v0.6` branches. Would you be able to rebase your branch against the latest `master` commit please to ensure that all tests pass?
I.e.
```
git remote add upstream https://github.com/JanusGraph/janusgraph
git fetch upstream master
git rebase upstream/master
git push -f
```

porunov:
@ministat it also looks like the benchmark class is often stuck in the first or second iteration for long time. I'm not sure but it could be flaky. I was waiting for 10 minutes for the second iteration to pass but it didn't finish. Sometimes it hangs on the warmup iteration. Do you know what is a correct way to run the benchmark you included?

porunov (approved):
LGTM. Thank you @ministat !
The PR looks good to me but I still suggest to rebase it against the latest `master` commit just in case (described [here](https://github.com/JanusGraph/janusgraph/pull/3188#issuecomment-1231447860)).

ministat:
@porunov, I have rebased per your suggestion.
For the benchmark, I run it through idea firstly. In order to run it in cmd line, I copied the idea's java classpath when it was running, pasted the long classpath into a file and run it again. I got the same perf results. I did not see hang in my test.

porunov:
> @porunov, I have rebased per your suggestion. For the benchmark, I run it through idea firstly. In order to run it in cmd line, I copied the idea's java classpath when it was running, pasted the long classpath into a file and run it again. I got the same perf results. I did not see hang in my test.

Thanks! Now to merge this PR we need to either wait for another commiter to approve this PR or wait 7 days to merge this PR with lazy consensus.

ministat:
@porunov I found the build "mvn clean package" failed in any janusgraph-xxx folder, for example, janusgraph-benchmark.
Before my CI, the build is successful. I'm still investigating. Do you have any suggestion?
```
~/Work/OpenSource/ministat_janusgraph/janusgraph-core @ LM-SHC-16507917 (hongjizhang) 
| => mvn clean package
[INFO] Scanning for projects...
[INFO] 
[INFO] -------------------< org.janusgraph:janusgraph-core >-------------------
[INFO] Building JanusGraph-Core: Core Library for JanusGraph 1.0.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:3.2.0:clean (default-clean) @ janusgraph-core ---
[INFO] Deleting /Users/hongjizhang/Work/OpenSource/ministat_janusgraph/janusgraph-core/target
[INFO] 
[INFO] --- maven-enforcer-plugin:3.0.0:enforce (enforce-dependency-convergence) @ janusgraph-core ---
[WARNING] 
Dependency convergence error for org.checkerframework:checker-qual:jar:3.12.0:compile paths to dependency are:
+-org.janusgraph:janusgraph-core:jar:1.0.0-SNAPSHOT
  +-org.janusgraph:janusgraph-driver:jar:1.0.0-SNAPSHOT:compile
    +-com.google.guava:guava:jar:31.1-jre:compile
      +-org.checkerframework:checker-qual:jar:3.12.0:compile
and
+-org.janusgraph:janusgraph-core:jar:1.0.0-SNAPSHOT
  +-com.github.ben-manes.caffeine:caffeine:jar:2.9.3:compile
    +-org.checkerframework:checker-qual:jar:3.19.0:compile

[WARNING] Rule 1: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability. See above detailed error message.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.278 s
[INFO] Finished at: 2022-08-30T23:13:27+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0:enforce (enforce-dependency-convergence) on project janusgraph-core: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

ministat:
After I "mvn install -DskipTests" from the root, the "mvn clean package" from janusgraph-core is successful. So, I think the previous failure is caused by referencing the wrong library.

li-boxuan:
>  I would suggest excluding this dependency from guava which is in janusgraph-driver and adding checker-qual:3.19.0 explicitly to the pom.xml

I think we should add `checker-qual:3.19.0` to the `<dependencyManagement>` section in root `pom.xml`. We don't want to introduce this dependency explicitly since we are not using it.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -319,7 +319,7 @@ public void close() {
                     config.getVertexCacheSize(), effectiveVertexCacheSize, MIN_VERTEX_CACHE_SIZE);
         }
 
-        vertexCache = new GuavaVertexCache(effectiveVertexCacheSize,concurrencyLevel,config.getDirtyVertexSize());
+        vertexCache = new CaffeineVertexCache(effectiveVertexCacheSize,config.getDirtyVertexSize());
```
li-boxuan, line 322: Can you also remove `GuavaVertexCache` class since it's not being used anymore?

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.concurrent.ConcurrentMap;
+
+public class CaffeineVertexCache implements VertexCache {
+    private static final Logger log =
+        LoggerFactory.getLogger(CaffeineVertexCache.class);
+
+    private final ConcurrentMap<Long, InternalVertex> volatileVertices;
+    private final Cache<Long, InternalVertex> cache;
+
+    private long createdTime; // the timestamp when the cache was created.
```
li-boxuan, line 42: (nitpick): I would remove this comment since your code perfectly explains itself.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return vertices;
+    }
+
+    @Override
+    public synchronized void close() {
+        if (log.isDebugEnabled()) {
+            long end = System.currentTimeMillis();
+            CacheStats stats = cache.stats();
+            log.debug("Caffeine cache (lifespan: {}ms) stats: {}", end - createdTime, stats.toString());
```
li-boxuan, line 131: (nitpick) Redundant `toString()` call.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
```diff
@@ -0,0 +1,137 @@
...
+public class CaffeineVertexCache implements VertexCache {
+    private static final Logger log =
+        LoggerFactory.getLogger(CaffeineVertexCache.class);
+
+    private final ConcurrentMap<Long, InternalVertex> volatileVertices;
+    private final Cache<Long, InternalVertex> cache;
+
+    private long createdTime; // the timestamp when the cache was created.
+
+    class CaffeineRemovalListener implements RemovalListener<Long, InternalVertex> {
```
li-boxuan, line 44: (nitpick) It is a common practice in JanusGraph to put private classes at the bottom of the file.

li-boxuan (commented):
Thanks for your contribution! I just have some nitpicks:

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
porunov: As far as I understand it is used in the benchmark test class. I guess we could move that class into testing module then

porunov: I guess we can remove that class with the benchmark test class in the follow up PRs but I'm OK either way.

ben-manes: Should the `GuavaVertexCache` be moved there rather than have it linger around as dead code?

porunov: Actually I think it's better to leave the benchmark test class in this PR for historical reasons. We can move `GuavaVertexCache` to test module and then later after the full transition to Caffeine cache is done we can remove those test classes with the benchmark classes. It's just good if those benchmark classes are in git history

li-boxuan: >  later after the full transition to Caffeine cache is done we can remove those test classes with the benchmark classes

That sounds good to me!

ministat: I agree to leave GuavaVertexCache there. I'm ok to compose several following PRs to clean it up.
Meanwhile, I found there are some other Guava cache, like https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/subquerycache/GuavaSubqueryCache.java, and it is supposed to be replaced by Caffeine cache also, but I have not yet benchmark it.

porunov: > I agree to leave GuavaVertexCache there. I'm ok to compose several following PRs to clean it up. Meanwhile, I found there are some other Guava cache, like https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/subquerycache/GuavaSubqueryCache.java, and it is supposed to be replaced by Caffeine cache also, but I have not yet benchmark it.

There are 8 caches in JanusGraph actually and all of them need to be replaced by Caffeine (not in the current PR but in follow up PRs). I posted those places here: https://github.com/JanusGraph/janusgraph/issues/871#issuecomment-1230207577

janusgraph-core/pom.xml
```diff
@@ -81,6 +81,17 @@
             <artifactId>jsr305</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>${caffeine.version}</version>
```
li-boxuan, line 87: (nitpick) You don't need this line because you have already specified the version in the `<dependencyManagement>` section of root `pom.xml`.

li-boxuan (approved):
Neat! I am okay with the failing codecov.

janusgraph-core/pom.xml
porunov: Right. I.e. this is enough when the version has been already specified in `dependencyManagement`:
```
<dependency>
            <groupId>com.github.ben-manes.caffeine</groupId>
            <artifactId>caffeine</artifactId>
</dependency>
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java
```diff
@@ -0,0 +1,137 @@
...
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentMap;
+
+public class CaffeineVertexCache implements VertexCache {
+    private static final Logger log =
+        LoggerFactory.getLogger(CaffeineVertexCache.class);
+
+    private final ConcurrentMap<Long, InternalVertex> volatileVertices;
```
mad, line 39: Caffeine has special feature for avoid this extra field - https://github.com/ben-manes/caffeine/issues/264

Implementation could be looks like https://gist.github.com/mad/9119f48812000a03dea16c7485c621fc

@ministat @porunov

porunov: > Caffeine has special feature for avoid this extra field - [ben-manes/caffeine#264](https://github.com/ben-manes/caffeine/issues/264)
> 
> Implementation could be looks like https://gist.github.com/mad/9119f48812000a03dea16c7485c621fc
> 
> @ministat @porunov

That's interesting! I would prefer refactoring this class to eliminate this extra field as you proposed. That said, as it is a performance related part of the code we should probably execute benhmarking for any improvement we make here, just to make sure we don't have a regression. Other than that your proposal is much cleaner, so that's great.
Should we open a separate ticket to track this?

porunov: Also, a related transitioning PR which replaces other Guava caches with Caffeine cache is: #3198 
After that transitioning is merged I believe we need to refactor [CaffeineVertexCache.java](https://github.com/JanusGraph/janusgraph/blob/c30c7071227de338a63d82336b34f749f2384b22/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java) and [ExpirationKCVSCache.java](https://github.com/JanusGraph/janusgraph/blob/0822cce3d7a50cdb530d13a363f076b18f9b14be/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/cache/ExpirationKCVSCache.java).

PR #3189, opened 2022-08-29
Add clear way to disable db-cache per transaction

porunov:
Not sure if we necessary need this or not but thought this could be helpful for someone who needs to enable / disable db-cache access per transaction.
Another way to control db-cache access would be to control that option in the current running transaction which could allow enabling / disabling db-cache access even per query if necessary:
```
StandardJanusGraphTx tx = (StandardJanusGraphTx) graph.newTransaction();
BackendTransaction backendTransaction = tx.getTxHandle();
backendTransaction.disableCache();
tx.traversal().V().has("name", "hello").next();
backendTransaction.enableCache();
tx.traversal().V().has("name", "world").next();
```

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/interactions/transactions.md
+++ b/docs/interactions/transactions.md
@@ -322,6 +322,10 @@ following aspects of a transaction to be configured:
     confirmation at the application level to avoid inconsistencies. USE
     WITH GREAT CARE!
 
+-   `skipDBCacheRead()` - Disables JanusGraph database level 
+    cache access during read operations. Doesn't have any effect if database 
+    level cache was disabled via config `cache.db-cache`.
+
 Once, the desired configuration options have been specified, the new
 transaction is started via `start()` which returns a
 `JanusGraphTransaction`.
--- a/janusgraph-core/src/main/java/org/janusgraph/core/TransactionBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/core/TransactionBuilder.java
@@ -138,6 +138,14 @@ public interface TransactionBuilder {
      */
     TransactionBuilder commitTime(Instant instant);
 
+    /**
+     * Skips usage of JanusGraph database level cache during read operations.
+     * <p>
+     * Doesn't have any effect if database level cache was disabled via config `cache.db-cache`.
+     *
+     * @return Object with the skip db-cache reads check settings
+     */
+    TransactionBuilder skipDBCacheRead();
 
     /**
      * Sets the group name for this transaction which provides a way for gathering
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java
@@ -80,12 +80,13 @@ public class BackendTransaction implements LoggableTransaction {
     private final Map<String, IndexTransaction> indexTx;
 
     private boolean acquiredLock = false;
-    private boolean cacheEnabled = true;
+    private boolean cacheEnabled;
 
     public BackendTransaction(CacheTransaction storeTx, BaseTransactionConfig txConfig,
                               StoreFeatures features, KCVSCache edgeStore, KCVSCache indexStore,
                               KCVSCache txLogStore, Duration maxReadTime,
-                              Map<String, IndexTransaction> indexTx, Executor threadPool) {
+                              Map<String, IndexTransaction> indexTx, Executor threadPool,
+                              boolean cacheEnabled) {
         this.storeTx = storeTx;
         this.txConfig = txConfig;
         this.storeFeatures = features;
@@ -95,6 +96,7 @@ public class BackendTransaction implements LoggableTransaction {
         this.maxReadTime = maxReadTime;
         this.indexTx = indexTx;
         this.threadPool = threadPool;
+        this.cacheEnabled = cacheEnabled;
     }
 
     public boolean hasAcquiredLock() {
@@ -130,6 +132,10 @@ public class BackendTransaction implements LoggableTransaction {
 
     public void enableCache() {
         this.cacheEnabled = true;
+    }
+
+    public boolean isCacheEnabled(){
+        return cacheEnabled;
     }
 
     public void commitStorage() throws BackendException {
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/TransactionConfiguration.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/TransactionConfiguration.java
@@ -190,4 +190,10 @@ public interface TransactionConfiguration extends BaseTransactionConfig {
      */
     boolean hasRestrictedPartitions();
 
+    /**
+     * Returns true if read queries should skip accessing JanusGraph database level cache (db-cache).
+     * Doesn't have any effect if database level cache was disabled via config `cache.db-cache`.
+     */
+    boolean isSkipDBCacheRead();
+
 }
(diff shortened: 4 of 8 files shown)


porunov:
Failed HBase tests are unrelated to the current change but caused by the issue #3190

janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java
```diff
@@ -132,6 +134,10 @@ public void enableCache() {
         this.cacheEnabled = true;
     }
 
+    public boolean isCacheEnabled(){
```
li-boxuan, line 137: Is this being used anywhere?

porunov: It's used in the added test and I also think it might be useful for anyone who wants to check if current transaction has cache enabled or not

li-boxuan: Did you forget to commit some code? I didn't see code logic that actually "disables" DB cache.

porunov: I added a test called `testSkipDBCacheRead` into the PR. It sets skipDBCacheRead to `true` which disables cache for the current transaction. Then for the current transaction we use this method to double check that the cache is actually disabled

janusgraph-test/src/test/java/org/janusgraph/graphdb/transaction/StandardTransactionBuilderTest.java
```diff
@@ -87,4 +88,26 @@ public void testConsistencyChecks() {
...
+        StandardJanusGraphTx tx = (StandardJanusGraphTx) graph.newTransaction();
+        BackendTransaction backendTransaction = tx.getTxHandle();
+        assertFalse(tx.getConfiguration().isSkipDBCacheRead());
+        assertTrue(backendTransaction.isCacheEnabled());
+        tx.rollback();
+
+        tx = (StandardJanusGraphTx) graph.buildTransaction().skipDBCacheRead(true).start();
+        backendTransaction = tx.getTxHandle();
+        assertTrue(tx.getConfiguration().isSkipDBCacheRead());
+        assertFalse(backendTransaction.isCacheEnabled());
```
porunov, line 103: Here we check that the cache is disabled

porunov: `ExpirationKCVSCache` has methods such as: `getSlice` and `getSliceNoCache`.
Basically, when the cache is disabled in the transaction then any time we are getting a slice we will call `getSliceNoCache` which reads directly from the storage backend. If cache is enabled then a transaction will call `getSlice`.

porunov: Here are the places where we check if we want to execute `getSlice` or `getSliceNoCache` depending if `cacheEnabled` is `true` or `false` in the current transaction:

https://github.com/JanusGraph/janusgraph/blob/2d4bbf81e5712db222a5f7ce4231cff90c276ee1/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java#L274

https://github.com/JanusGraph/janusgraph/blob/2d4bbf81e5712db222a5f7ce4231cff90c276ee1/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java#L290

https://github.com/JanusGraph/janusgraph/blob/2d4bbf81e5712db222a5f7ce4231cff90c276ee1/janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java#L401

Thus we can enable / disable cache access not only for the whole transaction but actually for part of the transaction. That said, this PR adds possibility to enable / disable cache for the whole transaction only. If users want to enable / disable cache for current transaction multiple times then they need to use the approach I posted in the PR description: https://github.com/JanusGraph/janusgraph/pull/3189#issue-1354789937

li-boxuan: Got it, I missed the current usage of `cacheEnabled` in `BackendTransaction`. I thought this was a new variable introduced by your PR. Now it's clear to me how it works. Thanks!

janusgraph-core/src/main/java/org/janusgraph/core/TransactionBuilder.java
```diff
@@ -138,6 +138,14 @@ public interface TransactionBuilder {
      */
     TransactionBuilder commitTime(Instant instant);
 
+    /**
+     * Enables/disables JanusGraph database level cache access during read operations.
+     * Doesn't have any effect if database level cache was disabled during JanusGraph instance creation.
+     *
+     * @param enabled Enable db-cache access during read operations
```
li-boxuan, line 145: The doc here is a bit ambiguous IMO. "Enable db-cache access during read operations" implies when this parameter is true, db-cache is enabled, while the fact is the opposite. Maybe we can simply change the method signature to `skipDBCacheRead()` just like `disableBatchLoading()` method?

porunov:
> Sweet! I wonder if it is possible to write an integration test for this change.

I added a unit test `testSkipDBCacheRead` which checks if cache is enabled or disabled in the transaction after the transaction is created.
To write an integration test I guess we could if there is a way to get information if a specific query hits a cache or not but I don't know the way to check that yet

li-boxuan:
> To write an integration test I guess we could if there is a way to get information if a specific query hits a cache or not but I don't know the way to check that yet

I am not sure but `IndexMetricTest` might be of help here. We can check the cache hit counter to see if a query hits the cache or not.

porunov:
> I am not sure but IndexMetricTest might be of help here. We can check the cache hit counter to see if a query hits the cache or not.

Sounds like a good plan. I will try to add that test later today

porunov:
@li-boxuan changed signature of `skipDBCacheRead` as you proposed above and added the integration test (`shouldSkipDBCacheAccess`) as you suggested. If you try to remove `.skipDBCacheRead()` from tx creation of that test it will fail due to additional retrieval being recorded by metrics manager.

docs/interactions/transactions.md
```diff
@@ -322,6 +322,10 @@ following aspects of a transaction to be configured:
     confirmation at the application level to avoid inconsistencies. USE
     WITH GREAT CARE!
 
+-   `skipDBCacheRead(boolean)` - Enables/disables JanusGraph database level 
+    cache access during read operations. Doesn't have any effect if database 
+    level cache was disabled during JanusGraph instance creation.
```
li-boxuan, line 327: (nitpick): `JanusGraph instance creation` might confuse some users - "instance creation" sounds like the time when the graph is first created. What about "if database level cache was disabled via config `cache.db-cache`"?

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/TransactionConfiguration.java
```diff
@@ -190,4 +190,10 @@ public interface TransactionConfiguration extends BaseTransactionConfig {
      */
     boolean hasRestrictedPartitions();
 
+    /**
+     * Returns true if read queries should skip accessing JanusGraph database level cache (db-cache).
+     * Doesn't have any effect if database level cache was disabled during JanusGraph instance creation.
```
li-boxuan, line 195: Same as above

li-boxuan (approved):
LGTM 👍  I just have some minor comments on the doc wording

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/TransactionConfiguration.java
porunov: Thank you. Updated documentation as you suggested

PR #3197, opened 2022-08-31
Replace Guava caches with Caffeine caches

### LINKED ISSUE #871: TinyLFU cache (vs LRU)

porunov:
This is a follow-up commit to the 9a22a1adbc7294a00b93101d7a0c801dd94aa0a5 commit

Migrates the rest 7 caches from Guava implementation to Caffeine implementation

This commit also removes a previously added Benchmark class with the obsolate GuavaVertexCache.class. The benchmark is provided in the next GitHub PR: https://github.com/JanusGraph/janusgraph/pull/3188 which shows that Caffeine has better performance characteristics in all scenarious.

ConcurencyLevel is not provided for the cache implementation anymore. The reason for this can be found by the following link https://groups.google.com/g/guava-discuss/c/BnoN9q9fDgw/m/6kF6H2aLAgAJ

Related to #3188 and #3185

Partially related to #2033

Fixes #871

-----

### For all changes:

### For code changes:


porunov:
Suppressed by #3198

PR #3203, opened 2022-09-13
Implement mixed index aggregations

### LINKED ISSUE #3202: Add strategy that will optimize numeric aggregations
The PR #2200 add a strategy that asks the mixed index to count the number of matching elements. This optimization can be improved and add aggregation functions on numeric fields (`min()`, `max()`, `mean()` and `sum()`).

The strategy can optimize queries that look like:
 - `g.V().has(...).values('num_field').min()`
 - `g.V().has(...).values('num_field').max()`
 - `g.V().has(...).values('num_field').avg()`
 - `g.V().has(...).values('num_field').sum()`

The storage backend is then not touched and the performance should be highly increased.

To-om:
Fixes #3202

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -156,6 +156,12 @@ JanusGraph now officially supports Java 11 in addition to Java 8. We encourage e
 !!! note
     The distribution zip archives are however still built with Java 8 since the full distribution includes Cassandra which will only support Java 11 in version 4.
 
+##### Mixed index aggregation optimization
+
+A new optimization has been added to compute aggregations (min, max, sum and avg) using mixed index engine (if the aggregation function follows an indexed query).
+If the index backend is Elasticsearch, a `double` value is used to hold the result. As a result, aggregations on long numbers greater than 2^53 are approximate.
+In this case, if the accurate result is essential, the optimization can be disabled by removing the strategy `JanusGraphMixedIndexAggStrategy`: `g.traversal().withoutStrategies(JanusGraphMixedIndexAggStrategy.class)`.
+
 ### Version 0.6.3 (Release Date: ???)
 
 ```xml tab='Maven'
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -77,7 +77,7 @@ import org.janusgraph.graphdb.query.index.ApproximateIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.index.BruteForceIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.index.ThresholdBasedIndexSelectionStrategy;
 import org.janusgraph.graphdb.query.profile.QueryProfiler;
-import org.janusgraph.graphdb.tinkerpop.optimize.step.JanusGraphMixedIndexCountStep;
+import org.janusgraph.graphdb.tinkerpop.optimize.step.JanusGraphMixedIndexAggStep;
 import org.janusgraph.graphdb.tinkerpop.optimize.step.JanusGraphStep;
 import org.janusgraph.graphdb.tinkerpop.optimize.strategy.JanusGraphMixedIndexCountStrategy;
 import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
@@ -1115,14 +1115,20 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         final PropertyKey prop = makeKey("prop", String.class);
         final PropertyKey description = makeKey("desc", String.class);
         final PropertyKey pet = makeKey("pet", String.class);
-        mgmt.buildIndex("mixed", Vertex.class).addKey(name, Mapping.STRING.asParameter())
-            .addKey(prop, Mapping.STRING.asParameter()).buildMixedIndex(INDEX);
+        final PropertyKey height = makeKey("height", Integer.class);
+        final PropertyKey weight = makeKey("weight", Float.class);
+        mgmt.buildIndex("mixed", Vertex.class)
+            .addKey(name, Mapping.STRING.asParameter())
+            .addKey(prop, Mapping.STRING.asParameter())
+            .addKey(height)
+            .addKey(weight)
+            .buildMixedIndex(INDEX);
         mgmt.buildIndex("mi", Vertex.class).addKey(description).addKey(pet).buildMixedIndex(INDEX2);
         finishSchema();
 
-        tx.addVertex("name", "bob", "prop", "val", "desc", "he likes coding", "pet", "he likes dogs", "age", 20);
-        tx.addVertex("name", "bob", "prop", "val2", "desc", "he likes coding", "pet", "he likes cats", "age", 25);
-        tx.addVertex("name", "alex", "prop", "val", "desc", "he likes debugging", "pet", "he likes cats", "age", 20);
+        tx.addVertex("name", "bob", "prop", "val", "desc", "he likes coding", "pet", "he likes dogs", "age", 20, "height", 170, "weight", 72.3);
+        tx.addVertex("name", "bob", "prop", "val2", "desc", "he likes coding", "pet", "he likes cats", "age", 25, "height", 175, "weight", 102.5);
+        tx.addVertex("name", "alex", "prop", "val", "desc", "he likes debugging", "pet", "he likes cats", "age", 20, "height", 190, "weight", 55.3);
         tx.commit();
 
         // satisfied by a single graph-centric query which is satisfied by a single mixed index query
@@ -1130,6 +1136,11 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
             newTx();
             assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).count().next());
             assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).toList().size());
+            assertEquals(190, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("height").max().next());
+            assertEquals(25, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("age").max().next());
+            assertEquals(170, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("height").min().next());
+            assertEquals(535L, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("height").sum().next());
+            assertEquals(535.0/3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("height").mean().next());
             Metrics mMixedOr = tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val"))
                 .profile().next().getMetrics(0);
             assertEquals("Or(JanusGraphStep([],[name.eq(bob)]),JanusGraphStep([],[prop.eq(val)]))", mMixedOr.getName());
@@ -1223,6 +1234,10 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         newTx();
         assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).count().next());
         assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).toList().size());
+        assertEquals(190, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).values("height").max().next());
+        assertEquals(170, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).values("height").min().next());
+        assertEquals(535L, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).values("height").sum().next());
+        assertEquals(535.0/3, tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20)).values("height").mean().next());
         Metrics mMixedOr = tx.traversal().V().or(__.has("name", "bob"), __.has("age", 20))
             .profile().next().getMetrics(0);
         assertEquals("Or(JanusGraphStep([],[name.eq(bob)]),JanusGraphStep([],[age.eq(20)]))", mMixedOr.getName());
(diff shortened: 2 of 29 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexAggStrategy.java
```diff
@@ -19,29 +19,35 @@
...
 import org.janusgraph.graphdb.tinkerpop.optimize.step.JanusGraphStep;
 
 import java.util.Collections;
 import java.util.Set;
 
 /**
  * If the query can be satisfied by a single mixed index query, and the query is followed by a count step, then
- * this strategy replaces original step with {@link JanusGraphMixedIndexCountStep}, which fires a count query against
+ * this strategy replaces original step with {@link JanusGraphMixedIndexAggStep}, which fires a count query against
  * mixed index backend without retrieving all elements
```
porunov, line 41: `count` > `aggregation`

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexAggStrategy.java
```diff
@@ -60,48 +66,72 @@ public void apply(final Traversal.Admin<?, ?> traversal) {
...
+                    else if (currentStep instanceof SumGlobalStep) return Aggregation.SUM(fieldName);
+                }
+            }
+        }
+        return null;
     }
 
     /**
-     * Apply mixed index count step in the traversal and remove "has" steps that already folded in the mixedIndexCountStep
+     * Apply mixed index count step in the traversal and remove "has" steps that already folded in the mixedIndexAggStep
```
porunov, line 116: `count` > `aggregation`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -426,6 +427,58 @@ public long countTotal(String indexName, Map<String, Object> requestData) throws
         }
     }
 
+    private double executeAggs(String indexName, Map<String, Object> requestData, String agg, String fieldName) throws IOException {
```
porunov, line 430: I think we should also add documentation about aggregation loose of precision for big numbers when using ElasticSearch. As noted here, for any number more then `2^53` the result is approximate: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html#limits-for-long-values

I think users should now about this limitation and if they want precise values they need to disable this optimization in their queries. I.e.:
```
g.traversal().withoutStrategies(JanusGraphMixedIndexAggStrategy.class).has("someProperty", 123).sum()
```

porunov (commented):
Thank you for this feature contribution @To-om !
I glanced over the PR very quickly and found some Javadoc are using `count query` where the logic is changed for `aggregation query`. I also think we need to add doc explaining a precision problem when using ElasticSearch for aggregation.
I didn't check `JanusGraphMixedIndexAggStep` and `JanusGraphMixedIndexAggStrategy` too deep yet, but hopefully @li-boxuan could take a look at these two classes because we was adding the count optimization strategy previously.
I also didn't check tests and Solr with Lucene implementations yet but will take a look at them soon.
I left small nitpicks below and will continue the review soon

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
porunov: I'm also thinking that maybe it would make sense to actually keep Count aggregation strategy separate from other aggregations optimization strategy. If so, users might want to opt-in using and accept aggregation limits it or not using it and to not have any limits.

li-boxuan: > I'm also thinking that maybe it would make sense to actually keep Count aggregation strategy separate from other aggregations optimization strategy

That makes sense because even ES count (without aggregation) itself is an approximate...

janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java
```diff
@@ -943,16 +945,74 @@ public Long queryCount(IndexQuery query, KeyInformation.IndexRetriever informati
...
-            final TopDocs docs = searcher.search(q, 1);
-            log.debug("Executed query [{}] in {} ms", q, System.currentTimeMillis() - time);
-            return docs.totalHits.value;
+            switch (aggregation.getType()) {
+                case COUNT: return executeCount(searcher, q);
+                case MIN: return executeMin(searcher, q, aggregation.getFieldName(), aggregation.getDataType());
+                case MAX: return executeMax(searcher, q, aggregation.getFieldName(), aggregation.getDataType());
+                case AVG: return executeAvg(searcher, q, aggregation.getFieldName());
+                case SUM: return executeSum(searcher, q, aggregation.getFieldName(), aggregation.getDataType());
+                default: throw new IOException();
```
li-boxuan, line 954: Use `UnsupportedOperationException`

li-boxuan (commented):
Looks great! We also need to:
1) Add a changelog explaining that `JanusGraphMixedIndexCountStrategy` has been renamed `JanusGraphMixedIndexAggStrategy` since it's a breaking change.
2) (Optional) Make this strategy configurable as suggested [here](https://github.com/JanusGraph/janusgraph/pull/3203#discussion_r979960051). We can add it afterward if you don't have time.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
porunov: > That makes sense because even ES count (without aggregation) itself is an approximate...

I think you are referring to the old implementation (probably in 0.2 or 0.3 version of JanusGraph) where we were using [total hits](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-your-data.html#track-total-hits).

Right now for ElasticSearch implementation we are using [Count API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-count.html) which suppose to return accurate value unless I'm missing something. At least I didn't find information that it's not accurate.

So, my point is that users most likely want to have `count` optimization in their queries all the times because it's accurate but aggregation optimization isn't accurate when the resulting value is more then 2^53. So, some users may want to exclude aggregation optimization in some cases but have count optimization enabled in their queries.

That's why I suggested to have both optimizations for `count` and for `aggregation`.

Another reason is that:
`g.V().has("foo", "bar").limit(1000).count()` - can use optimization with a specific logic like `Math.min(limit, countResultFromIndex)` but I can't say the same for other aggregation queries.

Moreover, later (not necessary in this PR) I think we will need to add a configuration parameter which excludes this aggregation optimization by default - otherwise it may be a breaking change for some users due to approximate results in some cases.

li-boxuan: > Right now for ElasticSearch implementation we are using [Count API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-count.html) which suppose to return accurate value unless I'm missing something. At least I didn't find information that it's not accurate.

You are right! Never mind, I misremembered it :)

To-om: ElasticSearch count API is accurate but the Lucene one is not. It only counts the 1000 first elements ([TotalHits](https://lucene.apache.org/core/8_11_2/core/org/apache/lucene/search/TotalHits.html)). It was not the case in Lucene 7.7.2 (JanusGraph 5.x).

I agree with you, accurate optimizations should not be mixed with approximate one. I can split it.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -426,6 +427,58 @@ public long countTotal(String indexName, Map<String, Object> requestData) throws
...
+    }
+
+    private Number adaptNumberType(double value, Class<? extends Number> expectedType) {
+        if (expectedType == null) return value;
+        else if (Byte.class.isAssignableFrom(expectedType)) return (byte)value;
+        else if (Short.class.isAssignableFrom(expectedType)) return (short)value;
+        else if (Integer.class.isAssignableFrom(expectedType)) return (int)value;
+        else if (Long.class.isAssignableFrom(expectedType)) return (long)value;
+        else if (Float.class.isAssignableFrom(expectedType)) return (float)value;
+        //else if (Double.class.isAssignableFrom(expectedType)) return value;
```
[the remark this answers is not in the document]
To-om: are you sure ?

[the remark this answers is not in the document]
li-boxuan: I mean, you commented out line 454 already, then why not just remove it?

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/JanusGraphMixedIndexAggStep.java
```diff
@@ -25,54 +25,86 @@
...
+                    .filter(Number.class::isAssignableFrom)
+                    .findFirst();
+                if (dataType.isPresent()) {
+                    isIndexed = true;
+                    aggregation.setDataType(dataType.get());
+                    RelationType rt = tx.getRelationType(fieldName);
+                    if (rt instanceof PropertyKey)
+                        aggregation.setFieldName(key2Field(indexType, (PropertyKey) rt));
+                } else {
+                    isIndexed = false;
```
li-boxuan, line 86: Maybe I am missing something, but is this branch covered by any test case? I didn't find any but not sure why codecov didn't complain.

li-boxuan (commented):
Thanks for the refactoring! Could you please also add a changelog/user doc to briefly introduce `JanusGraphMixedIndexAggStrategy` so that people know how to disable it when they need accurate counts?

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -1115,21 +1115,29 @@ public void testGraphCentricQueryProfiling() {
...
+        tx.addVertex("name", "bob", "prop", "val2", "desc", "he likes coding", "pet", "he likes cats", "age", 25, "height", 175);
+        tx.addVertex("name", "alex", "prop", "val", "desc", "he likes debugging", "pet", "he likes cats", "age", 20, "height", 190);
         tx.commit();
 
         // satisfied by a single graph-centric query which is satisfied by a single mixed index query
         if (indexFeatures.supportNotQueryNormalForm()) {
             newTx();
             assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).count().next());
             assertEquals(3, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).toList().size());
+            assertEquals(190, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("height").max().next());
```
li-boxuan, line 1137: Could you please add one more test case to cover the branch mentioned [here](https://github.com/JanusGraph/janusgraph/pull/3203#discussion_r985127606)? I know the branch has been removed, but right now that default case is still not tested. An example would be:
```java
assertEquals(25, tx.traversal().V().or(__.has("name", "bob"), __.has("prop", "val")).values("age").max().next());
```
i.e. A mixed index is matched, but the index itself does not cover the property of interest. In other words, the mixed index is not a covering index.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AbstractJanusGraphMixedIndexAggStrategy.java
```diff
@@ -0,0 +1,96 @@
...
+    abstract Aggregation getAggregation(final GraphStep originalGraphStep);
+
+    /**
+     * Apply mixed index aggregation step in the traversal and remove "has" steps that already folded in the
+     * mixedIndexAggStep
+     *
+     * @param originalGraphStep
+     * @param mixedIndexAggStep
+     * @param traversal
+     */
```
li-boxuan, line 84: missing param `hasField` in javadoc

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexAggStrategy.java
```diff
@@ -0,0 +1,75 @@
...
+            if (propertyKeys.length == 1) {
+                final String fieldName = propertyKeys[0];
+                currentStep = currentStep.getNextStep();
+                while (isEligibleToSkip(currentStep)) {
+                    currentStep = currentStep.getNextStep();
+                }
+                if (currentStep instanceof MaxGlobalStep) return Aggregation.MAX(fieldName);
+                else if (currentStep instanceof MinGlobalStep) return Aggregation.MIN(fieldName);
+                else if (currentStep instanceof MeanGlobalStep) return Aggregation.AVG(fieldName);
+                else if (currentStep instanceof SumGlobalStep) return Aggregation.SUM(fieldName);
```
li-boxuan, line 61: Not necessary to be included in this PR, but I guess a good next step is to also add support for `values(key).count()`.

li-boxuan (commented):
Thanks for the great contribution! I just have some nitpicks, otherwise LGTM.

docs/changelog.md
```diff
@@ -156,6 +156,13 @@ JanusGraph now officially supports Java 11 in addition to Java 8. We encourage e
 !!! note
     The distribution zip archives are however still built with Java 8 since the full distribution includes Cassandra which will only support Java 11 in version 4.
 
+##### Mixed index aggregation optimization
+
+A new optimization has been added to compute aggregations (min, max, sum and avg) using mixed index engine (if the aggregation function follows an indexed query).
+If the index backend is Elasticsearch, a `double` value is used to hold the result. As a result, aggregations on long numbers greater than 2^53 are approximate.
+In this case, if the accurate result is essential, the optimization can be disabled by removing the strategy `JanusGraphMixedIndexAggStrategy`: `g.traversal().withoutStrategies(JanusGraphMixedIndexAggStrategy.class)`   
+may be approximate
```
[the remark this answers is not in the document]
To-om: Of course. I will remove it.

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/JanusGraphMixedIndexAggStrategy.java
To-om: Yes I add support of this query. It is equivalent to `has(key).count()` which is optimized.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
To-om: I've added a [test](https://github.com/StrangeBeeCorp/janusgraph/blob/aggregations/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/IndexMetricTest.java#L71) that cover the default case, but only for metrics (the result is not checked).
I'll add this test.

li-boxuan (approved):
LGTM 👍

analytically:
@To-om 

janusgraph-solr/src/main/java/org/janusgraph/diskstorage/solr/SolrIndex.java#L781
Added line #L781 was not covered by tests

porunov (approved):
LGTM. Thank you for this feature @To-om !

porunov:
@To-om could you please squash your commits into a single commit?
If it's complicated then you can just hard reset from this branch: https://github.com/porunov/janusgraph/tree/aggregations
Alternatively you can click `Allow edits and access to secrets by maintainers` on this PR and I will force push into your branch.

PR #3212, opened 2022-09-19
Remove GryoMessageSerializerV3d0

### LINKED ISSUE #2921: Stop using GryoMessageSerializer
TinkerPop will remove the GryoMessageSerializers completely in version 3.6.0: [TINKERPOP-2639](https://issues.apache.org/jira/browse/TINKERPOP-2639) and instead recommends to use GraphBinary.
Due to the breaking change in serialization of Geoshapes via GraphBinary coming in #2876, I suggest that we only use GraphBinary by default in a version that already contains that change, so probably 1.0.0.

We could also switch to GraphSON instead if we want to move away from Gryo earlier, but it's probably fine to just directly move to GraphBinary in 1.0.0. So, I'm adding this to the v1.0.0 milestone for now.

farodin91:
Fixes #2921

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -156,6 +156,10 @@ JanusGraph now officially supports Java 11 in addition to Java 8. We encourage e
 !!! note
     The distribution zip archives are however still built with Java 8 since the full distribution includes Cassandra which will only support Java 11 in version 4.
 
+##### Drop support for 
+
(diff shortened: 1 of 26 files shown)


porunov (approved):
LGTM. Thank you @farodin91 !

We can improve docs later with follow-up PRs or with CTR.

PR #3245, opened 2022-10-10
Remove cassandra-all dependency

farodin91:
Done
1. cleanup janusgraph-cql. Move cassandra hadoop impl to us. 

Todo:
1. switch to v4 driver
2. Merge and refactor

-----

### For all changes:

### For code changes:


FlorianHockmann:
I haven't really reviewed the PR, but since you seem to have copied files here completely from Apache Cassandra, I think that we have to include this in our `NOTICE.txt`. The TinkerPop `NOTICE` file contains an example where a source file was also copied from another project: https://github.com/apache/tinkerpop/blob/master/NOTICE#L23

farodin91:
@FlorianHockmann  I added a section to the NOTICE.txt

farodin91:
@FlorianHockmann  I added a section to the NOTICE.txt

the codecov/patch will fail.

.gitignore
```diff
@@ -28,6 +28,7 @@ target/
 /janusgraph-*/janusgraph-*/.classpath
 /janusgraph-*/janusgraph-*/.project
 /janusgraph-*/janusgraph-*/.settings/
+#/janusgraph-cql/src/main/java/org/apache
```
FlorianHockmann, line 31: Was this left by accident?

janusgraph-cql/src/main/java/org/janusgraph/hadoop/formats/cql/CqlBinaryInputFormat.java
```diff
@@ -79,19 +76,5 @@ public void setConf(final Configuration config) {
         ConfigHelper.setInputColumnFamily(config, janusgraphConf.get(CQLConfigOptions.KEYSPACE),
             mrConf.get(JanusGraphHadoopConfiguration.COLUMN_FAMILY_NAME), wideRows);
         log.debug("Set keyspace: {}", janusgraphConf.get(CQLConfigOptions.KEYSPACE));
-
```
FlorianHockmann, line 82: Everything in this PR is really just copying files from Apache Cassandra to us and updating a dependency, except for these lines that you remove here. Can you give a short explanation of why they can be removed now?

farodin91: We didn't had a test for it. SlicePredicate doesn't exists anymore. If we want this feature again, we should test it.

FlorianHockmann: I see, thanks for the explanation!

FlorianHockmann (approved):

porunov (commented):
Why do we need to copy so many classes from cassandra dependency to `janusgraph-cql`? 
Why can't we just introduce a necessary dependency?
There are too many classes we will need to support after this PR.

farodin91:
> Why do we need to copy so many classes from cassandra dependency to janusgraph-cql?

I copied as aleast as possible. It seems that the hadoop part is really support in cassandra. We could switch to datastax variant which uses spark instead of hadoop. That would require much more work.

> Why can't we just introduce a necessary dependency?

Their was a dependency cassandra-all which a depedency hell.

porunov:
> Their was a dependency cassandra-all which a depedency hell.

I'm just thinking maybe it makes sense to refactor cassandra to provide a small dependency with necessary classes we are using. I.e. to create something like `cassandra-hadoop-util` library or something like that and publish it as a Maven artifact. If so, we won't need to support these classes by our own and anytime we bump cassandra version we would be able to get all features and bugfixes without copy pasting source code every time there is a new Cassandra version. Also by copy pasting these classes user's won't be able to replace transitive dependencies with the newer / older cassandra versions because these classes are hard coded into JanusGraph logic now.
If it's complicated to introduce a new dependency on cassandra side then I would prefer just creating our own library (or at least another module) and make `janusgraph-cql` module depend on that module with these copy pasted classes. Otherwise, people who start to learn `janusgraph-cql` module will be overwhelmed with some classes they don't really need because those classes are meant to be copied from cassandra and not developed by JanusGraph developers.

farodin91:
I could move the classes into a own module. Afterward, i will open issue on cassandra. 

Besides, i don't think we have to update on each cassandra update as they just depending open the datastax driver.

porunov (approved):
Assuming that `cassandra-hadoop-util` is copy paste of classes from `cassandra` and assuming that all tests pass it LGTM. Thank you @farodin91 !

farodin91:
@porunov @FlorianHockmann Should i try to create patch file with difference between cassandra and cassandra hadoop?

FlorianHockmann:
> Should i try to create patch file with difference between cassandra and cassandra-hadoop-util?

Doesn't [your second commit here](https://github.com/JanusGraph/janusgraph/pull/3245/commits/f4816036968295598a2665369a4c42b548999cb6) already show the full diff? That helped me at least to review this change and the PR still looks good to me.

It was definitely a good suggestion to move this into its own module by the way and asking the folks at Apache Cassandra whether they would be willing to move these files into its own module on their side of course also makes sense.

FlorianHockmann:
> I want show a diff between cassandra and cassandra-hadoop-util.

You mean there are changes between the original files and the ones added with this PR? I just assumed that you completely copied them. In that case, a diff would be good I guess.

porunov:
> I want show a diff between cassandra and cassandra-hadoop-util.

If that's only headers then I'm fine without the diff. If there are any other changes then it would be good to see the diff

farodin91:
@FlorianHockmann  @porunov i've added a extra commit.

FlorianHockmann (commented):
I just reviewed the diff. You seem to have mostly removed code that we aren't using which should be fine.
I'm wondering however about users who use JanusGraph embedded in their Java project and also Cassandra. They might need to exclude `cassandra-hadoop-util` as a dependency and add a dependency on `cassandra-all` themselves now if they use any of the removed code themselves. I guess this something we should mention in the upgrade docs.

Some changes are however also changing the behavior if I'm not mistaken, e.g., special handling for cases with a single datacenter. I would have preferred a follow-up PR for such changes as that would make the review easier and you could better explain these changes. But I'm also fine with keeping them in here if it's too much effort to extract them from the PR now.

farodin91:
@FlorianHockmann Changelog is a good idea.

farodin91:
> ### Removal of cassandra-all dependency
> We have extracted the hadoop part from cassandra to reduce the dependency tree of JanusGraph. If you are running JanusGraph with 
an embeded Cassandra, you have to exclude the `cassandra-hadoop-util` from `janusgraph-cql`.

@FlorianHockmann what do you think?

docs/changelog.md
```diff
@@ -88,6 +88,11 @@ For more information on features and bug fixes in 1.0.0, see the GitHub mileston
 
 #### Upgrade Instructions
 
+##### Removal of cassandra-all dependency
+
+We have extracted the hadoop part from cassandra to reduce the dependency tree of JanusGraph. If you are running JanusGraph with
```
FlorianHockmann, line 93: How about this:

> JanusGraph had a dependency on `cassandra-all` only for some Hadoop-related classes. We moved these few classes into a new module `cassandra-hadoop-util` to reduce the dependencies of JanusGraph. "If you are running JanusGraph [...]".

farodin91: rebased and changed

FlorianHockmann (approved):

PR #3264, opened 2022-10-24
Send vertex label to user log processor for all mutations

### LINKED ISSUE #3155: Global db-cache invalidation on mutation
Currently we invalidate db-cache only in the next 3 scenarios:
1) Enough time passed (`cache.db-cache-time`).
2) Evicted due to cache size limitation (`cache.db-cache-time`).
3) Evicted on current JanusGraph instance only due to being mutated on the current JanusGraph instance.

I would propose adding a feature which adds a possibility to evict JanusGraph cache on all JanusGraph instances during mutation on any of the instances (something like `cache.db-cache-global-eviction = true`).
This would be a good feature for situations when there are no many writes but many reads. Of course, if there are too many writes this would pressure other JanusGraph nodes which isn't good but then this feature shouldn't be used in such use-cases.

I didn't think on how to implement this feature but I guess maybe reusing JanusGraph messaging mechanism could be a good start.

### LINKED ISSUE #3263: Send vertex label to user log processor for all mutations
Right now using `LogProcessorFramework` there is no way to retrieve mutated vertex label. Because messages don't include vertex label but only vertex id. 
This is a blocker for #3155 . 
Without sending vertex label id in the mutation message we won't have vertex label information during log processing.
I.e. `vertexLabel()` method will always return default `vertex` label for all received vertices (unless this particular vertex was cached on the receiving instance, which usually isn't the case). 
Without knowing vertex label of the mutated vertex we won't be able to properly invalidate indices which are bound to specific vertex label. I.e., those indices which use `indexOnly(JanusGraphSchemaType schemaType)` method.
To be able to invalidate those indices (necessary for #3155) we should provide vertex label information.

**Describe the feature:**
Provide vertex label information for mutated vertices in log processor.

Implementing global db-cache invalidation feature (#3155).

porunov:
Fixes #3263
Related to #3155 

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -171,6 +171,24 @@ A new optimization has been added to compute aggregations (min, max, sum and avg
 If the index backend is Elasticsearch, a `double` value is used to hold the result. As a result, aggregations on long numbers greater than 2^53 are approximate.
 In this case, if the accurate result is essential, the optimization can be disabled by removing the strategy `JanusGraphMixedIndexAggStrategy`: `g.traversal().withoutStrategies(JanusGraphMixedIndexAggStrategy.class)`.
 
+##### Breaking change for transaction logs processing
+
+[Transaction Log](advanced-topics/transaction-log.md) processing has a breaking change.  
+Previously for all mutated vertices a default vertex label `vertex` has been used in `ChangeState` events. 
+The new approach stores vertex labels in logs as well which increases storage size for +8 bytes per each relation and 
+makes all previously stored transaction logs incompatible with the new structure.   
+To pass this breaking change it's necessary to ensure that transaction logs stored via previous JanusGraph versions are 
+not processed via JanusGraph version >= 1.0.0.   
+One of the possible ways to ensure previous logs are not processed is to use `setStartTimeNow` 
+to process only newest logs. Another way could be to process all previous logs, store log identifier state and start 
+processing logs from the latest log ensuring that the latest logs are send by the JanusGraph version >= 1.0.0.   
+For example:  
+```java
+LogProcessorBuilder logProcessorBuilder = logProcessorFramework
+  .addLogProcessor("myLogProcessorIdentifier")
+  .setStartTimeNow();
+```
+
 ### Version 0.6.3 (Release Date: ???)
 
 ```xml tab='Maven'
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -141,9 +141,11 @@ import org.janusgraph.graphdb.transaction.StandardJanusGraphTx;
 import org.janusgraph.graphdb.types.CompositeIndexType;
 import org.janusgraph.graphdb.types.StandardEdgeLabelMaker;
 import org.janusgraph.graphdb.types.StandardPropertyKeyMaker;
+import org.janusgraph.graphdb.types.VertexLabelVertex;
 import org.janusgraph.graphdb.types.system.BaseVertexLabel;
 import org.janusgraph.graphdb.types.system.ImplicitKey;
 import org.janusgraph.graphdb.types.vertices.JanusGraphSchemaVertex;
+import org.janusgraph.graphdb.vertices.AbstractVertexUtil;
 import org.janusgraph.graphdb.vertices.CacheVertex;
 import org.janusgraph.testutil.FeatureFlag;
 import org.janusgraph.testutil.JanusGraphFeature;
@@ -4871,6 +4873,8 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
 
         PropertyKey weight = tx.makePropertyKey("weight").dataType(Float.class).cardinality(Cardinality.SINGLE).make();
         EdgeLabel knows = tx.makeEdgeLabel("knows").make();
+        String testVertexLabel = "testVertex";
+        tx.makeVertexLabel(testVertexLabel).make();
         JanusGraphVertex n1 = tx.addVertex("weight", 10.5);
         tx.addProperties(knows, weight);
         newTx();
@@ -4885,7 +4889,7 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
         final long v1id = getId(v1);
         txTimes[1] = times.getTime();
         tx2 = graph.buildTransaction().logIdentifier(userLogName).start();
-        JanusGraphVertex v2 = tx2.addVertex("weight", 222.2);
+        JanusGraphVertex v2 = (JanusGraphVertex) tx2.traversal().addV(testVertexLabel).property("weight", 222.2).next();
         v2.addEdge("knows", getV(tx2, v1id));
         tx2.commit();
         final long v2id = getId(v2);
@@ -4998,8 +5002,8 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
             assertEquals(0, userLogMsgCounter.get());
         } else {
             assertEquals(4, userLogMsgCounter.get());
-            assertEquals(7, userChangeCounter.get(Change.ADDED).get());
-            assertEquals(4, userChangeCounter.get(Change.REMOVED).get());
+            assertEquals(8, userChangeCounter.get(Change.ADDED).get());
+            assertEquals(5, userChangeCounter.get(Change.REMOVED).get());
         }
 
         clopen(option(VERBOSE_TX_RECOVERY), true);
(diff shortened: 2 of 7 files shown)


janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -5057,6 +5059,8 @@ public void read(Message message) {
                         txNo = 2;
                         //v2 addition transaction
                         assertEquals(1, Iterables.size(changes.getVertices(Change.ADDED)));
+                        String vertexLabel = changes.getVertices(Change.ADDED).iterator().next().label();
+                        assertEquals(testVertexLabel, vertexLabel);
```
porunov, line 5063: This two lines would fail without the business logic change because `vertexLabel` here would be equal to `vertex` even so real label of the vertex label is `testVertex`. With this PR this vertex now returns correct vertex label.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -5077,6 +5081,8 @@ public void read(Message message) {
                         //v2 deletion transaction
                         assertEquals(0, Iterables.size(changes.getVertices(Change.ADDED)));
                         assertEquals(1, Iterables.size(changes.getVertices(Change.REMOVED)));
+                        String vertexLabel = changes.getVertices(Change.REMOVED).iterator().next().label();
+                        assertEquals(testVertexLabel, vertexLabel);
```
porunov, line 5085: This two lines would fail without the business logic change because `vertexLabel` here would be equal to `vertex` even so real label of the vertex label is `testVertex`. With this PR this vertex now returns correct vertex label.

porunov:
@JanusGraph/committers I have intention to merge this **breaking change** on Monday using lazy consensus if there are no reviews. In case you need more time, please, let me know.
This is a blocker for #3155

docs/changelog.md
```diff
@@ -171,6 +171,24 @@ A new optimization has been added to compute aggregations (min, max, sum and avg
 If the index backend is Elasticsearch, a `double` value is used to hold the result. As a result, aggregations on long numbers greater than 2^53 are approximate.
 In this case, if the accurate result is essential, the optimization can be disabled by removing the strategy `JanusGraphMixedIndexAggStrategy`: `g.traversal().withoutStrategies(JanusGraphMixedIndexAggStrategy.class)`.
 
+##### Breaking change for transaction logs processing
+
+[Transaction Log](advanced-topics/transaction-log.md) processing has a breaking change.
```
li-boxuan, line 176: This only affects transaction logs started by users, right? I think so but just wanted to make sure.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/log/TransactionLogHeader.java
```diff
@@ -85,7 +87,10 @@ public StaticBuffer serializeModifications(Serializer serializer, LogTxStatus st
     private static void logRelations(DataOutput out, final Collection<InternalRelation> relations, StandardJanusGraphTx tx) {
         VariableLong.writePositive(out,relations.size());
         for (InternalRelation rel : relations) {
-            VariableLong.writePositive(out,rel.getVertex(0).longId());
+            InternalVertex vertex = rel.getVertex(0);
+            VariableLong.writePositive(out,vertex.longId());
+            VertexLabel vertexLabel = vertex.vertexLabel();
+            VariableLong.writePositive(out, vertexLabel.hasId() ? vertexLabel.longId() : 0L);
```
li-boxuan, line 93: I think it's fine, and I believe we should cache vertex labels aggressively. But just to understand more about the intention here: cannot the receiver execute a backend query to fetch vertex label?

li-boxuan (commented):
I like this change 👍 I just have some quick questions

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/log/TransactionLogHeader.java
porunov: It's not guaranteed that there will be such vertex in the graph because the vertex might be already removed.
1) It's guaranteed to be removed if we receive the change for removed vertex.
2) Even if we received a change for updated / added vertex - it doesn't guarantee that this vertex still exists because someone could remove that vertex faster then the change is received by the processing JanusGraph instance.

farodin91:
Just one question, what happens during an upgrade? How old janusgraph instances handle these messages?

porunov:
> Just one question, what happens during an upgrade? How old janusgraph instances handle these messages?

After updated you will need to start consuming new messages only. Thus, it's necessary that you stop any mutations during upgrade, consume all old logs and only then upgrade JanusGraph nodes. After that you can enable consumption again on new logs.
I wish we used a better structure for logs (like protobuf maybe) to be able to extend logs with new fields without introducing a breaking change but it looks like preference was to use better compaction for logs instead of better extensibility.

farodin91:
> > Just one question, what happens during an upgrade? How old janusgraph instances handle these messages?
> 
> After updated you will need to start consuming new messages only. Thus, it's necessary that you stop any mutations during upgrade, consume all old logs and only then upgrade JanusGraph nodes. After that you can enable consumption again on new logs.
> If old JanusGraph instances consume these logs then the result is unpredictable.
> I wish we used a better structure for logs (like protobuf maybe) to be able to extend logs with new fields without introducing a breaking change but it looks like preference was to use better compaction for logs instead of better extensibility. 

We should add a note to that during upgrade you should stop mutation.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/log/TransactionLogHeader.java
li-boxuan: If the vertex is already removed, why do we need the vertex label on the receiver side? I guess my real question is, can you elaborate on the following statement?

> Without knowing vertex label of the mutated vertex we won't be able to properly invalidate indices which are bound to specific vertex label.

porunov: To invalidate an index which is constrained to a specific label we need to know the label of the removed / updated / added vertex

porunov: If we have `db-cache` enabled, and we have an index which was created with `indexOnly(vertexLabel)` and in case we cached some query results in indexStore then we will need vertex label + vertex id + vertex properties to invalidate related cached queries like: `g.V().hasLabel("someVertexLabel").has("myProperty", "propValue")`
We can't invalidate the above query knowing vertex id only

porunov: > Two questions:
> 
> 1. I thought vertex id + vertex properties would be sufficient to invalidate index entry in the cache, no? If you can give a pointer to the code or your POC, that would be great!
> 2. What if we don't invalidate this entry, will there be any problem?

1. No, vertex id + properties is only sufficient to invalidate indices which don't use 'indexOnly' feature. Those which use 'indexOnly' features won't be invalidated using vertex id + properties only. We need to have vertex label as well for that invalidation.
2. Yes, queries which touch the cached index query result will return removed vertex id which will cause an exception when trying to get properties for already removed vertex.

li-boxuan: > Those which use 'indexOnly' features won't be invalidated using vertex id + properties only. We need to have vertex label as well for that invalidation.

Gotcha. Do you have a code snippet showing how we could invalidate normal index entries in DB cache? I thought index entries with `indexOnly` constraints are no different from normal index entries (from cache perspective) but from what you said it's not the case.

li-boxuan: I am not 100% sure but I believe vertex label is not needed here. 

In `invalidateUpdatedVertexProperty`, you called 

```java
Collection<IndexUpdate> indexUpdates = indexSerializer.getIndexUpdates(cacheVertex, Arrays.asList(propertyPreviousVal, propertyNewVal));
```
to fetch a collection of index records.

In `IndexSerializer::getIndexUpdates`, it skips an index entry candidate if it does not conform to the index only constraint:

https://github.com/JanusGraph/janusgraph/blob/11e40fca4c3a9584163f36bf3aa5bacd232ed1a7/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java#L190

This makes sense when we want to generate a new index entry, but it is not necessary when we try to remove an existing index entry. If you temporarily comment out this line, I believe your test case would pass even if the LogEntry does not contain vertex label.

porunov: @li-boxuan you are right. I guess in this case it's better to add another method to `IndexSerializer` which won't filter indices by `indexOnly` constraint.

porunov: @li-boxuan I opened the PR which adds possibility to get Index updates with disabled constraints here #3279

porunov:
The PR is obsolete because we are able to invalidate cached indexes without knowing vertex label as stated here: https://github.com/JanusGraph/janusgraph/pull/3264#discussion_r1014220226

Thus, closing this PR without merging it.

PR #3268, opened 2022-10-29
Add support for ElasticSearch 8

### LINKED ISSUE #3015: Add support for ElasticSearch 8
ElasticSearch 8 was released a while ago (February 10, 2022). It's now safe to add support for ElasticSearch 8 into JanusGraph.

porunov:
- Added ElasticSearch 8.6.0 High Rest Client support
- Fixed ElasticSearch 7.17.8 version for janusgraph-full distribution because ElasticSearch 8 doesn't support Java 8 or Java 11 (see discussion here: #3467)
- Added BKD index mapping support for ElasticSearch (see [more info here](https://www.elastic.co/blog/bkd-backed-geo-shapes-in-elasticsearch-precision-efficiency-speed))
- Added BKD Circle Processor implementations to allow converting Circle to another shape when such mapping is used
- Moved BKD Circle Processors to general module `janusgraph-mixed-index-utils`, so that it could be reused by Lucene or Solr when BKD Geoshape mapping is added to those index backends. Currently 3 general circle processors are added and a possibility to provide custom circle processor for advanced use cases. 

Fixes #3015

Prehistory.
Initially I moved ElasticSearch internal Circle processor to public `elasticsearch-geo` library to reuse it JanusGraph (here is the PR: https://github.com/elastic/elasticsearch/pull/88088). That said, after a small investigation the next was found:
- `elasticsearch-geo` requires Java 17 to be used, but we use Java 8 in JanusGraph.
- The Circle Processor implementation ElasticSearch has right now has problems with small radius circles, circles touching north or south poles, slow `haversinMeters` is used instead of fast `haversinMeters` which is provided in `org.apache.lucene.util.SloppyMath`.
- `CircleUtils` in ElasticSearch uses internal ES Geometry implementation classes which prevents us from reusing that logic for Solr or Lucene in the future if we decide to add BKD support to Solr or Lucene.

Thus, the decision was made to copy math under main `CircleUtils` provided by ElasticSearch to a similar class `CircleUtils` provided by JanusGraph but make it more general and with `haversinMeters` usage. Reference to those classes are made in JavaDoc.
As such, the pros we gain are the next:
- We are not bound to ElasticSearch Geometry data structures which gives us ability to reuse this logic for Lucene and Solr without introducing `elasticsearch-geo` dependency there.
- We can use this logic for Java 8 (the version we are using right now)
- We have small performance improvement due to `org.apache.lucene.util.SloppyMath.haversinMeters` usage.

ElasticSearch Circle processor doc for reference: https://github.com/elastic/elasticsearch/blob/main/docs/reference/ingest/processors/circle.asciidoc

-----

### For all changes:

### For code changes:

### DIFF
--- a/.github/workflows/ci-index-es.yml
+++ b/.github/workflows/ci-index-es.yml
@@ -65,6 +65,10 @@ jobs:
       matrix:
         include:
           - module: es
+            args: "-Pelasticsearch8"
+            name: es8
+            java: 8
+          - module: es
             args: "-Pelasticsearch7"
             name: es7
             java: 8
@@ -76,6 +80,11 @@ jobs:
             args: "-Pelasticsearch60"
             name: es60
             java: 8
+          - module: es
+            install-args: "-Pjava-11"
+            args: "-Pelasticsearch8"
+            name: es8
+            java: 11
           - module: es
             install-args: "-Pjava-11"
             args: "-Pelasticsearch7"
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -28,7 +28,7 @@ All currently supported versions of JanusGraph are listed below.
 | JanusGraph | Storage Version | Cassandra | HBase | Bigtable | Elasticsearch | Solr | TinkerPop | Spark | Scala |
 | ----- | ---- | ---- | ---- | ---- | ---- | ---- | --- | ---- | ---- |
 | 0.6.z | 2 | 3.0.z, 3.11.z | 1.6.z, 2.2.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 7.y, 8.y | 3.5.z | 3.0.z | 2.12.z |
-| 1.0.z | 2 | 3.11.z, 4.0.z | 2.5.z | 1.3.0, 1.4.0, 1.5.z, 1.6.z, 1.7.z, 1.8.z, 1.9.z, 1.10.z, 1.11.z, 1.14.z | 6.y, 7.y | 8.y | 3.6.z | 3.2.z | 2.12.z |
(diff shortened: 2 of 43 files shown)


porunov:
Apparently `elasticsearch-geo-8.6.0` library is compiled with Java 17, but we use Java 8. Thus, we can't use `elasticsearch-geo-8.6.0`.  It means that https://github.com/elastic/elasticsearch/pull/88088 isn't solving the issue because we still can't reuse that ElasticSearch code. I don't have a good solution to this issue right now, but will try to think about other solutions.

porunov:
> Apparently `elasticsearch-geo-8.6.0` library is compiled with Java 17, but we use Java 8. Thus, we can't use `elasticsearch-geo-8.6.0`. It means that [elastic/elasticsearch#88088](https://github.com/elastic/elasticsearch/pull/88088) isn't solving the issue because we still can't reuse that ElasticSearch code. I don't have a good solution to this issue right now, but will try to think about other solutions.

The [CircleUtils](https://github.com/porunov/elasticsearch/blob/f4adbae6f08233c9c2c847b626e2034080d448b5/libs/geo/src/main/java/org/elasticsearch/geometry/utils/CircleUtils.java) is pure mathematics and ElasticSearch internal geo structures.
I believe we can re-implement the same logic internally at JanusGraph.
We could also probably add BKD mapping to Lucene / Solr later. Thus, it make sense to make that logic generalized, so that later we could reuse it.
I will work on adding this logic and update the PR.

porunov:
Codecov is wrong about `codecov/patch` test coverage. There are unit tests available and passing but seems codecov doesn't get covarage from some of them for some reason.

porunov:
@JanusGraph/committers I have intention to merge this PR using lazy consensus closer to Thursday. In case anyone wants to review this PR and need more time, please, let me know.

porunov:
Enabled auto-merge for lazy consensus merge

PR #3022, opened 2022-04-27
Add possibility to remove composite index entries

porunov:
With this change it will be possible to remove stale composite index entries from the graph.

Current limitations:
- Current approach didn't cover Mixed indices. I think it's possible to add this feature in the future PRs but I didn't dig into mixed index direct record changes yet.
- It's more challenging to change Edge composite index record but still possible in some cases. 
- Vertex centric indices were not covered. I guess it's more challenging but I didn't dig into changing entries of vertex centric indices yet. (not sure if that's needed at all)

Related to #1099

Signed-off-by: Oleksandr Porunov <alexandr.porunov@gmail.com>

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there an issue associated with this PR? Is it referenced in the commit message?
- [x] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [x] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered?

Issue #1099, opened 2018-06-06
Not able to delete ghost vertices!!


anamika121:
**Over the period of time we are observing some ghost vertices on our production setup where if we do gremlin query for having msid =18038893 shows ghost vertex.** 
gremlin> g.V().has('msid',18038893)
==>v[7244255240]

**but if we try to delete this vertex it shows that Vertex with id 7433638024 was removed.** 
gremlin> g.V().has('msid',18038893).drop()
Vertex with id 7244255240 was removed.
Type ':help' or ':h' for help.
Display stack trace? [

gremlin> g.V().has('msid',18869037).valueMap()
Vertex with id 7433638024 was removed.
gremlin> g.V().has('msid',18133305)
==>v[7228928032]

**I am not able to recreate vertex with given msid again in the graph. It seems that index vertex has already been deleted but index on msid has not been updated correctly and has been corrupted. Kindly suggest how to remove such cases.**

Miroka96:
Similar issue for me:

gremlin> v = g.V(2598101162).next()
==>v[2598101162]
gremlin> g.V(v).outE()
==>e[iq0ledl-16yubm2-3cwl-37s][2598101066-badHost->4168]
gremlin> g.V(v).outE().outV()
==>v[2598101066]
gremlin> g.V(v).outE().outV().next() == v
==>true

I think, my (unique) index is broken, because the same vertex is reachable from two different vertex IDs. And if I step from a vertex out to an edge and back to the outVertex, I get one which is equal but has a different vertex ID.

Maybe my unique Index prevents the duplicates from being found.
I have 400k vertices in my graph, created by only 160k unique data elements. They have been inserted through Spark in parallel with 96 threads.
My backend is Cassandra.

yosemiteAGM:
Have been having a similar issue:
We've been running a homegrown dataGrooming tool for over a year to look for ghost/phantom vertices and remove them from our graph.  Most of them seem to be caused by updates and deletes coming in on different db nodes (we run JanusGraph on HBase) at the same time.  

Up until recently, once the vertex was identified, we could just do vertex.remove() to get rid of it.  But now, while we get no errors running the remove() command and committing it, the vertices are still in the graph the next time I look.  (Like animika121's issue above).

I tried using g.V().drop() just as a change of pace, but that did not help.   Short of re-loading the db from a GraphSON snapshot, is there a way to get these vertices out of our graph?
[...]

PR #3327, opened 2022-11-18
Add possibility to remove stale mixed index entries for given vertices or edges

porunov:
Continuation of #3022

Related to #1099

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/advanced-topics/stale-index.md
+++ b/docs/advanced-topics/stale-index.md
@@ -18,13 +18,13 @@ As for now there is the utility tool which may be used to fix permanent stale in
 to fix permanent stale index entries. 
 
 `StaleIndexRecordUtil.forceRemoveVertexFromGraphIndex` can be used to force remove an index record for any vertex from 
-a graph index. Right now the limitations of this method is that it can be used with Composite indices only.  
+a graph index.   
 An example of using this method is below:
 ```java
 // Let's say we want to remove non-existent vertex from a stale index. 
 // We will assume the next constraints: 
 // Vertex id is: `12345`
-// Composite index name is: `nameAgeIndex`
+// Index name is: `nameAgeIndex`
 // There are two indexed properties: `name` and `age`
 // Value of the name property is: `HelloWorld`
 // Value of the age property is: `123`
@@ -41,7 +41,7 @@ StaleIndexRecordUtil.forceRemoveVertexFromGraphIndex(
     12345L, // vertex id of the index record to be removed
     indexRecordPropertyValues, // index record property values
     graph,
-    "nameAgeIndex" // composite index name for which to remove the index record
+    "nameAgeIndex" // graph index name for which to remove the index record
 );
 ```
-For more in-depth information about usage of this tool see JavaDoc for `StaleIndexRecordUtil`.
+For more in-depth information about usage of this tool as well as explanations of additional methods of this tool see JavaDoc for `StaleIndexRecordUtil`.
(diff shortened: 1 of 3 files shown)


li-boxuan (commented):
Looks good to me! I just have two comments:

1. Can you reword the PR title and the commit message? This PR makes it possible to remove certain (stale) mixed index entries for given elements (vertices & edges). The PR title itself reads like "now users can delete mixed indexes via JG".

2. Maybe I missed it, but do we have test cases where the forceRemoveElementFromMixedIndex would fail? For example, the given vertex is not indexed by the given mixed index. This can be due to: 1) someone already manually deleted the entry from the mixed index backend, 2) that index has a label constraint and the given vertex does not apply, or 3) the given vertex does not contain any property to be indexed under the given index. The latter two cases might be the same intrinsically.

porunov:
> Maybe I missed it, but do we have test cases where the forceRemoveElementFromMixedIndex would fail? For example, the given vertex is not indexed by the given mixed index. This can be due to: 1) someone already manually deleted the entry from the mixed index backend, 2) that index has a label constraint and the given vertex does not apply, or 3) the given vertex does not contain any property to be indexed under the given index. The latter two cases might be the same intrinsically.

I will add some tests to fail, but I'm not sure about the use-cases you suggest here. Explaining the cases below:

> For example, the given vertex is not indexed by the given mixed index.

It feels like an additional validation logic which we can validate and throw in case there is no entry. That said - I don't see much value in such logic. If users really need this validation logic then we can simply create a new method in this tool like `retrieveIndexRecords` which retrieves index records for a given element (possible for mixed indexes only) or `validateRecordExists` which validates that a record exists for a set of vertex id + provided properties (possible both for composite and mixed indexes). That said, this retrieval methods feels like a separate PR and I'm not sure what are the use cases of those methods.
At this moment - if a user tries to remove non-existing entries from the index - we don't throw anything. We just trigger removal of `nothing`.

> someone already manually deleted the entry from the mixed index backend.

For this we need to implement retrieval methods but again, it feels like a separate PR. Not 100% we need that but I guess in some cases some users may find it useful to verify there is a specific record in the index.

> that index has a label constraint and the given vertex does not apply.

This is similar to the above points. That said, if we are not talking about `fail` state but rather additional tests than it makes sense to add tests which remove records from indexes with label constraint. I will try to add such tests for mixed indices in this PR and will try to add more tests for composite indices in another PR.

 > the given vertex does not contain any property to be indexed under the given index.

I guess this case isn't fully valid because the main purpose of this util class is to add possibility to resolve those permanent inconsistency between storage backend and an index backend. Thus, even if the vertex doesn't have any indexed property it still may `had` those properties in the past which resulted in the stale index. 
In case you mean here that there is no index record exists for this id + properties then it's the same as your first use-case.

I will add some tests where I didn't cover a fail logic but not for the use cases you proposed. If you don't agree with any of the statements, let me know. I guess, I could implement verification / retrieval logic in this PR but I would prefer splitting that logic on multiple PRs.

li-boxuan:
> At this moment - if a user tries to remove non-existing entries from the index - we don't throw anything. We just trigger removal of nothing.

That (together with your subsequent points) sounds totally reasonable to me. I was just unsure what the behavior would be. The worst case would be the whole DB crash - but I am glad to hear it would simply be a no-op.

porunov:
@li-boxuan I renamed the commit and the PR title to `Add possibility to remove stale mixed index entries for given vertices or edges`.
I also added 3 additional tests:
`testForceRemoveElementFromMixedIndexShouldFailOnCompositeIndex` - fail test to check that we can't trigger `forceRemoveElementFromMixedIndex` for composite indices (i.e. only mixed indices).
`testForceRemoveElementFromMixedIndexShouldFailOnWrongPropertiesUsed` - fail test to check that we can use only those properties for mixed index records removal which exist in that index. I.e. you can't use properties which are not used in the index.
In this index I also demonstrate that multiple removals of non existent records are fine (they don't do anything other than making a redundant operation). 
Made `testStaleGraphIndexForceRemoveVertex` test parameterized so that it tests index record removal for both index with vertex label constraint and index without vertex label constraint.
Let me know if you think some more tests may be helpful as well.

li-boxuan (approved):
LGTM 👍

PR #3339, opened 2022-11-25
Use graph name for ES indices if index name is not configured

### LINKED ISSUE #3338: Elasticsearch should separate graphs by name when using ConfiguredGraphFactory
**Describe the feature:**
As stated in the [docs](https://docs.janusgraph.org/operations/configured-graph-factory/#important) JanusGraph's `ConfiguredGraphFactory.createTemplateConfiguration` is able to use the graph name as a storage directory, tablename or keyspace name for the storage backend. However, this is not the case for index backends. For example, when using elasticsearch, all indexed data ends up in one large index and therefore index queries from one graph also return matching vertices from other graphs.
Index backends should employ a mechanism similar to the one that is in use for storage backends.

Isolation of indices between multiple graphs.

rngcntr:
Fix #3338

The elasticsearch backend now uses the graph name by default if no index name is configured. This is relevant for users of `ConfiguredGraphFactory` who use template configurations and want to assign index names automatically upon graph creation.

Co-authored-by: Florian Hockmann <fh@florian-hockmann.de>

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -117,6 +117,16 @@ an older version will not be able to read a Geoshape created by a client that is
 versa.
 Users who do not use GraphBinary yet or who are not using Geoshapes are not affected by this change.
 
+##### ConfiguredGraphFactory now creates separate indices per graph in Elasticsearch
+
+If the ConfiguredGraphFactory is used together with Elasticsearch as the index backend, then the same Elasticsearch
+index is used for all graphs (if the same index names were used across different graphs).
+Now it is possible to let JanusGraph create the index names dynamically by using the `graph.graphname` if no `index.name`
+is provided in the template configuration. This is exactly like it was already the case for the CQL keyspace name for
+example.
+
+Users who don't want to use this feature can simply continue providing the `index.name` in the template configuration.
+
 ##### Removal of deprecated classes/methods/functionalities
 
 ###### Methods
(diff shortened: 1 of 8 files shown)


docs/changelog.md
```diff
@@ -117,6 +117,16 @@ an older version will not be able to read a Geoshape created by a client that is
 versa.
 Users who do not use GraphBinary yet or who are not using Geoshapes are not affected by this change.
 
+##### ConfiguredGraphFactory now creates separate indices per graph in Elasticsearch
+
+If the ConfiguredGraphFactory is used together with Elasticsearch as the index backend, then the same Elasticsearch
+index is used for all graphs (if the same index names were used across different graphs).
+Now it is possible to let JanusGraph create the index names dynamically by using the `graph.name` if no `index.name`
```
porunov, line 124: Isn't it `graph.graphname` instead of `graph.name`?

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -391,6 +392,11 @@ public ElasticSearchIndex(Configuration config) throws BackendException {
         setupStoredScripts();
     }
 
+    public static String determineIndexName(Configuration config) {
+        if ((!config.has(INDEX_NAME) && (config.has(GRAPH_NAME)))) return config.get(GRAPH_NAME);
```
porunov, line 396: _nitpick_
Looks like too many redundant braces. Could be simplified to:
```
if (!config.has(INDEX_NAME) && config.has(GRAPH_NAME)) return config.get(GRAPH_NAME);
return config.get(INDEX_NAME);
```
Or even using ternary operator: 
```
return !config.has(INDEX_NAME) && config.has(GRAPH_NAME) ? config.get(GRAPH_NAME) : config.get(INDEX_NAME);
```

porunov (commented):
Thank you @rngcntr ! Looks good to me but I have some nitpicks

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
rngcntr: Yes, you are right. I think we just copied this one from the CQL backend and we should fix it there as well.

porunov (approved):
LGTM. Thank you @rngcntr !

Issue #1333, opened 2018-11-30
Deadlock deleting 2 vertices in parallel


mbbeaubi:
To reproduce, have two vertices with an edge between them. Open 1 transaction. Create two threads. Have the first thread delete the first vertex while at the same time have the second thread delete the second vertex. This is a timing issue so you may need to set some breakpoints to get it to happen. Below is a thread dump of the issue happening in the wild:

```
"propagation--5" #113563 daemon prio=5 os_prio=0 tid=0x00007f93ac8fb000 nid=0x14186 waiting for monitor entry [0x00007f9351768000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.janusgraph.graphdb.vertices.StandardVertex.updateLifeCycle(StandardVertex.java:45)
	- waiting to lock <0x000000073e4a1438> (a org.janusgraph.graphdb.vertices.CacheVertex)
	at org.janusgraph.graphdb.vertices.StandardVertex.removeRelation(StandardVertex.java:51)
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.removeRelation(StandardJanusGraphTx.java:611)
	at org.janusgraph.graphdb.relations.CacheEdge.remove(CacheEdge.java:149)
	at org.janusgraph.graphdb.transaction.RelationConstructor$1$1.remove(RelationConstructor.java:67)
	at org.janusgraph.graphdb.vertices.AbstractVertex.remove(AbstractVertex.java:102)
	- locked <0x000000073e48f5e0> (a org.janusgraph.graphdb.vertices.CacheVertex)
	at org.janusgraph.graphdb.vertices.StandardVertex.remove(StandardVertex.java:101)
	- locked <0x000000073e48f5e0> (a org.janusgraph.graphdb.vertices.CacheVertex)
	at com.infa.products.ldm.janusgraph.nextgen.DeleteVertexHandler.handle(DeleteVertexHandler.java:35)
	at com.infa.products.ldm.janusgraph.nextgen.PostIngestHandler$HandlerCallable.call(PostIngestHandler.java:71)
	at com.infa.products.ldm.janusgraph.nextgen.PostIngestHandler$HandlerCallable.call(PostIngestHandler.java:1)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #3340, opened 2022-11-28
Optimize space for StandardVertex

mad:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/StandardVertex.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/StandardVertex.java
@@ -32,7 +32,6 @@ import org.janusgraph.util.datastructures.Retriever;
 
 public class StandardVertex extends AbstractVertex {
 
-    private final Object lifecycleMutex = new Object();
     private volatile byte lifecycle;
     private volatile AddedRelationsContainer addedRelations=AddedRelationsContainer.EMPTY;
 
@@ -41,10 +40,8 @@ public class StandardVertex extends AbstractVertex {
         this.lifecycle=lifecycle;
     }
 
-    public final void updateLifeCycle(ElementLifeCycle.Event event) {
-        synchronized(lifecycleMutex) {
-            this.lifecycle = ElementLifeCycle.update(lifecycle,event);
-        }
+    public final synchronized void updateLifeCycle(ElementLifeCycle.Event event) {
+        this.lifecycle = ElementLifeCycle.update(lifecycle, event);
     }
 
     @Override


rngcntr:
LGTM. Have you searched the repo for any other occurrences of this pattern?

porunov (approved):
LGTM. Thank you @mad !

porunov:
> LGTM. Have you searched the repo for any other occurrences of this pattern?

I search for other places of such pattern and found only a single place where we create a mutex object which isn't necessary (other than what changed in this PR). It's `JanusGraphManager.instantiateGraphLock`. That said, that place is relevant for schema management only. I don't think that would make a big difference.
In all other places where `synchronized` is used it uses reusable objects. Thus, should be fine.

li-boxuan:
Isn't this PR basically a revert of https://github.com/JanusGraph/janusgraph/pull/1380?

That said, I don't see why #1380 was necessary in the first place, so I think we could merge this PR.

rngcntr:
I took a look at the original issue (#1333) from the PR you mentioned. The reason for the original change seems to have been an issue caused by the two `synchronized` methods `updateLifeCycle` and `remove`. 

The issue mentions steps to reproduce the bug, so to ensure that the locking works as intended here, you could implement these steps as a test case and compare both locking mechanisms using the test.

rngcntr:
Upon further investigation I think the original problem occurred when two vertices connected by an edge were deleted concurrently by different threads. As `remove()` is also `synchronized` and calls `updateLifeCycle()`, both threads started the execution of `remove()` and were then unable to call each other's `updateLifeCycle()` because the other thread is already running within a synchronized environment on the same object.

The introduction of the `lifecycleMutex` object fixed this behavior by ensuring that concurrent access to `remove()` and `updateLifeCycle` can be granted without having to grant concurrent access on `updateLifeCycle()` itself.

porunov (changes requested):
Good @li-boxuan found the original issue because I totally missed that.
I agree with @rngcntr here. This optimization introduces a bug for concurrent vertices removal. Thus shouldn't be merged.

porunov:
On the other hand it's strange that this test didn't fail in this PR: #1486

rngcntr:
As far as I can see, the introduced bug is a race condition and since the test can not enforce the exact order of execution, it can not reliably detect the race condition.

li-boxuan:
> when two vertices connected by an edge were deleted concurrently by different threads. As remove() is also synchronized and calls updateLifeCycle(), both threads started the execution of remove() and were then unable to call each other's updateLifeCycle()

AFAIK this should never happen because `it()` in `remove()` function only refers to the **current vertex** in the new tx context, so it will not interfere with another vertex connected by the same edge.

rngcntr:
#1333 contains an example that supposedly causes a deadlock. I have not tried to reproduce it, however.

porunov:
> AFAIK this should never happen because it() in remove() function only refers to the current vertex in the new tx context, so it will not interfere with another vertex connected by the same edge.

Could be that something changed. I do remember I was able to reproduce the issues that time. 
I had a test here: https://github.com/JanusGraph/janusgraph/pull/1380#issuecomment-467964531
Which was added later in this PR: https://github.com/JanusGraph/janusgraph/pull/1486

That time I remember the main issue was that when you remove a vertex it first checks all the edges to be removed and removes those edges as well.
Now imagine if we have 2 vertices linked by a normal bi-directed edge. If we start removing both those edges in parallel using the same transaction (i.e. non thread bound transaction) then there is a chance to get a deadlock because vertex `A` wants to remove the edge (which should be removed in vertex `B`). Vertex `B` wants to perform the same operation but vice versa (i.e. remove an the from vertex `A`). 
As far as I remember this was the main problem which was caused by the deadlock, but I could be wrong.

So, basically here I would propose to take a look not at `updateLifeCycle` method but at `remove` method.
```
@Override
public synchronized void remove() {
    super.remove();
    ((StandardVertex)it()).updateLifeCycle(ElementLifeCycle.Event.REMOVED);
}
```

This method is `synchronized`. If we go to `super.remove();` we can see that it traverses all the relations of this vertex and removes them. When we are removing a relation we trigger `removeRelation` of another vertex and `removeRelation` may trigger `updateLifeCycle` of that vertex. If both `remove` and `updateLifeCycle` are synchronized on the same object then there could potentially be a deadlock unless I miss something.

I don't know why the added test isn't failing - could be a flaky/wrong test or something was changed which makes a deadlock impossible in this situation.

-----

Another interesting thing, unrelated to this issue is the implementation of that `super.remove()`:
```
Iterator<JanusGraphRelation> iterator = it().query().noPartitionRestriction().relations().iterator();
while (iterator.hasNext()) {
    iterator.next();
    iterator.remove();
}
```
Even so, eventually `iterator.remove()` triggers removal it's not obvious at all because it's just how our `ResultSetIterator` implements `remove` method. Usually `iterator.remove()` just removes an element from the collection and not triggering anything in the current object. In `ResultSetIterator` we made that `remove()` actually triggers `remove()` of the current element. Thus, `iterator.remove()` magically works but it feels strange for some reason. Maybe I'm wrong but it feels that it should be `iterator.next().remove();`.

mad:
This test catch this deadlock

```
    @RepeatedTest(10)
    public void test() throws ExecutionException, InterruptedException {
        try (JanusGraph g = JanusGraphFactory.open("inmemory")) {
            final JanusGraphVertex v1 = g.addVertex("V1");
            final JanusGraphVertex v2 = g.addVertex("V2");
            v1.addEdge("E", v2);
            g.tx().commit();

            final JanusGraphTransaction tx = g.newTransaction();
            List<CompletableFuture<Void>> futures = new ArrayList<>();
            for (int i = 0; i < 2; i++) {
                final CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
                    final List<Vertex> vertices = Lists.newArrayList(tx.vertices());
                    if (RandomUtils.nextBoolean())
                        Collections.reverse(vertices);
                    for (Vertex vertex : vertices) {
                        vertex.remove();
                    }
                });
                futures.add(future);
            }

            CompletableFuture.allOf(futures.toArray(new CompletableFuture[] { })).get();
        }
    }
```

image
image

PR #3341, opened 2022-11-28
Fix NO_LIMIT handling for key-value stores

mad:
Without fix could not reindex store with more than Integer.MAX_VALUE entries because `KeySelector.limit` is integer

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/keyvalue/KeySelector.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/diskstorage/keycolumnvalue/keyvalue/KeySelector.java
@@ -18,6 +18,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import org.janusgraph.diskstorage.StaticBuffer;
+import org.janusgraph.graphdb.query.Query;
 
 /**
  * A {@link KeySelector} utility that can be generated out of a given {@link KVQuery}
@@ -50,7 +51,7 @@ public class KeySelector {
     }
 
     public boolean reachedLimit() {
-        return count >= limit;
+        return Query.NO_LIMIT != limit && count >= limit;
     }
 
 }
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/LimitAdjustingIterator.java
@@ -67,7 +67,7 @@ public abstract class LimitAdjustingIterator<R> implements CloseableIterator<R> 
     @Override
     public boolean hasNext() {
         if (iterator ==null) iterator = getNewIterator(currentLimit);
-        if (count < currentLimit)
+        if (currentLimit == Query.NO_LIMIT || count < currentLimit)
             return iterator.hasNext();
         if (currentLimit>=maxLimit) return false;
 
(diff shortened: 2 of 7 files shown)


porunov (approved):
LGTM. Thank you @mad !

FlorianHockmann:
> Without fix could not reindex store with more than Integer.MAX_VALUE entries because KeySelector.limit is integer

If that is the problem, can't we then simply switch from integer to long?

Also, is it possible (with reasonable effort) to write a test that shows this problem?

porunov:
> If that is the problem, can't we then simply switch from integer to long?

I was trying to quickly switch from `int` to `long` and it seems to be a little bit more complicated because too many JanusGraph interfaces and classes relay on `int` specifically. In some places `long` doesn't even make sense (in case when we want to compare `limit` to collection size or resize some array because in Java arrays and standard collections may have up to `int` elements and no more). I would say switching to `long` is possible and maybe we should do that but we need to revise all places where it's used and there is a chance it will introduce a breaking change in some places. That said, I didn't spend enough time investigating the issue.

> Also, is it possible (with reasonable effort) to write a test that shows this problem?

I personally don't know how to write a meaningful test to check reindexing of `2^32` elements. I guess could be possible to write unit tests for classes which are changed. 
I can't say I'm sure that in all changed places the check is necessary because I didn't investigate this issue enough, but it looks to me that those checks are either fixing the problem or they are simply redundant (but harmless) checks. 
Thus, I am OK with merging this PR without tests if @mad confirms it fixes the problem with re-indexing of > `2^32` elements.
I personally don't have any specific preference on how to resolve the issue. Thus, take my thoughts above with a graint of salt.

porunov:
@FlorianHockmann @mad 

I believe this PR fixes the issue. I didn't come up with a good test to verify the issue, but I was able to make a simple test which might be useful to understand the issue.
I tested the below logic in `master` branch (failed) and in this branch (passed).
The testing logic was the next:
1. In every file where @mad introduced additional check on `hasLimit()` I changed the assuming starting point to `Integer.MAX_VALUE - 1000`.
2. Made a test which adds 2000 vertices into the graph in 4 transactions (500 each).
3. Create an index which covers the added vertices.
4. Reindex data using the newly added index.
5. Check that count from index and full-scan count return the same amount of elements.
6. Check that count from index return 2000 elements.

The last step fails in master and returns only 1000 elements.
- Index aggregation count query returns only 1000 elements which verifies that REINDEX operation didn't process all vertices.
- Full-scan query returns 1000 elements which verifies that no full-scan queries can process more than `Integer.MAX_VALUE` elements.

Test which fails against latest `master` branch is available in [this branch](https://github.com/porunov/janusgraph/tree/test/master-no-limit-reindex-problem) ([commit](https://github.com/porunov/janusgraph/commit/01f532c84207ea6e3eae34aa23d6f6c2a3eb4bb9)).
Test which passes against this PR branch is available in [this branch](https://github.com/porunov/janusgraph/tree/test/mads-fix-for-no-limit-reindex-problem) ([commit](https://github.com/porunov/janusgraph/commit/a6f85f0e2b9e06f33c0e9c1f28ce3f7b38c7b801))

Test is called: `testReindexOfDataSetOutOfIntLimitSize()`.

Notice that we can't include that test into this PR because that test assumes we modified core code to have starting count as `Integer.MAX_VALUE - 1000` instead of `0`.

I believe we should change all limits to `long`, but as I noticed [here](https://github.com/JanusGraph/janusgraph/pull/3341#issuecomment-1420029769) it has some complexity. 
Thus, I propose to merge this PR and let the fix lend into `v0.6` branch.

FlorianHockmann:
Thanks for your detailed investigation and confirming that the changes here fix the problem with a test.

I really didn't want to cause so much trouble with my question whether we could add a test here 😅 

But we can definitely merge this from my side.

porunov:
> Thanks for your detailed investigation and confirming that the changes here fix the problem with a test.
> 
> I really didn't want to cause so much trouble with my question whether we could add a test here sweat_smile
> 
> But we can definitely merge this from my side.

You comment about a test case and `long` instead of `int` definitely make sense. I also created an issue to transition to `long` limit #3603

Issue #3344, opened 2022-11-29
Inefficient vertex deserialization by JanusGraphVertexDeserializer


ministat:
HadoopRecordReader uses JanusGraphVertexDeserializer to parse the data and deserialize it to TinkerGraph first, then create StarGraph by VertexWritable. In fact, this can be simplified to deserialize to StarGraph directly.

In other words, the deserialization process "parsing data -> TinkerGraph -> StarGraph" can be optimized to "parsing data -> StarGraph", which readuces the memory pressure.

For confirmed bugs, please report:

- Version: 0.6.2 and before 0.6.2 
- Storage Backend: Any backend which HadoopInputFormat can read
- Mixed Index Backend: None
- Link to discussed bug: 
- Expected Behavior: Remove the middle object: TinkerGraph
- Current Behavior: JanusGraphVertexDeserializer deserializes the data to TinkerGraph, and transforms it to StarGraph after that.
- Steps to Reproduce:
Use janusgraph console to load the data from, for example, HBase backend. The JanusGraphVertexDeserializer will be applied, and the above mentioned process will be triggered.

### Stack Trace (if you have one)

```
paste stack trace here
```

PR #3346, opened 2022-11-29
Fix the issue [#3344]

ministat:


li-boxuan:
> This process is like "parse data -> create TinkerVertex -> create StarVertex", which can be simplified to be "parse data -> create StarVertex".

I am not sure what you mean. Do you mind elaborating? To me, it seems like the only change is we use `StarGraph` and `StarVertex` rather than `TinkerGraph` and `TinkerVertex`.

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/HadoopRecordReader.java
```diff
@@ -65,7 +65,7 @@ public void initialize(final InputSplit inputSplit, final TaskAttemptContext tas
     public boolean nextKeyValue() throws IOException, InterruptedException {
         while (reader.nextKeyValue()) {
             // TODO janusgraph05 integration -- the duplicate() call may be unnecessary
-            final TinkerVertex maybeNullTinkerVertex =
+            final Vertex maybeNullTinkerVertex =
```
li-boxuan, line 68: nitpick: we should rename this variable as it is now not a "TinkerVertex".

ministat: ack

ministat:
retest please

li-boxuan (approved):
LGTM 👍  Don't worry about CI, I don't think it's related to your changes

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
```diff
@@ -77,7 +74,7 @@ private boolean edgeExists(Vertex vertex, RelationType type, RelationCache possi
 
     // Read a single row from the edgestore and create a TinkerVertex corresponding to the row
```
FlorianHockmann, line 75: (nitpick) Please also update this comment as it explicitly mentions that a `TinkerVertex` will be created:

```suggestion
    // Read a single row from the edgestore and create a Vertex corresponding to the row
```

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
```diff
@@ -92,9 +89,8 @@ public TinkerVertex readHadoopVertex(final StaticBuffer key, Iterable<Entry> ent
         }
 
         // Create TinkerVertex
```
FlorianHockmann, line 91: Same nit here:

```suggestion
        // Create vertex
```

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
```diff
@@ -105,19 +101,19 @@ public TinkerVertex readHadoopVertex(final StaticBuffer key, Iterable<Entry> ent
                 long vertexLabelId = ((Number) relation.getOtherVertexId()).longValue();
                 VertexLabel vl = typeManager.getExistingVertexLabel(vertexLabelId);
                 // Create TinkerVertex with this label
```
FlorianHockmann, line 103: and here:
```suggestion
                // Create vertex with this label
```

FlorianHockmann (commented):
Thanks for the contribution! The changes also look good to me, I just added some inline comments where I think the comments should also be updated as they are now a bit misleading.

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
ministat: Ack

janusgraph-hadoop/src/main/java/org/janusgraph/hadoop/formats/util/JanusGraphVertexDeserializer.java
ministat: Ack

ministat:
retest please

FlorianHockmann (approved):

PR #3362, opened 2022-12-02
Refactor Index Management and support removal of Mixed Indexes

### LINKED ISSUE #460: Improve Index Documentation

### LINKED ISSUE #857: Re-enable a disabled index
Currently, disabling an index is only a precursor to removing it. However, a user may want to turn on/off indexes for various reasons - performance testing, maintenance, etc.

While an index could be removed and recreated, the docs imply that it can not have the same name. The docs state `Index removal deletes everything associated with the index except its schema definition and its DISABLED state. This schema stub for the index remains even after deletion`. This is not a good alternative.

There should be a way to re-enable a disabled index and alternatively, remove every last remnant of the index so that it could be recreated with the same name.

rngcntr:
**tldr;** new index management, here it is
---

**Longer version:**
---

This PR fixes #857, fixes #460 and makes the following changes:
- Allows complete removal of indexes including their schema vertices.
- As a consequence, the names of previously deleted indices can now be reused.
- Adds automatic deletion of mixed indices in Elasticsearch
- Indices in Lucene and Solr have to be deleted manually, but once done, the schema management can take over again and handle the rest.
- Indices that have been disabled can be re-enabled again. Applies to both composite and mixed indices.
- Notable changes to the docs.

Potentially breaking changes:
- Schema actions for index deletion have changed.
- Introduced the `DISCARDED` index state. However, note that no state has been removed. Therefore reading states written by earlier versions of JanusGraph should work flawlessly.

Further work:
- As stated earlier, I only implemented index deletion for ES but not for Lucene and Solr so far. @porunov [noted](https://github.com/JanusGraph/janusgraph/issues/857#issuecomment-1314987013) that solutions for those backends should be feasible as well.

### For all changes:

### For code changes:


porunov (commented):
Thank you @rngcntr ! I love the new index management lifecycle. It is definitely more thoughtful and more developed than the one we have right now. 

I have just small comments / questions regarding documentation, some tests and `MARK_DISCARDED` action below. Otherwise LGTM.

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,75 @@ mgmt.commit();
...
+:   Removes all data from the index, preparing it for removal.
+    This operation is supported for all composite indices and for some mixed index backends.
+    For more information on index removal see [Removal](index-removal.md).
 
--   **ENABLE_INDEX** Enables the index so that it can be used by the query processing engine. An index must be registered before it can be enabled.
+**MARK_DISCARDED**
+:   All indices which do not support **DISCARD_INDEX** have to be discarded manually in the backend.
+    As this is not recognized by JanusGraph automatically, the state can be set to **DISCARDED** manually via this action.
+    This operation is supported for all composite indices and for some mixed index backends.
+    For more information on index removal see [Removal](index-removal.md).
```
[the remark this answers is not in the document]
rngcntr: `MARK_DISCARDED` can be used with all index backends, including Elasticsearch. I agree that the comment is misleading and I will fix that. Thanks for the heads-up

janusgraph-lucene/src/test/java/org/janusgraph/diskstorage/lucene/BerkeleyLuceneTest.java
```diff
@@ -71,6 +78,49 @@ protected boolean supportsGeoCollections() {
         return false;
     }
 
+    @Test
+    public void someTestThatIDoNotWantToKeep() throws Exception {
```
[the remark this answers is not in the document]
rngcntr: Ouch, the name of the test even tells that it was only meant for debugging purposes and I obviously forgot to implement it properly :P

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
```diff
@@ -3749,4 +3754,169 @@ private GraphTraversal<Vertex, Vertex> getV(String propertyName, Object value, S
         }
         return traversal.has(propertyName, value);
     }
+
+    @Test
+    public void someTestThatIDoNotWantToKeep() throws Exception {
```
porunov, line 3759: Probably should now be called `testMarkDiscardedRegisteredIndex` or something like that :)

docs/schema/index-management/index-removal.md
```diff
@@ -162,25 +164,100 @@ g.V().has('name', 'jupiter')
...
-```
\ No newline at end of file
+```
+
+## Executing a manual index removal for mixed indices
+
+!!! info
+    This method applies to the following types of indices:
+
+    - Mixed indices (Lucene and Solr)
```
porunov, line 198: Any mixed index can be used. I.e. ElasticSearch as well

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
rngcntr: I'm just too quick pushing commits I haven't reviewed myself...

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,72 @@ mgmt.commit();
 
 ## Index states and transitions
 
-![States and transitions](index-lifecycle.png)
+The diagram below shows the possible states of an index in JanusGraph.
+Transitions which are not recommended or should be used with caution are shown in grey.
```
FlorianHockmann, line 30: I don't like the formulation "are not recommended" very much to be honest. Maybe "are not recommended in production" or something like that?

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,72 @@ mgmt.commit();
...
+:   The index has been created on the current JanusGraph instance, but it is not yet known to all instances in the cluster.
+    This state is not shown in the diagram above because it is not stable in terms of synchronization between instances.
 
--   **ENABLED** The index is enabled and in use
+**REGISTERED**
 
--   **DISABLED** The index is disabled and no longer in use
+:   The index is known to all instances in the cluster.
+    However, it is not in use yet.
+    In this state, the index contains no data and is not eligible to be used in queries.
```
FlorianHockmann, line 46:
```suggestion
    In this state, the index contains no data and will not be used in queries.
```

(My reasoning here is that JanusGraph is used by many people who aren't native English speakers so we should make the docs easy to understand.)

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,72 @@ mgmt.commit();
...
--   **DISABLE_INDEX** Disables the index in the graph so that it is no longer used.
+:   The index has been decommissioned and its contents have been deleted.
+    The only remaining artifact is the schema vertex representing the index.
+
+## Actions (SchemaAction)
+The following actions can be performed on an index to change its state via `mgmt.updateIndex()`:
 
--   **REMOVE_INDEX** Removes the index from the graph (optional operation). Only on composite index.
+**REGISTER_INDEX**
+:   Registers the index with all instances in the graph cluster.
```
FlorianHockmann, line 68: I think graph cluster is a confusing term. The next sentence even speaks of _graph instances_ which should be _JanusGraph instances_.
Not sure what to do with _graph cluster_. Maybe just drop the _graph_?

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,72 @@ mgmt.commit();
...
+**DISCARD_INDEX**
+:   Removes all data from the index, preparing it for removal.
+    This operation is supported for all composite indices and for some mixed index backends.
+    For more information on index removal see [Removal](index-removal.md).
+
+**MARK_DISCARDED**
+:   All indices which can not be discarded using **DISCARD_INDEX** have to be discarded manually in the backend.
+    As this is not recognized by JanusGraph automatically, the state can be set to **DISCARDED** manually via this action.
+
+**DROP_INDEX**
```
FlorianHockmann, line 95: Wouldn't it be possible to merge the drop action into the two discard actions and then remove the `DISCARDED` state completely? Or what's the purpose of the `DISCARDED` state except of waiting for the user to execute this action?

docs/schema/index-management/index-removal.md
```diff
@@ -7,24 +7,20 @@
...
+In the first stage, one JanusGraph instance deactivates the index by setting its state to `DISABLED`.
+At that point, JanusGraph stops using the index to answer queries and stops incrementally updating the index.
+Index-related data in the storage backend remains present but ignored.
+
+The second stage depends on whether the index is mixed or composite.
+All composite indices can be deleted via JanusGraph.
+As with reindexing, removal can be done through either MapReduce or JanusGraphManagement.
+However, not all mixed index backends allow automatic removal.
+Index backends for which JanusGraph does not provide an automated deletion mechanism — currently Lucene and Solr — must be manually dropped in the index backend.
+For supported mixed index backends, the removal process is analogue to the composite index removal, with the exception that neither MapReduce nor multi threading in JanusGraphManagement is necessary.
```
FlorianHockmann, line 20: Please wrap long lines. We usually use 120 as the max length.

docs/schema/index-management/index-removal.md
```diff
@@ -98,41 +94,47 @@ g.V().has('name', 'jupiter')
...
-m.updateIndex(nameIndex, SchemaAction.DISABLE_INDEX).get()
+m.updateIndex(m.getGraphIndex('name'), SchemaAction.DISABLE_INDEX).get()
 m.commit()
 graph.tx().commit()
 
-// Block until the SchemaStatus transitions from INSTALLED to REGISTERED
+// Block until the SchemaStatus transitions from ENABLED to DISABLED
 ManagementSystem.awaitGraphIndexStatus(graph, 'name').status(SchemaStatus.DISABLED).call()
 
 // Delete the index using MapReduceIndexJobs
```
FlorianHockmann, line 104: nitpick, this doesn't delete the index itself, so maybe:

```suggestion
// Delete the indexed data using MapReduceIndexJobs
```

FlorianHockmann (commented):
I've only reviewed the docs so far and they look good so far. Only added a few minor comments.
Hope that I can come back to this next week to review the code changes.

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,72 @@ mgmt.commit();
...
+Reindexing is possible in all stable states and automatically enables the index once finished.
+
+**ENABLE_INDEX**
+:   Enables the index so that it can be used by the query processing engine.
+An index must be registered before it can be enabled.
+If enabling the index manually instead of performing a reindex, be aware that past modifications of the graph are not represented in the index.
+Enabling a previously disabled index without `REINDEX` may cause index queries to return ghost vertices, which have been deleted from the graph while the index was disabled.
+
+**DISABLE_INDEX**
+:   Disables the index temporarily so that it is no longer used to answer queries.
```
li-boxuan, line 83: Why is `DISABLE_INDEX` from `ENABLED` to `DISABLED` not in grey, but `ENABLE_INDEX` from `DISABLED` to `ENABLED` in grey?

docs/schema/index-management/index-removal.md
```diff
@@ -162,25 +164,100 @@ g.V().has('name', 'jupiter')
...
+```
+
+## Executing a manual index removal for mixed indices
+
+!!! info
+    This method applies to the following types of indices:
+
+    - Mixed indices which can not be removed by Janusgraph automatically
+
+If an index backend does not support automatic removal, you will receive an exception when trying to execute `DISCARD_INDEX`.
```
li-boxuan, line 200: IIUC, currently ES supports automatic removal, no? In that case can we document it here?

docs/schema/index-management/index-lifecycle.md
```diff
@@ -26,28 +26,72 @@ mgmt.commit();
...
+The index also stops receiving updates in the graph.
+
+**DISCARD_INDEX**
+:   Removes all data from the index, preparing it for removal.
+    This operation is supported for all composite indices and for some mixed index backends.
+    For more information on index removal see [Removal](index-removal.md).
+
+**MARK_DISCARDED**
```
li-boxuan, line 91: Two questions:
1. Why is this action marked grey?
2. What happens if the user doesn't delete the real indices from the index backend manually, but still runs this action?

li-boxuan (commented):
This is such a great and must-have functionality. Glad finally some progress is made. I haven't got a chance to review the code yet, so I just have some comments on the doc.

FlorianHockmann (commented):
Just reviewed the code changes. I added some inline comments, but the changes look really good to me. My comments are mostly nits or questions I had during the review. So, feel free to ignore those comments.

docs/schema/index-management/index-removal.md
rngcntr: As I may have stated somewhere else already, I'm a big fan of semantic wrapping (per sentence) because it makes it so much easier to spot the differences in reviews. With a hard cap at 120 characters, changing a single word and re-wrapping makes it look like the entire paragraph has changed.
But if that really is the project's default, I will redo this and wrap the lines.

docs/schema/index-management/index-lifecycle.md
rngcntr: The idea behind the grey transitions was to give an intuition for consequences to be aware of. Disabling an index does exactly what everybody probably expects, but plain enabling without reindexing could lead to faulty index responses. One has to keep in mind that ENABLE does not just reverse a prior DISABLE and that may not be obvious to everybody at first glance.

docs/schema/index-management/index-lifecycle.md
rngcntr: Let me answer 2. first:
If the user performs this step without deleting the indexed data, all of it remains in the index backend of course. At the same time, it looks to JanusGraph as there this index does not exist and may never have. Therefore, JanusGraph would happily allow creating a new index with the same name that already exists in the backend. Even though I have not yet bothered to find out exactly __how__ things misbehave, I still consider this highly unsafe.

Now to 1.:
Grey means that an operation has to be used with care and may carry unforeseen consequences. In the case of this action, the user must be aware of the implications I explained in the previous answer (2)

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -2116,7 +2242,7 @@ public void testIndexShouldRegisterWhenWeRemoveAnInstance() throws InterruptedEx
             mgmt.updateIndex(mgmt.getGraphIndex("theIndex"), SchemaAction.ENABLE_INDEX));
         finishSchema();
 
-        //close second graph instance, so index can move to REGISTERED
+        //close second graph instance, so index can move to CLEAN
```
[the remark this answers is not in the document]
rngcntr: You're right, I refactored the naming multiple times and seemingly forgot to update this line

docs/schema/index-management/index-removal.md
FlorianHockmann: I agree with semantic wrapping, but I think that also wrapping at 120 characters makes sense as longer lines aren't nice to work with in a diff since they are just too long then to be displayed (depending on the screen size & resolution of course).
We usually use 120 characters as a limit in JanusGraph so I think we should stay with that if we don't have good reasons against it. Additionally wrapping each sentence makes sense in my opinion and should already ensure that most lines will be below the 120 char limit.

docs/schema/index-management/index-lifecycle.md
rngcntr: I already tried that, but there is currently no mechanism in place to properly wait for the long running index removal and schedule the deletion of the internal vertex (aka `DROP_INDEX`) afterwards.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -1684,25 +1811,24 @@ public void testGotGIndexRemoval() throws InterruptedException, ExecutionExcepti
...
 
         // Remove name index
-        graphIndex = mgmt.getGraphIndex(name);
-        mgmt.updateIndex(graphIndex, SchemaAction.REMOVE_INDEX);
-        ScanJobFuture graphMetrics = mgmt.getIndexJobStatus(graphIndex);
-        finishSchema();
+        ScanJobFuture future = mgmt.updateIndex(mgmt.getGraphIndex(name), SchemaAction.DISCARD_INDEX);
+        mgmt.commit();
 
+        ManagementSystem.awaitGraphIndexStatus(graph, name).status(SchemaStatus.REGISTERED).timeout(10, ChronoUnit.SECONDS).call();
```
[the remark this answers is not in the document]
rngcntr: Yes. The test still passed because the await ran into the default timeout but the success wasn't checked.

rngcntr:
Thanks for your suggestions @FlorianHockmann! I'm always impressed by the depth and thoroughness of your reviews.
Added another commit to make it easier to review the changes.

docs/schema/index-management/index-lifecycle.md
FlorianHockmann: Thanks for the explanation. Then I'm of course good with this. We can still try to improve this at some point in the future, but then it's out of scope for this PR.

docs/schema/index-management/index-lifecycle.md
li-boxuan: That makes sense. Would be even better if you could add a note/warning that one should not do `MARK_DISCARDED` without physically removing the indices.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -910,22 +907,39 @@ public ScanJobFuture updateIndex(Index index, SchemaAction updateAction, int num
                 future = new EmptyScanJobFuture();
                 break;
             case DISABLE_INDEX:
-                setStatus(schemaVertex, SchemaStatus.INSTALLED, keySubset);
+                setStatus(schemaVertex, SchemaStatus.DISABLED, keySubset);
```
li-boxuan, line 910: Do you know why the old code set the schema vertex to `INSTALLED` first and then set an update trigger, and why we don't do it here?

rngcntr: I noticed that and failed to come up with reasons why it would be beneficial to set the state to `INSTALLED` first. A little bit more background:

- `setStatus` modifies the schema vertex to update the status property. This happens within a transaction and is not written to the storage backend until the management transaction is committed.
- The update trigger is applied during the commit phase of a management transaction. Therefore in the old version, the status is set to `INSTALLED` and immediately overwritten to `DISABLED` when committing the management transaction.

Consequently, the only difference is the state in which we see the index within our uncommitted transaction. I can only speculate about the intended effects of this:
- Maybe the authors wanted to prevent a subsequent call to `REMOVE_INDEX` within the same transaction. That would have been prevented because we see the index as `INSTALLED` after calling `DISABLE_INDEX` until the transaction is committed.
- That, however, would introduce the ability to erroneously call `REGISTER_INDEX` as long as the transaction is still open and have the state end up in `REGISTERED`, which is unexpected behavior and should not be possible according to the index lifecycle.

Both of these arguments still apply to the updated version in a similar way. Instead of exchanging one possibly unwanted effect for another, I opted for readability of the code.

FlorianHockmann (approved):

FlorianHockmann:
The changes look good to me, but some tests are now failing. Some of those are flaky tests and I restarted them, but the others are probably related to changes here.

rngcntr:
Thanks @FlorianHockmann. Permanent failures can only be related to changes from the last commit since all tests were passing before applying the suggestions from the review. I already noticed that `testDiscardAndDropIndexAfterRegistering` is repeatedly failing in the CI even though it passes locally. I'll have to take a closer look at that one.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
li-boxuan: That makes sense, thanks for the detailed explanation!

rngcntr:
Tests are now passing and commits are squashed.

FlorianHockmann:
> Tests are now passing and commits are squashed.

I just reviewed your recent updates and they look good to me. So, everything's good from my side here.

PR #3357, opened 2022-12-01
Publish snapshot packages to GitHub

porunov:
Discussion: https://lists.lfaidata.foundation/g/janusgraph-dev/topic/discussion_release_snapshot/84922650

Signed-off-by: Oleksandr Porunov <alexandr.porunov@gmail.com>

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered?

PR #3364, opened 2022-12-05
Publish commit packages to Sonatype

porunov:
Replaces the next PR: #3357

For the snapshot releases the next official Sonatype plugging is used: [nexus-maven-plugins](https://github.com/sonatype/nexus-maven-plugins/blob/main/staging/maven-plugin/README.md).
We should probably replace our deployment process for official releases to use this plugin as well, but it's better to do that in following PRs.

Discussion: https://lists.lfaidata.foundation/g/janusgraph-dev/topic/discussion_release_snapshot/84922650

-----

### For all changes:

### For code changes:


FlorianHockmann:
I'm unfortunately not really an expert on Java / Maven packaging and how snapshots are usually consumed in this environment.
But isn't the usual Java way to just deploy packages with the same groupId, just with a snapshot version, and then to a different repository? That way, users can switch from official releases to snapshot releases by configuring a different repository to consume the packages from. They don't have to change anything else.
That at least seems to be the usual way to handle snapshot packages from a quick search. That's also how it's for example described [here](https://www.baeldung.com/maven-snapshot-release-repository).

Is your main argument to deviate from this that users have to create a token to access GitHub packages?

Just to clarify: I'm not completely against using a different groupId for snapshots. I just want to make sure that we're not trying to build a solution that is very unusual for Java users.

Also do we need to replace the `SNAPSHOT` version ourselves? That means that users will have to explicitly specify that version, right? I thought that the Java way to handle snapshots is to simply use the special `SNAPSHOT` tag which allows to create new versions of it without users needing to update the version every time.

porunov:
> I'm unfortunately not really an expert on Java / Maven packaging and how snapshots are usually consumed in this environment. But isn't the usual Java way to just deploy packages with the same groupId, just with a snapshot version, and then to a different repository? That way, users can switch from official releases to snapshot releases by configuring a different repository to consume the packages from. They don't have to change anything else. That at least seems to be the usual way to handle snapshot packages from a quick search. That's also how it's for example described [here](https://www.baeldung.com/maven-snapshot-release-repository).
> 
> Is your main argument to deviate from this that users have to create a token to access GitHub packages?
> 
> Just to clarify: I'm not completely against using a different groupId for snapshots. I just want to make sure that we're not trying to build a solution that is very unusual for Java users.
> 
> Also do we need to replace the `SNAPSHOT` version ourselves? That means that users will have to explicitly specify that version, right? I thought that the Java way to handle snapshots is to simply use the special `SNAPSHOT` tag which allows to create new versions of it without users needing to update the version every time.

I think the confusion is driven by by the word `snapshot`. I guess we can replace this word to any other suitable word, but the idea was not to have SNAPSHOT releases which users could relay on (the usual way you are referencing to). Instead, the idea was to publish a new release per each commit. The user could just add that dependency and use it without worrying that tomorrow some new breaking change is introduced to SNAPSHOT release and their code breaks because of that. The idea was to have fixed releases per each commit. The user just select a commit they are interested in and simply use the release available for that commit (it could be latest commit or the older one).

Instead of the word `snapshot` we can use other words like: `commit-release`, `unstable`, `test-release`, `rc`, something else. I couldn't find a good word for that, so I used `snapshot` word.

> Is your main argument to deviate from this that users have to create a token to access GitHub packages?

I think for users it's easier to use repository which doesn't require authentication. Thus, my preference would be Sonatype over GitHub but I'm OK with GitHub as well.

> Just to clarify: I'm not completely against using a different groupId for snapshots. I just want to make sure that we're not trying to build a solution that is very unusual for Java users.

If we use the same (maven central) repository, then we will need to use a different groupId. Otherwise we will just mix different types of releases under the same groupId which seems strange.
This PR isn't building a general SNAPSHOT solution but instead building a solution which makes per-commit releases.
We can add SNAPSHOT solution as well if users interested in that.
If so, we could have the next releases:
- Official releases (our normal releases). I.e. release such as `0.6.3`.
- Per-commit releases with different groupId (or different repository). I.e. release such as: `0.6.3-commit-abcdefg12345`
- Snapshot releases which don't require users to specify a commit hash in their versions (snapshot repository). I.e. release such as: `0.6.3-SNAPSHOT` which automatically downloads the latest snapshot release when you don't have this lib locally. 

----

Should we maybe change `org.janusgraph.snapshot` groupId to `org.janusgraph.commit`? Also we can change `0.6.3-SNAPSHOT-abcdefg12345` artifact ids to `0.6.3-commit-abcdefg12345` or even to `0.6.3-abcdefg12345`.

FlorianHockmann:
Thanks for the detailed explanation! I think I now have a better understanding of the use case and why the usual maven `-SNAPSHOT` version isn't a good solution here.

> I think for users it's easier to use repository which doesn't require authentication. Thus, my preference would be Sonatype over GitHub but I'm OK with GitHub as well.

Agreed, then let's use Sonatype for this. GitHub just makes it harder to use and will therefore turn users away from these packages.

> We can add SNAPSHOT solution as well if users interested in that.

I'd say we start with this approach like you have it here in the PR and we can then still add such a SNAPSHOT deployment as well later if users really ask for it and have good reasons why it's more suitable for them than the per-commit version you're proposing here. The per-commit version is at least already usable for you.

> Should we maybe change org.janusgraph.snapshot groupId to org.janusgraph.commit? 

`org.janusgraph.commit` sounds good to me, yes.

> Also we can change 0.6.3-SNAPSHOT-abcdefg12345 artifact ids to 0.6.3-commit-abcdefg12345 or even to 0.6.3-abcdefg12345.

How about `0.6.3-20221207-100250.abcdefg12345` where the format is: `MAJOR.MINOR.PATCH-DATE-TIME.COMMIT` (`DATE `format is `yyyyMMdd` and `TIME` format: `HHmmss`)?
That way, it's still possible to see which deployment is newer which isn't possible solely based on the commit hash.
I've found this format described [in this blog post](https://phauer.com/2016/version-numbers-continuous-delivery-maven-docker/).

porunov:
> How about 0.6.3-20221207-100250.abcdefg12345 where the format is: MAJOR.MINOR.PATCH-DATE-TIME.COMMIT (DATE format is yyyyMMdd and TIME format: HHmmss)?

I love this solution. In addition to having an ability to choose any commit the default version sorting will be ordered.
I will change version format to the one you suggested.

porunov:
@FlorianHockmann pushed [changes](https://github.com/JanusGraph/janusgraph/compare/1c2e7d4a81a6d8ddef7c80e6624cc06d02fd0df0..fad6a4b9877efc998ce44f7f4f8d2e01390fd3ae). Apparently last time I forgot to push `pom.xml` changes (pushed them now as well).
 
I would also prefer to publish a single release manually to verify that it released as expected in Maven Central, normally accessible, and works. Would that be OK? I'm OK to use different group id for this single manual release (maybe something like `org.janusgraph.test.commit.release`)

FlorianHockmann (commented):
The changes look good to me. I just did a quick search in our docs to see whether we mention `SNAPSHOT` releases already somewhere and we actually do: https://docs.janusgraph.org/development/#depending-on-janusgraph-snapshots

I think you can simply remove that section completely.

> I would also prefer to publish a single release manually to verify that it released as expected in Maven Central, normally accessible, and works. Would that be OK? I'm OK to use different group id for this single manual release (maybe something like org.janusgraph.test.commit.release)

Sure, makes sense to test this. I would also be ok with using the same groupId (`org.janusgraph.commit`), but that's up to you.

porunov:
> I think you can simply remove that section completely.

Removed

FlorianHockmann (approved):

porunov:
Confirm that the release is normally released and is accessible. 
Locally built release is available [here](https://search.maven.org/artifact/org.janusgraph.commit/janusgraph-core/0.6.3-20221207-135206.1d324078a/jar). Next releases will be automatic.
This particular release is build with the next Java:
```
openjdk version "1.8.0_352"
OpenJDK Runtime Environment (build 1.8.0_352-b08)
OpenJDK 64-Bit Server VM (build 25.352-b08, mixed mode)
```
Going to merge this PR

Issue #415, opened 2017-07-16
Add testing of Bigtable support via Bigtable emulator


mbrukman:
PR https://github.com/JanusGraph/janusgraph/issues/103 added support for Google Cloud Bigtable via the HBase backend, but the tests are only being run using HBase. We should add tests to run against the [Bigtable emulator](https://cloud.google.com/bigtable/docs/emulator).

@lindsayismith, I believe you ran the HBase test suite manually with the Bigtable emulator when you made the changes in PR #103; would you mind adding an automated way to run it so that they run every time with Travis CI to ensure that everything continues to work properly with the Bigtable backend?

It would also be good to verify that any new changes, as well as new client library versions, continue to work well with JanusGraph + Bigtable, and this would be the way to do it.

burukuru:
If I was to set up our own tests against Janusgraph/Bigtable for [Grakn](https://github.com/graknlabs/grakn), would I just need to set the following settings with Bigtable emulator running on localhost?
```
storage.backend=hbase
storage.hbase.ext.hbase.client.connection.impl=com.google.cloud.bigtable.hbase1_0.BigtableConnection
storage.hostname=localhost
```
Anything else I might be missing?

mbrukman:
@burukuru — you need to run the Bigtable emulator and set the appropriate environment variables, as per the [instructions](https://cloud.google.com/bigtable/docs/emulator):

```bash
> gcloud beta emulators bigtable start
> $(gcloud beta emulators bigtable env-init)
```

which will be automatically picked up by the Bigtable client library when run in the context of JanusGraph, which is specified via the `storage.hbase.ext.hbase.client.connection.impln` parameter.

I don't think you should set the `storage.hostname` as it should be picked up from the environment variables (it also needs the port).

burukuru:
Thanks @mbrukman. It's working with those commands and without `storage.hostname` defined but it seems to be be asking for project.id and instance.id.
```
Caused by: java.lang.IllegalArgumentException: Project ID must be supplied via google.bigtable.project.id
        at com.google.bigtable.repackaged.com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
        at com.google.cloud.bigtable.hbase.BigtableOptionsFactory.getValue(BigtableOptionsFactory.java:283)
        at com.google.cloud.bigtable.hbase.BigtableOptionsFactory.fromConfiguration(BigtableOptionsFactory.java:246)
[...]

PR #2352, opened 2021-01-11
Make any passed cassandra image consider as compatible

fruch:
since `scylladb/scylla` were comptible out of the box for
testcontinars, we now enable any docker image to be cosider as
cassandra comptible

Fixes: #2351

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [x] Is there an issue associated with this PR? Is it referenced in the commit message?
- [x] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [x] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [x] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

PR #3372, opened 2022-12-07
Remove Bigtable as tested compatibility

### LINKED ISSUE #3371: Remove Bigtable from list of tested compatibility versions
The changelog currently lists quite a few Bigtable versions under "Tested Compatibility": 
> Google Bigtable 1.3.0, 1.4.0, 1.5.0, 1.6.0, 1.7.0, 1.8.0, 1.9.0, 1.10.0, 1.11.0, 1.14.0

but we have no automated tests for Bigtable: #415

I think we should remove Bigtable completely from the list under _tested compatibility_ until we have some automated tests for this backend in place. We might not even notice if we break compatibility with Bigtable right now.

We also don't list Scylla there for the same reason: #2451

FlorianHockmann:
Fixes #3371

We can add it back there as soon as we have some automated tests in place: #415

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -62,13 +62,22 @@ compile "org.janusgraph:janusgraph-core:1.0.0"
 
 * Apache Cassandra 3.11.10, 4.0.6
 * Apache HBase 2.5.0
-* Google Bigtable 1.3.0, 1.4.0, 1.5.0, 1.6.0, 1.7.0, 1.8.0, 1.9.0, 1.10.0, 1.11.0, 1.14.0
 * Oracle BerkeleyJE 7.5.11
 * Elasticsearch 6.0.1, 6.6.0, 7.17.5
 * Apache Lucene 8.11.1
 * Apache Solr 8.11.1
 * Apache TinkerPop 3.6.1
 * Java 8, 11
+
+!!! note
+    Google Bigtable was removed from this list because there is no automatic testing in place specifically for that
+    backend.
+    Since the adapter for Bigtable is however just using the HBase adapter, it is also covered by the tests for HBase.
+
+    We invite anyone who is interested in the Bigtable storage adapter to help with this by contributing so that the
+    tests for HBase are also automatically executed for Bigtable.
+    More information can be found in this GitHub issue:
+    [janusgraph/janusgraph#415](https://github.com/JanusGraph/janusgraph/issues/415).
 
 **Installed versions in the Pre-Packaged Distribution:**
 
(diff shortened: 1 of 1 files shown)


farodin91 (approved):

rngcntr:
What about the previous versions? If I understand #415 correctly, we never had Bigtable tests in place. Right now, one may interpret the docs saying Bigtable was tested in 0.6.2 for example but we dropped support in 0.6.3. That could fool users into thinking they should opt for 0.6.2 when operating JanusGraph on Bigtable.

FlorianHockmann:
> What about the previous versions? If I understand https://github.com/JanusGraph/janusgraph/issues/415 correctly, we never had Bigtable tests in place. Right now, one may interpret the docs saying Bigtable was tested in 0.6.2 for example but we dropped support in 0.6.3. That could fool users into thinking they should opt for 0.6.2 when operating JanusGraph on Bigtable.

That's a good point. The problem is however that these versions are already released and I think we should not modify past releases, including their release notes. We also distribute the docs as a zip archive for each release. So, if we modify the changelog of a past release then the online docs for that release will differ from the docs shipped with the release.

farodin91:
We could upgrade instruction with the note that we are looking for help.

rngcntr:
@FlorianHockmann As a first measure we could simply append a line in the changelog for the new versions stating that the tests have not been removed but rather never existed.

FlorianHockmann:
> As a first measure we could simply append a line in the changelog for the new versions stating that the tests have not been removed but rather never existed.

I just updated the commit to include such a remark.

farodin91 (approved):

FlorianHockmann:
> As an organization that uses JanusGraph over BigTable - what do you recommend we do going forward?
Migrate away from BigTable? How much confidence does running our traversals against the BigTable emulator should provide?
Many thanks!

Using JanusGraph with the Bigtable backend uses the same code that JanusGraph uses for HBase and that is tested extensively, just with HBase, not with Bigtable. This means that as long as Bigtable is compatible with HBase, JanusGraph should work with Bigtable just the same as with HBase.
But we unfortunately don't know whether changes we make to the HBase storage adapter might break compatibility with Bigtable if that backend behaves slightly different than HBase in cases that we didn't have before, but got due to the changes in our HBase adapter.
We also don't know whether updates in Bigtable break compatibility with HBase which is why we don't update the Bigtable client library any more.

Overall, I think if you test your own traversals with the Bigtable emulator, then that should provide good confidence that it will also work in production. (I don't know much about Bigtable and it's emulator. I'm assuming here that the emulator won't behave differently than Bigtable itself.)

The best way to get confidence in the Bigtable storage adapter would of course be if you try to contribute to JanusGraph to get automatic testing in place for the backend. I haven't looked into it, but that may be as simple as allowing the HBase tests to be executed with a different Docker image (Bitable emulator instead of HBase) and then adding GitHub actions with that image for the HBase tests (basically copy all HBase GitHub actions and then apply config to use the Bigtable image). #2352 did that for Scylla and the CQL backend.

Over the long term, we will probably need some automatic testing for Bigtable or we will have to drop official support for it completely as it doesn't make much sense to stay on an outdated client library for Bigtable. Every change to the HBase storage adapter we make basically also increases the probability of getting compatibility problems.

li-boxuan (approved):
Ideally, we could set up some CI job that runs the test suite against a real BigTable instance before release, but it will require quite some work and I am not sure who's gonna pay for the cost.

The same applies to Amazon Keyspace. We don't document Amazon keyspace as tested compatibility even though I manually test it against recent releases.

FlorianHockmann:
> Ideally, we could set up some CI job that runs the test suite against a real BigTable instance before release, but it will require quite some work and I am not sure who's gonna pay for the cost.

You mean a real instance instead of the emulator that was mentioned in #415?

Having automated testing in place would of course be ideal and then we wouldn't need to remove Bigtable from the tested compatibility list. But since #415 has been open for so long I don't expect this to be implemented soon. Or are you planning to work on it before the next release?
Otherwise, I'd say we merge this now and then we can later just revert it as soon as we have automated testing in place.

> We don't document Amazon keyspace as tested compatibility even though I manually test it against recent releases.

Good to know that you at least test it manually. But yes, automated testing would also be great there and then we could also add it to the list.

li-boxuan:
Yeah I meant by a real instance rather than an emulator, ultimately. Having an emulator in place would be great, but unfortunately I don’t have a plan to implement that in recent future.

I agree with merging this PR.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

FlorianHockmann:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|v0.6|[](https://github.com/JanusGraph/janusgraph/pull/3443)|

### Questions ?

PR #3297, opened 2022-11-07
refactor backport

farodin91:
part of #3247

Signed-off-by: Jan Jansen <jan.jansen@gdata.de>

-----

Thank you for contributing to JanusGraph!

In order to streamline the review of the contribution we ask you
to ensure the following steps have been taken:

### For all changes:
- [ ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [ ] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you written and/or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered?

PR #3409, opened 2022-12-21
Document new backporting strategy

### LINKED ISSUE #3404: Document new backporting based strategy
This is also part of #3247

We have decided that we want to officially change our strategy to accept PRs for different release branches: https://github.com/JanusGraph/janusgraph/discussions/3373

The main idea is that we now want to let all PRs target `master` and then use [a tool](https://github.com/sqren/backport) to create another PR to backport the changes to release branches for older releases that are still maintained, like `v0.6` right now.
This is a change from our previous approach where PRs should target `v0.6` for example if they are non-breaking and we would then manually merge `v0.6` into `master` after having merged the PR into `v0.6`.

What we should document:

- PRs should target `master`.
- Add labels for backport automation if the change is not a breaking change.
- How to manually handle conflicts in case automatic backporting fails.
- Reviewing a backport PR shouldn't be necessary if they changes were trivial..

FlorianHockmann:
Fixes #3404

Unfortunately, adding the backport labels is a manual process right now that we have to do as committers because:

* Labels cannot be added already by a pull request template: https://github.com/community/community/discussions/12815
* Contributors cannot add them themselves as they lack the required permissions to add labels.

We can however look into adding some action that automatically adds labels to all PRs: https://github.com/actions/labeler Then we would only have to remove the label again on PRs that should not be backported (as that should be fewer than the PRs that we want to backport).

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -83,7 +83,7 @@ include:
 
 - Documentation typo or small documentation addition
 - Addition of a new test case
-- Merging approved fixes into an upstream branch
+- Backporting approved changes into other release branches
 
 Any commit that is made following CTR shall include the fact that it is
 a CTR in the commit comments. Community members who wish to review CTRs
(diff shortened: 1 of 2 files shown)


docs/development.md
```diff
@@ -92,27 +92,50 @@ list so that they will see CTRs as they come through. If another
...
+It can either be merged manually with `git` commands or through the GitHub UI.
+
+For pull requests that should be backported (see [Enable Automatic Backporting](#enable-automatic-backporting)), the
+backporting action should create a pull request to backport the changes to other release branches after the pull
+request has been merged.
+The committer who merges the pull request should ensure that this automatic backporting succeeds.
+It may fail in case of merge conflicts.
+In that case the backporting needs to be [performed manually](#manual-backporting).
+
+### Manual Backporting
```
li-boxuan, line 119: I just tried out this tool and I kind of feel this is even more cumbersome than doing cherry-pick on my own. For example, it checks out a git repo under `.backport/` folder and I have to fix the conflicts there. I would prefer doing so manually on my working directory.

li-boxuan (approved):

docs/development.md
```diff
@@ -92,27 +92,50 @@ list so that they will see CTRs as they come through. If another
...
 
+### Enable Automatic Backporting
+
+Before merging the pull request, it should be decided whether the contribution should be backported to other release
+branches.
+This should be the case for most non-breaking changes.
+If the change applies to something that is not present at all on other still supported release branches, like updating
+a dependency that was only introduced on `master` or code that was added / heavily modified on `master`, then
+backporting cannot work or would require too much effort.
+Otherwise, the appropriate labels for backporting should be added before the pull requests gets merged as that allows
```
li-boxuan, line 103: What milestone shall we use? Shall we use the backported branch's milestone (right now it's 0.6 milestone)?

docs/development.md
FlorianHockmann: I personally like the tool because I don't have to think much about what to do exactly, I only need to resolve the merge conflicts which is the only work that really needs to be done manually. I think is also less error prone than manual cherry picking, especially if there are multiple commits in the PR. But I have to say that I never really do cherry picking manually so I don't have much experience there.
But how about if we just add an additional sentence here that mentions that manual cherry picking can also be done instead of using the tool? Then people can decide on their own which approach they want to use.

docs/development.md
FlorianHockmann: When I wrote this, two labels where necessary: `backport` and `backport/v0.6`. After #3297 is merged, only `backport/v0.6` needs to be added. I'll amend this to include that as an example and mention that the label(s) depend on which branches we currently support.

docs/development.md
li-boxuan: Sounds good!

docs/development.md
li-boxuan: Oh sorry I was not talking about labels. Maybe I should have started the conversation somewhere else, but I meant to ask about **milestones**.

li-boxuan: In other words, for a new PR that has backport label, shall we mark it as 1.0 milestone or 0.6 milestone?

FlorianHockmann: The PR should get the milestone for its target branch. So, a PR on `master` should get the 1.0.0 milestone and a PR for `v0.6` should get the 0.6.3 milestone. This also applies to backporting PRs.
That way, each milestone will really list all PRs that are included in that release and we don't need this "This release also includes all changes from the 0.6.z release" any more.

I'm however not sure about issues that are closed by a PR that will be backported. We can only add an issue to one milestone. Should that be 1.0.0 or 0.6.3?

FlorianHockmann:
I just started a discussion about how we want to merge backporting PRs: #3460. I'll incorporate the results of that discussion here when we have a consensus.

FlorianHockmann:
@li-boxuan I just pushed an update which addresses your comments. Would you like to review again?

I didn't document the topic of auto-merging for these backporting PRs because I guess that this is mostly self explanatory (?) and can also easily change any time. We could for example also enable auto-merging for other types of PRs like from Dependabot. Not sure if we need to document this as it shouldn't really affect contributors in their workflows.

li-boxuan (approved):

porunov (approved):
All the doc changes makes sense to me. Thank you @FlorianHockmann !

PR #3418, opened 2022-12-22
Support string-type custom vertex id [cql-tests]

### LINKED ISSUE #1221: Support String-type custom vertex ID

li-boxuan:
Currently, JanusGraph only supports long-type vertex ids. Vertex ids
are automatically generated, though users could provide their own long-type
ids. There is a high demand that string-type custom vertex id could be
supported by JanusGraph. This also makes migration from some other graph
databases to JanusGraph easier.

A backward incompatibility is in RelationIdentifierGraphBinarySerializer,
where the binary representation of relation (property/edge) identifier has changed.
This means if users are using GraphBinary format for JanusGraph (gremlin) server,
they need to upgrade to 1.0.0 at once, for both clients and servers. Otherwise,
clients won't understand servers and vice versa.

Now users could turn on the GLOBAL_OFFLINE option `graph.allow-custom-vid-types` to assign
custom string-type vertex ids. This functionality is experimental, so users should use it
at their own risks. Note that even if users now can provide custom string vertex
ids, the ids of edges and properties are still of long-type and are auto-generated.
Also notice that existing graph instances can also start leveraging this feature - but
once you turn on this feature, you must ALWAYS generate vertex ID by yourself, otherwise
JanusGraph will reject your vertex creation request. You can turn this feature off again
at anytime.

By default, we don't allow users to use the character '-' in their custom IDs because
it is a reserved keyword used as relation delimiter in JanusGraph. However, users could
overwrite System property "JANUSGRAPH_RELATION_DELIMITER" to some other ASCII
character, e.g. "@". Then users could use '-' in their custom IDs but never character '@'.
This system property setting must be fixed across all JanusGraph servers and clients,
otherwise, it would cause data corruption.

Closes #1221

Demo website: Citegraph (http://www.citegraph.io/) with 9+ million vertices.
Side-note to reviewers: Don't get scared by the number of code changes. Most of them are just test cases 😆 

TODO:

- [ ] Document breaking changes
- [ ] Test and document serialization's backward compatibility (graphson should be backward compatible but graphbinary does not)
- [ ] Test and document vertex id creation backward compatibility (upgrade to 1.0.0 shall not cause any trouble except string ids like `"123"` may not be auto-converted to long ids anymore)

-----

### For all changes:

### For code changes:


tormodhau:
Hei! 

You mention that "_One important limitation is that we don't allow the usage of character '-' in vertex ids_". Does this mean that GUID format IDs are not supported by this PR? It seems to me like this is a fairly common use case.

li-boxuan:
> Hei!
> 
> You mention that "_One important limitation is that we don't allow the usage of character '-' in vertex ids_". Does this mean that GUID format IDs are not supported by this PR? It seems to me like this is a fairly common use case.

Not supported by this current PR, but next PR. I am not sure which approach would be better so I can also use your input:

1. Support any ASCII characters. Since dash ‘-‘ is reserved in JanusGraph, we would use base64 encoding for user-provided IDs. That’s about 33% overhead.
2. Allow users to replace dash character with another user-specified ASCII character. For example, if you will be using UUID, you know you would need ‘-‘ but not character ‘@‘. In that case, we can automatically replace all ‘-‘ with ‘@‘. This is transparent to users and has no overhead. The drawback is you would not be allowed to use ‘@‘ in your user-generated ID forever.

We could of course implement both above, but I would love to hear some early feedback.

rngcntr:
Without having looked at your code changes so far, using Base64 internally sounds to me like the cleanest approach. I'm curious about the overhead of 33% that you mentioned. Is that 33% of the entire query runtime or 33% of only the part of the evaluation that is concerned about mapping ID handling?
In my naive mind, that conversion would only ever have to be performed for user-facing interaction. Therefore, queries which do not handle IDs as an input or output would never have to convert between Base64 and literal strings.

rngcntr (commented):
Hey @li-boxuan, thanks for the great effort to support string IDs!
I did not yet have the time to review the PR in its entirety, but instead I left some stylistic comments and as soon as I began to understand the greater picture, I decided to take a step back and discuss the general idea first.
Even though the adaptions in each class are rather small, they introduce a certain level of complexity to be aware of. Many classes now contain variables like `allowStringVertexId` and use if statements to decide whether one behavior or another should be applied. From my point of view, this could make it harder to maintain features like this because the complexity is spread over many classes. Thus, I would like to hear your opinion on generifying all ID-type related functionality into an `interface IdUtil<IdType>` and providing classes that implement `IdUtil<Long>` and `IdUtil<String>` for instance. The goal here would then be to isolate ID handling form the rest of the code such that the choice of the ID type becomes a plug-and-play between different implementations.
When implemented, that would open up the possibility to enforce UUIDs for example, whereas the current approach relies on the user to provide valid UUIDs. Somewhere in the code I noticed a comment that certain features are only available with fixed-length IDs which — combined with string IDs — would open up another use case for custom ID types.

I lack the expertise to see right away if this is a feasible approach, but since you have put such a significant effort into the ID code base, you are probably the one to tell good and bad ideas apart.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
```diff
@@ -32,6 +32,10 @@ public final class RelationIdentifier implements Serializable {
     private final long relationId;
     private final Object inVertexId;
 
+    // this does not appear in LongEncoding.BASE_SYMBOLS, so as long as we see this marker,
+    // we know the id must be of string type rather than a string-encoded long type
+    public static final char STRING_MARKER = 'S';
```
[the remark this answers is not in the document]
li-boxuan: Good catch. We have LongEncodingTest::testConflictWithStringMarker to prevent that from happening.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idhandling/VariableString.java
```diff
@@ -0,0 +1,82 @@
...
+            byte b = (byte)c;
+            if (i == value.length() - 1) b |= 0x80; //End marker
+            out.putByte(b);
+        }
+    }
+
+    public static String read(ReadBuffer in) {
+        StringBuilder sb = new StringBuilder();
+        while (true) {
+            int c = 0xFF & in.getByte();
```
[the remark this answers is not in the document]
li-boxuan: This was actually also copied from somewhere else. Interestingly, doing `0xFF & byte` before assigning to an integer is a common practice in Java to avoid byte-to-int cast bugs (ref: https://stackoverflow.com/a/32452883). So we have to do `0xFF` here.

li-boxuan:
> Many classes now contain variables like allowStringVertexId and use if statements to decide whether one behavior or another should be applied. From my point of view, this could make it harder to maintain features like this because the complexity is spread over many classes. Thus, I would like to hear your opinion on generifying all ID-type related functionality into an interface IdUtil<IdType> and providing classes that implement IdUtil<Long> and IdUtil<String> for instance. 

Yeah I, unfortunately, increased code complexity in many places due to the inelegant usage of `allowStringVertexId` and `if` statements. To keep backward compatibility, this seems to be inevitable to me. I am not sure how generalization could be implemented here to make the code cleaner. One curse is that the option of using `string` is not decided at compile time but at runtime. Would love to hear if you have some examples.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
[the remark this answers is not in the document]
rngcntr: Ah, thanks! I haven't reviewed all tests yet

li-boxuan:
I created a website to demo this feature: http://www.citegraph.io/

Citegraph contains ~9.4 million vertices and all of them have string ids. Some of them are from external datasets, and some of them are generated as UUID-like strings. I've done write, read, and a few SparkGraphComputer traversals. So far, so good.

jianhuang1648678534:
can you explain the steps to migrate the graph data to replace the  long vertex id with this new string id?

li-boxuan:
> can you explain the steps to migrate the graph data to replace the long vertex id with this new string id?

Good question, I'll do some experiments and write a doc.

li-boxuan:
> Many classes now contain variables like allowStringVertexId and use if statements to decide whether one behavior or another should be applied. 

@rngcntr I refactored the encoding scheme which now has less overhead (saving 1 byte in most places) and less ugly code like before.

docs/advanced-topics/custom-vertex-id.md
```diff
@@ -0,0 +1,89 @@
...
+g.addV().property(T.id, UUID.randomUUID().toString()).next()
+g.addV().property(T.id, "custom_vid_001").next()
+g.tx().commit()
+```
+
+!!! warning
+    If you decide to replace default `JANUSGRAPH_RELATION_DELIMITER`, you must
+    always set the system property value consistently across all JanusGraph
+    instances, including servers and clients. Otherwise, data corruption could
+    happen.
```
porunov, line 89: Thank you @li-boxuan for the great work!
I didn't start the review yet, but wanted to leave thoughts / questions below.
1. I didn't research it myself, but as you worked in this direction already, I wonder if adding possibility to use raw bytes as vertex ids would make sense or possible at all.  I know that many databases support using raw bytes as primary keys (for example, `blob` in Cassandra / ScyllaDB or  `Binary` in HBase). I suppose using raw bytes could have more benefit then using String as users might literally serialize anything into such ids (UUIDs, integers, custom strings, custom objects, etc.). I suppose, having stored ids as bytes eliminates the need of using specific string characters and also gives ability to serialize more efficiently.
2. I just wander what is the reason to have it as a system property and not JanusGraph `GLOBAL_OFFLINE` configuration property?

li-boxuan: > I wonder if adding possibility to use raw bytes as vertex ids would make sense or possible at all

Good question. Unfortunately, it's not possible because of the data layout in JanusGraph. 

image

JanusGraph needs a way to mark the boundary of different components. For example, it needs to know the end of "adjacent vertex id" and the start of "edge id". When using a long-type ID (which is the status quo), JanusGraph uses a special bit as a marker. Specifically, any byte with the first bit being 1 is a boundary marker.

When using a string-type ID (which is in this PR), we use a special byte to mark the boundary. Since we require the string ID to consist of ASCII characters, we are able to easily use a special byte to mark the boundary.

What about byte IDs? It's impossible unless we require users to avoid using a specific byte.

Moreover, IIRC, Elasticsearch requires document ID to be of string type. We could use base64 to encode the user-given ID but it will be a huge breaking change in `RelationalIdentifier`. Your second question is related.

> what is the reason to have it as a system property and not JanusGraph GLOBAL_OFFLINE configuration property

Unfortunately, the delimiter is not only used by JanusGraph server code but also by driver code. In the driver code, you don't have access to any JanusGraph configuration. Using system property to control is definitely not ideal but I am afraid there is nothing else we could do unless we introduce a breaking change (e.g. always use base64 to encode the relation identifier), which will make JanusGraph NOT backward-compatible :(

porunov: Thank you for the great explanation! Definitely makes sense. 
I assume that the default delimiter `-` is chosen to not have breaking changes in relationship ids encoding, but in case I'm wrong and it's not a breaking change then I guess some less popular character would make more sense to be used as a default one (like `'` or `*`). 
I will try to check the PR later this week.

li-boxuan: > I assume that the default delimiter - is chosen to not have breaking changes in relationship ids encoding

You are right - we have to keep it for backward compatibility.

porunov: > What about byte IDs? It's impossible unless we require users to avoid using a specific byte.

I guess if we encode byte array size (4 bytes at the beginning of adjacent vertex id byte array) then it could be possible to read those 4 bytes and understand when the specific key is finishing. This would, of course, increase the key size +3 bytes. It just some thoughts for future in case we ever want to implement that vertex ids as `byte[]`.

li-boxuan: > I guess if we encode byte array size (4 bytes at the beginning of adjacent vertex id byte array) then it could be possible to read those 4 bytes and understand when the specific key is finishing

That makes sense, yes, I think it should work. And yes, we could make it happen in the future. Currently, we leverage the fact that a long-type ID always have the first bit of the first byte being 0. I use an arbitrary byte that starts with bit 1 to mark that it's a string-type ID. In the future we could use another byte to mark that it's a different type.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +64,39 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+        if (s.length() == longSize) {
+            buffer = ByteBuffer.allocate(longSize + byteSize);
+        } else {
+            buffer = ByteBuffer.allocate(s.length());
+        }
+        for (int i = 0; i < s.length(); i++) {
+            int c = s.charAt(i);
+            assert c <= 127;
+            byte b = (byte)c;
+            if (i+1==s.length()) b |= 0x80; //End marker
```
porunov, line 90: Why we need this `End marker`? Not sure I understand the difference between this `End marker` and `STOP_MASK`.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idhandling/VariableString.java
```diff
@@ -0,0 +1,98 @@
...
+ * to assume all IDs are positive long values. To maintain backward compatibility, we have to make sure
+ * JanusGraph knows how to distinguish between string ID and long ID when decoding binary data. The
+ * convention is, if the first byte is STOP_MASK, then it's a string ID, otherwise it's a long ID. This
+ * leverages the fact that IDs are always positive, so the first byte would never be STOP_MASK for long
+ * IDs.
+ *
+ * This class uses {@link StringSerializer} as a reference.
+ *
+ * @author Boxuan Li (liboxuan@connect.hku.hk)
+ */
```
porunov, line 42: Is it the byte which tells that the key is String or Long? (https://github.com/JanusGraph/janusgraph/pull/3418#discussion_r1171987782)
If so, I just wander why we need it exactly. Unless users want to use a combination of keys (i.e. `Long` for some vertices and `String` for other vertices), this marker won't be useful that much unless I'm missing something.
Is it here so that a single JanusGraph instance could work with both ID types simultaneously? 
I.e. JanusGraph instance already knows weather it works with String ids or Long ids. Not sure we need to check the key type for each key we retrieve. 
I definitely not against using this byte, I just not sure I understand why we need to use it.

porunov (commented):
I didn't check all the files yet, thus I might figure out some of my questions letter on, but decided to drop some questions for now.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +64,39 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+    /**
+     * If string is not 8 bytes, then we write it as it is. Otherwise, we write a special
+     * marker at the end, so that when JanusGraph sees the buffer, it knows it should
+     * reinterpret the byte buffer as string buffer. See {@link IDManager#getKeyID(StaticBuffer)}
+     * for more details.
+     *
+     * @param s
+     * @return
+     */
+    public static StaticBuffer getStringBuffer(String s) {
```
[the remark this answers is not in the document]
li-boxuan: If we don't write a special marker at the end for a string with 8 bytes, `IDManager#getKeyID(StaticBuffer)` will break because it uses the length of the buffer to judge whether it's string-type or long-type.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: Great observation! Here, we don't necessarily need this `0x80` as an end marker because, upon read time, we know the length of the buffer. This is more to keep it consistent with `StringEncoding#writeAsciiString` where an end marker is necessary.

The additional benefit of using an end marker (which effectively changes the first bit to 1) is that we can use it for corruption detection in the future.

The variable names and magic constants are indeed confusing. I'll rename them.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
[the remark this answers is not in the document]
porunov: Got it. I think the next question could be related: https://github.com/JanusGraph/janusgraph/pull/3418#discussion_r1172041147

I'm just thinking that JanusGraph instance should probably have some runtime information about type of id to be used (String ID, Long ID, or any other sort of ids). In such case, we probably won't need to use that marker.
I just think some of the logic regarding ID could be abstracted away and during JanusGraph instance creation we could configure it to use one or another logic for ids via some Strategy patterns or else.

Btw. I suspect if we decide on the above solution (i.e. without special marker) then some refactoring might be needed, but I can potentially help on weekends with it if you need a hand on this.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idhandling/VariableString.java
li-boxuan: > Is it the byte which tells that the key is String or Long? (https://github.com/JanusGraph/janusgraph/pull/3418#discussion_r1171987782)

Yes.

>  Unless users want to use a combination of keys (i.e. Long for some vertices and String for other vertices), this marker won't be useful that much unless I'm missing something

To some extent, you are right. This marker is helpful in:
I. Allow users to use `Long` for some vertices and `String` for other vertices.
II. (same as I., but from a different perspective) Allow existing JanusGraph users to start using string-type IDs without creating a new graph from scratch (a.k.a. ability to "upgrade"). This opens a door to those who would like to "gradually" abandon long-type IDs.
III. (same as I., but from a different perspective) Allow users to stop using string-type IDs without creating a new graph from scratch (a.k.a. ability to "downgrade").
IV. Make the codebase cleaner. Otherwise, we would need a boolean/enum variable across the codebase to control whether JanusGraph is dealing with long-type IDs or string-type IDs.

So who benefits from this design and who suffers from it?
- Users who stick to long-type IDs: Neither overhead nor benefit for them.
- Existing users who would like to start using string-type IDs: This is a must-have feature for them.
- New users who are only interested in using string-type IDs: One byte overhead for ID :(
- Developers: minimal benefit as code is cleaner this way.

li-boxuan: Actually, the most important reason is we MUST do it (not 100% sure, though). Schema definitions are modeled as vertices and edges in JanusGraph, too. Since these vertices are of long-type, there is currently no way of getting rid of long-type IDs completely.

porunov: I do agree that this byte give users ability to slowly migrate to String id or to migrate back to Long, but I think there are some additional cons we need to consider:
- Additional runtime branching logic. Even so JIT might optimize many of those branches they will give some overhead (probably negligible, but still). In case a single JanusGraph instance support only a single type of ID then we could make refactoring and configure JanusGraph to use only the specific ID strategy.
- Additional logic to understand for a developer. In case of a separate strategies abstracted away by a common interface, the developer won't need to read and understand code for both String IDs and Long IDs. Instead, it's enough to read a single strategy which developer is interested in.
- In case we add any other ID types the byte might be not very useful as it is designed specifically for String type. I.e. it basically means that any 9 bytes array with a STOP_MASK is an 8 bytes array. It won't help if we want to have 16 bytes String and UUID type or byte array type as ID type. 

> Actually, the most important reason is we have to do it (not 100% sure, though). Schema definitions are modeled as vertices and edges in JanusGraph, too. Since these vertices are long-type, there is currently no way of getting rid of long-type IDs completely. That being said, it's possible to add special control options in the codebase so that we know that we are reading normal vertices or schema vertices (assuming we would like to get rid of the marker byte).

Agree that we have to keep logic for Long ids always due to Schema vertices. This does seems to be a use case when we need to use `Long` type always and `another type` optionally.
In such case, I think we could leave this byte as is, but I would say that it makes sense to think about this byte not as `String` marker byte but as `non-default type` marker byte.
Meaning that this marker means that the type of this key is equal to the one, configured in JanusGraph instance (could be String, UUID, byte array, or something else serializable).
That said, in such case, we need to always store this extra byte for any byte array grater then 7 bytes. Otherwise it won't be possible to tell weather this 9 bytes array is actually of size 9 or 8. Currently you store this byte only for a String with length == 8, but it's possible just because you use ASCII characters in the String and non-ASCII character for a marker. For byte arrays it won't be a good idea to restrict usage of the first byte.
What do you think about it? Do you think it makes sense to store this byte only for String.length == 8 (improves efficiency of Strings which are bigger then 8 bytes) or store this byte for String.length >=8 (adds possibility of implementing other data types later, but sacrifices 1 byte for each type >= 8 bytes)?

li-boxuan: Just to be clear, my current approach uses two different encoding schemes for STRING:
1. In `VariableString` class, it uses a special STRING marker byte at the beginning.
2. In `IDManager::getKeyID` class, it relies on the fact that the length of STRING id is not 8 bytes (we add one dummy byte if its length is 8).

Note that the first scenario applies to encoding vertex IDs in edges and indexes, whereas the latter scenario applies to encoding vertex IDs as partition keys.

For the latter case, you brought up a good point: if we use `length == 8` to judge whether it's LONG or STRING, it leaves no room for any future extension where we might allow byte array as ID. In that case, I think we should always add a marker at the beginning when it's not long-type.

li-boxuan: Updated `IDManager::getKeyID`. Now if JanusGraph sees the buffer has 8 bytes, it knows it's a long-type ID. Otherwise:
1) If the first byte is `STRING_ID_MARKER`, then it's a long ASCII string. The end of the string is determined by the `STOP_MASK`. (We add a dummy padding byte if applicable, to make sure the byte buffer is always not 8 bytes).
2) Otherwise, it's neither long-type nor string-type. This leaves room for future extensions, e.g. byte array ID.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +65,51 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+             * to stop reading the string. This is more to keep it consistent
+             * with {@link StringEncoding#writeAsciiString(byte[], int, String)}
+             * where we use STOP_MASK to mark the end of the string.
+             *
+             * An additional benefit of doing so is to enable corruption check
+             * in the future. Note: this approach has no overhead.
+             */
+            if (i+1==s.length()) b |= STOP_MASK;
+            buffer.put(b);
+        }
```
porunov, line 104: It feels that we could use UTF-8 characters instead of ASCII characters if we get rid of `STOP_MASK` here.
We lose 1 bit per character here anyway, so why not let users decide if they want to use ASCII characters or some other unicodes using UTF-8 encoding?
In such case, if we use UTF-8, we lose the ability to implement data corruption check in the future, but it seems to be more related to the storage backend features rather then to JanusGraph layer anyway (not sure here actually).
Moreover, this will simplify this method copy part to something like:
`System.arraycopy(s.getBytes(), 0, arr, 0, s.length());`

Moreover, UTF-8 might potentially be better here because if we want to use non-ASCII characters latter - we won't be able due to `String` type being reserved for ASCII character only. With UTF-8 it's up to the user what characters to use.

Notice, I personally think ASCII characters along are usually enough for IDs (I haven't seen people use other characters for IDs previously), but I just try to view on this problem from different perspective.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +65,51 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+        if (s.length() + byteSize == longSize) {
+            // this could be any dummy byte
+            buffer.put(STOP_MASK);
+        }
```
porunov, line 108: This means Strings with size == 7 and Strings with size == 8 both will be converted to the String with size == 9.
I told about that 4 bits of `STRING_ID_MARKER` could be used as a data type. I now think 1 bit from that could also be used to tell if the encoded size of the value if less then 8 bits or not. If yes, then we know that the string is of size 7. Otherwise it's 8.
It will help to use UTF-8 characters here. This is can also be applicable for `byte[]` data type in the future.
Thus the first byte marker could have 3 reserved bits, 1 bit for value size detection (less then 8 bits or not), and 4 bits for data type detection (String, byte[], UUID, etc.).

janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java
```diff
@@ -102,10 +103,19 @@ public class GraphDatabaseConfiguration {
...
+            "are unique to avoid duplication. Must use `graph.getIDManager().toVertexId(long)` to convert your id first. Once " +
+            "this is enabled, you have to provide vertex id when creating new vertices. EXPERT FEATURE - USE WITH GREAT CARE.",
+            ConfigOption.Type.GLOBAL_OFFLINE, false);
+
+    public static final ConfigOption<Boolean> ALLOW_STRING_VERTEX_ID = new ConfigOption<>(GRAPH_NS, "allow-string-vid",
+            "Whether string-type vertex ids are allowed. " + ALLOW_SETTING_VERTEX_ID.getName() +
+            " must be enabled in order to use this functionality. This does not prevent users from using custom ids with " +
+            "long type - both string ids and long ids are supported. If your storage backend does not support unordered " +
+            "scan, then some operations will be disabled. You cannot use this feature with Berkeley DB. EXPERT FEATURE - USE WITH GREAT CARE.",
+            ConfigOption.Type.GLOBAL_OFFLINE, false);
```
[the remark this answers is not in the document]
li-boxuan: Good idea. I think we don't need to explicitly ask users to specify a type - we can use a boolean and infer the type if needed.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idhandling/IDHandler.java
```diff
@@ -38,6 +39,15 @@ public class IDHandler {
...
     public static final StaticBuffer MAX_KEY = BufferUtil.getLongBuffer(-1);
 
+    public static final byte STOP_MASK = -128;
+
+    /**
+     * See {@link VariableString} for more detail. This marker has highest
+     * bit being one. When JanusGraph sees it, it knows the buffer stores
+     * a string ID rather than a positive long ID (whose highest bit is zero).
+     */
+    public static final byte STRING_ID_MARKER = -128;
```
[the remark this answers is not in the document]
li-boxuan: > then it seems it doesn't matter what this byte has

It does matter. I understand your confusion. Unfortunately, yes, this is indeed confusing. You might want to read the javadoc of `IDManager::getKeyID` to learn more about the background.

Briefly speaking, due to historical reasons, when a long-type vertex id appears as a partition key (`key` column in the data layout), it's always 8 bytes. It does use some weird encoding (which seems unnecessary to me) so that the first bit can be either 0 or 1 even though the value is always positive. Anyways, it always uses an 8-byte buffer to store the vertex ID (we could have just stored the value as it is without worrying about encoding and decoding - I don't know why this decision was made). In this case, the only way to distinguish a long-type ID and a non-long-type ID is to count the buffer size. Thus, we rely on the fact that 8 bytes are always `Long`. If we only support string type, then we don't need a marker at all. However, since we want to leave some space for future extensions, we use an arbitrary byte here. You are right the first bit can be anything.

When a long-type vertex id appears in the `column` column or `value` column, it no longer has a fixed length. It uses another encoding scheme that makes every but the last byte have the highest bit being 0. You might recall the usage of `b | STOP_MASK` or equivalently, `b | 0x80` that makes the last byte have a leading 1. Since long-type IDs are no longer 8 bytes, the above trick does not work anymore. Luckily, we leverage the fact that the highest bit of the first byte is always 0. Therefore, we use this `STRING_ID_MARKER` whose highest bit is 0. For now, this is just an arbitrary byte (0b1000000000000000). The last 15 bits are reserved for future usage.

To conclude, in the first case, the first byte can be anything. In the second case, the first byte can be anything except the highest bit MUST be 1. I use the same byte coz why not?

Back to your point:

> We probably could use lowest 4 bits to represent data type

I agree. That being said, we could defer this discussion until we actually start to support other types.

docs/advanced-topics/custom-vertex-id.md
```diff
@@ -0,0 +1,92 @@
...
+A string vertex ID can have arbitrary length as long as the underlying storage backend
+permits. The string value must consist of ASCII characters only, and it cannot contain
+JanusGraph reserved relation delimiter. By default, this reserved character is `-` (dash),
+which means you cannot use UUID string as a vertex ID since it contains a `-` character.
+However, JanusGraph allows you to set `JANUSGRAPH_RELATION_DELIMITER` system property
+which can be any visible ASCII character. Once it is set, JanusGraph will prohibit that
+specific character instead of the default `-` character. Alternatively, you can set
+`JANUSGRAPH_RELATION_DELIMITER` environment variable, which is evaluated if and only if
+the `JANUSGRAPH_RELATION_DELIMITER` system property is not present. For example, you can
+use the following command to set system property when opening gremlin console:
```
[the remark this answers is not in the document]
li-boxuan: > but if user uses non-printable characters for ID and printable character for delimiter wouldn't it wrongly printer in the Gremlin Console anyway

Good point. In the beginning, I thought we should allow all ASCII characters but later I changed my mind, leaving the requirements inconsistent. I am thinking of requiring readable ASCII characters only.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: Good point! My primary design principle is to reduce storage overhead as much as possible. An ordinary Java string uses 2 bytes for 1 character. UTF-8, even worse, uses 1 to 4 bytes per character.

If one would like to use any character other than visible ASCII characters, they could use base64 encoding on the app side, or they could just use a byte array (once we support it). Since ASCII characters are usually enough for IDs, this approach makes most use cases more efficient.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: > I now think 1 bit from that could also be used to tell if the encoded size of the value if less then 8 bits or not. If yes, then we know that the string is of size 7. Otherwise it's 8.

Make sense, perfect.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
porunov: > Good point! My primary design principle is to reduce storage overhead as much as possible. An ordinary Java string uses 2 bytes for 1 character. UTF-8, even worse, uses 1 to 4 bytes per character. Since we only allow ASCII characters, we only use 1 byte per character. That's at least 50% cheaper than other encoding schemes.
> 
> If one would like to use any character other than visible ASCII characters, they could use base64 encoding on the app side, or they could just use a byte array (once we support it). Since ASCII characters are usually enough for IDs, this approach makes most use cases more efficient.

Yes, but in UTF-8 if you use only ASCII characters in your String then every character will take 1 byte. As far as I understand the the main purpose of using the highest bit.
I.e. ASCII originally uses 7 bits for 128 characters.
UTF-8 also uses 8 bits for ASCII characters because one bit is used to understand that it’s non-ASCII character which requires a second byte or more.
In other words:
“abcd” - takes 4 bytes in ASCII encoding.
“abcd” - takes 4 bytes in UTF-8 encoding.
base64 string will take the same size if encoded in ASCII or UTF-8. 
I hardly understand the reason of using ASCII encoding over UTF-8 encoding in any case (unless you specifically want to forbid characters from non-ASCII table).
Again, I’m OK with ASCII characters. It’s just interesting that we specifically forbid UTF encoding. Most likely I am just missing some requirements regarding data keys encoding due to lack of knowledge of that part of the codebase.

porunov: About ordinary Java String. It uses UTF-16 by default. I definitely not suggesting using UTF-16, but UTF-8 is a better alternative then ASCII encoding in my understanding almost in any case due to taking the same amount of storage for ASCII characters, but allowing to use more characters (with less optimal encoding) in case the user needs to use those characters.

docs/advanced-topics/custom-vertex-id.md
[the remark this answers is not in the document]
porunov: I’m just thinking if this requirement is only for better visibility in Gremlin console then it could be we don’t need this restriction actually. TinkerPop supports Object as id for vertices, meaning that it can be a String with non-printable characters. The fact they will be harder to copy paste manually via Gremlin console isn’t a deal breaker I think. Users need to understand that if they use non-printable characters - they logically won’t see those characters, but it shouldn’t stop them from using those characters because they can still be correctly used inside JanusGraph. As far as I understand it doesn’t matter for JanusGraph if key characters or relation delimiter is printable or non-printable. The only requirement for the delimiter is to be different from the rest of the characters.
I might be wrong though. Just brainstorming.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idhandling/IDHandler.java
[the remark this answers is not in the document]
porunov: Agree, but it could be a breaking change then. I thought that we could agree on the bits usage of this market and start using correct bits for that.
As you say the highest bit should always be 1 for “other type” and “0” for Long - so be it. We have plenty of bits in the market to be used.
We could do the next encoding for the marker:
highest bit - always 1.
next 2 bits - reserved for some latter features / extensions.
1 bit - tells if the original value is 7 bits or less. “1” - value is 7 bits or less. “0” - value is 8 bits or more.
4 bits - encoded data type. “0000” - is for String lets say.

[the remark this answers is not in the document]
porunov: > At this point, I realize and regret that I should have written a design doc for better understanding at the beginning. Sorry for that.

No worries at all. I just ask some dummy questions because I didn’t work with that part of the codebase and simply don’t understand some parts sometimes. It’s great explanation though 👍 Thank you

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
porunov: > Thanks, that's something I wasn't familiar with and very interesting.
> 
> It seems to me that the biggest problem of using UTF-8 (rather than ASCII) is we don't know the end of the string. Recall that when vertex id is stored in `column` column or `value` column, it's only part of the whole byte buffer and JanusGraph must find out the boundary by itself. When we use ASCII only, we know the first bit is always 0, thus we can mark the first bit of the last byte to be 1 (the same trick as encoding a long value). This trick does not work for UTF-8 because the first bit of a UTF-8 character can be 1.

Yeah, I now see what you mean. I assume this could be needed for index store (probably but not sure).
As for edges `column` key, to get `adjacent vertex id` from an edge, I thought it could be possible to determine key by getting bytes between `sort key` (or label id + direction) and `edge id`, because we know that `edge id` is of size 8 bytes, so the `adjacent vertex id` should be everything after `sort key` and up to `edge id`. That said, this will work to retrieve key from an edge, but I'm not sure it will work to retrieve key from other stores.
I don't have immediate solution on top of my head on how to allow using UTF-8  encoding for index store without using additional 4 bytes at the beginning to determine string length. Thus, I'm good with ASCII encoding.
I think if we implement `byte[]` later then users can use that structure to store UTF-8 strings if they really want to use non-ASCII chars for their vertex ids (which is probably a rare case).

porunov:
@li-boxuan I created an Epic issue to track all the data types we want to add for vertex ids: #3732 
I added information about the first byte end what each bit of that first byte means for custom ids. Also added information about how we determine if the key is `Long` or something else.
In case I missed something or added incorrectly, please, feel free to edit it or comment. Thanks you!

docs/advanced-topics/custom-vertex-id.md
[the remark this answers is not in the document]
porunov: I'm now thinking, as this actually not a breaking change on the storage layer, we could really allow using only printable ASCII characters for Strings as for now, but if users claim they need to use non-printable ASCII characters we can always relax the checks and allow any ASCII characters. 
This way we will not deal with potential problems printing those ids on Gremlin Console side, and have room to relax id requirements later if users want to.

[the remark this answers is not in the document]
li-boxuan: My arguments:

1) Users might accidentally introduce non-printable ASCII characters and this would make debugging difficult. For example, they might accidentally enter a new line on the gremlin console when they enter the vertex ID and still succeed.
2) We could easily extend support for printable ASCII characters to all ASCII characters, without introducing a breaking change.
3) Possible implications on indexing backends. Some secondary indexing backends might not allow non-printable ASCII characters (I didn't verify, so this might not actually hold).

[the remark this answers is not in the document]
porunov: Agree with your points here. Let's use only printable characters as a safety measure as for now (both for delimiter and for actual id). Later on if we receive feedback that anyone needs to use non-printable characters for some use-cases then we could introduce a special feature flag which relaxes String requirements and allows any characters to be used.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: > I thought it could be possible to determine key by getting bytes between sort key (or label id + direction) and edge id

This may or may not work; in some cases (depending on multiplicity and constraints), the adjacent vertex id is part of `value` column followed by properties. The above trick might fail in this case.

Another perspective is: what if we allow custom edge id as the next step? Then you cannot rely on the fact that edge id is 8 bytes.

> I think if we implement byte[] later then users can use that structure to store UTF-8 strings if they really want to use non-ASCII chars for their vertex ids (which is probably a rare case).

Yep, or they can base64 encode their vertex ids first.

porunov: > > I thought it could be possible to determine key by getting bytes between sort key (or label id + direction) and edge id
> 
> This may or may not work; in some cases (depending on multiplicity and constraints), the adjacent vertex id is part of `value` column followed by properties. The above trick might fail in this case.
> 
> Another perspective is: what if we allow custom edge id as the next step? Then you cannot rely on the fact that edge id is 8 bytes.
> 
> > I think if we implement byte[] later then users can use that structure to store UTF-8 strings if they really want to use non-ASCII chars for their vertex ids (which is probably a rare case).
> 
> Yep, or they can base64 encode their vertex ids first.

Yeah, that make sense definitely and relying on `edge id` size is a bad idea as you mentioned because we may change edge ids later on as well.
I think using ASCII only characters is a good solution here.

docs/changelog.md
```diff
@@ -105,6 +105,13 @@ For more information on features and bug fixes in 1.0.0, see the GitHub mileston
 
 #### Upgrade Instructions
 
+##### String vertex ID support
+
+Users now can use custom string vertex ids. See `Custom Vertex ID` chapter in documentation. Prior
```
porunov, line 110: (nitpick) Add link to documentation page:
`[Custom Vertex ID](advanced-topics/custom-vertex-id.md)`

janusgraph-core/src/main/java/org/janusgraph/diskstorage/BackendTransaction.java
```diff
@@ -373,9 +377,10 @@ public KeyIterator edgeStoreKeys(final SliceQuery sliceQuery) {
         return executeRead(new Callable<KeyIterator>() {
             @Override
             public KeyIterator call() throws Exception {
-                return (storeFeatures.isKeyOrdered())
-                        ? edgeStore.getKeys(new KeyRangeQuery(EDGESTORE_MIN_KEY, EDGESTORE_MAX_KEY, sliceQuery), storeTx)
-                        : edgeStore.getKeys(sliceQuery, storeTx);
+                return (storeFeatures.isKeyOrdered() && !allowCustomVertexIdType)
+                    // if string vertex id is allowed, then min key and max key are unknown
```
porunov, line 381: (nitpick)
`if string vertex id is allowed` -> `if non long-type vertex id is allowed`

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +65,50 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+     * in the future. Note: this approach has no overhead.
+     *
+     * @param s
+     * @return
+     */
+    public static StaticBuffer getStringBuffer(String s) {
+        if (StringUtils.isEmpty(s) || !StringUtils.isAsciiPrintable(s)) {
+            throw new IllegalArgumentException("value must be non-empty printable ASCII string but received: " + s);
+        }
+        ByteBuffer buffer;
```
porunov, line 87: (nitpick)
Doesn't look like any of `ByteBuffer`  specific functionality is used in this method. I.e. you could potentially used simple `byte[]` instead. That said, I'm fine with `ByteBuffer`.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +65,50 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+            buffer = ByteBuffer.allocate(s.length() + byteSize);
+        }
+        buffer.put(STRING_ID_MARKER);
+        for (int i = 0; i < s.length(); i++) {
+            int c = s.charAt(i);
+            assert c <= 127;
+            byte b = (byte)c;
+            /**
+
+             */
```
porunov, line 100: (nitpick) Empty comment

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
```diff
@@ -57,6 +65,50 @@ public static StaticBuffer getIntBuffer(int[] ids) {
...
+        if (StringUtils.isEmpty(s) || !StringUtils.isAsciiPrintable(s)) {
+            throw new IllegalArgumentException("value must be non-empty printable ASCII string but received: " + s);
+        }
+        ByteBuffer buffer;
+        if (s.length() + byteSize == longSize) {
+            buffer = ByteBuffer.allocate(longSize + byteSize);
+        } else {
+            buffer = ByteBuffer.allocate(s.length() + byteSize);
+        }
+        buffer.put(STRING_ID_MARKER);
```
porunov, line 93: Ping me when this comment is implemented: https://github.com/JanusGraph/janusgraph/pull/3418/files#r1173201627  (more discussion about what each of those 8 bits will mean is here: https://github.com/JanusGraph/janusgraph/issues/3732#issue-1678683473 )

If you added that logic somewhere, but I missed it, please, ping me and I will take a look at it.

docs/changelog.md
li-boxuan: 👍  Thanks, I forgot we could include link to other doc in changelog

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: I noticed our code base has been using `ByteBuffer` consistently even if as you said, many places could have just used plain byte array.

I am guessing the reason is that if we decide to switch to off-heap memory, we can do so without too many code changes.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: I didn't add that logic yet. For ASCII string type, we don't need to rely on a value size detection bit (because the final non-trivial byte is masked (`if (i+1==s.length()) b |= STOP_MASK;`).

I agree that can be very useful for other non-fixed-length types, e.g. byte array. I just feel like before we actually implement that functionality, there's no need to make the contract fixed. In other words, right now I feel like a value size detection bit is a very good idea for a byte array type, but once we really implement support for byte array ID, we may come up with a different approach.

porunov: That make sense. I think this bit will be necessary for byte[] only type. Thus, should be fine. 
I initially thought that we need this bit for String as well, but got it that we don’t need it. 
I think it would be nice to add 2 tests just in case where we decode two strings of size == 9 with one string being 7 in original size and another one being 8 in original size. 
This is just in case we decide to make some breaking changes and forget about this case.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
porunov: That’s a good point. I think we should use ByteBuffer as you already do in such case just for consistency reasons.

janusgraph-core/src/main/java/org/janusgraph/diskstorage/util/BufferUtil.java
li-boxuan: This is covered in `JanusGraphCustomIdTest.java#testMixedStringAndLongVertexId` but I'll add two additional tests to be safe.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/idhandling/VariableString.java
```diff
@@ -0,0 +1,96 @@
...
+ * leverages the fact that IDs are always positive, so the first byte would never be STRING_MARKER for long
+ * IDs.
+ *
+ * This class uses {@link StringSerializer} as a reference.
+ *
+ * @author Boxuan Li (liboxuan@connect.hku.hk)
+ */
+public class VariableString {
+
+    public static void write(WriteBuffer out, final String value) {
```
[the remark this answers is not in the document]
li-boxuan: >  it seems that we can still generalize common logic and reuse it in both methods

Make sense 👍 

>  in BufferUtil you are adding the last dummy byte in case the combined string is of size == 8, but you don't do it here. Is it expected?

Yes. This is unfortunately, again, due to https://github.com/JanusGraph/janusgraph/issues/3732#issuecomment-1519395193. Briefly speaking, when `vertexId` is `key` column, we need to make sure it's not 8 bytes. When it's `column` column or `value` column, we don't have this restriction (because of the difference in long-encoding in these two different places).

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/util/IndexRecordUtil.java
```diff
@@ -107,23 +127,36 @@ public static String key2Field(ParameterIndexField field) {
...
 
-    public static String longID2Name(long id) {
-        Preconditions.checkArgument(id > 0);
-        return LongEncoding.encode(id);
+    public static String id2Name(Object id) {
+        IDUtils.checkId(id);
+        if (id instanceof Number) {
+            return LongEncoding.encode(((Number) id).longValue());
+        } else {
+            return STRING_ENCODING_MARKER + id.toString();
```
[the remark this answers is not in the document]
li-boxuan: Right. In mixed index stores, the encoding scheme is (again!!!) different.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BaseVertexCentricQueryBuilder.java
```diff
@@ -79,8 +80,8 @@
 
     private final SchemaInspector schemaInspector;
 
-    protected BaseVertexCentricQueryBuilder(SchemaInspector schemaInspector) {
-        this.schemaInspector = schemaInspector;
+    protected BaseVertexCentricQueryBuilder(StandardJanusGraphTx tx) {
+        this.schemaInspector = tx;
     }
```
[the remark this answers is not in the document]
li-boxuan: lol I don't remember. Maybe an accidental change. Anyways, it doesn't matter. I'll revert it.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
```diff
@@ -98,10 +126,22 @@ public boolean equals(Object other) {
...
-        s.append(LongEncoding.encode(relationId)).append(TOSTRING_DELIMITER).append(LongEncoding.encode(((Number) outVertexId).longValue()))
-            .append(TOSTRING_DELIMITER).append(LongEncoding.encode(typeId));
-        if (inVertexId != null && ((Number) inVertexId).longValue() > 0) {
-            s.append(TOSTRING_DELIMITER).append(LongEncoding.encode(((Number) inVertexId).longValue()));
+        s.append(LongEncoding.encode(relationId)).append(TOSTRING_DELIMITER);
+        if (outVertexId instanceof Number) {
+            s.append(LongEncoding.encode(((Number) outVertexId).longValue()));
+        } else {
+            assert outVertexId instanceof String;
+            s.append(STRING_ENCODING_MARKER).append(outVertexId);
```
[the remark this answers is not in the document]
li-boxuan: Right. `RelationIdentifier` (again!!!!!!!) uses a different encoding scheme.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
```diff
@@ -14,24 +14,52 @@
...
 /**
  * @author Matthias Broecheler (me@matthiasb.com)
  */
 
 public final class RelationIdentifier implements Serializable {
 
-    public static final String TOSTRING_DELIMITER = "-";
+    private static final Logger LOGGER = LoggerFactory.getLogger(RelationIdentifier.class);
+    public static final String JANUSGRAPH_RELATION_DELIMITER = "JANUSGRAPH_RELATION_DELIMITER";
+    public static final String TOSTRING_DELIMITER;
```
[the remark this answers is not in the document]
li-boxuan: It indeed must have only one character (otherwise `id.split(TOSTRING_DELIMITER)` could lead to wrong results). We have a check for that in the static initializer of this class. The variable is a String for convenience reasons - no need to cast string to char when we load from env variable (or system properties), plus no need to cast char back to string when we do `id.split(TOSTRING_DELIMITER);`.

Environment variables and System properties are always of String type, so there shouldn't be ambiguity/confusion.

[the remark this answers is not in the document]
li-boxuan: I'll explain in the doc that it should be a string with a single character.

janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
```diff
@@ -472,6 +489,16 @@ public Iterable<JanusGraphVertex> getVertices(Object... ids) {
...
+            assert id instanceof String || id instanceof Number;
+            if (!allowCustomVertexIdType && id instanceof String) {
+                // Convert string to long to keep backward compatibility prior to 1.0.0
+                // prior to JanusGraph 1.0.0, vertices always have ids of long types. However, even if
+                // a vertex has id 100L, both g.V(100L) and g.V("100") will return this vertex.
+                // Since JanusGraph 1.0.0, vertex id can be of either long type or string type. To
+                // keep backward compatibility, we hereby explicitly cast string id to long ids, if
+                // string custom vertex id functionality is disabled.
+                id = Long.valueOf((String) id);
+            }
```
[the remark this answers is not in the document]
li-boxuan: I totally forgot about this, thanks for the reminder. I added a one-liner test in `JanusGraphTest#testBasic`. Also verified the test passes on the master branch.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/tinkerpop/io/binary/RelationIdentifierGraphBinarySerializer.java
```diff
@@ -21,29 +21,93 @@
...
+ * To distinguish between string and long IDs, we use an extra byte as marker.
+ * Technically, we could save this byte by leveraging the fact that long-type IDs
+ * are always non-negative, but it would require changes from TinkerPop side as
+ * well.
+ */
 public class RelationIdentifierGraphBinarySerializer extends JanusGraphTypeSerializer<RelationIdentifier> {
+
+    public static final byte LONG_MARKER = 0;
+    public static final byte STRING_MARKER = 1;
+
```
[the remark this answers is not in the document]
li-boxuan: Thanks - it's actually straight-forward. I totally forgot to write tests for this class.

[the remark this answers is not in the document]
li-boxuan: Oh, wait it's already covered by `JanusGraphSerializerBaseIT`. I don't know why CodeCov still complains about it.

Anyways, I added it in the unit test: `RelationIdentifierGraphBinarySerializerTest`.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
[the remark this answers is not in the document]
porunov: This is funny how many different encoding schemes are there :smiling_face_with_tear: 
I think after this PR we might actually want to document them somewhere in the `Data model` or else.

janusgraph-driver/src/main/java/org/janusgraph/graphdb/tinkerpop/io/binary/RelationIdentifierGraphBinarySerializer.java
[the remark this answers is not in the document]
porunov: Yeah, Codecov often wrongly assumes that some part of the code isn't covered by tests and I get confused sometimes. Sorry for the false negative.

porunov:
For some reason 2 TP jobs are running more than 6 hours and then are canceled by the GitHub Actions. I'm restarting them, but I doubt they will pass

porunov:
@li-boxuan I think long-running TP tests are not related to this PR. Most likely we have a bug in `master` branch because I have the same problem here: in #3743 
For some reason TP tests are not executed in the PR but they are executed in my branch and you can see the same 2 jobs were canceled after 6 hours: https://github.com/porunov/janusgraph/actions/runs/4809004095/jobs/8559714661

li-boxuan:
Also, I don't know why `timeout-minutes: 480` defined in `ci-release.yml` does not work... Sometimes it's frustrating working with GitHub Actions :(

Anyway, I'll remove my naive attempt to increase the timeout from this PR. I'll also remove the `[tp-tests]` tag from this PR.

porunov:
> > Update: Scheduled TP tests are failing for 2 months now.
> 
> In my PR you can see sometimes the TP tests pass (e.g. [5c23fa1](https://github.com/JanusGraph/janusgraph/commit/5c23fa1c989d1d25ab9e5aa4c7c0be0fe277e3c1)). And when it fails, it fails due to different tasks. Thus I suspect it might be an issue with GitHub Actions rather than a bug of our own.

Hmm. I thought I identified the trouble commit in #3747 but now that you shared you commit which was later and TP tests passed for that commit - it's now strange. 
Could be that the problem is indeed with the GitHub Actions now and my investigation was influenced by that. That said, it could be that this particular commit was somehow lucky and TP tests passed for that commit.

Nevertheless, I created a draft PR #3749 which may potentially help with TP tests. I started TP tests for that PR and will see if it actually works or not.

porunov (approved):
LGTM. Thank you @li-boxuan for this great feature! :tada: 
I think it will help a number of people who can't use `Long` vertex ids due to migration complexity or ids allocation problem.

As for TinkerPop tests - I think this PR shouldn't influence them, but we don't have a good solution to execute TinkerPop tests at the moment ( #3747 ).

janusgraph-driver/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifier.java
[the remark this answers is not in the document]
li-boxuan: Yeah or maybe a dev doc

li-boxuan:
@porunov Thank you for reviewing this complicated PR! You caught a lot of points that I missed. Really appreciate it :)

li-boxuan:
@jianhuang-2 Hard to say because we have some blockers: https://github.com/JanusGraph/janusgraph/milestone/21

I am working on https://github.com/JanusGraph/janusgraph/issues/3763 right now and after that, I'll continue testing the string-type ID feature on my demo website. I will announce after testing is done, and then users could start trying this out without a formal release (we do build artifacts and publish on Maven for each commit, so an official release is not necessary for early trial/testing).

jianhuang-2:
@li-boxuan Thanks for the  quick update.

li-boxuan:
http://www.citegraph.io/ is back. This time, instead of using JanusGraph embedded (which, in my use case, has better performance because I only have one app running on one VM), I choose to use java gremlin driver + JanusGraph server which represents a more common use of JanusGraph. So far, so good.

I realize that setting `JANUSGRAPH_RELATION_DELIMITER` for the gremlin client is not necessary, but I'll keep the doc as it is (which suggests setting that parameter for both clients and servers) to avoid any confusion.

PR #3450, opened 2023-01-04
Enable Limited Batch Processing by default

### LINKED ISSUE #3186: Enable query.batch by default
AFAIK, most if not all users benefit from `query.batch` option (https://docs.janusgraph.org/configs/configuration-reference/#query). I don't see why we shouldn't enable it by default. A lot of new users won't bother reading docs carefully or tuning their configs, and they are likely frustrated by the poor performance without `query.batch`.

FlorianHockmann:
Batch processing has been disabled by default, probably because it can lead to an unreasonable high batch size which negatively affects performance.
Since we have limited batch processing now which limits the size of these batches, we can enable it by default as most users should benefit from this option.

Fixes #3186

-----

### For all changes:

### For code changes:


docs/changelog.md
```diff
@@ -134,6 +134,22 @@ Support for Gryo MessageSerializer [has been dropped in TinkerPop 3.6.0](https:/
 and we therefore also no longer support it in JanusGraph.
 GraphBinary is now used as the default MessageSerializer.hb
 
+##### Batch Processing enabled by default
```
li-boxuan, line 137: Shall we call it `limited batch processing enabled by default`, since we now also enable limited batch processing?

FlorianHockmann: I don't have a strong opinion on this, but I'd opted to just mention that batch processing itself is enabled by default as that is the major change in my opinion. I think the _limited_ part is only a configuration to let the batch processing behave better.

rngcntr (approved):
LGTM

docs/changelog.md
li-boxuan: Sounds fair!

li-boxuan (approved):

FlorianHockmann:
Looks like some tests now fail because of the changed default. Example:
```
HBaseGraphTest>JanusGraphTest.testVertexCentricQueryProfiling:6362 expected: <JanusGraphVertexStep([time.lt(10)])> but was: <JanusGraphMultiQueryStep>
```

I'll have to look into that and probably change the tests a bit.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/IndexMetricTest.java
```diff
@@ -75,8 +75,10 @@ public void testIndexMetrics() {
         Assertions.assertThrows(JanusGraphException.class,  () ->
             graph.indexQuery("idx", "!@#$%^").vertexTotals());
 
-        verifyIndexMetrics("search", METRICS_INDEX_PREFIX, ImmutableMap.of(M_MUTATE, 1L, M_QUERY, 1L, M_MIXED_AGG_QUERY, 5L, M_TOTALS, 2L, M_RAW_QUERY, 1L));
-        assertEquals(1, metric.getCounter(METRICS_INDEX_PREFIX, "search", M_TOTALS, MetricInstrumentedStore.M_EXCEPTIONS).getCount());
+        verifyIndexMetrics("search", METRICS_INDEX_PREFIX, ImmutableMap.of(M_MUTATE, 1L, M_QUERY, 1L,
+            M_MIXED_AGG_QUERY, 1L, M_TOTALS, 2L, M_RAW_QUERY, 1L));
```
FlorianHockmann, line 79: To be honest, it's not clear to me at all why `M_MIXED_AGG_QUERY` is now 1 instead of 5. Does this really mean that JanusGraph only performed one query to the index backend for the 5 max, min, sum, mean, and max traversals?
@rngcntr: Maybe you have a better understanding of this, given that you implemented the limited batch processing?

FlorianHockmann:
I think I've fixed all tests. (The failed tests passed on my machine, but I haven't executed all tests locally. So, I'm waiting for CI for confirmation.)
They basically required adding some `barrier()` / multi-query steps to expected traversals after optimizations.
Maybe you want to review this again to see whether the changes I made to the tests make sense, @li-boxuan @rngcntr.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/IndexMetricTest.java
rngcntr: I can't give an immediate answer to this either. After all, batching should not work across the boundaries of traversals :stuck_out_tongue: 
This could be utter nonsense, but I vaguely remember the elasticsearch API returning different kinds of aggregations (e.g. max, min, avg) even though a query only requires one of them. Together with result caching, that could in theory explain the first four queries being answered by a single index query. **However**, that does neither explain the fifth query, nor does should this behavior have changed with the introduction of batch processing.

FlorianHockmann: `p4` however isn't backed by an index. So, it shouldn't use a `MIXED_AGG_QUERY` for this traversal anyway, right?

Overall, I'm not sure to be honest what this test should exactly test. Do we expect to get certain metrics back or do we just want to test that we get somehow meaningful values for the different metrics?
To me, it looks more like the latter and I would therefore be OK with changing the test like this.
If we however want to verify that we get certain metrics back, then we might need to disable batch processing for just this test (?)

porunov: - Neither LuceneIndex (which is used for this test) nor ElasticSearch (not used for this test) returns multiple aggregations. They return a single aggregation you used.
- For this test the MixedAggQuery aggregation optimization isn't used for min/max/mean/sum when query batch is enabled, but they are used when query batch is disabled.

I didn't research the reason why the optimization is not used for `min/max/mean/sum` when query.batch is enabled, but below are the profiles (for `count` query optimization is used, but for `min/max/mean/sum` it's not used).

Profile for `graph.traversal().V().has("p1", "value").values("p3").max().profile().next()` (query.batch disabled):
```
Traversal Metrics
JanusGraphMixedIndexAggStep                                            1           1          24.265   100.00
  MixedIndexCountQuery                                                                        21.351
    \_query=[(p1 = value)]:idx
                                            >TOTAL                     -           -          24.265        -
```

Profile for `graph.traversal().V().has("p1", "value").values("p3").max().profile().next()` (query.batch enabled):
```
Traversal Metrics
JanusGraphStep([],[p1.eq(value)])                                      1           1           4.598    41.20
  constructGraphCentricQuery                                                                   0.578
  GraphCentricQuery                                                                            3.974
    \_condition=(p1 = value)
    \_orders=[]
    \_isFitted=true
    \_isOrdered=true
    \_query=[(p1 = value)]:idx
    \_index=idx
    \_index_impl=search
JanusGraphMultiQueryStep                                               1           1           0.832     7.46
NoOpBarrierStep(null)                                                  1           1           0.195     1.75
JanusGraphPropertiesStep([p3],value)                                   1           1           3.476    31.16
  \_condition=type[p3]
  \_orders=[]
  \_isFitted=true
  \_isOrdered=true
  \_query=p3:SliceQuery[0x50E0,0x50E1)
  \_multi=true
  \_vertices=1
  optimization                                                                                 0.317
MaxGlobalStep                                                          1           1           2.056    18.43
                                            >TOTAL                     -           -          11.159        -
```

FlorianHockmann: Thanks for investigating and the detailed explanation, @porunov! Your explanation made it quite clear why the batch processing (implemented in the `JanusGraphLocalQueryOptimizerStrategy`) effectively disabled the mixed index aggregations: The `JanusGraphMultiQueryStep` that was added by the `JanusGraphLocalQueryOptimizerStrategy` also includes the aggregation query that we can perform with the mixed index backend, but the mixed index aggregation optimization isn't able to work on a `JanusGraphMultiQueryStep`.
I fixed this by making sure that the mixed index aggregation optimizations are always applied before the `JanusGraphLocalQueryOptimizerStrategy`.

porunov (approved):
LGTM! Thank you @FlorianHockmann for tuning and the fix. I see the failed tests should be resolved after #3525 is merged.

FlorianHockmann:
@li-boxuan and @rngcntr do you want to review the latest changes here again? Otherwise, I'd merge this early next week.

docs/operations/batch-processing.md
```diff
@@ -122,14 +130,14 @@ traversal can only aggregate traversers once they enter the
...
+- ... you have a mixture of traversals that traverse a high number of vertices
+  and traversals that only access few vertices of the graph.
 
 ### Possible limitations
-- Increased memory consumption
+- Increased memory consumption (compared to no batch processing)
 - The performance of queries depends on the configured batch
   size.
-  If you switch to this configuration, make sure that the
+  If you use to this configuration, make sure that the
```
li-boxuan, line 140:
```suggestion
  If you use this configuration, make sure that the
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AbstractJanusGraphMixedIndexAggStrategy.java
```diff
@@ -47,6 +50,13 @@ public void apply(final Traversal.Admin<?, ?> traversal) {
         });
     }
 
+    @Override
+    public Set<Class<? extends ProviderOptimizationStrategy>> applyPost() {
+        Set<Class<? extends ProviderOptimizationStrategy>> postStrategies = new HashSet<>();
+        postStrategies.add(JanusGraphLocalQueryOptimizerStrategy.class);
```
li-boxuan, line 56: nitpick: we can use a static final variable with value `Collections.singleton(JanusGraphLocalQueryOptimizerStrategy.class)`.

li-boxuan (approved):
LGTM except for two nitpicks!

janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/strategy/AbstractJanusGraphMixedIndexAggStrategy.java
FlorianHockmann: Good suggestion, I also changed it in a different strategy from which I copied this implementation.

porunov: I noticed this as well but I checked that in every JanusGraph strategy we use this approach. I don't know exactly why we don't use constant in JanusGraph strategies here.
Nevertheless, I think it would make sense to change that in each strategy and not only in the current strategy. Thus, I think it's OK to do it in a separate PR

FlorianHockmann: Ah, I think I misunderstood your suggestion a bit, @li-boxuan. I just changed the implementation to use `Collections.singleton()` directly here, but didn't add a final variable. Let me add that and I'll also do it for the other strategies.
If we're already discussing this here, then I think it's the easiest if we also implement it here so you can directly review it.

FlorianHockmann: done

porunov (approved):

li-boxuan (approved):

PR #3451, opened 2023-01-05
Improve docker section in installation guide

li-boxuan:
There are a few wording that could be confusing to newcomers. This commit fixes them and also add a warning on usage of ARM containers.

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -5,9 +5,10 @@
 For virtualization and easy access, JanusGraph provides a [Docker image](https://hub.docker.com/r/janusgraph/janusgraph).
 Docker makes it easier to run servers and clients on a single machine without dealing with multiple installations.
 For instructions on installing and using Docker, please refer to the [docker guide](https://docker.com/get-started).
-With JanusGraph server running in one container, we can either run the client in another container, or on our bare-metal machine.
+With JanusGraph Server running in one container, we can either run the client in another container, or on our bare-metal machine.
+In the examples below, we use Gremlin Console as the client.
 
-### Running Client inside another Docker container
+### Running Gremlin Console inside another Docker container
 
 Let's start by running a simple JanusGraph instance in Docker:
 
@@ -39,7 +40,7 @@ The server may need a few seconds to start up so be patient and wait for the cor
     ```
 
 We can now instruct Docker to start a second container for the client and link it to the already running server. Here we
-use gremlin console (`gremlin.sh`) as the client.
+use Gremlin Console (`gremlin.sh`) as the client.
 
 ```bash
 $ docker run --rm --link janusgraph-default:janusgraph -e GREMLIN_REMOTE_HOSTS=janusgraph \
@@ -64,10 +65,10 @@ gremlin> :remote connect tinkerpop.server conf/remote.yaml
     Due to a known [issue](https://github.com/JanusGraph/janusgraph-docker/issues/123), `gremlin.sh` won't work in ARM
     containers.
 
-### Running Client on Bare-metal
+### Running Gremlin Console on Bare-metal
 
-We can also run the client on our bare-metal machine. To make the server able to communicate with the client, we need
-to expose the 8182 port when launching JanusGraph server:
+We can also run the Gremlin Console on our bare-metal machine. To make the server able to communicate with the gremlin
+console, we need to expose the 8182 port when launching JanusGraph server:
 
 ```bash
 $ docker run -it -p 8182:8182 janusgraph/janusgraph
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -16,7 +16,7 @@ Let's start by running a simple JanusGraph instance in Docker:
 $ docker run --name janusgraph-default janusgraph/janusgraph:latest
 ```
 
-The above command launches a JanusGraph server instance. We name the container so that we can link a second container to it.
+The above command launches a JanusGraph Server instance. We name the container so that we can link a second container to it.
 The server may need a few seconds to start up so be patient and wait for the corresponding log messages to appear.
 
 ??? note "Example log"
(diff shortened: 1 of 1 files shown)


docs/getting-started/installation.md
```diff
@@ -42,7 +43,6 @@ $ bin/gremlin.sh
 gremlin> :remote connect tinkerpop.server conf/remote.yaml
 ==>Configured localhost/127.0.0.1:8182
 ```
-Notice that the client side of this works exactly the same as before when running both the client and server locally without Docker.
```
li-boxuan, line 45: This sentence can be misleading - not sure what "as before" means. Seems this section was after the local installation guide but now it's before the local installation guide. Anyways, I felt it was not important and thus removed it.

docs/getting-started/installation.md
```diff
@@ -32,7 +32,8 @@ The server may need a few seconds to start up so be patient and wait for the cor
     3965 [gremlin-server-boss-1] INFO  org.apache.tinkerpop.gremlin.server.GremlinServer  - Channel started at port 8182.
     ```
 
-We can now start a Gremlin Console on our local device and try to connect to the new server:
+We can now start a Gremlin Console and try to connect to the new server. We can do so either on our local device (see local
+installation section below) or inside the Docker container.
```
FlorianHockmann, line 36: _inside the Docker container_ <- Do we really want to document that? I think it's easier and a best practice in general with Docker containers to not start a second process inside the Docker container, but to instead start a different container for the console.

docs/getting-started/installation.md
```diff
@@ -63,6 +63,10 @@ gremlin> :remote connect tinkerpop.server conf/remote.yaml
 Notice how it's not necessary to bind any ports in order to make this example work.
 For further reading, see the [JanusGraph Server](../operations/server.md) section as well as the [JanusGraph Docker documentation](https://github.com/JanusGraph/janusgraph-docker/blob/master/README.md).
 
+!!! warning
+    Due to a known [issue](https://github.com/JanusGraph/janusgraph-docker/issues/123), `gremlin.sh` won't work in ARM
```
FlorianHockmann, line 67: We should link from the issue to this line once this is merged so we don't forget to remove the warning once we have a fix for the issue.

docs/getting-started/installation.md
FlorianHockmann: I understand the confusion and it probably also originated from this section having been below the local installation originally.
But my problem with this approach is that it's recommending something that is generally considered to not be good practice for Docker. See for example [this part of the official Docker docs](https://docs.docker.com/config/containers/multi-service_container/):

> It is generally recommended that you separate areas of concern by using one service per container.

or [this tutorial](https://www.tutorialworks.com/containers-single-or-multiple-processes/): 

> One of these principles is that there should just be one process running in a container. That is to say, a Docker container should have just one program running inside it.

How about if we avoid the confusion by only linking to the local installation section for the Gremlin Console here and then directly continuing with the approach that starts a second JanusGraph container for the console?

Or if you really want to document starting the console within the same container, then I would mention that this is only meant for quick testing or something like that.

li-boxuan: > But my problem with this approach is that it's recommending something that is generally considered to not be good practice for Docker

Well said. Let me try to paraphrase this section and see if we can make it more clear to users.

docs/getting-started/installation.md
```diff
@@ -32,35 +38,54 @@ The server may need a few seconds to start up so be patient and wait for the cor
     3965 [gremlin-server-boss-1] INFO  org.apache.tinkerpop.gremlin.server.GremlinServer  - Channel started at port 8182.
     ```
 
-We can now start a Gremlin Console on our local device and try to connect to the new server:
+We can now instruct Docker to start a second container for the client and link it to the already running server. Here we
+use gremlin console (`gremlin.sh`) as the client.
```
FlorianHockmann, line 42:
```suggestion
use Gremlin Console (`gremlin.sh`) as the client.
```

(Gremlin Console is a name)

docs/getting-started/installation.md
```diff
@@ -5,11 +5,17 @@
 For virtualization and easy access, JanusGraph provides a [Docker image](https://hub.docker.com/r/janusgraph/janusgraph).
 Docker makes it easier to run servers and clients on a single machine without dealing with multiple installations.
 For instructions on installing and using Docker, please refer to the [docker guide](https://docker.com/get-started).
-Let's try running a simple JanusGraph instance in Docker:
+With JanusGraph server running in one container, we can either run the client in another container, or on our bare-metal machine.
+
+### Running Client inside another Docker container
```
FlorianHockmann, line 10: I think _Client_ isn't very specific as this could be any client that can access JanusGraph Server. So, I would say "the Gremlin Console" instead and use the same in other places in this article.

docs/getting-started/installation.md
```diff
@@ -5,11 +5,17 @@
 For virtualization and easy access, JanusGraph provides a [Docker image](https://hub.docker.com/r/janusgraph/janusgraph).
 Docker makes it easier to run servers and clients on a single machine without dealing with multiple installations.
 For instructions on installing and using Docker, please refer to the [docker guide](https://docker.com/get-started).
-Let's try running a simple JanusGraph instance in Docker:
+With JanusGraph server running in one container, we can either run the client in another container, or on our bare-metal machine.
```
FlorianHockmann, line 8:
```suggestion
With JanusGraph Server running in one container, we can either run the client in another container, or on our bare-metal machine.
```

JanusGraph Server is also a name, just like Gremlin Console.

FlorianHockmann (commented):
This looks really good to me. I only added some comments, but those are only small linguistic details.

docs/getting-started/installation.md
```diff
@@ -5,11 +5,18 @@
...
+
+Let's start by running a simple JanusGraph instance in Docker:
+
 ```bash
-$ docker run -it -p 8182:8182 janusgraph/janusgraph
+$ docker run --name janusgraph-default janusgraph/janusgraph:latest
 ```
-We run the image interactively and request Docker to make the container's port `8182` available for us to see.
+
+The above command launches a JanusGraph server instance. We name the container so that we can link a second container to it.
```
FlorianHockmann, line 19:
```suggestion
The above command launches a JanusGraph Server instance. We name the container so that we can link a second container to it.
```

FlorianHockmann (approved):

li-boxuan:
backported [c0b8e1d](https://github.com/JanusGraph/janusgraph/commit/c0b8e1d77df87a394e91d06ff5d651ceea636194) to v0.6

Issue #3199, opened 2022-09-02
Inconsistent result for query when whether to use JanusGraphMixedIndexCountStrategy


dh-cloud:
- Version:   0.6.0, 0.6.1, 0.6.2
- Storage Backend: hbase
- Mixed Index Backend:  elasticsearch
- Expected Behavior: I would expect the result to stay the same even when whether use JanusGraphMixedIndexCountStrategy
- Current Behavior: The result is different when use JanusGraphMixedIndexCountStrategy or not
- Steps to Reproduce:
```
GraphTraversalSource g = graph.traversal();
JanusGraphManagement m = graph.openManagement();

VertexLabel person = m.makeVertexLabel("person").make();
PropertyKey sex = m.makePropertyKey("sex").dataType(String.class).make();
m.buildIndex("personIndex", Vertex.class).addKey(sex).indexOnly(person).buildMixedIndex("search");
m.commit();

g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();
g.addV("person").property("sex", "man").next();

g.tx().commit();

System.out.println(g.V().has('person','sex','man').limit(2).count().next()); // 10

g = graph.traversal().withoutStrategies(JanusGraphMixedIndexCountStrategy.class)
System.out.println(g.V().has('person','sex','man').limit(2).count().next()); // 2
```

The ultimate reason is when query.hasLimit() the /_search API should be used instead of the /_count API

PR #3452, opened 2023-01-06
Fixes issue #3199

dh-cloud:
The root cause of the problem is that the count query result is returned directly without respect the limit setting.

Fixes issue #3199

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -1664,6 +1664,7 @@ public abstract class JanusGraphIndexTest extends JanusGraphBaseTest {
         }};
         checkMixedIndexCountProfiling(profile, annotations);
         assertEquals(total, graph.traversal().V().has("text", Text.textContains("ducks")).count().next());
+        assertEquals(1, graph.traversal().V().has("text", Text.textContains("ducks")).limit(1).count().next());
 
         // edge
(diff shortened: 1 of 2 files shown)


li-boxuan (commented):
Thank you for the fix! Do you mind adding some test cases for this bug fix?

Also, please remember to sign-off your name in the commit message :)

porunov (commented):
@dh-cloud thank you for the fix and sorry for the late reply.
It looks like this fix is applied to `ElasticSearch` index provider only and not to all indexes.
I think it would make sense to fix it on some level above to fix it for all index providers and not only for ElasticSearch.
I would probably look somewhere at `IndexTransaction` at method `totals(RawQuery query)`. I think it might make more sense to fix it there rather then for specific index provider. That said, fixing it for all index providers should work as well.
I didn't check but I assume the following changes to that method would fix the issue for all index providers:
```
public Long totals(RawQuery query) throws BackendException {
    Long totals = index.totals(query, keyInformation,indexTx);
    if (query.hasLimit()){
        totals = Math.min(totals, query.getLimit());
    }
    return totals;
}
```

I also didn't looks at test changes yet but noticed that the changes in `JanusGraphIndexTest` lead to failures for all index providers.

P.S. you can retarget your PR to `master` and we will handle backporting to `v0.6`. Targeting this PR to `v0.6` branch is OK as well.

porunov:
I have created a replacement PR which fixes count for each index provider here #3570 
Thus, closing this PR. Thank you @dh-cloud for finding the bug and initiating a fix for this issue.

PR #3463, opened 2023-01-12
[v0.6] Fix edgeID inconsistency between OLAP and OLTP queries

FlorianHockmann:
# Backport

This will backport the following commits from `master` to `v0.6`:
 - [Fix edgeID inconsistency between OLAP and OLTP queries](https://github.com/JanusGraph/janusgraph/pull/3377)

### Questions ?

### DIFF
--- a/janusgraph-hadoop/pom.xml
+++ b/janusgraph-hadoop/pom.xml
@@ -116,6 +116,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.janusgraph</groupId>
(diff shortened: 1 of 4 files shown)


FlorianHockmann:
Backporting this contribution seems non-trivial. There are some issues related to vertex ID types which need to be `long` in `v0.6` instead of `object`. These are easy to resolve, but the added code also uses functionality that was only introduced to `master` in #3184.
I think it's OK to only let this land in 1.0.0. If someone really wants this in 0.6.3, then we can of course still reconsider, but I personally don't see much value in putting the effort now to backport this.

PR #3469, opened 2023-01-18
Document JanusGraph.Net

### LINKED ISSUE #1276: Create docs for JanusGraph GLV clients

FlorianHockmann:
The getting started part for .NET still only uses Gremlin.Net to keep things as simple as possible. A second section then informs about how to get support for JanusGraph specific things through JanusGraph.Net.

Fixes #1276. #1276 also talks about Python but since we still don't have an official Python library, I'd like to reduce the scope of that issue to just .NET. We can later create another issue for Python once we haven an official library for that.

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/interactions/connecting/dotnet.md
+++ b/docs/interactions/connecting/dotnet.md
@@ -32,7 +32,7 @@ var g = Traversal().WithRemote(new DriverRemoteConnection(client));
         // Reuse 'g' across the application
 ```
 
-2.  Execute a simple traversal:
+1.  Execute a simple traversal:
 ```csharp
 var herculesAge = g.V().Has("name", "hercules").Values<int>("age").Next();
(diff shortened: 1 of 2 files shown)


porunov (approved):
Not using .NET, but the doc changes look good to me

PR #3501, opened 2023-01-26
SubqueryIterator should ignore invalid elements

### LINKED ISSUE #3500: SubqueryIterator sometimes returns incomplete results

- Version: 0.6.2
- Storage Backend: cql
- Mixed Index Backend: elasticsearch
- Link to discussed bug: N/A 
- Expected Behavior: All valid results should be returned
- Current Behavior: Valid results are returned up until the first invalid result
- Steps to Reproduce: Insert entry into mixed index that matches the subquery with an invalid element ID.

For my deployment, this bug originated with a data corruption. An obvious solution is to purge the mixed index of these invalid entries, but for a large graph, this purge could take considerable resources and time. In the interim, there is not much justification for the SubqueryIterator to fail to return the valid elements.

This issue originates here: https://github.com/janusgraph/janusgraph/blob/v0.6.2/janusgraph-core/src/main/java/org/janusgraph/graphdb/util/SubqueryIterator.java#L70. An invalid element will map through `function` to `null`, and the implementation of `hasNext()` for the converted stream will return `false` on the first `null` result. There is no valid use-case for terminating the iterator early in this case, so these `null` values should be logged and ignored.

geekmug:
If the mixed index contains invalid elements, then the map will return null, which causes the stream to end, giving incomplete results. This change logs the invalid element identifier, but also filters the null elements out of the stream to allow the iterator to continue.

Fixes #3500

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java
@@ -45,6 +45,7 @@ import org.janusgraph.core.JanusGraphQuery;
 import org.janusgraph.core.JanusGraphTransaction;
 import org.janusgraph.core.JanusGraphVertex;
 import org.janusgraph.core.JanusGraphVertexProperty;
+import org.janusgraph.core.Multiplicity;
 import org.janusgraph.core.PropertyKey;
 import org.janusgraph.core.VertexLabel;
 import org.janusgraph.core.attribute.Cmp;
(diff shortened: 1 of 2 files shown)


porunov (commented):
Do you know a situation when a mixed index contains invalid elements? Is it due to manual intervention or is it a regular behavior you see? Would you be able to write a test to show this problem?

geekmug:
The only scenario where this came up was a backend failure. The two scenarios that are possible is that 1) you have a database failure on the storage database that causes a silent loss of records, or 2) you delete something from the storage database, but the request to the mixed index fails (either temporary or permanent, because the mixed index failing does not cause a rollback of the transaction). Neither of those scenarios are "normal", so I found it difficult to imagine how to write a unit test for it. The easiest method would be to write a test for a specific mixed index backend (e.g., explicitly inserting a bogus document into Elasticsearch as part of the test setup), but the code to pull that off would be considerably more complex than this null-check.

porunov:
> The only scenario where this came up was a backend failure. The two scenarios that are possible is that 1) you have a database failure on the storage database that causes a silent loss of records, or 2) you delete something from the storage database, but the request to the mixed index fails (either temporary or permanent, because the mixed index failing does not cause a rollback of the transaction). Neither of those scenarios are "normal", so I found it difficult to imagine how to write a unit test for it. The easiest method would be to write a test for a specific mixed index backend (e.g., explicitly inserting a bogus document into Elasticsearch as part of the test setup), but the code to pull that off would be considerably more complex than this null-check.

I see. In that case, I think your issue would be related to #1099 .
There was a PR #3327 which added an util class to remove stale index records. I would expect this PR to break those tests which were added in that PR, but quickly checking I see it didn't break anything.
Checking it more, I see that the log which is expected to be triggered on a broken element (i.e. stale element) is never logged.
A little deeper investigation shows that `JanusGraphElement r = function.apply(e);` never returns `null` element (at least for normal data vertices).
Even in case of a cache miss it will create a `CacheVertex` (which is simply a lazy loading vertex). Thus, a `CacheVertex` will be returned and not `null`.
Of course, CacheVertex won't be able to load any properties (because there are none for a stale element), but it itself will be seen as a valid vertex.

A test example to reproduce this issue is below (you can copy and paste it into `JanusGraphIndexTest.java`).
A test flow is the next:
1) Create a normal vertex which is indexed.
2) Remove that vertex.
3) Inject a mixed index record. This step introduces a stale index situation when a vertex is removed from the database but then failed to be removed from the mixed index backend.
4) Check that this PR fixes an issue by not returning that element. This step (last line) fails.

```
@Test
public void testElementsAreSkippedFromStaleMixedIndex() throws BackendException {
    clopen();
    String namePropKeyStr = "name";
    String indexName = "mixed";
    PropertyKey nameProp = mgmt.makePropertyKey(namePropKeyStr).dataType(String.class).cardinality(Cardinality.SINGLE).make();
    mgmt.buildIndex(indexName, Vertex.class).addKey(nameProp, getStringMapping()).buildMixedIndex(INDEX);
    finishSchema();

    String nameValue = "NameTestValue";

    tx.traversal().addV().property(namePropKeyStr, nameValue).iterate();
    newTx();

    Vertex testVertex = tx.traversal().V().has(namePropKeyStr, nameValue).next();

    ManagementSystem managementSystem = (ManagementSystem) graph.openManagement();
    JanusGraphIndex janusGraphIndex = managementSystem.getGraphIndex(indexName);
    JanusGraphSchemaVertex indexChangeVertex = managementSystem.getSchemaVertex(janusGraphIndex);
    MixedIndexType index = (MixedIndexType) indexChangeVertex.asIndexType();

    JanusGraphVertexProperty indexedNameProperty = (JanusGraphVertexProperty) testVertex.property(namePropKeyStr);
    PropertyKey namePropertyKey = indexedNameProperty.propertyKey();

    tx.traversal().V().has(namePropKeyStr, nameValue).drop().iterate();
    newTx();

    JanusGraphElement element = new CacheVertex((StandardJanusGraphTx) tx, ((JanusGraphElement) testVertex).longId(), ElementLifeCycle.New);

    IndexUpdate<String, IndexEntry> nameUpdate = IndexRecordUtil.getMixedIndexUpdate(
        element,
        namePropertyKey,
        nameValue,
        index,
        IndexMutationType.ADD
    );

    BackendTransaction transaction = ((StandardJanusGraphTx) tx).getTxHandle();
    IndexTransaction indexTransaction = transaction.getIndexTransaction(index.getBackingIndexName());

    // Adds index entry of the non-existing vertex to the index. I.e. this line corrupts index.
    // This is a possible situation of a stale index being introduced.
    indexTransaction.add(index.getStoreName(), nameUpdate.getKey(), nameUpdate.getEntry(), true);
    newTx();

    assertFalse(tx.traversal().V(testVertex.id()).hasNext());
    assertTrue(tx.traversal().V().has(namePropKeyStr, nameValue).hasNext());

    // A vertex with `namePropKeyStr == nameValue` is removed from the storage backend, but it's stale record
    // exists in the mixed index.
    // Below we would expect that the new logic would skip the broken element (i.e. stale element), but instead
    // it returns this element (just with no properties).
    assertEmpty(tx.traversal().V().has(namePropKeyStr, nameValue).valueMap().with(WithOptions.tokens, WithOptions.ids).toList());
}
```

porunov:
@geekmug let me know if you find a way of reproducing that `log` which is added in that PR. It can come not only in form of a test but in any other form. I.e. like a step guide when I add / remove an element directly to ElasticSearch or add / drop an element directly to Cassandra, etc. In case I see it's possible to reproduce the log you added we can reconsider this fix, but for now I don't see a way how this fix helps in any of the situations.

geekmug:
I see what you are saying. This fix was applied to a database that is in use and is currently producing this log entry, so I will investigate it closer. I am thinking this may only be an issue for a mixed index on edges. I appreciate the help with bootstrapping a unit test, I think I can adapt the unit test to do the same for an edge elements.

porunov:
> I see what you are saying. This fix was applied to a database that is in use and is currently producing this log entry, so I will investigate it closer. I am thinking this may only be an issue for a mixed index on edges. I appreciate the help with bootstrapping a unit test, I think I can adapt the unit test to do the same for an edge elements.
> 
> Yep, it's an issue with a mixed index on the edge, because `RelationIdentifierUtils.findRelation(id)` will return `null` if the vertex does not exist, in which case the conversion function for edges will also return `null`.

Thank you for explaining the problem. It does make sense now. I rebased this PR against later `master` branch and added a test case to reproduce an issue you described.

porunov (approved):
LGTM. Thank you @geekmug !
I verified that the added test fails without the provided fix and works with it.

porunov:
Merging with lazy consensus

PR #3503, opened 2023-01-26
Should not retry CQL ServerErrors for large key values

geekmug:
For Cassandra, the QueryValidationException was wrapped in an ExecutionException causing the EXCEPTION_MAPPER to not match, but for SycllaDB, the failure is reported as a general ServerError, so we can only match on this condition by inspectng the detail message.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
+++ b/janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
@@ -14,6 +14,7 @@
 
 package org.janusgraph.diskstorage.cql;
 
+import com.datastax.oss.driver.api.core.AllNodesFailedException;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.cql.BatchableStatement;
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
@@ -24,6 +25,7 @@ import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.metadata.TokenMap;
 import com.datastax.oss.driver.api.core.servererrors.QueryValidationException;
+import com.datastax.oss.driver.api.core.servererrors.ServerError;
 import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.querybuilder.delete.DeleteSelection;
 import com.datastax.oss.driver.api.querybuilder.insert.Insert;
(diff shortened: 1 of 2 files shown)


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -117,9 +121,49 @@ public class CQLKeyColumnValueStore implements KeyColumnValueStore {
...
+                    ServerError error = (ServerError) throwable;
+                    if (error.getMessage().startsWith("Key size too large:")) {
+                        return true;
+                    }
+                    return false;
+                }
+                return false;
+            });
+        });
+    }
```
porunov, line 150: As Codacy commented in this PR.
1) it's better to declare all methods after all fields are declared. I.e. it's better to move this method down after all fields are declared.
2) It's possible to make boolean statements easier without if / else statements. I.e. something like below:
```
private static boolean isKeySizeTooLarge(Throwable cause) {
    AllNodesFailedException failed = ExceptionUtils.throwableOfType(cause, AllNodesFailedException.class);
    if (failed == null) {
        return false;
    }
    return failed.getAllErrors().values().stream().anyMatch(errors ->
        errors.stream().anyMatch(error -> (error instanceof ServerError)
            && error.getMessage().startsWith("Key size too large:")));
}
```

porunov (approved):
I have only a single nitpick about `isKeySizeTooLarge` as Codacy commented. Otherwise the fix looks good to me. Thank you @geekmug !

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -117,9 +121,37 @@ public class CQLKeyColumnValueStore implements KeyColumnValueStore {
...
+     * This type of ServerError is a permanent failure, since the same query with
+     * the same value will always produce this error. Properly treating this as a
+     * permanent failure avoids excessive retrying of bad CQL queries and allows
+     * upstream code to properly report the exception.
+     */
+    private static boolean isKeySizeTooLarge(Throwable cause) {
+        AllNodesFailedException failed = ExceptionUtils.throwableOfType(cause, AllNodesFailedException.class);
+        return failed != null && failed.getAllErrors().values().stream().anyMatch(errors ->
+            errors.stream().anyMatch(error -> (error instanceof ServerError)
+                && error.getMessage().startsWith("Key size too large:")));
```
geekmug, line 153: For reference, this originates from this bit of code in ScyllaDB: https://github.com/scylladb/scylladb/blob/8d7e35caefecc053ba5495c519cc270632c6db1a/compound.hh#L106

geekmug:
@porunov Thanks for the feedback and code review! Your changes looked good to me and I re-tested everything locally. Thanks!

porunov (approved):

porunov:

PR #3504, opened 2023-01-26
Avoid copying all properties and edges into memory for IndexRepairJob

geekmug:
This is especially important if the collection of edges is very large, which could cause heap exhaustion, preventing any indexing from being possible.

-----

### For all changes:
- [ x Is your initial contribution a single, squashed commit?

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
@@ -54,6 +54,7 @@ import org.janusgraph.util.StringUtils;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
```diff
@@ -182,18 +183,16 @@ public void process(JanusGraphVertex vertex, ScanMetrics metrics) {
                 assert indexType!=null;
                 IndexSerializer indexSerializer = graph.getIndexSerializer();
                 //Gather elements to index
-                List<JanusGraphElement> elements;
+                Iterable<JanusGraphElement> elements;
```
porunov, line 186: I think it could be used as `Iterator` and traversed below because we access elements one by one only once below. 
```
Iterator<? extends JanusGraphElement> elements;
switch (indexType.getElement()) {
    case VERTEX:
        elements = Collections.singletonList(vertex).iterator();
        break;
    case PROPERTY:
        elements = addIndexSchemaConstraint(vertex.query(),indexType).properties().iterator();
        break;
    case EDGE:
        elements = addIndexSchemaConstraint(vertex.query().direction(Direction.OUT),indexType).edges().iterator();
        break;
    default: throw new AssertionError("Unexpected category: " + indexType.getElement());
}
```
And then below instead of traversing an Iterable we could traverse Iterator.
I.e. 
```
elements.forEachRemaining(element -> {
    if (indexSerializer.reindexElement(element, (MixedIndexType) indexType, documentsPerStore)) {
        metrics.incrementCustom(DOCUMENT_UPDATES_COUNT);
    }
});
```

porunov (commented):
Thank you @geekmug for contribution!
The change makes sense to me, but I usually deemed to avoid using Guava when native Java replacement exists (see ticket #2033). I think in this case a similar approach could be done with using Iterator (see attached comment).

janusgraph-core/src/main/java/org/janusgraph/graphdb/olap/job/IndexRepairJob.java
geekmug: Good call. There really is no reason it can't be an Iterator, and that also avoids the type-safety issue that was causing the need to use `Iterables.transform(..., v -> v)` to unwrap and rewrap the element.

porunov (approved):
LGTM. Thank you @geekmug !

porunov:
Merging following lazy consensus

PR #3572, opened 2023-02-10
Apply limit and offset to IndexProvider.totals method

### LINKED ISSUE #3571: Apply limit and offset to IndexProvider.totals method
As for now `IndexProvider.totals` don't take into account `limit` and `offset`, but it makes sense to use for consistency between `stream` equivalent methods. Foe example, the below queries should return the same count but they return different count because `vertexTotals` doesn't apply `offset` and `limit` but `vertexStream` does:
https://github.com/JanusGraph/janusgraph/blob/87360779795e7441c4cf7e5cd4ce21eacd3b0afd/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java#L1941-L1942

porunov:
Fixes #3571

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -110,6 +110,34 @@ For more information on features and bug fixes in 1.0.0, see the GitHub mileston
 The index management has received an overhaul which enables proper index removal.
 The schema action `REMOVE_INDEX` is no longer available and has been replaced by `DISCARD_INDEX`.
 
+##### `totals` for direct index queries now applies provided offset and limit
+
+Direct index queries which search count for totals (`vertexTotals`, `edgeTotals`, `propertyTotals` and direct execution of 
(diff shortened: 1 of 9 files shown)


porunov:
@JanusGraph/committers I think it makes sense to introduce `totals` breaking change for direct index queries in JanusGraph 1.0.0 and make the logic taking `limit` and `offset` into account instead of ignoring those parameters. 
That said, if anyone has any concerns or thoughts about this breaking change. Please, let me know.
I personally don't have any specific reason to introduce this breaking change. I initiated it just because it looked strange to me that `totals` ignore `limit` and `offset`, but `stream` doesn't. That said, maybe I just missed some points why it was designed like that.

porunov:
@JanusGraph/committers I have intention to merge this PR by following lazy consensus tomorrow. In case anyone needs more time to review it, please, let me know. This is a breaking change.

li-boxuan (approved):
LGTM 👍

Issue #2505, opened 2021-03-07
Hadoop integration tests are failing for scylladb


li-boxuan:
See https://github.com/JanusGraph/janusgraph/pull/2352#discussion_r587617149

The same tests pass for Cassandra but fail for Scylladb.

porunov:
I confirm that locally running those tests are very slow when using Scylla and fast when using Cassandra.
I don't know the reason for this behavior but here are my observations:
- In [this blog post](https://docs.scylladb.com/stable/kb/scylla-and-spark-integration.html#scylla-limitations) they are talking about limitations and they said: `Scylla doesn’t populate system.size_estimates, and therefore the connector won’t be able to perform automatic split sizing optimally.`. Not sure if that affect tests or not but could be.
- The test which is running 7 minutes when Cassandra is used ([here](https://github.com/JanusGraph/janusgraph/actions/runs/4155096401/jobs/7187951184)) takes more than 3 hours to finish when ScyllaDB is used ([here](https://github.com/JanusGraph/janusgraph/actions/runs/4155096422/jobs/7187958366)).
- For Hadoop we are using Cassandra's [CqlInputFormat](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hadoop/cql3/CqlInputFormat.java) which is using datastax 3.x driver instead of datastax 4.x driver. I don't think that's a driver problem because it works fine with Cassandra, but something to know about just in case.

I also think it might make sense to warn users about using ScyllaDB with JanusGraph Hadoop integration. I don't know if Hadoop integration will be performing normally when used with ScyllaDB (at least I didn't test ScyllaDB performance for that integration myself).

cc @dorlaor

dorlaor:
Hmm, does Scylla in this test use the same bindings as C*? It should work too.
It might be the docker setup of Scylla or it can be related to the size estimates
although this https://www.scylladb.com/2018/08/21/spark-scylla-2/ shows Scylla does
support it and we do have it in our docs. Could be that the estimate may not be accurate.

Is it possible to run in debug mode and figure out the differences?

porunov:
> Hmm, does Scylla in this test use the same bindings as C*? It should work too. It might be the docker setup of Scylla or it can be related to the size estimates although this https://www.scylladb.com/2018/08/21/spark-scylla-2/ shows Scylla does support it and we do have it in our docs. Could be that the estimate may not be accurate.
> 
> Is it possible to run in debug mode and figure out the differences?

[...]

PR #3578, opened 2023-02-11
Add ScyllaDB driver support

### LINKED ISSUE #1778: Dedicated Scylla backend
Scylla now has [its own Java driver](https://github.com/scylladb/java-driver/) which is a fork of DataStax's CQL driver we currently use for the CQL backend with some Scylla specific optimizations. These optimizations are described in [this presentation from the 2018 Scylla summit](https://www.slideshare.net/ScyllaDB/scylla-summit-2018-scylla-feature-talks-gains-by-using-scyllaspecific-drivers) which hints at significant latency reductions thanks to shard awareness in the driver.

It would be great if we could use this driver together with our existing CQL backend, so we get the performance improvements without having to maintain yet another full backend.

### LINKED ISSUE #2451: Officially support Scylla
We basically already support running JanusGraph with Scylla but we don't communicate that anywhere. I suggest that we add Scylla to our test suite running with GitHub actions and then add it to the list of tested projects [in our changelog](https://docs.janusgraph.org/changelog/) and in the [release notes](https://github.com/JanusGraph/janusgraph/releases). Afterwards we could add a small section to the docs explaining how to use Scylla with the CQL backend.

- [ ] Test Scylla automatically with GitHub actions
- [ ] Document tested versions in changelog and release notes
- [ ] Document how to use Scylla with the CQL backend

porunov:
- Add ScyllaDB support via `janusgraph-scylla`
- `janusgraph-cql` contains tests for Cassandra3, Cassandra4, ScyllaDB. `janusgraph-scylla` contains tests for ScyllaDB only.

ScyllaDB driver has the same classpath as DataStax driver. Thus, we can't have both `janusgraph-cql` and `janusgraph-scylla` dependencies in the classpath at the same time. 
It means that `janusgraph-all` can have only one of them. This is fine I think because anyone who is using that dependency should be able to exclude `janusgraph-cql` and include `janusgraph-scylla` when necessary.

That said, it also means that we can't have `janusgraph-scylla` in any of the JanusGraph distributions because those distributions already have `janusgraph-cql` dependency. Right now the only option to add `scylla` as a storage adapter to JanusGraph distribution is by manually replacing libraries. 
The issue to make it more easier for users to replace `cql` drivers with `scylla` drivers is here: #3580 (has a solution description).

I checked that the optimized ScyllaDB driver is used for `scylla` storage option.

Fixes #1778
Related to #2451 
Fixes to #2505 (comment about observations of Hadoop tests with ScyllaDB is [here](https://github.com/JanusGraph/janusgraph/issues/2505#issuecomment-1427215950))

-----

### For all changes:

### For code changes:


porunov:
If anyone knows why jacoco complains when `-Pcovarage` is used for ScyllaDB tests, please let me know because I couldn't figure out the problem yet.
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (default-cli) on project janusgraph-scylla: An Ant BuildException has occured: Unable to read execution data file /home/runner/work/janusgraph/janusgraph/target/jacoco-combined.exec
Error:  around Ant part ...<report>... @ 5:13 in /home/runner/work/janusgraph/janusgraph/janusgraph-scylla/target/antrun/build-main.xml
```
If I omit `-Pcovarage` parameter then it builds fine.

farodin91:
@porunov  Would it be possible to use java shadowing for instead of library coping?

porunov:
> @porunov Would it be possible to use java shadowing for instead of library coping?

I don't have experience with shading Java libraries, so can't say for sure. I have created an issue #3580 to be able to avoid copying libraries. In that issue I proposed making diff between `cql` and `scylla` dependencies, split them to multiple folders and choose the folder to include based on environment variable. 
I think shading should be another approach to solve #3580 but I don't have a clear picture of how to accomplish it with shading due to lack of experience.

docs/changelog.md
```diff
@@ -236,6 +237,18 @@ ElasticSearch 8 doesn't allow creating new indexes with `Mapping.PREFIX_TREE` ma
 using `Mapping.PREFIX_TREE` will work in ElasticSearch 8 after migration. See 
 [ElasticSearch 8 migration guide](https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.0.html#geo-shape-strategy).
 
+##### Add support for ScyllaDB driver
```
FlorianHockmann, line 240: Not that important, but I would move this section and the one above about ES 8 at least above the deprecation notices. Those deprecations should be the last part of the changelog and I would order the changelog in general by priority. Maybe move both sections right behind the breaking changes?

docs/storage-backend/scylladb.md
```diff
@@ -27,3 +27,69 @@ ScyllaDB is fully compatible with Cassandra. To use it as the data storage layer
...
+}
+/* Include `janusgraph-scylla` */
+compile "org.janusgraph:janusgraph-scylla:1.0.0"
+```
+
+#### Using Scylla driver in JanusGraph Server provided via the standard distribution builds
+
+!!! note
+    The manual process described below is temporary and should be replaced by automated process after the issue
+    [janusgraph/janusgraph#3580](https://github.com/JanusGraph/janusgraph/issues/3580) is resolved.
```
FlorianHockmann, line 80: Would be really great if we could get this resolved until the 1.0.0 release.

janusgraph-scylla/src/main/java/org/janusgraph/diskstorage/cql/ScyllaStoreManager.java
```diff
@@ -0,0 +1,33 @@
...
+import org.janusgraph.diskstorage.configuration.Configuration;
+import org.janusgraph.diskstorage.cql.builder.CQLMutateManyFunctionBuilder;
+import org.janusgraph.diskstorage.cql.builder.CQLProgrammaticConfigurationLoaderBuilder;
+import org.janusgraph.diskstorage.cql.builder.CQLSessionBuilder;
+import org.janusgraph.diskstorage.cql.builder.CQLStoreFeaturesBuilder;
+
+public class ScyllaStoreManager extends CQLStoreManager {
+
+    public ScyllaStoreManager(Configuration configuration) throws BackendException {
+        this(configuration, DEFAULT_MUTATE_MANY_FUNCTION_BUILDER, DEFAULT_STORE_FEATURES_BUILDER, DEFAULT_CQL_SESSION_BUILDER, DEFAULT_PROGRAMMATIC_CONFIGURATION_LOADER_BUILDER);
```
FlorianHockmann, line 27: Haven't tested it, but wouldn't this also already work as that has exactly the same implementation?
```suggestion
        super(configuration);
```

janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java
```diff
@@ -140,6 +155,11 @@ public JanusGraphCassandraContainer(boolean bindDefaultPort) {
         }
     }
 
+    private void applyCassandraServerConfiguration(String resourcePath){
+        withClasspathResourceMapping(resourcePath, "/etc/cassandra/cassandra.yaml", BindMode.READ_WRITE);
```
FlorianHockmann, line 159: Just for my own understanding: These two lines will mount the config file into `/etc/cassandra/cassandra.yaml` AND `/etc/scylla/scylla.yaml` for the container, irrespective of whether it's Cassandra or Scylla, so each service can use its config file in the expected location, right?
And you probably did it like this because you wanted to avoid the complexity of having to distinguish again between Scylla and Cassandra only to save one more mount that isn't necessary?

If yes, then we can of course keep it like this. I'm really just asking to better understand this. (But maybe a small comment would help here in case anyone else later wonders about this.)

FlorianHockmann (commented):
Thanks a lot for tackling this, @porunov! I also tried that a while back, but then stopped when I ran into the issue you're describing in #3580 as I had no idea how to resolve this. But your approach of first only adding this as a module and resolve the problem with selecting the right library in the distribution at runtime in a follow-up issue definitely makes sense to get us moving here :-)

I just reviewed the diff here and everything looks good basically. I only added some small nitpick comments. The logs of the added GH Actions also show that the Scylla driver is really being used there:

```
Using Scylla optimized driver!!!
```

I would like to also build the distribution and try out the approach you've documented with manually replacing the JARs if I find some time in the next few days but wanted to already provide my review comments so far.

docs/storage-backend/scylladb.md
porunov: Agree. I don't really want users to handle manual copy/paste process. I think we should just change the distribution build process to have `lib`, `cqlLib`, `scyllaLib` folders as noted in #3580 
It should be fairly simple for users to switch between CQL or Scylla drivers by adding an optional environment variable. 
@farodin91 also [mentioned](https://github.com/JanusGraph/janusgraph/pull/3578#issuecomment-1428032096) using Java Shadowing approach which could potentially be an alternative version to resolve #3580 , but I haven't had a chance to investigate the approach with Java Shadowing.

janusgraph-cql/src/test/java/org/janusgraph/JanusGraphCassandraContainer.java
porunov: Yes, initially I just decided to add the same config to both places for Cassandra and for Scylla even so one will be redundant because the `JanusGraphCassandraContainer` is for testing purposes, so it shouldn't be a problem to have redundant configs there.
That said, as you mentioned it I realized that it makes sense to simply reuse `isScylla()` method here and mount only one configuration depending on the container type. I will change this method here.

porunov:
Thanks @FlorianHockmann for the review! I will experiment tomorrow or on Wednesday with ScyllaDB parameters suggested here https://github.com/JanusGraph/janusgraph/issues/2505#issuecomment-1428379299 and update the PR with your suggestions and probably modify the JanusGraphCassandraContainer a bit to use those parameters.

I don't personally like the manual approach of replacing libs and also think we should handle #3580 before we release 1.0.0, but thought it makes sense to ship Scylla module first.

docs/changelog.md
porunov: Moving both sections above the deprecation notices, but we probably need to reorder that list before the release.

porunov:
@FlorianHockmann I fixed the comments you left. I also fixed Hadoop tests. 
That said, I noticed that we didn't apply configuration for ScyllaDB. After fixing that I see that `ssl` tests are failing. The reason is described here: https://github.com/JanusGraph/janusgraph/issues/2505#issuecomment-1429165384 
Couldn't figure out the cause why ScyllaDB is failing to start yet :/

porunov:
I'm stuck with the issue about ssl certificates ([here](https://github.com/JanusGraph/janusgraph/issues/2505#issuecomment-1431330743)). I'm not sure I'll be able to resolve that issue. Thus, I can't just comment out tests related to ScyllaDB ssl certificates and open a new issue for the tests. Let me know what you think about it.

porunov:
Commented the failing tests. The PR is ready for re-review

FlorianHockmann (approved):

PR #3581, opened 2023-02-13
Add log processor for global db-cache invalidation

### LINKED ISSUE #3155: Global db-cache invalidation on mutation

porunov:
TODO:
- Fix `buildCacheInvalidationChangeProcessor` to be able to invalidate updated vertices which are already removed by the time processor is started
- Add parameter to JanusGraph which will allow automatically starting this log processor
- Cover added logic with tests
- Document global db-cache invalidation logic (like: lag time between transaction is committed and cache entry is evicted, log configuration for db-cache invalidation log processor, transactions usage for it to work)

Fixes #3155

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/util/LogProcessorCacheInvalidationUtil.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/util/LogProcessorCacheInvalidationUtil.java
@@ -1,0 +1,164 @@
+// Copyright 2023 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.database.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.janusgraph.core.JanusGraphEdge;
+import org.janusgraph.core.JanusGraphElement;
+import org.janusgraph.core.JanusGraphFactory;
+import org.janusgraph.core.JanusGraphRelation;
+import org.janusgraph.core.JanusGraphVertex;
+import org.janusgraph.core.JanusGraphVertexProperty;
+import org.janusgraph.core.log.Change;
+import org.janusgraph.core.log.ChangeProcessor;
+import org.janusgraph.core.log.LogProcessorBuilder;
+import org.janusgraph.core.log.LogProcessorFramework;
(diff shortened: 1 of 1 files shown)


jekramer:
Hi @porunov  

I've been referencing this and have some more questions. For cache invalidation purposes, should it matter whether an IndexUpdate is new or removed? I've noticed that `getIndexUpdatesNoConstraints` for the properties pulled from `changeState.getRelations` [throws an AssertionError](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java#L239), because their ElementLifeCycle is "Loaded" instead of New or Removed. 

I can avoid this by creating a new StandardVertexProperty with one of the accepted statuses, like you do with `reverseProperty` in this PR. From what I can tell, we only need the key from IndexUpdate for invalidation, and that is the same no matter what lifecycle is. Is there any reason I shouldn't do that? I wanted to check because you add both the property and reverseProperty to the internalRelations list here, and I didn't understand why.

porunov:
> Hi @porunov  
> 
> I've been referencing this and have some more questions. For cache invalidation purposes, should it matter whether an IndexUpdate is new or removed? I've noticed that `getIndexUpdatesNoConstraints` for the properties pulled from `changeState.getRelations` [throws an AssertionError](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java#L239), because their ElementLifeCycle is "Loaded" instead of New or Removed. 
> 
> I can avoid this by creating a new StandardVertexProperty with one of the accepted statuses, like you do with `reverseProperty` in this PR. From what I can tell, we only need the key from IndexUpdate for invalidation, and that is the same no matter what lifecycle is. Is there any reason I shouldn't do that? I wanted to check because you add both the property and reverseProperty to the internalRelations list here, and I didn't understand why.

I actually didn't check that myself and added that code quickly. Initially I thought the lifecycle may matter, but now I think it doesn't actually matter for cache invalidation purposes. Thus, that code which I shared should probably be changed without reverseProperty, but I didn't investigate that much yet.

jekramer:
>Initially I thought the lifecycle may matter, but now I think it doesn't actually matter for cache invalidation purposes. Thus, that code which I shared should probably be changed without reverseProperty, but I didn't investigate that much yet.

ah okay, that addresses my concerns then, thank you! Actually reverseProperty is the one that works, it's the initial property that throws an error. So I'm glad you included it!

PR #3687, opened 2023-03-31
Profiler: Display multiKSQ detail

li-boxuan:
Currently, multiKSQ (multi-key slice query) does not show concrete queries fired in profiling results. This sometimes makes debugging difficult.

This work is inspired by a user question on the Discord channel.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/MultiKeySliceQuery.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/MultiKeySliceQuery.java
@@ -75,6 +75,14 @@ public class MultiKeySliceQuery extends BaseQuery implements BackendQuery<MultiK
         StringBuilder builder = new StringBuilder("multiKSQ[");
         builder.append(queries.size()).append("]");
         if (hasLimit()) builder.append("@").append(getLimit());
+        builder.append("{");
+        for (int i = 0; i < queries.size(); i++) {
+            if (i > 0) {
+                builder.append(",");
+            }
+            builder.append(queries.get(i));
+        }
+        builder.append("}");
         return builder.toString();
     }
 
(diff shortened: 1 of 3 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/query/graph/MultiKeySliceQuery.java
```diff
@@ -75,6 +75,14 @@ public String toString() {
         StringBuilder builder = new StringBuilder("multiKSQ[");
         builder.append(queries.size()).append("]");
         if (hasLimit()) builder.append("@").append(getLimit());
+        builder.append("{");
+        for (int i = 0; i < queries.size(); i++) {
+            if (i > 0) {
+                builder.append(",");
+            }
+            builder.append(queries.get(i));
+        }
```
cdegroc, line 84: **❓ question:** Won't this result in very long strings when there are many queries?
If so,
- should we truncate the final string to make the output easier to read?
- or should the queries be added as "nested" elements in the profile?‏

li-boxuan: Good question! My feelings are:
1) People usually don't profile unless they have performance problems. So, how bad could a profiler be? We can easily make progressive efforts if people say they need it to be shorter. Truncation sounds reasonable. But then we should consider: should we and how could we make this truncation optional and configurable? [With Step](https://tinkerpop.apache.org/docs/current/reference/#with-step) might be a good place to apply that ad-hoc config.
2) I personally cannot think of a case where this results in very long strings. An example would be `g.V().has("index", P.within("a", "b", .... ))`. I guess that won't happen a lot. I don't like IN queries myself, and if people have to do an "IN" with a long list, they know what they are doing and they need to more or less just live with it.
3) I agree and I think the best solution is to make the queries as nested elements. I just couldn't find out a way to do this in a reasonably short time.

Assuming we can only do one thing for now: either display all OR always do truncation. Which do you think is better? If we do truncation, what should be the fixed (hard-coded) limit?

cdegroc: You're making a good point in 1. that users call `profile()` to debug queries. So displaying the whole string is probably better than truncating. Also agree that this could be improved over time in any case 👍

cdegroc (approved):

PR #3743, opened 2023-04-26
Remove unused JanusGraphMultiQueryStep when possible [tp-tests][cql-tests]

### LINKED ISSUE #3737: Do not inject JanusGraphMultiQueryStep before Parent step which has not any child step benefiting from this optimization
`JanusGraphMultiQueryStep` will be always inserted before the next steps right now: `BranchStep.class, OptionalStep.class, RepeatStep.class, TraversalFilterStep.class, ConnectiveStep.class` (these steps can be found in `JanusGraphTraversalUtil`).

That said, it doesn't mean that those steps can actually benefit from the `JanusGraphMultiQueryStep`. In case none of their children benefit from the pre-cached vertices this step not only doesn't give any benefits for such parents but also may cause harm in some situations.

Probably a bad example, but let's imagine a user wants to get any vertex from the graph but if the returned vertex has any adjacent vertex then better to return the adjacent vertex instead. So, we used the next query (for some reason):
`graph.traversal().V().optional(__.both().limit(1)).limit(1).profile().next()`

In such situation `JanusGraphMultiQueryStep` will be injected before `OptionalStep`. This leads to pre-caching of vertices before passing them next. I.e. this step receives a vertex for future process, potentially pre-register this vertex with the future `MultiQueriable` steps and then passes this vertex next (to the `barrier` step). 
This basically leads fetching **all** vertices into the barrier step before passing a single vertex into the `optional` step. This is because the default injected barrier step has no limit (`Integer.MAX_VALUE` means no limit in `NoOpBarrierStep`).
Of course, we could re-write the above query to have `limit(1)` or  `barrier(1)` steps at the beginning, but this wouldn't show the real issue then. I wouldn't be surprised if users simple get OOM exception for such query.
I think the more parent steps we add as supported `MultiQueriable` parents (like `ConnectiveStep` step here #3736) the bigger issue it becomes. Thus, I would suggest dealing with this issue sooner than later.

Some of my thoughts which could potentially fix the issue (didn't verify those hypothesis yet):
- Add a separate optimization strategy which removes any `JanusGraphMultiQueryStep` without any clients attached to this step. If `JanusGraphMultiQueryStep` is getting removed, then automatically created barrier should be removed as well.
- When we check the that the specific step may benefit from inserting `JanusGraphMultiQueryStep` before it, we could check if any of the step's children traversals have any starting steps which may benefit from `JanusGraphMultiQueryStep`. If there are non - we shouldn't insert this step. If there are any - we should insert it. We could check if any nested steps have MultiQuery compatible steps with DFS or BFS.

porunov:
Fixes #3737

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
@@ -65,6 +65,7 @@ import org.janusgraph.graphdb.database.cache.KCVSCacheInvalidationService;
 import org.janusgraph.graphdb.database.cache.SchemaCache;
 import org.janusgraph.graphdb.database.idassigner.VertexIDAssigner;
 import org.janusgraph.graphdb.database.idhandling.IDHandler;
+import org.janusgraph.graphdb.tinkerpop.optimize.strategy.JanusGraphUnusedMultiQueryRemovalStrategy;
 import org.janusgraph.util.IDUtils;
 import org.janusgraph.graphdb.database.index.IndexInfoRetriever;
 import org.janusgraph.graphdb.database.index.IndexUpdate;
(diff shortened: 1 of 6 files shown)


porunov:
For some reason TP tests were not executed in this PR even so I added `[tp-tests]` into the commit message.
Nevertheless `master` branch have a bug executing TP tests, thus they won't pass here anyways. See: #3747

porunov:
I have intention of merging this PR via lazy consensus. The new behavior can be disabled if `JanusGraphUnusedMultiQueryRemovalStrategy` is excluded from the traversal.

PR #3775, opened 2023-05-12
Use default limited batch size even when LazyBarrierStrategy not applies

### LINKED ISSUE #3773: Ensure automatically created barrier steps are always limited
JanusGraph [documentation](https://docs.janusgraph.org/operations/batch-processing/#limited-batch-processing) says:
> Although batch size does not necessarily need to be configured, it can provide an additional tuning parameter to improve the performance of a query. By default, the batch size of [TinkerPop's barrier step](https://tinkerpop.apache.org/docs/current/reference/#barrier-step) will be used, which is currently at 2500. The batch size of each vertex step can be individually configured by prepending a barrier() step. For example, in the query below, the first out() step would use the default batch size of 2500 and the second out() step would use a manually configured batch size of 1234:

This isn't fully true right because TinkerPop may insert some barrier steps using `LazyBarrierStrategy` with default size of `2500` in some places, but not to all places. After that in `JanusGraphMultiQueryStrategy` we may insert more `.barrier` steps with unlimited size to places where neither TinkerPop not user included `.barrier` step. Thus, it will mean that for such requests there will be unlimited batches. 

- Version: master
- Link to discussed bug: https://github.com/JanusGraph/janusgraph/issues/3733#issuecomment-1545215620
- Expected Behavior: created `barrier` steps are limited to the default `2500`
- Current Behavior: some of the `barrier` steps are unlimited 
- Steps to Reproduce:

`graph.traversal().V(a).emit().repeat(__.in("knows")).count().profile().next()` - this will insert unlimited `.barrier` step before `.in` step and before `.repeat` step.

porunov:
Fixes #3773

-----

### For all changes:

### For code changes:


rngcntr (approved):
Looks very good, thank you!

docs/operations/batch-processing.md
```diff
@@ -90,8 +90,11 @@ This is the default configuration of JanusGraph since version 1.0.0.
 Although batch size does not necessarily need to be configured,
 it can provide an additional tuning parameter to improve the
 performance of a query.
-By default, the batch size of [TinkerPop's barrier step](https://tinkerpop.apache.org/docs/current/reference/#barrier-step)
-will be used, which is currently at 2500.
+By default, the batch size for [TinkerPop's barrier step](https://tinkerpop.apache.org/docs/current/reference/#barrier-step)
+will be provided by `LazyBarrierStrategy`, which is currently at `2500`. 
+For batchable cases where `LazyBarrierStrategy` doesn't inject any `barrier` steps,
+the barrier step will be ingected with the size configured via `query.limited-batch-size` 
+(which is defaults to `2500`, same as with `LazyBarrierStrategy`).
```
FlorianHockmann, line 97:
```suggestion
(which defaults to `2500`, same as with `LazyBarrierStrategy`).  
```

FlorianHockmann (approved):

PR #3853, opened 2023-06-23
Remove 2 unused dependencies

FlorianHockmann:
These dependencies have been added 10 years ago and I can't find any place where they are still used:
https://github.com/JanusGraph/janusgraph/commit/94e75d3fa968c0d2cf890c1883ccbb2f21a391e3

PR #3867, opened 2023-06-27
[v0.6] Remove 2 unused dependencies

FlorianHockmann:
# Backport

This will backport the following commits from `master` to `v0.6`:
 - [Remove 2 unused dependencies](https://github.com/JanusGraph/janusgraph/pull/3853)

### Questions ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1093,16 +1093,6 @@
                 <version>${gson.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.apache.ant</groupId>
-                <artifactId>ant</artifactId>
-                <version>1.10.13</version>
-            </dependency>
-            <dependency>
-                <groupId>javax.servlet</groupId>
-                <artifactId>servlet-api</artifactId>
-                <version>2.5</version>
-            </dependency>
-            <dependency>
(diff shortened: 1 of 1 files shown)


FlorianHockmann:
This is failing for `v0.6` with a dependency convergence error:

```
Dependency convergence error for org.apache.ant:ant:1.7.0 paths to dependency are:
+-org.janusgraph:janusgraph-solr:0.6.4-SNAPSHOT
  +-org.janusgraph:janusgraph-cql:0.6.4-SNAPSHOT
    +-org.apache.cassandra:cassandra-all:3.11.10
      +-org.apache.cassandra:cassandra-thrift:3.11.10
        +-de.jflex:jflex:1.6.0
          +-org.apache.ant:ant:1.7.0
and
+-org.janusgraph:janusgraph-solr:0.6.4-SNAPSHOT
  +-org.apache.solr:solr-test-framework:8.11.0
    +-org.apache.ant:ant:1.8.2
```

Looks like we at least need the `ant` dependency here. I'll just close the PR and not backport this change then as it's not really important and apart from that, it might also be a breaking change if users rely on getting these dependencies from us.
So, I don't see a good reason in general why we should backport this.

PR #3879, opened 2023-07-05
Group CQL keys into a single query [cql-tests] [tp-tests]

### LINKED ISSUE #3863: Group keys by token ranges for CQL storage backend and execute a single SliceQuery for them via IN operator
Right now CQL storage backend executes a separate CQL query per each partition key (vertex id) when executing `SliceQuery` queries.
This comes with additional network overhead, channels clogging, as well as potential overhead on the Cassandra storage side. 
Usually, usage of `IN` operator isn't a good choice due to potential touch of multiple nodes ([blog link](https://lostechies.com/ryansvihla/2014/09/22/cassandra-query-patterns-not-using-the-in-query-for-multiple-partitions/), [doc link](https://docs.datastax.com/en/dse/6.8/cql/cql/cql_reference/cql_commands/cqlSelect.html#cqlSelect__selectInNot)).
However, I suspect that touching multiple partition withing the same token range (i.e. same data shard) may have positive influence on performance due to less network overhead (**needs to be verified**).

As such, if we group keys by token ranges then it makes sense to execute all (or some amount) of grouped keys together.
For example, let's imagine we have 2 token ranges and we want to execute a query for 10 keys (4 from the first token range, and 6 from the second token range).
As such, after grouping we will have 2 CQL queries (instead of 10 CQL queries as it is right now). The first query will use keys 1-4, and the second query will use keys 5-10. 

However, if we execute a query touching multiple partition keys we will need to return a requested `key` back with a query (i.e. +8 bytes of data). 
I believe the overhead is quite small and it worth returning keys with such queries if grouped queries improve performance.

## How to implement it?

WIP PR can be found here: #3879

An example of grouping keys by same partition could be found in `CQLSliceQueryUtil` in this unfinished commit:
https://github.com/porunov/janusgraph/commit/f8efa3fe74adb6129e27186dab11e3fbc4d184a7

porunov:
Fixes #3863

#### Local single instance Cassandra node

CQL benchmarks for local single instance Cassandra node is located here: https://gist.github.com/porunov/b0804e789ad6af56625230785ae8698b

There 2 tests for `master` branch (with enabled and disabled `slice` query grouping) as well as 4 tests with different configurations of this PR.

The observations from `local` testing is that generally performance increases 10-20% for cases when there are only `keys` grouping used (i.e. multi-key cases but not multi-slice cases). 
I noticed that when there are few keys to be grouped (less than 500) then performance decreases (up to 20% performance decrease). I imagine it's due to less parallelism on deserialization side (1 thread instead of `CPUs * 2` threads) as well as increased amount of data to be fetched (when grouping keys we need to return `key` beck per each column fetched).
Another observation is that when both `slice` and `keys` grouping is triggered for a specific query the performance also drops about 5%. It could be that it's harder for Cassandra to process a query with two `IN` operators, but it's hard to verify the real reason.

Overall, from the above tests it's clearly seen that:
- When `keys` grouping is disabled - the performance is the same as in `master` branch (i.e. no regression and no improvements).
- When `keys` grouping is enabled most of the queries with more than 700 keys gain additional performance improvements (mostly 10-20%, but goes up to 40% for some cases). 
- There are some cases when performance drops with enabled `keys` grouping.

#### Distributed cluster

At home I started a distributed ScyllaDB cluster on 3 different machines and started JanusGraph benchmarking on the 4th machine via:
```
# Machine 1:
docker run --name scyllaTest --network host --volume /var/lib/scylla:/var/lib/scylla -d scylladb/scylla:5.2.4 --developer-mode=0 --smp 4 --memory 28G --api-address 0.0.0.0 --listen-address 192.168.68.100
# Machine 2:
docker run --name scyllaTest --network host --volume /var/lib/scylla:/var/lib/scylla -d scylladb/scylla:5.2.4 --developer-mode=0 --smp 4 --memory 28G --api-address 0.0.0.0 --seeds=192.168.68.100,192.168.68.65,192.168.68.54 --listen-address 192.168.68.65
# Machine 3
docker run --name scyllaTest --network host --volume /var/lib/scylla:/var/lib/scylla -d scylladb/scylla:5.2.4 --developer-mode=0 --smp 4 --memory 28G --api-address 0.0.0.0 --seeds=192.168.68.100,192.168.68.65,192.168.68.54 --listen-address 192.168.68.54
```

Unfortunately, processors, disks, and RAM are different and have different performance characteristics. Moreover this was a home WIFI network up to 1800 Mbps with possible network congestion from neighbor devices. Thus, this testing should be taken with a grain of salt because it could be influenced by a number of factors. 

I couldn't finish all the tests yet, but quick first tests of `getAdjacentVerticesLocalCounts` show that when `keys` grouping is enabled the performance actually drops 5-8%. I did run this test multiple times and the results are consistent (5%-8% percent performance decrease). 
It seems that the amount of queries in the distributed environment becomes less important that the amount of data fetched. Seems the additional `key` fetching per each `column` influence performance more than I expected.
It's hard to verify if the performance decrease is really related to the fact that additional data is being fetched or the way `IN` operator is processed on the storage node side.
After a meeting with AstraDB developers they ensured me that this way of grouping partition keys together is legit and they usually don't recommend using `IN` operator because users usually don't group partition keys by same token ranges. However, in our case, it should be OK. Nevertheless, they warned that the checksum verification by coordinator node could still potentially be executed by multiple queries which could potentially affect performance, but should be verified case by case.
Another grey area for now is internal parallelism on the `IN` query when executed by a node. It isn't obvious weather Cassandra / ScyllaDB utilizes multiple cores for processing a single `IN` query or not. In case only a single CPU is utilized for `IN` query but multiple CPUs are utilized to process separate queries then the internal parallelism of separate queries could potentially be higher (_in non overloaded clusters_).

## Conclusion

Performance improvement of the `keys` grouping feature is questionable. It might improve performance in some cases and decrease performance in other cases.
However, the obvious benefit is that this gives users a choice of enabling keys grouping for Serverless cases when users pay per read for each CQL query. In some cases performance might be not the main factor and the price reduction could be more needed.

I propose to have this option be disabled by default and users should enable it only for cases when they are sure they will gain benefits from grouping CQL queries.    

-----

### For all changes:

### For code changes:


docs/configs/janusgraph-cfg.md
```diff
@@ -482,6 +478,32 @@ Configuration options for CQL executor service which is used to process deserial
...
+If this option if `false` then each Slice query will be executed in a separate asynchronous CQL query even when grouping is allowed. | Boolean | true | MASKABLE |
+| storage.cql.grouping.slice-limit | Maximum amount of grouped together slice queries into a single CQL query.
+Notice, for ScyllaDB this option should not exceed the maximum number of distinct clustering key restrictions per query which can be changed by ScyllaDB configuration option `max-partition-key-restrictions-per-query` (https://enterprise.docs.scylladb.com/branch-2022.2/faq.html#how-can-i-change-the-maximum-number-of-in-restrictions). For AstraDB this limit is set to 20 and should be asked to be changed by AstraDB side via `partition_keys_in_select_failure_threshold` and `in_select_cartesian_product_failure_threshold` threshold configurations (https://docs.datastax.com/en/astra-serverless/docs/plan/planning.html#_cassandra_yaml).
```
li-boxuan, line 503: Users might be confused about `should be asked to be changed by AstraDB side`. My first impression was one can ask AstraDB customer support to change this, but it seems the value is a fixed number and cannot be configured.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/service/GroupingAsyncQueryExecutionService.java
```diff
@@ -0,0 +1,484 @@
...
+                                              String keyspaceName,
+                                              String edgeStoreTableName,
+                                              Function<Select, Select> addTTLFunction,
+                                              Function<Select, Select> addTimestampFunction,
+                                              CQLColValGetter singleKeyGetter,
+                                              CQLColValGetter multiKeysGetter,
+                                              ExecutorService executorService,
+                                              QueryBackPressure queryBackPressure) {
+        this.session = session;
+        sliceGroupingLimit = configuration.get(SLICE_GROUPING_LIMIT) < 1 ? 1 : configuration.get(SLICE_GROUPING_LIMIT);
```
li-boxuan, line 106:
```suggestion
        sliceGroupingLimit = Math.min(configuration.get(SLICE_GROUPING_LIMIT), 1);
```

janusgraph-cql/src/test/java/org/janusgraph/graphdb/cql/CQLGraphTest.java
```diff
@@ -252,6 +359,42 @@ public void testLimitBatchSizeForMultiQueryMultiCardinalityProperties(boolean sl
...
+            cs[i].property(VertexProperty.Cardinality.list, "listProperty", "listValue3");
+            cs[i].property(VertexProperty.Cardinality.list, "listProperty", "listValue4");
+        }
+
+        newTx();
+
+        clopen(configuration);
+
+        // test batching for `values()`
+        TraversalMetrics profile = testLimitedBatch(() -> graph.traversal().V(cs).barrier(elementsAmount).values("singleProperty1", "singleProperty2", "singleProperty3", "setProperty", "listProperty"), configuration);
```
li-boxuan, line 394: I have two questions:

1. `testLimitedBatch` only compares results with and without batching. In the context of this PR, we would want to compare the results with and without key/column grouping.
2. Can one have one query that requests different properties (columns) from different vertices (keys)? For example, `g.V('v1').values('prop1').as('v1Prop1').V('v2').values('prop2').as('v2Prop2').select('v1Prop1', 'v2Prop2')`. I am not sure if a query like this will trigger your new logic at all, but I would like to make sure the grouping doesn't assume we request the same columns for different keys.

li-boxuan:
We should test this feature against as many cloud providers as possible. I can test it with Amazon Keyspace.

janusgraph-cql/src/test/java/org/janusgraph/graphdb/cql/CQLGraphTest.java
porunov: 1. That makes sense. I will change to use a new method which compares results also with and without keys grouping.
2. This logic is applied to a single query like ‘values(“prop1”, “prop2”)’ only because that’s how multi-query is working. However, we may have some data already cached for some vertices in transaction level cache of db-level cache which will modify what we are actually querying from the storage backend. Nevertheless, those will be different query groups. The grouping logic is aware that there might be different columns asked for different set of keys (thats what query group means there). Grouping logic won’t ever fetch columns which were not explicitly requested for specific keys.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/service/GroupingAsyncQueryExecutionService.java
porunov: I think it should be ‘max’ instead of ‘min’ here, but I agree that it looks better like you suggested. I will change it.

li-boxuan: Oh yeah it should be max

janusgraph-cql/src/test/java/org/janusgraph/graphdb/cql/CQLGraphTest.java
li-boxuan: > Nevertheless, those will be different query groups. The grouping logic is aware that there might be different columns asked for different set of keys (thats what query group means there)

Cool 👍

docs/configs/janusgraph-cfg.md
porunov: The value is generally fixed, but you may ask AstraDB customer support and they may modify some threshold restrictions for your own use case. 
That’s of course case by case, and it’s not guaranteed they will increase those numbers for your specific case. 
However, it’s possible. 
I will try to make it more clear in this message that it’s generally fixed, but there is a chance they will increase those limits per your request.

porunov:
> We should test this feature against as many cloud providers as possible. I can test it with Amazon Keyspace.

Thank you! That will be very helpful!
On my side I will test it with AstraDB side. 
I had a meeting with AstraDB developers and they said that generally they don’t recommend using ‘IN’ query because users often abuse its usage by providing partition keys of different nodes inside ‘IN’ query which results in poor performance. 
However, in this case, as we are grouping keys ourselves by the same token ranges, it should be good enough to reduce pricing and don’t affect performance too much. 

The one thing to notice, is queries with small number of keys. Generally you queries could be routed to different replicas. If you request 5 keys (for example) and assuming each separate CQL query takes 1 ms, then grouped query could potentially take longer than 1 ms because it is executed on the single node. I’m not sure how IN queries is executed inside that node (i.e. if it uses more than one CPU or not for evaluation), but this also could affect performance. Nevertheless, with queries where same token range keys amount be more than ‘CPUs over all replicas of that token range * grouping limit size’ - performance should not suffer theoretically.
I would say it’s a balance which users should find and decide which ‘keys-min’ and ‘keys-limit’ to use for their case. 

Now I also realised that for ‘janusgraph-scylla’ driver, token range only grouping won’t be enough because ScyllaDB driver is also shard aware. We should provide different grouping strategy for ScyllaDB and for token-aware only implementations (like Cassandra, AstraDB, etc.). Thus, I will update the code slightly to be able to select different grouping strategies (i.e. Token-Aware strategy or Token-And-Shard-Aware strategy or user provided custom strategy). I.e. when ScyllaDB driver is used we should group keys by token range + shard id.
This change won’t affect Cassandra / AstraDB / Amazon Keyspaces performance testing or grouping approach.

docs/configs/janusgraph-cfg.md
porunov: Changed the description, so that it would be clear for AstraDB side.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/service/GroupingAsyncQueryExecutionService.java
porunov: Applied `Math.max`

janusgraph-cql/src/test/java/org/janusgraph/graphdb/cql/CQLGraphTest.java
porunov: Updated this test to test each configuration with enabled / disabled grouping.

porunov:
@li-boxuan I added a new configuration option `storage.cql.grouping.keys-class` to be able to provide custom grouping strategies. There is only one default strategy right now which groups keys by token ranges only. This strategy won't be efficient for ScyllaDB (unless nodes with `1` CPU only are used) because this strategy ignores Shards concept which ScyllaDB have. I mentioned it in the documentation. When https://github.com/scylladb/java-driver/issues/232 is resolved then we will be able to add a more optimized grouping for `janusgraph-scylla` storage implementation.

porunov:
> The feature `storage.cql.grouping.keys-allowed` does not work for Amazon Keyspace:
> 
> ```
> PER PARTITION LIMIT is not supported
> Type ':help' or ':h' for help.
> Display stack trace? [yN]y
> com.datastax.oss.driver.api.core.servererrors.InvalidQueryException: PER PARTITION LIMIT is not supported
> 	at com.datastax.oss.driver.api.core.servererrors.InvalidQueryException.copy(InvalidQueryException.java:48)
> 	at com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures.getUninterruptibly(CompletableFutures.java:149)
> 	at com.datastax.oss.driver.internal.core.cql.CqlPrepareSyncProcessor.process(CqlPrepareSyncProcessor.java:59)
> 	at com.datastax.oss.driver.internal.core.cql.CqlPrepareSyncProcessor.process(CqlPrepareSyncProcessor.java:31)
> 	at com.datastax.oss.driver.internal.core.session.DefaultSession.execute(DefaultSession.java:230)
> 	at com.datastax.oss.driver.api.core.cql.SyncCqlSession.prepare(SyncCqlSession.java:206)
> 	at org.janusgraph.diskstorage.cql.service.GroupingAsyncQueryExecutionService.<init>(GroupingAsyncQueryExecutionService.java:150)
> ```
> 
> but as long as this option is turned off, everything seems to be working as usual, so I am happy with this PR. It would be even better if you could mention that this feature does not work for Amazon Keyspace.

Thank you @li-boxuan for checking it with Amazon Keyspace. I will surely reference that this feature doesn't work with Amazon Keyspaces.
I tested it with AstraDB (with keys group limit = 20). 
When grouping keys by up to 20 keys via `IN` operator the individual query becomes up to 5 times slower. However, the query returns data for 20 keys together. The thing to notice, is that the amount of data returned for our use-cases increases almost 80% (I thought it will be smaller for some reason). This is due to the fact that we now return partition-key with each row. The amount of data sent decreases 29% (which is good), however, as the amount of data received is 13 times bigger than the amount of data sent - thus, it's not that noticeable.
For this statistics you might have impression that the queries are now 5 times slower and return more redundant data. However it's actually opposite. As our use-cases usually involve querying data for many vertices, we see that the overall performance of our heavy queries increased 22%! Even so, the performance improvement isn't the main goal here, but it's a nice bonus. The main reason of grouping partition keys together is cost optimization for Serverless deployments like AstraDB. Even so we do return 80% more data from AstraDB now, it actually costs much less due to the fact that we simply have up to 20 times less CQL queries for some of the use-cases. 
Overall this feature brings positive influence for some of the use-cases , but I think this feature isn't for everyone and it really depends from use-case to use-case if this feature makes sense or not.
I think for less loaded use-cases users would prefer to disable this feature to achieve better query parallelism with potentially better performance, but for loaded use-cases where users struggle with exhausted connection channels or reaching the back-pressure limit this feature might help.

Screenshots below show the moment of enabling  grouping feature.
 
Screenshot from 2023-07-21 19-00-18

Screenshot from 2023-07-21 19-01-18

porunov:
@li-boxuan I added information about Amazon Keyspace as well as added one more grouping strategy which groups by same replica sets. The difference between grouping partition keys by token ranges and replica sets doesn't have a big difference, but knowing that a single node may have more than 1 token range grouping by same nodes may be slightly better (i.e. less groups will be created). However grouping by same token ranges means that those tokens leave close to each other on a disk (potentially less disk seeks). Overall the difference between these grouping strategy is small, but one grouping strategy might be preferred to another in some situations.
I still want to add Shard-aware grouping strategy (i.e. 3rd grouping strategy) when ScyllaDB opens this information in their driver (https://github.com/scylladb/java-driver/issues/232), but it will probably be in the follow up PRs.

docs/configs/janusgraph-cfg.md
```diff
@@ -482,6 +478,39 @@ Configuration options for CQL executor service which is used to process deserial
...
+| storage.cql.grouping.keys-limit | Maximum amount of the keys which belong to the same token range to be grouped together into a single CQL query. If more keys for the same token range are queried, they are going to be grouped into separate CQL queries.
+Notice, for ScyllaDB this option should not exceed the maximum number of distinct clustering key restrictions per query which can be changed by ScyllaDB configuration option `max-partition-key-restrictions-per-query` (https://enterprise.docs.scylladb.com/branch-2022.2/faq.html#how-can-i-change-the-maximum-number-of-in-restrictions). For AstraDB this limit is set to 20 and usually it's fixed. However, you can ask customer support for a possibility to change the default threshold to your desired configuration via `partition_keys_in_select_failure_threshold` and `in_select_cartesian_product_failure_threshold` threshold configurations (https://docs.datastax.com/en/astra-serverless/docs/plan/planning.html#_cassandra_yaml).
+Ensure that your storage backends allows more IN selectors than the one set via this configuration.
```
li-boxuan, line 502:
```suggestion
Ensure that your storage backend allows more IN selectors than the one set via this configuration.
```

docs/configs/janusgraph-cfg.md
```diff
@@ -482,6 +478,39 @@ Configuration options for CQL executor service which is used to process deserial
...
+Usually this configuration should always be set to `2`. It is useful to increase the value only in cases when queries with more keys should not be grouped, but be performed separately to increase parallelism in expense of the network overhead.
+This option takes effect only when `storage.cql.grouping.keys-allowed` is `true`. | Integer | 2 | MASKABLE |
+| storage.cql.grouping.slice-allowed | If `true` this allows multiple Slice queries which are allowed to be performed as non-range queries (i.e. direct equality operation) to be grouped together into a single CQL query via `IN` operator. Notice, currently only operations to fetch properties with Cardinality.SINGLE are allowed to be performed as non-range queries (edges fetching or properties with Cardinality SET or LIST won't be grouped together).
+If this option if `false` then each Slice query will be executed in a separate asynchronous CQL query even when grouping is allowed. | Boolean | true | MASKABLE |
```
li-boxuan, line 508:
```suggestion
If this option is `false` then each Slice query will be executed in a separate asynchronous CQL query even when grouping is allowed. | Boolean | true | MASKABLE |
```

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/function/slice/AsyncCQLMultiKeyMultiColumnFunction.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    BoundStatementBuilder bindMarkers(MultiKeysMultiColumnQuery query, BoundStatementBuilder statementBuilder) {
+        return statementBuilder
+            .setList(CQLKeyColumnValueStore.KEY_BINDING, query.getKeys(), ByteBuffer.class)
+            .setList(CQLKeyColumnValueStore.COLUMN_BINDING, query.getColumns(), ByteBuffer.class)
+            .setInt(CQLKeyColumnValueStore.LIMIT_BINDING, query.getLimit())
+            .setRoutingToken(query.getRoutingToken())
+            // usually routing key isn't needed when routingToken is specified,
+            // but in some cases different driver implementations (like ScyllaDB)
+            // or load balancing strategies may relay on routing key instead.
```
li-boxuan, line 42:
```suggestion
            // or load balancing strategies may rely on routing key instead.
```

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/function/slice/AsyncCQLMultiKeySliceFunction.java
```diff
@@ -0,0 +1,48 @@
...
+    BoundStatementBuilder bindMarkers(MultiKeysSingleSliceQuery query, BoundStatementBuilder statementBuilder) {
+        return statementBuilder
+            .setList(CQLKeyColumnValueStore.KEY_BINDING, query.getKeys(), ByteBuffer.class)
+            .setByteBuffer(CQLKeyColumnValueStore.SLICE_START_BINDING, query.getQuery().getSliceStart().asByteBuffer())
+            .setByteBuffer(CQLKeyColumnValueStore.SLICE_END_BINDING, query.getQuery().getSliceEnd().asByteBuffer())
+            .setInt(CQLKeyColumnValueStore.LIMIT_BINDING, query.getLimit())
+            .setRoutingToken(query.getRoutingToken())
+            // usually routing key isn't needed when routingToken is specified,
+            // but in some cases different driver implementations (like ScyllaDB)
+            // or load balancing strategies may relay on routing key instead.
```
li-boxuan, line 43:
```suggestion
            // or load balancing strategies may rely on routing key instead.
```

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/service/GroupingAsyncQueryExecutionService.java
```diff
@@ -0,0 +1,433 @@
...
+                                              final CQLStoreManager storeManager,
+                                              String tableName,
+                                              Function<Select, Select> addTTLFunction,
+                                              Function<Select, Select> addTimestampFunction,
+                                              CQLColValGetter singleKeyGetter,
+                                              CQLColValGetter multiKeysGetter) {
+
+        sliceGroupingLimit = Math.max(1, configuration.get(SLICE_GROUPING_LIMIT));
+        keysGroupingLimit = Math.max(1, configuration.get(KEYS_GROUPING_LIMIT));
+        keysGroupingMin = Math.max(2, configuration.get(KEYS_GROUPING_MIN));
```
li-boxuan, line 103: (nit) shall we throw exceptions or add a warning log when the user-specified number is less than 1 or 2?

li-boxuan (commented):
I did another pass on the comments and most of them are just nitpicks.

docs/configs/janusgraph-cfg.md
```diff
@@ -482,6 +478,39 @@ Configuration options for CQL executor service which is used to process deserial
...
+If this option is `false` then each partition key will be executed in a separate asynchronous CQL query even when multiple keys from the same token range are queried.
+Notice, the default grouping strategy is not taking into account shards. Thus, this might be inefficient with ScyllaDB storage backend. ScyllaDB specific keys grouping strategy should be implemented after the resolution of the [ticket #232](https://github.com/scylladb/java-driver/issues/232). | Boolean | false | MASKABLE |
+| storage.cql.grouping.keys-class | Full class path of the keys grouping execution strategy. The class should implement `org.janusgraph.diskstorage.cql.strategy.GroupedExecutionStrategy` interface and have a public constructor with two arguments `org.janusgraph.diskstorage.configuration.Configuration` and `org.janusgraph.diskstorage.cql.CQLStoreManager`.
+Shortcuts available:
+- `tokenRangeAware` - groups partition keys which belong to the same token range. Notice, this strategy is not taking into account shards. Thus, this might be inefficient with ScyllaDB storage backend.
```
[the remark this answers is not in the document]
porunov: ScyllaDB default load balancing strategy routes query using `RoutingKey` to a specific shard which is responsible to answer data for that shard. 
Each shard is served by a separate CPU. I.e. if your node has 8 CPUs than it means this node has 8 shards. 
ScyllaDB shard-aware load balancing strategy calculates ShardID from the provided routing key. If some of your partition keys provided in the query are not related to the specific shard which currently serves your query then part of your query will be routed to a separate shard (separate CPU). This sometimes might be less efficient, but it's hard to say how much less efficient it can be. 
Unfortunately, ShardInfo is currently private in the ScyllaDB Java Driver and it's complicated to figure our ShardId from PartitionKey + Node (without using Reflection API). When this information is open then we could group by Node + ShardId which should be more optimal than grouping by Node only.

docs/configs/janusgraph-cfg.md
```diff
@@ -482,6 +478,39 @@ Configuration options for CQL executor service which is used to process deserial
...
+- `replicasAware` - groups partition keys which belong to the same replica sets (same nodes). Notice, this strategy is not taking into account shards. Thus, this might be inefficient with ScyllaDB storage backend.
```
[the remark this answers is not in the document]
porunov: You are right. In this case key 2 must also be replicated to nodes A, B, and C to be grouped with key 1. If even a single node is different - those keys won't be grouped together.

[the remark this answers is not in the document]
porunov: Each node can have multiple token ranges (for example 256 token ranges). In such case `tokenRangeAware` grouping will be more spread. Most likely you will have more smaller groups with this grouping strategy if compared to `replicasAware`.
If you configured 1 token range per each node than `tokenRangeAware` strategy is equal to `replicasAware`.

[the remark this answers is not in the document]
porunov: I also added explanation into this configuration as the next line, so that it would be clear what's the difference between these grouping strategies.

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/service/GroupingAsyncQueryExecutionService.java
porunov: Make sense. I added warning in case user selects a value which is less than needed.

porunov:
Thank you @li-boxuan ! I force pushed the changes to resolve your comments here: https://github.com/JanusGraph/janusgraph/compare/93d0c4357e0af0d3072c89ba050c907b605c2f6e..f984038d6d7c056d4d9ad7397bf4153c683031cf

li-boxuan (approved):
LGTM 👍

PR #3918, opened 2023-08-24
Let Dependabot update JUnit dependencies together

FlorianHockmann:
junit and junit-platform need to be updated together which is why we always had to manually do the update since Dependabot could only create one PR per dependency.
The new grouping feature should solve this as it lets Dependabot create a single PR for both packages:
https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/


porunov (approved):
LGTM

Issue #3960, opened 2023-09-06
Flaky index test: indexShouldNotExistAfterDeletion


FlorianHockmann:
## Flaky Test

* **Test Name:** BerkeleyElasticsearchTest.indexShouldNotExistAfterDeletion
* **Link:** https://github.com/JanusGraph/janusgraph/blob/e2b0bed82f95817f2656606ce2a62225aed2a9f0/janusgraph-es/src/test/java/org/janusgraph/diskstorage/es/ElasticsearchJanusGraphIndexTest.java#L81
* **Branch:** master
* **Notes:** 

### Stack Trace

```
 Error:  Failures: 
Error:    BerkeleyElasticsearchTest>ElasticsearchJanusGraphIndexTest.indexShouldNotExistAfterDeletion:98->JanusGraphIndexTest.registerIndex:3943 expected: <true> but was: <false>
```

Issue #3959, opened 2023-09-06
Flaky CQL ConfiguredGraphFactoryTest dropShouldCleanUpTraversalSourceAndBindings


FlorianHockmann:
## Flaky Test

* **Test Name:** CQLConfiguredGraphFactoryTest.dropShouldCleanUpTraversalSourceAndBindings
* **Link:** https://github.com/JanusGraph/janusgraph/blob/e2b0bed82f95817f2656606ce2a62225aed2a9f0/janusgraph-backend-testutils/src/main/java/org/janusgraph/core/AbstractConfiguredGraphFactoryTest.java#L381
* **Branch:** master
* **Notes:** I saw it failing on `master` for two recent commits

Failed jobs:
* tests (cql, -Pcassandra4-murmur -Dtest="**/core/cql/*", murmur-core, -Pjava-11, 11)
* tests (scylla, -Pscylla-murmur -Dtest="**/core/cql/*", murmur-core, 8)

### Stack Trace

```java
Error:  Failures: 
Error:    CQLConfiguredGraphFactoryTest>AbstractConfiguredGraphFactoryTest.dropShouldCleanUpTraversalSourceAndBindings:393 expected: not <null>
```

Issue #3931, opened 2023-08-30
Flaky index test testDiscardAndDropRegisteredIndex


FlorianHockmann:
## Flaky Test

* **Test Name:** BerkeleyElasticsearchTest>JanusGraphIndexTest.testDiscardAndDropRegisteredIndex
* **Link:** https://github.com/JanusGraph/janusgraph/blob/d4cffaa415bb642b3855b741b29900dbb8c3bb0b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphIndexTest.java#L3825
* **Branch:** `master`
* **Notes:** 

### Stack Trace

```java
Error:  Tests run: 72, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 495.105 s <<< FAILURE! - in org.janusgraph.diskstorage.es.BerkeleyElasticsearchTest
Error:  org.janusgraph.diskstorage.es.BerkeleyElasticsearchTest.testDiscardAndDropRegisteredIndex  Time elapsed: 60.932 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
```

Issue #3841, opened 2023-06-16
Flaky TinkerPop test g_VX1X_repeatXbothEXcreatedX_whereXwithoutXeXX_aggregateXeX_otherVX_emit_path


porunov:
## Flaky Test

* **Test Name:** `org.apache.tinkerpop.gremlin.process.traversal.step.filter.WhereTest$Traversals
Error:  org.apache.tinkerpop.gremlin.process.traversal.step.filter.WhereTest$Traversals.g_VX1X_repeatXbothEXcreatedX_whereXwithoutXeXX_aggregateXeX_otherVX_emit_path`
* **Branch:** `master`
* **Notes:** I often see this test failing for quite some time now. It was happening even before we updated to TinkerPop 3.6.4. It also happens with any storage backend (I have seen it was failing for `inmemory`, `hbase`, `cassandra`, `berkeleyje` storage backends). For example, `tp-tests (inmemory, -Pjava-11, -Dtest.skip.tp=false -DskipTests=true, 11)` job sometimes fails because of that.

### Stack Trace 1

```
Tests run: 24, Failures: 0, Errors: 1, Skipped: 9, Time elapsed: 28.534 s <<< FAILURE! - in org.apache.tinkerpop.gremlin.process.traversal.step.filter.WhereTest$Traversals
Error:  org.apache.tinkerpop.gremlin.process.traversal.step.filter.WhereTest$Traversals.g_VX1X_repeatXbothEXcreatedX_whereXwithoutXeXX_aggregateXeX_otherVX_emit_path  Time elapsed: 1.145 s  <<< ERROR!
java.lang.IllegalStateException: org.janusgraph.core.JanusGraphException: Exception in JanusGraph
	at org.apache.tinkerpop.gremlin.process.computer.traversal.step.map.VertexProgramStep.processNextStart(VertexProgramStep.java:88)
	at org.apache.tinkerpop.gremlin.process.traversal.step.util.AbstractStep.hasNext(AbstractStep.java:150)
	at org.apache.tinkerpop.gremlin.process.traversal.step.util.ExpandableStepIterator.next(ExpandableStepIterator.java:55)
	at org.apache.tinkerpop.gremlin.process.computer.traversal.step.map.ComputerResultStep.processNextStart(ComputerResultStep.java:68)
	at org.apache.tinkerpop.gremlin.process.traversal.step.util.AbstractStep.next(AbstractStep.java:135)
[...]

PR #3961, opened 2023-09-07
Mark more tests as flaky

FlorianHockmann:
This adds the necessary annotation to repeat flaky tests if necessary for tests described in these issues:
* #1457
* #1459
* #1462
* #1464
* #1465
* #1497
* #2272
* #3142
* #3356
* #3392
* #3393
* #3651
* #3931
* #3959
* #3960

I went through the recent builds on `master` and checked all failed builds. I think this should handle all flaky tests that occurred in these builds, apart from the TinkerPop test where I just don't know how to deal with it: #3841, but we can of course just address that in another PR.
We however also sometimes get failed jobs because of some intermediate issues where a backend cannot be properly set up or teared down for some reason or downloading some artifact just fails. I don't know what we can do to improve this, but I would also consider that out of scope of this PR where I just wanted to deal with failing jobs due to specific tests being flaky.

Hopefully, this will already lead to significantly fewer cases where we have to manually restart failed jobs (which I had to do a lot lately for all these Dependabot PRs).


porunov (commented):
The changes look good to me. However, I'm not sure why we need `minSuccess = 2`.
It means that the test will be executed at least 2 times. I'm OK with it, however, it seems to me that `minSuccess = 1` is good as well.
Apart from that, it would be great if you could add a comment with a link to GitHub Issue which is being closed to each affected test. In such case, later it will be easier to track back to the reason why a specific test is marked as Flaky.

FlorianHockmann:
> However, I'm not sure why we need minSuccess = 2.
It means that the test will be executed at least 2 times. I'm OK with it, however, it seems to me that minSuccess = 1 is good as well.

I simply took that [from our docs on flaky tests](https://github.com/JanusGraph/janusgraph/blob/master/TESTING.md#marking-tests-as-flaky).
I guess the reasoning here was that a flaky test should at least pass in 50% of the cases, but I see your point. If the test is really flaky, then this might not be the case and it doesn't really help us if we then let the build fail again. It would only slow down our  development in general.
I'll change this config and also update the docs accordingly.

> Apart from that, it would be great if you could add a comment with a link to GitHub Issue which is being closed to each affected test. In such case, later it will be easier to track back to the reason why a specific test is marked as Flaky.

Good idea! I'll add that. Just note that our current policy is to leave flaky test issues open until the test is improved in a way so that it isn't flaky any more.

FlorianHockmann:
Done, I also added links to the issues for all flaky tests that already had the annotation.

porunov (approved):
Perfect, thank you @FlorianHockmann !

janusgraph-automations:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|v0.6|Backport failed because of merge conflicts|

### Manual backport
To create the backport manually run:
```
backport --pr 3961
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/JanusGraph/janusgraph/actions/runs/6181946905) for details

FlorianHockmann:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|v0.6|[](https://github.com/JanusGraph/janusgraph/pull/3982)|

### Questions ?

PR #3981, opened 2023-09-14
Remove minSuccess arguments from remaining tests

FlorianHockmann:
Follow-up to #3961 where I forgot to remove this argument from these 2 tests because they are using `ParameterizedRepeatedIfExceptionsTest` instead of `RepeatedIfExceptionsTest`.

This could have been a CTR commit, but unfortunately pushing directly to `master` doesn't work any more due to the required checks we have in place (which we need for the auto-merge feature).


porunov (approved):
LGTM

Issue #3580, opened 2023-02-12
Add possibility to select `scylla` or `cql` dependencies in JanusGraph Server


porunov:
ScyllaDB and DataStax Java Drivers cannot be together in JanusGraph Server because they have the same classpath. Thus, the only possibility to include Scylla drivers into JanusGraph Server right now is by manually replacing DataStax Java Drivers with Scylla Java Drivers.
It will be much more convenient is we add possibility to switch between `cql` and `scylla` drivers in JanusGraph Server using either environment variables or some configurations.
I don't know which approach is better but I would imagine something like below.

We ship all incompatible dependencies between `janusgraph-cql` and `janusgraph-scylla` in separate folders `libsScylla` and `libsCQL`. 
We add an additional option to `janusgraph-server.sh` to include dependencies from `libsScylla` if there is an environment variable exists like the following: `JANUSGRAPH_SCYLLA_DRIVER_ENABLED=true`. Otherwise we include dependencies from `libsCQL` by default.

To make the feature described above we will most likely need to change `janusgraph-dist` flow of building JanusGraph distributions by changing the flow to something like below:
- Build JanusGraph distribution `libs` folder as usual using `janusgraph-cql` dependency
- Build JanusGraph distribution `libs` folder but now excluding `janusgraph-cql` dependency and including `janusgraph-scylla` dependency.
- Run comparison between jars in both `libs` folders. If the a jar from one folder `libs` has an identical jar from another folder `libs` (same `groupId`, same `artifactId`, same `version`) then this jar goes into general `libs` folder. Otherwise this lib goes to `libsCQL` or `libsScylla` respectively.
- After we placed all identical jars in `libs` folder and all different jars in `libsCQL` and `libsScylla` folder we continue with our usual build process (i.e. zipping the resulting archive).

If we change the process to be like above then it should be easy for any user to switch between `CQL` and `Scylla` driver implementations for JanusGraph Server quite easy by providing an environment variable. As Docker supports passing environment variables as well it should be easy to switch drivers in JanusGraph Docker distributions as well.

The initial problem to this issue was described by @FlorianHockmann [here](https://github.com/JanusGraph/janusgraph/issues/1778#issuecomment-667161302).

This issue is a continuation of ScyllaDB Driver integration PR: #3578

cdegroc:
> ScyllaDB and DataStax Java Drivers cannot be together in JanusGraph Server because they have the same classpath.

[...]

PR #4050, opened 2023-10-13
Add possibility to build JanusGraph Scylla Release

porunov:
The idea in this PR is to add another profile which replaces `janusgraph-cql` with `janusgraph-scylla` when activated. Thus, allowing users to build a distribution release either with DataStax drivers or Scylla drivers. 

Building distribution with CQL libs in this PR is the same as we have in our `release` workflow:
```
mvn clean install -Pjanusgraph-release -DskipTests=true --batch-mode --also-make -Dgpg.skip=true
```

Building distribution with Scylla libs in this PR requires activating additional profile `use-scylla`:
```
mvn clean install -Pjanusgraph-release -Puse-scylla -DskipTests=true --batch-mode --also-make -Dgpg.skip=true
```

**Update**: moved previous discussion message into the [following comment](https://github.com/JanusGraph/janusgraph/issues/3580#issuecomment-1763176986), so that this PR focuses on adding possibility to build a distribution release with Scylla libraries instead of copy pasting them manually.
This PR is not about improving our current published JanusGraph distribution builds or adding support to Docker to run Scylla. Docker and published distribution builds can be improved in follow up PRs.

Related to #3580

-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/storage-backend/scylladb.md
+++ b/docs/storage-backend/scylladb.md
@@ -79,8 +79,20 @@ compile "org.janusgraph:janusgraph-scylla:1.0.0"
     The manual process described below is temporary and should be replaced by automated process after the issue
     [janusgraph/janusgraph#3580](https://github.com/JanusGraph/janusgraph/issues/3580) is resolved.
 
-In case `scylla` storage option is needed to be used in JanusGraph distribution then it's necessary to replace the next libraries 
-in `lib` directory (all libraries can be downloaded via Maven Central Repository). 
+In case `scylla` storage option is needed to be used in JanusGraph distribution then it's necessary to replace the next 
+DataStax provided CQL driver with Scylla provided CQL driver.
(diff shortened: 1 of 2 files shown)


farodin91:
It would be interesting to build both docker images with a different tag or?

porunov:
> It would be interesting to build both docker images with a different tag or?

I think we don’t necessarily need 2 docker releases (one for Scylla and another one for CQL). Instead, as I described in the referenced issue we can have an environment variable which controls if Scylla drivers or CQL drivers are used. 
However, I’m OK to have a separate Docker image with ScyllaDB. 
In this PR I’m just adding the ability to manually build 2 different distributions (one with DayaStax drivers and another one with Scylla drivers).

farodin91 (approved):
LGTM. Thank you.

Issue #2624, opened 2021-05-19
Usage of null deviates from TinkerPop 3.5.0


li-boxuan:
TinkerPop 3.5.0 makes the usage of null consistent: http://tinkerpop.apache.org/docs/3.5.0-SNAPSHOT/upgrade/#_use_of_null

This means the semantics of null has changed: http://tinkerpop.apache.org/docs/3.5.0-SNAPSHOT/upgrade/#_graph_system_providers

This blocks the [upgrade to 3.5.0 ](https://github.com/JanusGraph/janusgraph/pull/2619)because the following test is failing

https://github.com/apache/tinkerpop/blob/7025ee2a9abf10cef10749394476061e289cbf70/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/VertexPropertyTest.java#L174-L188

```java
// null property values are not supported so the value should not be added as set or list cardinality,
// but single will remove it
assertEquals(VertexProperty.empty(), v.property(VertexProperty.Cardinality.list, "name", null));
tryCommit(graph, g -> {
    assertEquals(3, IteratorUtils.count(v.properties("name")));
    assertEquals(4, IteratorUtils.count(v.properties()));
    assertVertexEdgeCounts(graph, 1, 0);
});
assertEquals(VertexProperty.empty(), v.property(VertexProperty.Cardinality.single, "name", null));
tryCommit(graph, g -> {
    assertEquals(0, IteratorUtils.count(v.properties("name")));
    assertEquals(1, IteratorUtils.count(v.properties()));
    assertVertexEdgeCounts(graph, 1, 0);
});
```

porunov:
Can't we just disallow `null` values?
Tinkerpop 3.5.0 has a feature method `supportsNullPropertyValues()` which returns `true` if a graph provider supports `null` values and `false` if it doesn't.
JanusGraph should probably set `supportsNullPropertyValues` to `false` because we won't be able to support `null` values in the near future (probably).
That said, I didn't check the implementation yet.

li-boxuan:
This test is exactly for the case when supportsNullPropertyValues is false. If the null value is supported, the null value will be inserted as a legitimate value. If the null value is not supported and null appears in mutation, providers are expected to ignore it if the property has set or list cardinality and remove the property if it has single cardinality.

porunov:
Ah, I see now from the action logs that we are producing:
```
org.janusgraph.core.SchemaViolationException: Property value cannot be null
[...]

PR #4061, opened 2023-10-17
Allow removal of configuration properties

cdegroc:
Fixes https://github.com/JanusGraph/janusgraph/issues/555

-----

### For all changes:

### For code changes:


janusgraph-backend-testutils/src/main/java/org/janusgraph/core/AbstractConfiguredGraphFactoryTest.java
```diff
@@ -247,14 +249,14 @@ public void updateConfigurationShouldRemoveGraphFromCache() throws Exception {
...
-            map.put("storage.hostname", "localhost,localhost");
+            map.put(STORAGE_HOSTS.toStringWithoutRoot(), "localhost,localhost");
             ConfiguredGraphFactory.updateConfiguration(graphName, new MapConfiguration(map));
             assertNull(gm.getGraph(graphName));
             assertNotNull(ConfiguredGraphFactory.open(graphName));
 
             // bogus backend will prevent the graph from being opened
             final Map<String, Object> map2 = graphConfig.getMap();
-            map2.put("storage.backend", "bogusBackend");
+            map2.put(STORAGE_BACKEND.toStringWithoutRoot(), "bogusBackend");
```
cdegroc, line 259: 📝 drive-by change to reuse constants

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
```diff
@@ -335,6 +351,15 @@ private void updateVertexWithProperties(String propertyKey, Object propertyValue
...
 
+    private void removeVertexProperties(String propertyKey, Object propertyValue, Set<String> set) {
+        final GraphTraversalSource g = getTraversal();
+        if (g.V().has(propertyKey, propertyValue).hasNext()) {
+            final Vertex v = g.V().has(propertyKey, propertyValue).next();
+            set.forEach(key -> v.property(key, null));
+            graph.tx().commit();
+        }
+    }
+
```
porunov, line 362: _(nitpick)_ 
if JanusGraph starts to support `null` values for properties (as TinkerPop does) this logic might potentially break.
I think we could potentially use something like this:
```
private void removeVertexProperties(String propertyKey, Object propertyValue, Set<String> set) {
    final GraphTraversalSource g = getTraversal();
    GraphTraversal<Vertex, Vertex> traversal = g.V().has(propertyKey, propertyValue);
    if (traversal.hasNext()) {
        final Vertex v = traversal.next();
        set.forEach(key -> v.property(key).remove());
        g.tx().commit();
    }
}
```
Or maybe even something simpler (if committing after all properties are removed is OK):
```
private void removeVertexProperties(String propertyKey, Object propertyValue, Set<String> set) {
    final GraphTraversalSource g = getTraversal();
    g.V().has(propertyKey, propertyValue).properties(set.toArray(new String[0])).drop().iterate();
    g.tx().commit();
}
``` 

I didn't check neither of the two methods I suggested in this comment, so can't guarantee they will work, but they should I think.

porunov (approved):
LGTM. Thank you @cdegroc !
I only have a nitpick about one of the methods you added, but that's not critical.

porunov (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
cdegroc: Thanks! I didn't know `null` support was on the radar. I've applied your suggestion (which is cleaner than what I had).

cdegroc:
I've deployed this code and tried to update an existing graph configuration. I could remove one or more configuration keys. Both `ConfigurationGraphFactory.removeConfiguration()` and `ManagementSystem.remove()` worked.

janusgraph-core/src/main/java/org/janusgraph/graphdb/management/ConfigurationManagementGraph.java
porunov: It's not on the radar for JanusGraph yet, but TinkerPop already supports `null` values from TinkerPop 3.5.0. Not sure if we are going to support nulls anytime soon, but that's an option for future releases I think.

Related:
https://github.com/JanusGraph/janusgraph/issues/2624 
https://github.com/JanusGraph/janusgraph/pull/2619

PR #4070, opened 2023-10-19
Update ElasticSearch to 8.10.4

porunov:
-----

### For all changes:

### For code changes:


pom.xml
```diff
@@ -75,7 +75,7 @@
         <jackson2.version>2.15.3</jackson2.version>
         <lucene-solr.version>8.11.2</lucene-solr.version>
         <!-- When this version updated please also update sha512 -->
-        <elasticsearch.version>8.6.0</elasticsearch.version>
```
FlorianHockmann, line 78: The comment above mentions some `sha512` which should be updated. That was introduced [here](https://github.com/JanusGraph/janusgraph/commit/cea70aaada2c3f3899c0f2fdd1952ea8e2ea8fb9#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8), but looks like the sha has been removed again already.
So we probably should also remove this comment here.
(This is not really of this PR, just something I just noticed while reviewing the changes. We can also address this with a CTR commit outside of the PR.)

FlorianHockmann (approved):

pom.xml
porunov: Thank you for finding this! Yes, I forgot to remove it previously during ElasticSearch 8 upgrade. I will remove it.

PR #3100, opened 2022-06-24
Redis integration #3018

washerath:
#3018
https://github.com/JanusGraph/janusgraph/issues/3018

Issue #3018, opened 2022-04-23
Centralized Redis cache


washerath:
**Describe the feature:**
In a multi node Janusgraph cluster, data modification done from one instance does not sync with others until it reaches the given expiry time (cache.db-cache-time)

As per the documentation[1] it does not recommends to enable database level cache in a distributed setup as cached data does not share amoung instances.

Once we enable the db.cache, what ever the modification done for particular vertex only visible for that JG instance untill the cache expires. So if we have multiple JG instances, the modifications done from one instance does not reflect on other immediately. 

If we can have centralized cache which syncs up with all JG instances this can be avoided.

**Describe a specific use case for the feature:**

We were able to overcome sync issue on multiple JG instance after doing modifications on void invalidate(StaticBuffer key, List<CachableStaticBuffer> entries) as suggested on here [2]

I have replaced existing guava cache implementation to connect with remote Redis db. So that Redis DB will be act as centralized cache which can connects with all other JG instances.

I would like to contribute my implementation to community. Please guide. 

**NOTE -** Although cache sync between JG instances happens on real time, it can expect slight delays due to network latencies and redis performance. So this implemented Redis cache approach will **compromise the DB consistancy**. 

[1] https://docs.janusgraph.org/operations/cache/#cache-expiration-time
[2] https://lists.lfaidata.foundation/g/janusgraph-users/topic/88877425

li-boxuan:
Hi @washerath You may want to check out the [contributing guide](https://github.com/JanusGraph/janusgraph/blob/master/CONTRIBUTING.md). Looking forward to seeing your PR!

li-boxuan:
Hi, @washerath Just checking if you need any help/guidance to contribute.

n5nk:
@li-boxuan  / @washerath are we planning to merge this Redis PR and make it available for all? Is anything stopping here for review ? 

I recently started realizing we also need a distributed cache, since we run atlas in distributed mode. And querying Cassandra turns out to be costly for us since we ours is a read-heavy system.

PR #4180, opened 2023-12-08
New Redis integration #3018 #3100

washerath:
-----

### For all changes:

### For code changes:

### DIFF
--- a/docs/configs/janusgraph-cfg.md
+++ b/docs/configs/janusgraph-cfg.md
@@ -18,10 +18,17 @@ Configuration options that modify JanusGraph's caching behavior
 
 | Name | Description | Datatype | Default Value | Mutability |
 | ---- | ---- | ---- | ---- | ---- |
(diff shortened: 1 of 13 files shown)


FlorianHockmann (commented):
Thanks a lot for contributing this, @washerath! Before diving too deep into reviewing the changes, I have some high-level questions:

1. Would it make sense to move this into an independent module like `janusgraph-redis`? That way, we wouldn't have to let `janusgraph-core` depend on a Redis library and I think it would also make it easier to maybe add other cache implementations in the future.
2. Are you already using this implementation yourself? Maybe even in production? Could you share some experiences with it? (This is of course not necessary. I'm just curios here.) 
3. @li-boxuan asked [in your other PR](https://github.com/JanusGraph/janusgraph/pull/3100#issuecomment-1166324024) about performance. Do you have anything to share about that?

I also think that we need some documentation for this. Otherwise users won't even know that it exists. Maybe we should even add a new section "Cache Backends" similar to our "Storage Backends" and "Index Backends".
But we can also do that in a follow-up PR I think.

PR #4258, opened 2024-02-08
Removed missing external reference in Elasticsearch index backend docs

pm-osc:
The following section of the JanusGraph docs https://docs.janusgraph.org/index-backend/elasticsearch/#write-optimization contains an external reference that is not available any more. Following the suggestion from @li-boxuan on Discord, I open this PR to remove the missing external reference.


li-boxuan (approved):
LGTM, thank you!

FlorianHockmann (approved):

PR #4262, opened 2024-02-15
Fix missing setuptools package for regression test

### LINKED ISSUE #4154: Benchmark Performance regression check failing regularly
I've noticed that this workflow is currently failing for some backporting PRs:
* #4143
* #4144
* #4145
* #4146

But it also failed for many recent commits on `master` and on `v1.0`.

Looks like there is some dependency missing:

```
12:55:57 ERROR com.datastax.oss.driver.api.testinfra.ccm.CcmBridge$2.processLine - ccmerr> Traceback (most recent call last):
12:55:57 ERROR com.datastax.oss.driver.api.testinfra.ccm.CcmBridge$2.processLine - ccmerr>   File "/opt/hostedtoolcache/Python/3.12.0/x64/bin/ccm", line 7, in <module>
12:55:57 ERROR com.datastax.oss.driver.api.testinfra.ccm.CcmBridge$2.processLine - ccmerr>     import pkg_resources
12:55:57 ERROR com.datastax.oss.driver.api.testinfra.ccm.CcmBridge$2.processLine - ccmerr> ModuleNotFoundError: No module named 'pkg_resources'
```

li-boxuan:
Our regression test CI does not pin python version - it simply uses 3.x. Since python 3.12, setuptools package, which is needed by ccm package we use, is no longer included by default. This fixes the problem by explicitly installing setuptools package.

Fixes #4154

-----

### For all changes:

### For code changes:

### DIFF
--- a/.github/workflows/ci-benchmark.yml
+++ b/.github/workflows/ci-benchmark.yml
@@ -58,6 +58,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.x'
+      - run: pip install setuptools
       - run: pip install git+https://github.com/li-boxuan/ccm.git
 
       - uses: actions/checkout@v4


li-boxuan:
Note: I can also fix it in my forked li-boxuan/ccm package, but I figure fixing it here would probably be more appropriate.

I don't think submitting a PR to upstream ccm repo would work. I've tried that in the past and didn't get any response.

FlorianHockmann (approved):
Thanks for the quick fix :-)

(There are some failing tests here, but that must be unrelated to your changes here.)

PR #4409, opened 2024-04-22
Added retry logic if client request returns an error code that is configured for retrying

### LINKED ISSUE #4408: Implement Retry Logic For Elasticsearch Requests
**Describe the feature:**
JG views writes to mixed indices as a "best effort" and [documents](https://docs.janusgraph.org/operations/recovery/#transaction-failure) that failures in transactions with regards to mixed indices are left to a separate periodic clean up process after enabling JG's WAL. 

However this results in clients not getting feedback for more mundane write failures to mixed indices like Elasticsearch like a `429 Too Many Requests` that just need to respect the backpressure being conveyed by Elasticsearch and retry again.

[Elasticsearch's guidance](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-indexing-speed.html#multiple-workers-threads) states that 429 errors in particular are the client's responsibility.

To implement this feature would require implementing retry logic around [this callsite](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java#L555). Additionally supporting some new config settings for Elasticsearch:

- `retry-limit`: How many times to attempt a retry. Default to 0 to maintain current behavior.
- `retry-initial-wait`: How long (in milliseconds) to wait before applying the initial retry. Defaults to 1ms.
- `retry-max-wait`: How long (in milliseconds) to wait at most for each retry attempt, operates as a ceiling if a high number of retries are configured since exponential backoff will grow exceptionally quickly. Defaults to 1000ms.
- `retry-error-codes`: Which error codes contained in a thrown Elasticsearch `ResponseException` to apply retries upon.

Once this feature is implemented it'll make writes to Elasticsearch better handle write errors like 429 during high write periods instead of silently failing the write and users of JanusGraph needing to be aware of the WAL & separate periodic repair application.

criminosis:
Closes #4408

To wait between retries since the code being called into Elasticsearch's client appears to be the synchronous API I opted for just using `Thread.sleep`. If there's a better utility or if a different approach is desired happy change.

Additionally I marked the new configuration options with the `LOCAL` flag. I noticed `retry_on_conflict` went with `MASKABLE` but it didn't seem readily apparent what further obligations would be needed to fulfill the "global" portion of that contract, so just figured it'd be best to start with `LOCAL` and ask about it in the PR.

---

### For all changes:

### For code changes:

- [X] Have you ensured that format looks appropriate for the output in which it is rendered?

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -304,6 +304,26 @@ public class ElasticSearchIndex implements IndexProvider {
             "Sets the maximum socket timeout (in milliseconds).", ConfigOption.Type.MASKABLE,
             Integer.class, RestClientBuilder.DEFAULT_SOCKET_TIMEOUT_MILLIS);
 
+    public static final ConfigOption<Integer> RETRY_LIMIT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry-limit",
+            "Sets the number of attempts for configured retryable error codes.", ConfigOption.Type.LOCAL,
+            Integer.class, 0);
+
+    public static final ConfigOption<Long> RETRY_INITIAL_WAIT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry-initial-wait",
+            "Sets the initial retry wait time (in milliseconds) before exponential backoff.",
+            ConfigOption.Type.LOCAL, Long.class, 1L);
+
+    public static final ConfigOption<Long> RETRY_MAX_WAIT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry-max-wait",
+            "Sets the max retry wait time (in milliseconds).", ConfigOption.Type.LOCAL,
+            Long.class, 1000L);
+
+    public static final ConfigOption<String[]> RETRY_ERROR_CODES =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry-error-codes",
+            "Comma separated list of Elasticsearch REST client ResponseException error codes to retry. " +
+                "E.g. \"408,429\"", ConfigOption.Type.LOCAL, String[].class, new String[0]);
+
     public static final int HOST_PORT_DEFAULT = 9200;
 
     /**
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
@@ -47,10 +47,12 @@ import org.slf4j.LoggerFactory;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -110,14 +112,28 @@ public class RestElasticSearchClient implements ElasticSearchClient {
     private Integer retryOnConflict;
 
     private final String retryOnConflictKey;
-    
-    public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean useMappingTypesForES7) {
+
+    private final int retryAttemptLimit;
+
+    private final Set<Integer> retryOnErrorCodes;
+
+    private final long retryInitialWaitMs;
+
+    private final long retryMaxWaitMs;
+
+public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean useMappingTypesForES7,
+                               int retryAttemptLimit, Set<Integer> retryOnErrorCodes, long retryInitialWaitMs,
+                               long retryMaxWaitMs) {
         this.delegate = delegate;
         majorVersion = getMajorVersion();
         this.scrollKeepAlive = scrollKeepAlive+"s";
         esVersion7 = ElasticMajorVersion.SEVEN.equals(majorVersion);
         useMappingTypes = majorVersion.getValue() < 7 || (useMappingTypesForES7 && esVersion7);
         retryOnConflictKey = majorVersion.getValue() >= 7 ? "retry_on_conflict" : "_retry_on_conflict";
+        this.retryAttemptLimit = retryAttemptLimit;
+        this.retryOnErrorCodes = Collections.unmodifiableSet(retryOnErrorCodes);
+        this.retryInitialWaitMs = retryInitialWaitMs;
+        this.retryMaxWaitMs = retryMaxWaitMs;
     }
 
     @Override
@@ -546,13 +562,35 @@ public class RestElasticSearchClient implements ElasticSearchClient {
         return performRequest(new Request(method, path), requestData);
     }
 
+    private Response performRequestWithRetry(Request request) throws IOException {
+        int retryCount = 0;
+        while (true) {
+            try {
+                return delegate.performRequest(request);
+            } catch (ResponseException e) {
+                if (!retryOnErrorCodes.contains(e.getResponse().getStatusLine().getStatusCode()) || retryCount >= retryAttemptLimit) {
+                    throw e;
+                }
+                //Wait before trying again
+                long waitDurationMs = Math.min((long) (retryInitialWaitMs * Math.pow(10, retryCount)), retryMaxWaitMs);
+                log.warn("Retrying Elasticsearch request in {} ms. Attempt {} of {}", waitDurationMs, retryCount, retryAttemptLimit);
+                try {
+                    Thread.sleep(waitDurationMs);
+                } catch (InterruptedException interruptedException) {
+                    throw new RuntimeException(String.format("Thread interrupted while waiting for retry attempt %d of %d", retryCount, retryAttemptLimit), interruptedException);
+                }
+            }
+            retryCount++;
+        }
+    }
+
     private Response performRequest(Request request, byte[] requestData) throws IOException {
 
         final HttpEntity entity = requestData != null ? new ByteArrayEntity(requestData, ContentType.APPLICATION_JSON) : null;
 
         request.setEntity(entity);
 
-        final Response response = delegate.performRequest(request);
+        final Response response = performRequestWithRetry(request);
 
         if (response.getStatusLine().getStatusCode() >= 400) {
             throw new IOException("Error executing request: " + response.getStatusLine().getReasonPhrase());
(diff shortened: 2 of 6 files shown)


li-boxuan (commented):
Thanks for taking care of this! My two cents:

1) Either local or maskable is fine. Maskable means the change will be (automatically) propagated to different instances, but every instance could override the value locally. Local is simpler and has less overhead (as no propagation is needed). I would prefer to use LOCAL here.
2) It's okay to write our own exponential backoff code.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -546,13 +572,42 @@ private Response performRequest(String method, String path, byte[] requestData)
...
+                if (!retryOnErrorCodes.contains(e.getResponse().getStatusLine().getStatusCode())) {
+                    throw e;
+                }
+                lastException = e;
+            }
+            retryCount++;
+        } while (retryCount <= retryAttemptLimit);
+
+        //If we've exceeded our retry limit then throw the last exception that was captured from retrying
+        throw lastException;
```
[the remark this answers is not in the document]
criminosis: I don't follow how it's a bug @li-boxuan.

> If the delegate.performRequest(request) does not throw a ResponseException on the first attempt and there are no retry attempts (because the response is successful or an error not suitable for retry occurs), the do-while loop will not enter or will exit early, and lastException will never be assigned a value.

ChatGPT seems to have missed the semantic implication that other exception types are not caught and thus would be thrown _immediately_ out of the method. In that scenario `lastException` never being assigned a value is moot.

The four cases should be:

1. The request is successful - we return out of the method.
2. The request is not successful - but it's an exception that's different than a `ResponseException` so it immediately gets thrown up out of the method.
3. The request is not successful - it is a `ResponseException` _but_ it is not a configured error code for retry and thus will be immediately thrown up out of the method.
4. The request is not successful - it is a `ResponseException` _and_ it *is* a configured error code for retry, which will then cause `lastException` to be assigned and possibly later thrown if it is the last (or only) attempt if retries are set to 0.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -304,6 +304,26 @@ public class ElasticSearchIndex implements IndexProvider {
...
+            ConfigOption.Type.LOCAL, Long.class, 1L);
+
+    public static final ConfigOption<Long> RETRY_MAX_WAIT =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry-max-wait",
+            "Sets the max retry wait time (in milliseconds).", ConfigOption.Type.LOCAL,
+            Long.class, 1000L);
+
+    public static final ConfigOption<String[]> RETRY_ERROR_CODES =
+        new ConfigOption<>(ELASTICSEARCH_NS, "retry-error-codes",
+            "Elasticsearch REST client ResponseException error codes To Retry.", ConfigOption.Type.LOCAL,
```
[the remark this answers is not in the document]
criminosis: Added in 5354c7dc7 👍

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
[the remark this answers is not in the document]
criminosis: Added some more tests in 9b57c0349 to verify what ChatGPT is claiming is a non-issue.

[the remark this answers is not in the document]
li-boxuan: Oh lol thank you so much for the detailed explanation. I rely on AI too much these days... and got cheated by its hallucination that at the first glance, seemed reasonable.

li-boxuan (approved):
This is great, thanks!

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -419,6 +429,22 @@ public void setRetryOnConflict(Integer retryOnConflict) {
...
+        this.retryInitialWaitMs = retryInitialWaitMs;
+    }
+
+    public void setRetryMaxWaitMs(long retryMaxWaitMs) {
+        this.retryMaxWaitMs = retryMaxWaitMs;
+    }
+
+    public void setRetryOnErrorCodes(Set<Integer> retryOnErrorCodes) {
+        this.retryOnErrorCodes.addAll(retryOnErrorCodes);
+    }
```
porunov, line 446: The method name is `setRetryOnErrorCodes`, but instead of `set` operation it `adds`.
You can fix it by actually setting `this.retryOnErrorCodes = retryOnErrorCodes` instead of mutating existing collection.
However, it seems there is no need to execute this method more than once during the lifetime of this class.
Thus, I would suggest moving all of these parameters into the constructor instead. I know that `setRetryOnConflict` and `setBulkRefresh` have the same problem, but I think it should be refactored.

Feel free to ignore this comment if you see any complications in this refactoring.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -546,13 +572,42 @@ private Response performRequest(String method, String path, byte[] requestData)
...
+                }
+            }
+            try {
+                return delegate.performRequest(request);
+            } catch (ResponseException e) {
+                if (!retryOnErrorCodes.contains(e.getResponse().getStatusLine().getStatusCode())) {
+                    throw e;
+                }
+                lastException = e;
+            }
```
porunov, line 596: it seems logic under the if statement `if (retryCount > 0)` needs to be executed on exception only. I would suggest moving the whole logic under that if statement under the `catch`. In this case you avoid having `if (retryCount > 0)`.
I.e. : 
```
} catch (ResponseException e) {
  if (!retryOnErrorCodes.contains(e.getResponse().getStatusLine().getStatusCode())) {
    throw e;
  }
  lastException = e;
  // put here that `sleep` logic.
}
```

porunov (commented):
Thank you @criminosis ! I have just some nitpicks below. Otherwise looks good.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: I was going to comment back that doing so would mean we'd do a wait when we'd end up breaking out of the loop anyways if we don't have that check. But I realized we could condense the loop's condition into the `if` statement and then get the best of both worlds at the trade off of doing the loop as a `while (true)`.

Done in 13e65bf87

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: Happy to make the change but for what it's worth that's going to turn the signature into:

```
public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean useMappingTypesForES7,
                                   int retryAttemptLimit, Set<Integer> retryOnErrorCodes, long retryInitialWaitMs,
                                   long retryMaxWaitMs) {
```

Which may be pushing a parameter parameter count for transitioning to a builder. Especially if `setRetryOnConflict` and `setBulkRefresh` are added one day.

criminosis: @porunov Moved the new configuration settings into the constructor as suggested. Isolated that into commit c3961fc22. 

If that looks acceptable then I'll do the squash & sign-off to get the PR into a mergable state again. Figured I'd wait before squashing it in case you had second thoughts upon seeing it given my aforementioned comment.

porunov: > Happy to make the change but for what it's worth that's going to turn the signature into:
> 
> ```
> public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean useMappingTypesForES7,
>                                    int retryAttemptLimit, Set<Integer> retryOnErrorCodes, long retryInitialWaitMs,
>                                    long retryMaxWaitMs) {
> ```
> 
> Which may be pushing a parameter parameter count for transitioning to a builder. Especially if `setRetryOnConflict` and `setBulkRefresh` are added one day.

Yes, that's a refactoring issue, I agree. However, I think it still better to have a clear contract for this class creation. Later on this contract can be simplified by moving all these parameters to a separate "configuration" POJO for easier extensibility and clearer constructor.
I like it better how it looks right now than it was before. Thank you!

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
porunov: Thank you. Looks great!

porunov (commented):
@criminosis please, feel free to squash the commits. However, notice that ElasticSearch tests are now failing:
```
Failures: 
Error:    RestClientSetupTest.testConnectBasicHttpConfigurationAllOptions:257 expected: <[408, 429]> but was: <null>
Error:  Errors: 
Error:    RestClientSetupTest.testSSLDisableHostNameVerifierDefaultOff » InvalidUseOfMatchers 
```

criminosis:
@porunov yeah, I just focused on compilation passing to present the new signatures in case you had feedback or further modification, etc.

Fixed it all up, squashed and re-signed off again. Should be ready for review again 👍

porunov (approved):
LGTM. Thank you for your contribution @criminosis !

criminosis:
Thank you @porunov & @li-boxuan , glad to finally get to find something to contribute back to JanusGraph!

Out of curiosity I see it's been earmarked for the 1.1.0 milestone. Assuming the build completely passes and it gets merged will it go out into nightly build / release prior to the 1.1.0 release?

Looking forward to getting this my cluster to see if it helps the issue that originally prompted this.

porunov:
> Thank you @porunov & @li-boxuan , glad to finally get to find something to contribute back to JanusGraph!
> 
> Out of curiosity I see it's been earmarked for the 1.1.0 milestone. Assuming the build completely passes and it gets merged will it go out into nightly build / release prior to the 1.1.0 release?
> 
> Looking forward to getting this my cluster to see if it helps the issue that originally prompted this.

Yes. If everything passes the snapshot release is automatically created for each merged commit for any supported branches (`master` and `v1.0` at the moment). You can read about snapshot releases here: https://docs.janusgraph.org/advanced-topics/commit-releases/

PR #4426, opened 2024-04-29
Drop optimization

ntisseyre:
Created a new MultiVertexQuery `drop` API that efficiently drops many vertices.

### DIFF
--- a/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
+++ b/janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
@@ -10011,6 +10011,48 @@ public abstract class JanusGraphTest extends JanusGraphBaseTest {
         graph2.close();
     }
 
+    @Test
+    public void testMultiQueryDropsVertices() {
+
+        mgmt.makePropertyKey("id").dataType(Integer.class).cardinality(Cardinality.SINGLE).make();
+        PropertyKey nameProp = mgmt.makePropertyKey("name").dataType(String.class).cardinality(Cardinality.SINGLE).make();
+        mgmt.makePropertyKey("details").dataType(String.class).cardinality(Cardinality.SINGLE).make();
+        mgmt.buildIndex("nameIndex", Vertex.class).addKey(nameProp).buildCompositeIndex();
+
+        finishSchema();
+
+        int verticesAmount = 42;
+
+        for (int i = 0; i < verticesAmount; i++) {
+            Vertex vertex = tx.addVertex("id", i);
+            vertex.property("name", "name_test");
+            vertex.property("details", "details_" + i);
+        }
+
+        clopen();
+
+        List<JanusGraphVertex> vertices = tx.traversal()
+            .V()
+            .has("name", "name_test")
+            .toList()
+            .stream()
+            .map(v -> (JanusGraphVertex) v)
+            .collect(Collectors.toList());
+
+        int actualCount = tx.multiQuery(vertices).drop();
+        clopen();
+
+        assertEquals(verticesAmount, actualCount);
+
+        int afterDropCount = tx.traversal()
+            .V()
+            .has("name", "name_test")
+            .toList()
+            .size();
+
+        assertEquals(0, afterDropCount);
+    }
+
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
     public void testParallelBackendOps(boolean parallelBackendOpsEnabled) {
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
@@ -30,9 +30,11 @@ import org.janusgraph.core.schema.SchemaStatus;
 import org.janusgraph.diskstorage.EntryList;
 import org.janusgraph.diskstorage.keycolumnvalue.SliceQuery;
 import org.janusgraph.graphdb.database.EdgeSerializer;
+import org.janusgraph.graphdb.database.idhandling.IDHandler;
 import org.janusgraph.graphdb.internal.ElementLifeCycle;
 import org.janusgraph.graphdb.internal.InternalRelationType;
 import org.janusgraph.graphdb.internal.InternalVertex;
+import org.janusgraph.graphdb.internal.OrderList;
 import org.janusgraph.graphdb.internal.RelationCategory;
 import org.janusgraph.graphdb.query.BackendQueryHolder;
 import org.janusgraph.graphdb.query.JanusGraphPredicate;
@@ -98,6 +100,12 @@ public abstract class BasicVertexCentricQueryBuilder<Q extends BaseVertexQuery<Q
      * Whether to query for system relations only
      */
     private boolean querySystem = false;
+
+    /**
+     * Whether to query for all relations, including system.
+     */
+    private boolean queryAll = false;
+
     /**
      Whether to query only for persisted edges, i.e. ignore any modifications to the vertex made in this transaction.
      This is achieved by using the {@link SimpleVertexQueryProcessor} for execution.
@@ -160,6 +168,15 @@ public abstract class BasicVertexCentricQueryBuilder<Q extends BaseVertexQuery<Q
      */
     public Q system() {
         this.querySystem = true;
+        return getThis();
+    }
+
+    /**
+     * Query all relations, including system types.
+     * @return
+     */
+    public Q all() {
+        this.queryAll = true;
         return getThis();
     }
 
@@ -500,10 +517,24 @@ public abstract class BasicVertexCentricQueryBuilder<Q extends BaseVertexQuery<Q
         return query;
     }
 
+    private BaseVertexCentricQuery getAllRelationsQuery() {
+        SliceQuery fullQuery = new SliceQuery(
+            IDHandler.getBounds(RelationCategory.RELATION, true)[0],
+            IDHandler.getBounds(RelationCategory.RELATION, false)[1]);
+
+        List<BackendQueryHolder<SliceQuery>> queries = Collections.singletonList(new BackendQueryHolder<>(fullQuery, true, true));
+        return new BaseVertexCentricQuery(RelationCategory.RELATION, Direction.BOTH, queries, OrderList.NO_ORDER, Query.NO_LIMIT);
+    }
+
     protected BaseVertexCentricQuery constructQueryWithoutProfile(RelationCategory returnType) {
         assert returnType != null;
         Preconditions.checkArgument(adjacentVertex==null || returnType == RelationCategory.EDGE,
                 "Vertex constraints only apply to edges");
+
+        if(queryAll) {
+            return getAllRelationsQuery();
+        }
+
         if (limit <= 0)
             return BaseVertexCentricQuery.emptyQuery();
 
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
@@ -57,7 +57,7 @@ public class MultiVertexCentricQueryBuilder extends BasicVertexCentricQueryBuild
 
     public MultiVertexCentricQueryBuilder(final StandardJanusGraphTx tx, Integer initialVerticesCapacity) {
         super(tx);
-        vertices = initialVerticesCapacity!=null ? new HashSet<>(initialVerticesCapacity) : new HashSet<>();
+        vertices = initialVerticesCapacity != null ? new HashSet<>(initialVerticesCapacity) : new HashSet<>();
     }
 
     @Override
@@ -74,7 +74,7 @@ public class MultiVertexCentricQueryBuilder extends BasicVertexCentricQueryBuild
     public JanusGraphMultiVertexQuery addVertex(Vertex vertex) {
         assert vertex != null;
         assert vertex instanceof InternalVertex;
-        vertices.add(((InternalVertex)vertex).it());
+        vertices.add(((InternalVertex) vertex).it());
         return this;
     }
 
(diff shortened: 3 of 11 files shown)


janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
```diff
@@ -0,0 +1,146 @@
...
+                .drop()
+                .toList()
+                .size();
+        }
+
+        tx.rollback();
+
+        /*if (dropCount != verticesAmount) {
+            throw new AssertionError("Only " + dropCount + " vertices were dropped, expected " + verticesAmount);
+        }*/
```
porunov, line 116: Seems this is better suited for unit testing than benchmark testing.

porunov (commented):
I would prefer adding a simple unit test also which tests the new API `tx.multiQuery(vertices).drop()` for better test coverage.
Otherwise this PR looks good to me. 
After this PR is merged I think we could also overwrite TinkerPop's `DropStep` to leverage this new optimization in Gremlin queries without the need to using `multuQuery` directly (i.e. similar as we did with `JanusGraphHasStep` as it is also a `filter` step).

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
ntisseyre: Will add unit test to validate deletion

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
```diff
@@ -0,0 +1,146 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph;
+
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.janusgraph.core.*;
```
porunov, line 18: I had to replace it with direct imports because lint doesn't allow star imports

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
```diff
@@ -0,0 +1,146 @@
...
+import org.janusgraph.core.schema.JanusGraphManagement;
+import org.janusgraph.diskstorage.configuration.ConfigElement;
+import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
+import org.janusgraph.diskstorage.configuration.WriteConfiguration;
+import org.janusgraph.diskstorage.cql.CQLConfigOptions;
+import org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration;
+import org.openjdk.jmh.annotations.*;
```
porunov, line 25: I had to replace it with direct imports because lint doesn't allow star imports

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
```diff
@@ -0,0 +1,146 @@
...
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.ADJUST_LIMIT;
+import static org.janusgraph.graphdb.configuration.GraphDatabaseConfiguration.HARD_MAX_LIMIT;
```
porunov, line 36: Ensure you don't add any unused imports because lint doesn't allow them (these two and above)

ntisseyre:
> I would prefer adding a simple unit test also which tests the new API `tx.multiQuery(vertices).drop()` for better test coverage. Otherwise this PR looks good to me. After this PR is merged I think we could also overwrite TinkerPop's `DropStep` to leverage this new optimization in Gremlin queries without the need to using `multuQuery` directly (i.e. similar as we did with `JanusGraphHasStep` as it is also a `filter` step).

I'm not sure it will work the same way on individual `DropStep`, since the benefit of multi-query that it constructs a query once and applies to all vertices.
For individual `DropStep`, it constructs a query every time for each individual vertex to fetch relations (system + standard), and that takes so much time.

porunov:
> > I would prefer adding a simple unit test also which tests the new API `tx.multiQuery(vertices).drop()` for better test coverage. Otherwise this PR looks good to me. After this PR is merged I think we could also overwrite TinkerPop's `DropStep` to leverage this new optimization in Gremlin queries without the need to using `multuQuery` directly (i.e. similar as we did with `JanusGraphHasStep` as it is also a `filter` step).
> 
> I'm not sure it will work the same way on individual `DropStep`, since the benefit of multi-query that it constructs a query once and applies to all vertices. For individual `DropStep`, it constructs a query every time for each individual vertex to fetch relations (system + standard), and that takes so much time.

Yes, the default `DropStep` processes vertices one by one, but our multiquariable custom JanusGraph steps which overwrite TinkerPop steps extend the interface with additional methods to "pre-register vertices which will be processed later". Thus, the overwritten version of the `DropStep` will simply execute `multiQuery` for a batch of vertices and any other execution of `filter` method and the vertices which are already processed will be skipped until a new batch is registered.
We use this technique for many custom steps (JanusGraphHasStep, JanusGraphPropertiesStep, JanusGraphVertexStep, etc.). After this PR is merged I can implement `JanusGraphDropStep` as it is quite easy to do if we already have the necessary `multiQuery` API and I already done that for multiple steps. So, I think it shouldn't be a problem unless I am missing something.

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
ntisseyre: Done

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
ntisseyre: Done

janusgraph-benchmark/src/main/java/org/janusgraph/MultiDropBenchmark.java
ntisseyre: Done

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphTest.java
```diff
@@ -10011,6 +10011,48 @@ public void testFullDBCacheInvalidation() throws InterruptedException, Execution
...
+
+        assertEquals(verticesAmount, actualCount);
+
+        int afterDropCount = tx.traversal()
+            .V()
+            .has("name", "name_test")
+            .toList()
+            .size();
+
+        assertEquals(afterDropCount, 0);
```
porunov, line 10053: (nitpick)
```
assertEquals(0, afterDropCount);
```

porunov (approved):
LGTM. Thank you @ntisseyre !

porunov (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
```diff
@@ -202,4 +203,19 @@ public Map<JanusGraphVertex, VertexList> vertexIds() {
...
+
+        return relations
+            .keySet()
+            .stream()
+            .mapToInt(vertex -> {
+                Iterable<JanusGraphRelation> allRelations = Iterables.concat(relations.get(vertex), systemRelations.get(vertex));
+                ((InternalVertex) vertex).remove(allRelations);
+                return 1;
+            })
+            .sum();
```
li-boxuan, line 219: perf nit: this seems unnecessary computation from performance perspective. I would guess JVM will be able to optimize this, but I'd prefer not to rely on that. We could simply return `relations.size()` without doing summation.

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
```diff
@@ -202,4 +203,19 @@ public Map<JanusGraphVertex, VertexList> vertexIds() {
         return execute(RelationCategory.EDGE, new VertexIdConstructor());
     }
 
+    @Override
+    public Integer drop() {
+        Map<JanusGraphVertex, Iterable<JanusGraphRelation>> relations = this.noPartitionRestriction().relations();
```
mad, line 208: Here it is possible to reduce queries by half

```
Subject: [PATCH]
---
Index: janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java	(revision 42faab99d80a68a49454102e0e30e2d1f2cba57b)
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java	(revision 6c7ce256fa20fa13f4612ab84eda0d05853902f0)
@@ -30,6 +30,7 @@
 import org.janusgraph.core.schema.SchemaStatus;
 import org.janusgraph.diskstorage.keycolumnvalue.SliceQuery;
 import org.janusgraph.graphdb.database.EdgeSerializer;
+import org.janusgraph.graphdb.database.idhandling.IDHandler;
 import org.janusgraph.graphdb.internal.ElementLifeCycle;
 import org.janusgraph.graphdb.internal.InternalRelationType;
 import org.janusgraph.graphdb.internal.InternalVertex;
@@ -101,6 +102,7 @@
      * Whether to query for system relations only
      */
     private boolean querySystem = false;
+    private boolean all = false;
     /**
      Whether to query only for persisted edges, i.e. ignore any modifications to the vertex made in this transaction.
      This is achieved by using the {@link SimpleVertexQueryProcessor} for execution.
@@ -165,6 +167,11 @@
         return getThis();
     }
 
+    public Q all() {
+        this.all = true;
+        return getThis();
+    }
+
     /**
      * Calling this method will cause this query to only included loaded (i.e. unmodified) relations in the
      * result set.
@@ -467,10 +474,22 @@
         return new BaseVertexCentricQuery(condition, Direction.OUT, queries, OrderList.NO_ORDER, Query.NO_LIMIT);
     }
 
+    private BaseVertexCentricQuery getFullQuery() {
+        SliceQuery fullQuery = new SliceQuery(
+            IDHandler.getBounds(RelationCategory.RELATION, true)[0],
+            IDHandler.getBounds(RelationCategory.RELATION,false)[1]);
+
+        List<BackendQueryHolder<SliceQuery>> queries = Collections.singletonList(new BackendQueryHolder<>(fullQuery, true, true));
+        return new BaseVertexCentricQuery(RelationCategory.RELATION, Direction.BOTH, queries, OrderList.NO_ORDER, Query.NO_LIMIT);
+    }
+
     protected BaseVertexCentricQuery constructQueryWithoutProfile(RelationCategory returnType) {
         assert returnType != null;
         Preconditions.checkArgument(adjacentVertex==null || returnType == RelationCategory.EDGE,
                 "Vertex constraints only apply to edges");
+        if (all) {
+            return getFullQuery();
+        }
         if (isVertexLabelType()) {
             return queryForLabel();
         }
Index: janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/AbstractVertex.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/AbstractVertex.java b/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/AbstractVertex.java
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/AbstractVertex.java	(revision 42faab99d80a68a49454102e0e30e2d1f2cba57b)
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/vertices/AbstractVertex.java	(revision 6c7ce256fa20fa13f4612ab84eda0d05853902f0)
@@ -103,13 +103,7 @@
     public synchronized void remove() {
         verifyAccess();
 //        if (isRemoved()) return; //Remove() is idempotent
-        Iterator<JanusGraphRelation> iterator = it().query().noPartitionRestriction().relations().iterator();
-        while (iterator.hasNext()) {
-            iterator.next();
-            iterator.remove();
-        }
-        //Remove all system types on the vertex
-        for (JanusGraphRelation r : it().query().noPartitionRestriction().system().relations()) {
+        for (JanusGraphRelation r : it().query().noPartitionRestriction().all().relations()) {
             r.remove();
         }
     }
```

ntisseyre: nice!

ntisseyre: i have tested this approach, it doesn't seem to return all relations, it returns only half of it

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
porunov: Yeah, seems can be just:
```
relations.keySet().foreach(vertex -> ((InternalVertex) vertex).remove(Iterables.concat(relations.get(vertex), systemRelations.get(vertex))));
return relations.size();
```

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
ntisseyre: never mind, i had a wrong check. The code works as expected.
I have integrated this change into mine, thanks again @mad !

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
ntisseyre: Fixed, thanks @li-boxuan

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/BasicVertexCentricQueryBuilder.java
```diff
@@ -500,10 +517,25 @@ protected BaseVertexCentricQuery constructQuery(RelationCategory returnType) {
         return query;
     }
 
+    private BaseVertexCentricQuery getAllRelationsQuery() {
+        System.out.println("here");
```
porunov, line 521: Unnecessary line

ntisseyre: it is a crucial line; the whole boost is coming from here!

porunov: My mistake! Bring this line back :smile:

li-boxuan (approved):
LGTM, just one nit:

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
mad: With a new query, this part `AbstractVertex.remove()` can be done in one pass

mad (approved):

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
```diff
@@ -202,4 +202,12 @@ public Map<JanusGraphVertex, VertexList> vertexIds() {
         return execute(RelationCategory.EDGE, new VertexIdConstructor());
     }
 
+    @Override
+    public Integer drop() {
+        Map<JanusGraphVertex, Iterable<JanusGraphRelation>> allRelations = this.noPartitionRestriction().all().relations();
+        allRelations.keySet().forEach(vertex -> {
+            ((InternalVertex) vertex).remove(allRelations.get(vertex));
```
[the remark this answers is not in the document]
porunov: > ```java
> allRelations.entrySet().forEach(entry -> {
>     JanusGraphVertex vertex = entry.getKey();
>     Iterable<JanusGraphRelation> relations = entry.getValue();
>     ((InternalVertex) vertex).remove(relations);
> });
> ```

No need to get `entrySet` in this case. I.e.:
```
allRelations.forEach((vertex, relations) -> ((InternalVertex) vertex).remove(relations));
```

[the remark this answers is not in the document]
ntisseyre: Updated

janusgraph-core/src/main/java/org/janusgraph/graphdb/query/vertex/MultiVertexCentricQueryBuilder.java
ntisseyre: Updated

porunov (approved):

PR #4428, opened 2024-04-30
fix: chown fail when OS create folders or files with root only permission

tien:
I'm having issue with Janusgraph whenever the OS create files or folders in the Janusgraph owned directory. For example, when the system restart unexpectedly and a `lost+found` folder is created. The `lost+found` created by `fsck` have root only access, so `chown` will fail.

I don't think `chown` need to be recursive because files that are created & used by Janusgraph will all be created by the default user (`janusgraph`) anyway.

### DIFF
--- a/janusgraph-dist/docker/docker-entrypoint.sh
+++ b/janusgraph-dist/docker/docker-entrypoint.sh
@@ -20,7 +20,7 @@ JANUSGRAPH_SERVER_YAML="${JANUS_CONFIG_DIR}/janusgraph-server.yaml"
 # running as root; step down to run as janusgraph user
 if [ "$1" == 'janusgraph' ] && [ "$(id -u)" == "0" ]; then
   mkdir -p ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
-  chown -R janusgraph:janusgraph ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
+  chown janusgraph:janusgraph ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
   chmod 700 ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
 
   exec chroot --skip-chdir --userspec janusgraph:janusgraph / "${BASH_SOURCE}" "$@"
@@ -32,7 +32,7 @@ if [ "$1" == 'janusgraph' ]; then
   mkdir -p ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
   cp conf/janusgraph-${JANUS_PROPS_TEMPLATE}-server.properties ${JANUS_PROPS}
   cp conf/janusgraph-server.yaml ${JANUSGRAPH_SERVER_YAML}
-  chown -R "$(id -u):$(id -g)" ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
+  chown "$(id -u):$(id -g)" ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
   chmod 700 ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
   chmod -R 600 ${JANUS_CONFIG_DIR}/*
 


li-boxuan:
@FlorianHockmann tagging you as `git blame` says you are probably most familiar with this :)

FlorianHockmann (approved):
I don't know why we had this recursive flag, but it has been like that since the very first version of our Docker images: https://github.com/JanusGraph/janusgraph-docker/commit/1bb998c3efff3a23fdaf05061a2131ae8ab70bba

But I agree that it shouldn't be needed so I'm OK with removing it.

janusgraph-dist/docker/Dockerfile
```diff
@@ -73,7 +73,7 @@ RUN chmod 755 /usr/local/bin/docker-entrypoint.sh && \
     chmod 755 /usr/local/bin/load-initdb.sh && \
     chmod 755 /usr/bin/yq && \
     mkdir -p ${JANUS_INITDB_DIR} ${JANUS_CONFIG_DIR} ${JANUS_DATA_DIR} && \
-    chown -R janusgraph:janusgraph ${JANUS_HOME} ${JANUS_INITDB_DIR} ${JANUS_CONFIG_DIR} ${JANUS_DATA_DIR}
+    chown janusgraph:janusgraph ${JANUS_HOME} ${JANUS_INITDB_DIR} ${JANUS_CONFIG_DIR} ${JANUS_DATA_DIR}
```
porunov, line 76: (nitpick) Not sure if this change is needed at all. Won't it be executed only once during image build process anyways? That said, I'm OK with removing `-R` flag here if that helps anyhow.

janusgraph-dist/docker/docker-entrypoint.sh
```diff
@@ -32,7 +32,7 @@ if [ "$1" == 'janusgraph' ]; then
   mkdir -p ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
   cp conf/janusgraph-${JANUS_PROPS_TEMPLATE}-server.properties ${JANUS_PROPS}
   cp conf/janusgraph-server.yaml ${JANUSGRAPH_SERVER_YAML}
-  chown -R "$(id -u):$(id -g)" ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
+  chown "$(id -u):$(id -g)" ${JANUS_DATA_DIR} ${JANUS_CONFIG_DIR}
```
porunov, line 35: I think I agree with the solution here.
However, I wanted to notice that the configuration is going to be overwritten here anyways. So, if the user changes some configs manually then they will be lost after the re-deployment. Not sure if that's the right behavior or not, but alternatively we could skip this whole section if  `${JANUS_DATA_DIR}` is present already (or maybe just use some file flag like `isInitialized`). In such case we won't run this section after the re-deployment at all. It just some thoughts, not sure if the solution I propose is viable or not.

porunov (approved):
Thank you @tien !
I left some thoughts below, but generally it LGTM. 
I personally am not using JanusGraph Docker image, so I will rely more on reviews from @FlorianHockmann and @farodin91 here.

janusgraph-dist/docker/Dockerfile
tien: It probably doesn't make any different, cuz in this case this will always be an empty folder, right?

janusgraph-dist/docker/docker-entrypoint.sh
tien: I'm not very proficient with shell scripting 😅, so wouldn't be comfortable adding extra logic tbh. Especially since this PR just aim at a quick fix. Can help to create something better in a separate PR of course, if you think it's worth doing.

janusgraph-dist/docker/Dockerfile
porunov: `${JANUS_HOME}` is not an empty folder on this step. It does contain some files when line 76 is executed

janusgraph-dist/docker/docker-entrypoint.sh
porunov: The current solution if fine by me. I'm not sure if users of this docker image need to have an ability to change configuration files manually or not. I don't use this image, so can't speak much about how it's better to be made.

janusgraph-dist/docker/Dockerfile
tien: Okay, thanks for pointing that out, I've re-added the `-R`.

porunov (approved):

janusgraph-dist/docker/docker-entrypoint.sh
FlorianHockmann: Users have 2 ways to configure the container:
1. use environment variables like `janusgraph.storage.hostname` which will be used to create the config (documented [here](https://docs.janusgraph.org/operations/container/#docker-environment-variables))
2. mount a directory to `/etc/opt/janusgraph`.

The mount option is documented [here](https://docs.janusgraph.org/operations/container/#mounted-configuration) and it mentions:
> You'll need to bind the files as read-only, however, if you do not wish to have the environment variables override the values in that file.

Does that clear things up for you, @porunov?

I have also just built the image locally with these changes and tried out both approaches just to make sure that it doesn't break anything.

porunov: In that case the current solution in this PR is complete. Thank you for explanation.

FlorianHockmann:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|v1.0|[](https://github.com/JanusGraph/janusgraph/pull/4451)|

### Questions ?

PR #4489, opened 2024-05-30
Retry bulk request if all errors are configured retry error codes & Split large bulk requests

### LINKED ISSUE #4488: Retry Bulk Request If Bulk Request Item Fails
**Describe the feature:**
Follow up to https://github.com/JanusGraph/janusgraph/issues/4408. Since implementing that feature I've noticed bulk requests do not get retried. In particular I periodically will have `circuit_breaking_exception` error messages mentioned by the thrown IOException [here](https://github.com/JanusGraph/janusgraph/blob/487e10ca276678862fd8fb369d6d17188703ba67/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java#L429).

It appears the bulk API doesn't participate as expected with regards to `ResponseException`.

It appears it can "successfully" return a Bulk Response, that contains items within it that failed. This PR applies the retry logic originally added in https://github.com/JanusGraph/janusgraph/issues/4408 if all the failures within the bulk response set returned error codes that are configured for retry.

Circuit Breaker Exceptions in particular are coded [on the server side](https://github.com/elastic/elasticsearch/blob/db1c76bef514fa36935e14f51a9430a528f8c94e/server/src/main/java/org/elasticsearch/common/breaker/CircuitBreakingException.java#L67) as being another type of 429 Too Many Requests that would be appropriate for back pressure to be applied. Operationally some circuit breaker errors (like the parent circuit breaker) are OOM preventative measures to the calling client just saying your request _right now_ may cause an OOM, try again later.

This feature would allow bulk request item failures to be reactive to backpressure via error codes.

criminosis:
Closes #4488

Reused the retry logic from https://github.com/JanusGraph/janusgraph/pull/4409 to also apply to bulk requests if an inner bulk item request failed.

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
@@ -324,6 +324,11 @@ public class ElasticSearchIndex implements IndexProvider {
             "Comma separated list of Elasticsearch REST client ResponseException error codes to retry. " +
                 "E.g. \"408,429\"", ConfigOption.Type.LOCAL, String[].class, new String[0]);
 
+    public static final ConfigOption<Integer> BULK_CHUNK_SIZE_LIMIT_BYTES =
+        new ConfigOption<>(ELASTICSEARCH_NS, "bulk-chunk-size-limit-bytes",
+            "The total size limit in bytes of a bulk request. Mutation batches in excess of this limit will be " +
+                "chunked to this size.", ConfigOption.Type.LOCAL, Integer.class, 100_000_000);
+
     public static final int HOST_PORT_DEFAULT = 9200;
 
     /**
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
@@ -16,11 +16,15 @@ package org.janusgraph.diskstorage.es.rest;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.PeekingIterator;
 import org.apache.http.HttpEntity;
+import org.apache.http.HttpStatus;
 import org.apache.http.entity.ByteArrayEntity;
 import org.apache.http.entity.ContentType;
 import org.apache.tinkerpop.shaded.jackson.annotation.JsonIgnoreProperties;
 import org.apache.tinkerpop.shaded.jackson.core.JsonParseException;
+import org.apache.tinkerpop.shaded.jackson.core.JsonProcessingException;
 import org.apache.tinkerpop.shaded.jackson.core.type.TypeReference;
 import org.apache.tinkerpop.shaded.jackson.databind.JsonMappingException;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
@@ -39,16 +43,21 @@ import org.janusgraph.diskstorage.es.ElasticSearchMutation;
 import org.janusgraph.diskstorage.es.mapping.IndexMapping;
 import org.janusgraph.diskstorage.es.mapping.TypedIndexMappings;
 import org.janusgraph.diskstorage.es.mapping.TypelessIndexMappings;
-import org.janusgraph.diskstorage.es.rest.RestBulkResponse.RestBulkItemResponse;
 import org.janusgraph.diskstorage.es.script.ESScriptResponse;
+import org.javatuples.Pair;
+import org.javatuples.Triplet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -121,9 +130,11 @@ public class RestElasticSearchClient implements ElasticSearchClient {
 
     private final long retryMaxWaitMs;
 
+    private final int bulkChunkSerializedLimitBytes;
+
 public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean useMappingTypesForES7,
                                int retryAttemptLimit, Set<Integer> retryOnErrorCodes, long retryInitialWaitMs,
-                               long retryMaxWaitMs) {
+                               long retryMaxWaitMs, int bulkChunkSerializedLimitBytes) {
         this.delegate = delegate;
         majorVersion = getMajorVersion();
         this.scrollKeepAlive = scrollKeepAlive+"s";
@@ -134,6 +145,7 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
         this.retryOnErrorCodes = Collections.unmodifiableSet(retryOnErrorCodes);
         this.retryInitialWaitMs = retryInitialWaitMs;
         this.retryMaxWaitMs = retryMaxWaitMs;
+        this.bulkChunkSerializedLimitBytes = bulkChunkSerializedLimitBytes;
     }
 
     @Override
@@ -241,7 +253,7 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
 
             final Response response = e.getResponse();
 
-            if(e.getResponse().getStatusLine().getStatusCode() == 404){
+            if(e.getResponse().getStatusLine().getStatusCode() == HttpStatus.SC_NOT_FOUND){
                 ESScriptResponse esScriptResponse = new ESScriptResponse();
                 esScriptResponse.setFound(false);
                 return esScriptResponse;
@@ -380,10 +392,11 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
         }
     }
 
-    @Override
-    public void bulkRequest(List<ElasticSearchMutation> requests, String ingestPipeline) throws IOException {
-        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        for (final ElasticSearchMutation request : requests) {
+    private class RequestBytes {
+        final byte [] requestBytes;
+        final byte [] requestSource;
+
+        private RequestBytes(final ElasticSearchMutation request) throws JsonProcessingException {
             Map<String, Object> requestData = new HashMap<>();
             if (useMappingTypes) {
                 requestData.put("_index", request.getIndex());
@@ -398,14 +411,38 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
                 requestData.put(retryOnConflictKey, retryOnConflict);
             }
 
-            outputStream.write(mapWriter.writeValueAsBytes(
-                ImmutableMap.of(request.getRequestType().name().toLowerCase(), requestData))
-            );
-            outputStream.write(NEW_LINE_BYTES);
+            this.requestBytes =  mapWriter.writeValueAsBytes(ImmutableMap.of(request.getRequestType().name().toLowerCase(), requestData));
             if (request.getSource() != null) {
-                outputStream.write(mapWriter.writeValueAsBytes(request.getSource()));
+                this.requestSource = mapWriter.writeValueAsBytes(request.getSource());
+            } else {
+                this.requestSource = null;
+            }
+        }
+
+        private int getSerializedSize() {
+            int serializedSize = this.requestBytes.length;
+            serializedSize+= 1; //For follow-up NEW_LINE_BYTES
+            if (this.requestSource != null) {
+                serializedSize += this.requestSource.length;
+                serializedSize+= 1; //For follow-up NEW_LINE_BYTES
+            }
+            return serializedSize;
+        }
+
+        private void writeTo(OutputStream outputStream) throws IOException {
+            outputStream.write(this.requestBytes);
+            outputStream.write(NEW_LINE_BYTES);
+            if (this.requestSource != null) {
+                outputStream.write(requestSource);
                 outputStream.write(NEW_LINE_BYTES);
             }
+        }
+    }
+
+    private Pair<String, byte[]> buildBulkRequestInput(List<RequestBytes> requests, String ingestPipeline) throws IOException {
+        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        for (final RequestBytes request : requests) {
+            request.writeTo(outputStream);
         }
 
         final StringBuilder builder = new StringBuilder();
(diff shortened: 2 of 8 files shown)


criminosis:
Had an idea for an improvement this morning to only retry the bulk items that returned an error. This is predicated upon the [Bulk API docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-response-body) stating the response's returned bulk item order is the same as the submission order.

> The bulk API’s response contains the individual results of each operation in the request, returned in the order submitted. The success or failure of an individual operation does not affect other operations in the request.

Wasn't sure if that would be preferable though, so I left it as an [unsquashed addition to the PR](https://github.com/JanusGraph/janusgraph/pull/4489/commits/5c2b132847d3d352e6a4af89475c8bf81e279c2a) if it'd be preferable to revert it. If it looks good happy to squash it down to true up the PR for merging.

criminosis:
Added a test for bulk retry in 81b07ac68.

While I was working on that though I noticed on the ES docs state there is a size limit for bulk requests:

>There is no "correct" number of actions to perform in a single bulk request. Experiment with different settings to find the optimal size for your particular workload. Note that Elasticsearch limits the maximum size of a HTTP request to 100mb by default ...

I've bumped into this in my JG deployment, ES pushes back saying the Payload is too large. I didn't have a meaningful knob to turn from what I could tell so I opted to work around the issue by more aggressively splitting up calls on my JG client side. So since I was already working in the bulk request logic I implemented a simple request chunker.

If a bulk request has a payload that is too large for a single call, it'll split it up into multiple bulk request calls.

criminosis:
@li-boxuan / @porunov might interest you since you reviewed the last one? I accidentally left the job half finished. Didn't realize bulk request errors were nested into the item level.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -416,17 +452,107 @@ public void bulkRequest(List<ElasticSearchMutation> requests, String ingestPipel
...
+        final List<Map<String, RestBulkResponse.RestBulkItemResponse>> bulkResponseItems,
+        final List<RequestBytes> submittedBulkRequestItems) {
+        final List<Triplet<Object, Integer, RequestBytes>> errors = new ArrayList<>(bulkResponseItems.size());
+        for (int itemIndex = 0; itemIndex < bulkResponseItems.size(); itemIndex++) {
+            Collection<RestBulkResponse.RestBulkItemResponse> bulkResponseItem = bulkResponseItems.get(itemIndex).values();
+            if (bulkResponseItem.size() > 1) {
+                throw new IllegalStateException("There should only be a single item per bulk reponse item entry");
+            }
+            RestBulkResponse.RestBulkItemResponse item = bulkResponseItem.iterator().next();
+            if (item.getError() != null && item.getStatus() != 404) {
```
li-boxuan, line 471: (nit) can we avoid explicitly using `404`? ES library should have a constant variable for 404.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -416,17 +452,107 @@ public void bulkRequest(List<ElasticSearchMutation> requests, String ingestPipel
...
+                        //Only retry the bulk request if *all* the bulk response item error codes are retry error codes
+                        final Set<Integer> errorCodes = bulkItemsThatFailed.stream().map(Triplet::getValue1).collect(Collectors.toSet());
+                        if (retryCount < retryAttemptLimit && retryOnErrorCodes.containsAll(errorCodes)) {
+                            //Build up the next request batch, of only the failed mutations
+                            bulkRequestChunk = bulkItemsThatFailed.stream().map(Triplet::getValue2).collect(Collectors.toList());
+                            performRetryWait(retryCount);
+                            retryCount++;
+                        } else {
+                            final List<Object> errorItems = bulkItemsThatFailed.stream().map(Triplet::getValue0).collect(Collectors.toList());
+                            errorItems.forEach(error -> log.error("Failed to execute ES query: {}", error));
```
li-boxuan, line 548: You don't need `{}` here
```suggestion
                            errorItems.forEach(error -> log.error("Failed to execute ES query:", error));
```

li-boxuan (commented):
Thank you! This looks good to me in general! I wonder if you could add an integration test to `ElasticsearchJanusGraphIndexTest` that demonstrates this chunking behavior? You could set BULK_CHUNK_SIZE_LIMIT_BYTES to different numbers to test different scenario. I would imagine the following:

1) A very small `BULK_CHUNK_SIZE_LIMIT_BYTES` that would fail the request - demonstrating the `we've encountered an element we cannot send to Elasticsearch given the configured limit` scenario
2) A relatively small `BULK_CHUNK_SIZE_LIMIT_BYTES` that would make the request split into chunks. We probably have no way to enforce that in the test (we lack observability) but I guess we could at least use IDE debugger to verify that manually.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: No objection from me it's just there because [it's what was there previously](https://github.com/JanusGraph/janusgraph/blob/e30b4d6ed84e243eddbc726e4304edd1091c3eb3/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java#L425).

Happy to change it. That's being compared against something from the Apache Http library, and it does provide status code constants like `HttpStatus.SC_NOT_FOUND`.

Switched in 09891809e

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
li-boxuan: Oh damn sorry for that. I thought it was an exception

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
li-boxuan: Oh I didn’t notice that. Thx for polishing it!

criminosis:
> A very small BULK_CHUNK_SIZE_LIMIT_BYTES that would fail the request - demonstrating the we've encountered an element we cannot send to Elasticsearch given the configured limit scenario

@li-boxuan I don't think it's possible to write such a test because JanusGraph doesn't treat such a scenario as a failure [in code](https://github.com/JanusGraph/janusgraph/blob/e30b4d6ed84e243eddbc726e4304edd1091c3eb3/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java#L1022-L1028) or in [documentation](https://docs.janusgraph.org/operations/recovery/#transaction-failure).

> If the primary persistence into the storage backend succeeds but secondary persistence into the indexing backends or the logging system fail, the transaction is still considered to be successful because the storage backend is the authoritative source of the graph.

So the response from the traversal comes back as a success. I guess I could assert that a query that uses the mixed index stored in ES fails to return the vertex because now that index is blind to the property...but that feels like a weird thing to test for, albeit it is congruent with intended behavior.

I've bumped into this asymmetric situation when I can query using an equality predicate (thus uses the data backend's composite index) returns the expected vertex, but because the write to the mixed index failed (successfully) if you use like a range predicate over the same value, it won't find the vertex.


> A relatively small BULK_CHUNK_SIZE_LIMIT_BYTES that would make the request split into chunks. We probably have no way to enforce that in the test (we lack observability) but I guess we could at least use IDE debugger to verify that manually.

Took a stab at this in af88b6a0a, set it as `@Disabled` so it doesn't contribute to the GH execution time given it has no assertions.

criminosis:
@li-boxuan sounded like there wasn't anything objectionable with the additions, so I went ahead and squashed & signed off on the commit so the PR was back in a merge ready state again pending further feedback.

li-boxuan:
> I guess I could assert that a query that uses the mixed index stored in ES fails to return the vertex because now that index is blind to the property...

A much more cumbersome is to read the transaction recovery log... but I won't set this as acceptance criteria.

I do think what you mentioned here seems promising. Doesn't seem weird to me :) we probably already have tests doing something similar.

criminosis:
@li-boxuan wrote `writingAnItemLargerThanPermittedChunkLimitFails` as requested to assert the expected data asymmetry between the composite and mixed index if an item exceeds the permitted bulk chunk limit.  

An alternative idea I had was rather than abort the bulk write at the time that is observed would be to maybe still write the other elements that aren't in excess of the limit. Get in at least what _could_ go in rather than have that data be lost too.

In other words, given a batch of mutations of an exceptionally large mutation followed by some number of small ones, fail the first, but don't throw an exception until later.

li-boxuan:
> In other words, given a batch of mutations of an exceptionally large mutation followed by some number of small ones, fail the first, but don't throw an exception until later.

i like that idea! Do you wanna open a follow-up PR for that 😉? Maybe a config to control ;)

criminosis:
Yeah, that'd probably be best done as a followup PR.

li-boxuan (approved):
Thank you for continuously investing in this field!

criminosis:
Happy to @li-boxuan. JanusGraph has been a critical piece to what I'm working on, hence why I've been bumping into these sharp edges 😅 .

Looking forward to getting a nightly build into my environment and put this change through its paces after whenever @porunov has time to give their review.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -416,17 +453,107 @@ public void bulkRequest(List<ElasticSearchMutation> requests, String ingestPipel
             APPEND_OP.apply(builder).append("refresh=").append(bulkRefresh);
         }
         builder.insert(0, REQUEST_SEPARATOR + "_bulk");
```
porunov, line 455: Not related to your PR, but it seems it makes sense to move this line up (to be the next line after `StringBuilder` is created) to avoid array shifting operation. I.e. Just:
```
final StringBuilder builder = new StringBuilder();
builder.append(REQUEST_SEPARATOR).append("_bulk");
```

porunov (approved):
Thank you @criminosis for this great contribution and sorry for the late review.
This PR looks great to me. I have added small nitpicks, but I think it's OK to fix them in follow up PRs.

criminosis:
@porunov I'd be happy to do this in the follow-up you mentioned. There's already a tentative PR in mind that I was talking about with @li-boxuan above.

Since you mentioned them as a follow-up is this PR valid for merging? I saw you tagged it for 1.1.0 so wasn't sure if that meant it was going to hang out unmerged for a period of time since there's also a live 1.0.1 milestone.

porunov:
@criminosis Just merged this PR into ‘master’ branch. Thus, these changes will be available in 1.1.0 release and in the following “commit” releases. However, it won’t be backported to 1.0.1 release since it’s based on the feature which isn’t available in 1.0.0 release. If you want to port these changes to 1.0.1 release for any reason then feel free to open a PR targeting ‘v1.0’ branch with these changes and your previous changes.

Allan-Clements:
Thanks @porunov! Having it 1.1.0 and the next commit release is perfectly fine. I was asking just as curiosity around merging logistics.

Excited to get this into my environment soon as I've been bumping into what prompted this PR fairly often.

I'll get rolling on the follow-ups. 👍

criminosis:
Thanks @porunov! Having it 1.1.0 and the next commit release is perfectly fine. I was asking just as curiosity around merging logistics.

Excited to get this into my environment soon as I've been bumping into what prompted this PR fairly often.

I'll get rolling on the follow-ups. 👍

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: Looked into this this for my follow-up, but it seems like this would change behavior?

https://github.com/JanusGraph/janusgraph/blob/d255396dcddf5c19317a873ff1ce16e31f77a304/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java#L105

`APPEND_OP` either prefixes a `?` or a `&` if the builder is empty. So as currently written I believe it's producing paths like `/_bulk?pipeline=foo&refresh=true`.

If we put `builder.append(REQUEST_SEPARATOR).append("_bulk");` further up, I think it'd produce paths like `/_bulk&pipeline=foo&refresh=true` which wouldn't be properly formatting the query parameters anymore.

How does this look if we just use the string builder for the query parameters, and then just directly build the path? I originally used another StringBuilder for the path variable, but IntelliJ's linter suggested just doing String concatenations.

```
final StringBuilder bulkRequestQueryParameters = new StringBuilder();
        if (ingestPipeline != null) {
            APPEND_OP.apply(bulkRequestQueryParameters).append("pipeline=").append(ingestPipeline);
        }
        if (bulkRefreshEnabled) {
            APPEND_OP.apply(bulkRequestQueryParameters).append("refresh=").append(bulkRefresh);
        }
        final String bulkRequestPath = REQUEST_SEPARATOR + "_bulk" + bulkRequestQueryParameters;
        return Pair.with(bulkRequestPath, outputStream.toByteArray());
```

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
```diff
@@ -324,6 +324,11 @@ public class ElasticSearchIndex implements IndexProvider {
             "Comma separated list of Elasticsearch REST client ResponseException error codes to retry. " +
                 "E.g. \"408,429\"", ConfigOption.Type.LOCAL, String[].class, new String[0]);
 
+    public static final ConfigOption<Integer> BULK_CHUNK_SIZE_LIMIT_BYTES =
+        new ConfigOption<>(ELASTICSEARCH_NS, "bulk-chunk-size-limit-bytes",
+            "The total size limit in bytes of a bulk request. Mutation batches in excess of this limit will be " +
+                "chunked to this size.", ConfigOption.Type.LOCAL, Integer.class, 100_000_000);
+
```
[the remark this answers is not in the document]
criminosis:
```
 public static final ConfigOption<Integer> BULK_CHUNK_SIZE_LIMIT_BYTES =
        new ConfigOption<>(ELASTICSEARCH_NS, "bulk-chunk-size-limit-bytes",
            "The total size limit in bytes of a bulk request. Mutation batches in excess of this limit will be " +
                "chunked to this size. Ensure that this limit is always less than or equal to the configured limit of " +
                "`http.max_content_length` on the Elasticsearch servers. For more information, refer to the " +
                "[Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/master/modules-network.html).",
            ConfigOption.Type.LOCAL, Integer.class, 100_000_000);
```

Got it ready for the follow-up👍

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
porunov: Oh. I missed that! I'm OK with leaving it as it is right now or updating it to the way you are suggesting. Both ways are OK by me.
btw. Usually `final` String concatenation is directly compiled to reuse `StringBuilder` by all modern Java compilers, so I believe that's why IntelliJ's suggested just using String concatenation here.
Anyways, the shift is probably too small to actually produce any meaningful performance differences, so I'm OK with either way as you prefer.

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/ElasticSearchIndex.java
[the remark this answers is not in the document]
porunov: Thanks! Would you be able to also add here or somewhere in the documentation description of what happens when part of the request produces a chunk which is grater than the one specified here?
I think this information will let users know that it's dangerous to reduce this chunk size to small sizes.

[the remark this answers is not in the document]
criminosis: Added `If a single bulk item exceeds this limit an exception will be thrown after the smaller bulk items are submitted.`  to the follow-up PR 👍 .

Figured we can continue the conversation of any further tweaking over there.

PR #4506, opened 2024-06-11
Disable vertex cache for cardinality checks

cdegroc:
In a service running JanusGraph embedded, I observed a number of sizeable allocations coming from [JanusGraphVertexFeatures#getCardinality](https://github.com/JanusGraph/janusgraph/blob/2c71b378339a3ab49b961eef29b5a042d018f513/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphFeatures.java#L161-L169).

From a profile (included below), it looks like a `StandardJanusGraphTx` transaction is created when [checking a vertex property's cardinality](https://github.com/JanusGraph/janusgraph/blob/2c71b378339a3ab49b961eef29b5a042d018f513/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphFeatures.java#L162). The transaction itself [allocates a `CaffeineVertexCache`](https://github.com/JanusGraph/janusgraph/blob/2c71b378339a3ab49b961eef29b5a042d018f513/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java#L335). The `CaffeineVertexCache` internally [creates a `NonBlockingHashMap`](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/vertexcache/CaffeineVertexCache.java#L44) which initial size is configured in the [`GraphDatabaseConfiguration` class](https://github.com/JanusGraph/janusgraph/blob/2c71b378339a3ab49b961eef29b5a042d018f513/janusgraph-core/src/main/java/org/janusgraph/graphdb/configuration/GraphDatabaseConfiguration.java#L1613-L1620) or the graph's dynamic configuration.

I think we don't need a vertex cache for cardinality checks and would suggest disabling it to release some memory.

Screenshot 2024-06-11 at 10 20 37

-----

### For all changes:

### For code changes:

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphFeatures.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphFeatures.java
@@ -159,7 +159,11 @@ public class JanusGraphFeatures implements Graph.Features {
 
         @Override
         public VertexProperty.Cardinality getCardinality(final String key) {
-            StandardJanusGraphTx tx = (StandardJanusGraphTx)JanusGraphFeatures.this.graph.newTransaction();
+            StandardJanusGraphTx tx = (StandardJanusGraphTx)JanusGraphFeatures.this.graph.buildTransaction()
+                .dirtyVertexSize(0)
+                .vertexCacheSize(0)
+                .readOnly()
+                .start();
             try {
                 if (!tx.containsPropertyKey(key)) return tx.getConfiguration().getAutoSchemaMaker().defaultPropertyCardinality(key).convert();
                 return tx.getPropertyKey(key).cardinality().convert();


li-boxuan (approved):
Great finding, thank you!

porunov (approved):
I usually have a default vertex cache size and dirty vertex size set to 0 and set a proper value per transaction when needed, thus I haven't bumped into this issue before.
However, the solution here perfectly makes sense to me as the transaction is closed immediately after we exiting this method.
We don't create any new mutations, thus dirty vertex size should be 0 and read-only transaction make sense. 
We return a single property key, and don't reuse this transaction for anything else, thus vertex cache size 0 makes sense here.
LGTM. Thank you @cdegroc !

cdegroc:
> I usually have a default vertex cache size and dirty vertex size set to 0 and set a proper value per transaction when needed, thus I haven't bumped into this issue before. However, the solution here perfectly makes sense to me as the transaction is closed immediately after we exiting this method. We don't create any new mutations, thus dirty vertex size should be 0 and read-only transaction make sense. We return a single property key, and don't reuse this transaction for anything else, thus vertex cache size 0 makes sense here. LGTM. Thank you @cdegroc !

Thank you so much for confirming @porunov 🙇

PR #4530, opened 2024-06-25
ES Bulk Retry Follow-up Items

### LINKED ISSUE #4529: ES Bulk Retry Follow-up Items
**Describe the feature:**
Implementing the follow-up items requested in https://github.com/JanusGraph/janusgraph/pull/4489. Namely:
- Reference the ElasticSearch documentation in the bulk chunker size limit config option
- Cleaning up the population of the ES bulk request path logic
- Submitting bulk request items that are below the configured limit, then throwing for overly large items

N/A

criminosis:
Closes #4529

- Reference the ElasticSearch documentation in the bulk chunker size limit config option
- Cleaning up the population of the ES bulk request path logic
- Submitting bulk request items that are below the configured limit, then throwing for overly large items

-----

### For all changes:

### For code changes:

- [X] Have you ensured that format looks appropriate for the output in which it is rendered?

### DIFF
--- a/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
+++ b/janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
@@ -14,6 +14,7 @@
 
 package org.janusgraph.diskstorage.es.rest;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterators;
@@ -54,6 +55,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -392,11 +394,13 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
         }
     }
 
-    private class RequestBytes {
+    @VisibleForTesting
+    class RequestBytes {
         final byte [] requestBytes;
         final byte [] requestSource;
 
-        private RequestBytes(final ElasticSearchMutation request) throws JsonProcessingException {
+        @VisibleForTesting
+        RequestBytes(final ElasticSearchMutation request) throws JsonProcessingException {
             Map<String, Object> requestData = new HashMap<>();
             if (useMappingTypes) {
                 requestData.put("_index", request.getIndex());
@@ -419,7 +423,8 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
             }
         }
 
-        private int getSerializedSize() {
+        @VisibleForTesting
+        int getSerializedSize() {
             int serializedSize = this.requestBytes.length;
             serializedSize+= 1; //For follow-up NEW_LINE_BYTES
             if (this.requestSource != null) {
@@ -445,15 +450,15 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
             request.writeTo(outputStream);
         }
 
-        final StringBuilder builder = new StringBuilder();
+        final StringBuilder bulkRequestQueryParameters = new StringBuilder();
         if (ingestPipeline != null) {
-            APPEND_OP.apply(builder).append("pipeline=").append(ingestPipeline);
+            APPEND_OP.apply(bulkRequestQueryParameters).append("pipeline=").append(ingestPipeline);
         }
         if (bulkRefreshEnabled) {
-            APPEND_OP.apply(builder).append("refresh=").append(bulkRefresh);
+            APPEND_OP.apply(bulkRequestQueryParameters).append("refresh=").append(bulkRefresh);
         }
-        builder.insert(0, REQUEST_SEPARATOR + "_bulk");
-        return Pair.with(builder.toString(), outputStream.toByteArray());
+        final String bulkRequestPath = REQUEST_SEPARATOR + "_bulk" + bulkRequestQueryParameters;
+        return Pair.with(bulkRequestPath, outputStream.toByteArray());
     }
 
     private List<Triplet<Object, Integer, RequestBytes>> pairErrorsWithSubmittedMutation(
@@ -476,7 +481,8 @@ public RestElasticSearchClient(RestClient delegate, int scrollKeepAlive, boolean
         return errors;
     }
 
-    private class BulkRequestChunker implements Iterator<List<RequestBytes>> {
+    @VisibleForTesting
+    class BulkRequestChunker implements Iterator<List<RequestBytes>> {
         //By default, Elasticsearch writes are limited to 100mb, so chunk a given batch of requests so they stay under
         //the specified limit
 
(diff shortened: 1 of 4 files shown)


criminosis:
@li-boxuan & @porunov follow-up PR as requested 👍

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -485,18 +491,32 @@ private class BulkRequestChunker implements Iterator<List<RequestBytes>> {
...
+                if (requestSerializedSize <= bulkChunkSerializedLimitBytes) {
+                    //Only keep items that we can actually send in memory
+                    serializedRequests.add(requestBytes);
+                } else {
+                    requestSizesThatWereTooLarge.add(requestSerializedSize);
+                }
             }
             this.requestIterator = Iterators.peekingIterator(serializedRequests.iterator());
+            //Condense request sizes that are too large into an int array to remove Boxed & List memory overhead
+            this.exceptionallyLargeRequests = requestSizesThatWereTooLarge.stream().mapToInt(Integer::intValue).toArray();
```
porunov, line 512: Not sure if it optimizes heap usage because we are essentially creating a duplicate array in heap. That said, it seems to me that such a case is really exceptional and normally we never face this case. Thus, maybe we don't even need to create these array in heap (unless we really have to?).
I.e. I would simply do something like:
```
List<RequestBytes> serializedRequests = new ArrayList<>(requests.size());
List<Integer> requestSizesThatWereTooLarge = null;
for (ElasticSearchMutation request : requests) {
                RequestBytes requestBytes = new RequestBytes(request);
                int requestSerializedSize = requestBytes.getSerializedSize();
                if (requestSerializedSize <= bulkChunkSerializedLimitBytes) {
                    //Only keep items that we can actually send in memory
                    serializedRequests.add(requestBytes);
                } else {
                    // it is better to not initialize `requestSizesThatWereTooLarge` in advance because normally it's never going to be used during proper queries usage.
                    if(requestSizesThatWereTooLarge == null){
                        requestSizesThatWereTooLarge = new ArrayList<>();
                    }
                    requestSizesThatWereTooLarge.add(requestSerializedSize);
                }
 }
 this.requestIterator = Iterators.peekingIterator(serializedRequests.iterator());
 //Condense request sizes that are too large into an int array to remove Boxed & List memory overhead
 this.exceptionallyLargeRequests = requestSizesThatWereTooLarge == null ? null : requestSizesThatWereTooLarge.stream().mapToInt(Integer::intValue).toArray();
```

After that below instead of using `exceptionallyLargeRequests.length > 0` we can use `exceptionallyLargeRequests != null`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
```diff
@@ -509,16 +529,18 @@ public List<RequestBytes> next() {
...
                     serializedRequests.add(requestIterator.next());
-                } else if (requestSerializedSize > bulkChunkSerializedLimitBytes) {
-                    //we've encountered an element we cannot send to Elasticsearch given the configured limit
-                    throw new IllegalArgumentException(String.format(
-                        "Bulk request item is larger than permitted chunk limit. Limit is %s. Serialized item size was %s",
-                        bulkChunkSerializedLimitBytes, requestSerializedSize));
                 } else {
                     //Adding this element would exceed the limit, so return the chunk
                     return serializedRequests;
                 }
```
porunov, line 535: Seems we can avoid duplicate arithmetical operation: 
```
                chunkSerializedTotal += peeked.getSerializedSize();
                if (chunkSerializedTotal <= bulkChunkSerializedLimitBytes) {
                    serializedRequests.add(requestIterator.next());
                } else {
                    //Adding this element would exceed the limit, so return the chunk
                    return serializedRequests;
                }
```

porunov (commented):
Thank you @criminosis ! I have just some small nitpicks regarding tiny optimizations

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: First and foremost I agree premature optimizations can quickly become evil costing readability and complexity and should be done judiciously.

The reason why I opted for it here was to allow the boxed forms to be collected as soon as the scope returns since they're so much heavier than their non-boxed form if it is truly the value we need and nothing about the boxed form is beneficial for Object purposes. But it's true, it should be very unlikely people would try to cause this to ever trip.

> Not sure if it optimizes heap usage because we are essentially creating a duplicate array in heap.

Setting aside the duplicate array for a moment, it should be more compact was the idea and the original array list should quickly be ripe for collection. We likely won't get any interning benefits for ints between -128 and 127 since these, at least by default, should be > 100MB.

Assuming a 64 bit JVM, iirc, a Boxed int is 24 bytes vs an unboxed int being just 4. Then there's the overhead of the list's extra wasted capacity, etc which is why I moved it into a fixed size array.

It looks like what you're suggesting is better in that it puts off the arraylist instantiation until it is used. Fwiw under the covers ArrayList is somewhat of a lazy instantiation of its internals, setting its state to point to an empty static global array instance until it's actually used. I found this interesting to know of "behind the curtains" and since you also knew about javac's optimization of `final String` to StringBuilder under the covers maybe you'll find this interesting too. I thought about mentioning that but for those that don't know about that it seems to cause more doubt than to just say "IntelliJ suggested this"  and appeal to the industry respect of Jet Brains😅 .

So you get that benefit somewhat in my code, admittedly with the overhead of an object header still but it won't allocate another array on the heap until the array list is actually used.

- https://github.com/openjdk/jdk/blob/57f8b91e558e5b9ff9c2000b8f74e3a1988ead2b/src/java.base/share/classes/java/util/ArrayList.java#L133-L139
- (Added in 2014) https://github.com/openjdk/jdk/commit/e0c3c7a07de23cca0ee938eb64af973d567fcdaf
- https://blogs.oracle.com/javamagazine/post/for-faster-java-collections-make-them-lazy

The main gain in your suggestion seems to be it ends with `exceptionallyLargeRequests ` being null if `requestSizesThatWereTooLarge ` is null by the end. That could be done without the extra branch for checking if `requestSizesThatWereTooLarge ` is empty but instead of assigning an empty array, just assigning it null and doing a null check for the `int[]` too.

How's this look instead?

`this.exceptionallyLargeRequests = requestSizesThatWereTooLarge.isEmpty() ? null : requestSizesThatWereTooLarge.stream().mapToInt(Integer::intValue).toArray();`

And then below we'd have to do:

In `hasNext()`: `return requestIterator.hasNext() || exceptionallyLargeRequests != null;`

Then in `next()`:  `if (serializedRequests.isEmpty() && this.exceptionallyLargeRequests != null) {`

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: Changed in the latest force push

janusgraph-es/src/main/java/org/janusgraph/diskstorage/es/rest/RestElasticSearchClient.java
criminosis: Went ahead and force pushed that up, on second thought figured that was easier to see than the lines being stated in a vacuum.

porunov: Thanks a lot for diving into `ArrayList` detailed implementation. I actually didn't know (or maybe forgot) that it has lazy initialization of underlying array, so that's definitely helpful! I like the solution you proposed here.

porunov (approved):
LGTM. Thank you @criminosis for a quick follow up PR!

criminosis:
@li-boxuan maybe worth your attention too since one of the follow-ups was your request?

li-boxuan (approved):
This is awesome, thank you!

PR #4689, opened 2024-10-08
Replace deprecated compression options in cql

To-om:
Closes discussion #4688 

-----
### For all changes:
- [x] Is there an ~issue~ discussion associated with this PR? Is it referenced in the commit message?
- [x] Does your PR body contain #xyz where xyz is the ~issue~ discussion number you are trying to resolve?

### For code changes:

### DIFF
--- a/docs/configs/janusgraph-cfg.md
+++ b/docs/configs/janusgraph-cfg.md
@@ -437,7 +437,7 @@ CQL storage backend options
 | storage.cql.compaction-strategy-options | Compaction strategy options.  This list is interpreted as a map.  It must have an even number of elements in [key,val,key,val,...] form. | String[] | (no default value) | FIXED |
 | storage.cql.compression | Whether the storage backend should use compression when storing the data | Boolean | true | FIXED |
 | storage.cql.compression-block-size | The size of the compression blocks in kilobytes | Integer | 64 | FIXED |
-| storage.cql.compression-type | The sstable_compression value JanusGraph uses when creating column families. This accepts any value allowed by Cassandra's sstable_compression option. Leave this unset to disable sstable_compression on JanusGraph-created CFs. | String | LZ4Compressor | MASKABLE |
+| storage.cql.compression-type | The compression class value JanusGraph uses when creating column families. This accepts any value allowed by Cassandra's compression class option. Leave this unset to disable compression on JanusGraph-created CFs. | String | LZ4Compressor | MASKABLE |
 | storage.cql.gc-grace-seconds | The number of seconds before tombstones (deletion markers) are eligible for garbage-collection. | Integer | (no default value) | FIXED |
 | storage.cql.heartbeat-interval | The connection heartbeat interval in milliseconds. | Long | (no default value) | MASKABLE |
 | storage.cql.heartbeat-timeout | How long the driver waits for the response (in milliseconds) to a heartbeat. | Long | (no default value) | MASKABLE |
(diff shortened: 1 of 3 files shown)


To-om:
ScyllaDB still uses the old options from Cassandra 2. Therefore, two different table initialization methods are needed to account for this difference.

The incompatibility of compression options has been reported here:
 - scylladb/scylladb#4200 
 - scylladb/scylladb/issues/2061

janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -331,7 +331,7 @@ private static CreateTableWithOptions compressionOptions(final CreateTableWithOp
         int chunkLengthInKb = configuration.get(CF_COMPRESSION_BLOCK_SIZE);
 
         return createTable.withOption("compression",
-            ImmutableMap.of("sstable_compression", compressionType, "chunk_length_kb", chunkLengthInKb));
+            ImmutableMap.of("class", compressionType, "chunk_length_in_kb", chunkLengthInKb));
```
porunov, line 334: Some of the solutions for compatibility I think:
1. We can fetch Cassandra version before creating any tables ([here](https://java-driver.docs.scylladb.com/stable/manual/core/#quick-start)): `select release_version from system.local`. I didn't check if ScyllaDB returns different versioning then Cassandra. If it returns only numbers then it's not useful too much, but if it returns something that defines it uniquely then we can determine if we are working with Cassandra 3, Cassandra 4, Cassandra 5, or ScyllaDB. In this case, if it's Cassandra and the version is >= 5 then we use new definition. Otherwise, we use previous definition.
2. We can provide an optional parameter which forces usage of one or another naming regardless of what version we are working with. If so, users can specify if they want to use `sstable_compression` or `class`. As such, they can work with Cassandra 5 or any other storage which has new names for options.

To-om: The latest version of ScyllaDB (6.2) returns the 3.0.8 when requesting the `release_version`. Selecting the right compression options based on this information seems to be a good option.

docs/configs/janusgraph-cfg.md
```diff
@@ -437,7 +437,7 @@ CQL storage backend options
 | storage.cql.compaction-strategy-options | Compaction strategy options.  This list is interpreted as a map.  It must have an even number of elements in [key,val,key,val,...] form. | String[] | (no default value) | FIXED |
 | storage.cql.compression | Whether the storage backend should use compression when storing the data | Boolean | true | FIXED |
 | storage.cql.compression-block-size | The size of the compression blocks in kilobytes | Integer | 64 | FIXED |
-| storage.cql.compression-type | The sstable_compression value JanusGraph uses when creating column families. This accepts any value allowed by Cassandra's sstable_compression option. Leave this unset to disable sstable_compression on JanusGraph-created CFs. | String | LZ4Compressor | MASKABLE |
+| storage.cql.compression-type | The compression class value JanusGraph uses when creating column families. This accepts any value allowed by Cassandra's compression class option. Leave this unset to disable compression class on JanusGraph-created CFs. | String | LZ4Compressor | MASKABLE |
```
porunov, line 440: build-doc CI fails because in the last sentence here you specified `compression class` but in configuration option only `compression` is specified. You can execute `mvn --quiet clean install -DskipTests=true -pl janusgraph-doc -am` to automatically re-generate this file, so that `compression class` automatically turns into `compression`.

porunov (commented):
Thank you @To-om ! Looks good to me. I have one commend below. Also, could you please squash your commits into a single commit?

porunov (approved):
LGTM. Thank you @To-om !

cdegroc (approved):
Thanks for contributing that change (+ in a backward compatible way!).

PR #4726, opened 2024-11-15
Reindex subset of vertices

ntisseyre:
### Summary
This PR introduces a significant optimization to the reindexing process in `JanusGraph` by allowing a subset of vertices to be reindexed instead of scanning the entire storage. 
This enhancement provides substantial performance improvements, primarily when the specific subset of vertices for indexing is already known.

### NOTE
This feature is currently supported only for `CQL` storage. Other storage backends still need to be implemented.

KeyColumnValueStore.java
```
KeyIterator getKeys(final List<StaticBuffer> keys, final SliceQuery query, final StoreTransaction txh) throws BackendException {
        throw new NotImplementedException();
    }
```

#### Motivation
Previously, reindexing required scanning all vertices in storage, which could be highly resource-intensive and time-consuming, particularly in large datasets. 
This update enables users to focus on a targeted subset of vertices, reducing the time and computational load for reindexing. This is especially beneficial in environments where only specific vertices are relevant to a given index or data update.

#### Changes

- Added the ability to specify a subset of vertices to include in the reindexing process.
- Optimized the indexing engine to skip unnecessary vertices, focusing only on those specified in the subset.

API in `JanusGraphManagement`

```
  /**
     * Updates the provided index according to the given {@link SchemaAction} for
     * the given subset of vertices.
     *
     * @param index
     * @param updateAction
     * @param vertexOnly Set of vertexIds that only should be considered for index update
     * @return a future that completes when the index action is done
     */
ScanJobFuture updateIndex(Index index, SchemaAction updateAction, List<Object> vertexOnly);
```

#### Benefits

- Improved Performance: By narrowing down the scope of vertices, the reindexing process is much faster and more efficient.
- Resource Optimization: Reduces CPU and memory usage during reindexing by avoiding a full scan.
Enhanced Flexibility: This feature allows users to update specific sections of the graph more easily without impacting the entire dataset.

### Backward Compatibility
This feature is backward compatible and does not impact existing functionality. Users not specifying a subset will still experience the previous behavior of scanning the entire storage.


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLSubsetIterator.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.common.collect.Iterables;
+import org.janusgraph.diskstorage.StaticBuffer;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Function;
+
+public class CQLSubsetIterator<TItem> implements Iterator<TItem> {
```
porunov, line 26: (nitpick)
Codacy suggests:
```
Generics names should be a one letter long and upper case.
```

I usually use all upper case, but not always one letter.

ntisseyre: fixed

porunov (approved):
LGTM. Thank you @ntisseyre !

PR #4754, opened 2025-01-23
Schema Management APIs for Enhanced Debugging in JanusGraph

ntisseyre:
This PR introduces new schema management APIs to JanusGraph to enhance debugging capabilities, especially for issues where information is derived from logs. These APIs provide utilities to map between vertex IDs and their hex string representation, retrieve index key hex strings, and obtain index information using hex strings.

The added methods are:

`getVertexId(String hexString)`

```
Purpose: Converts a hex string back to its corresponding vertex ID.
Parameters:
hexString: A hexadecimal string representing the vertex ID.
Returns:
Vertex ID as an object.

```

`getIndexKey(String indexName, Map<String, Object> fieldValues)`

```
Purpose: Generates a hex string representing a schema index's index key bytes for given index name and vertex properties values.
Parameters:
indexName: The name of the schema index.
fieldValues: A map of index fields and their corresponding values.
Returns:
A hex string representing the index key.
```

`getIndexInfo(String hexString)`

```
Purpose: Decodes a hex string to retrieve composite index information.
Parameters:
hexString: A hexadecimal string representing the index key.
Returns:
CompositeIndexType: The composite index details.
```

### Key Benefits
Improved Debugging: Facilitates troubleshooting by enabling a direct relationship between logs and schema elements.
Enhanced Index Management: Allows index information to be decoded for better schema debugging and analysis.

### Usage Example
1. Given a hex string from a log file, use getVertexId() to identify the associated vertex.
2. Decode the hex string with getIndexInfo() to retrieve comprehensive details about the composite index.

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
@@ -16,6 +16,8 @@ package org.janusgraph.graphdb.database.management;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterables;
+import org.apache.commons.codec.DecoderException;
+import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.tinkerpop.gremlin.structure.Edge;
@@ -51,6 +53,7 @@ import org.janusgraph.core.schema.SchemaAction;
 import org.janusgraph.core.schema.SchemaStatus;
 import org.janusgraph.core.schema.VertexLabelMaker;
 import org.janusgraph.diskstorage.BackendException;
+import org.janusgraph.diskstorage.StaticBuffer;
 import org.janusgraph.diskstorage.configuration.BasicConfiguration;
 import org.janusgraph.diskstorage.configuration.ConfigOption;
 import org.janusgraph.diskstorage.configuration.ModifiableConfiguration;
@@ -62,6 +65,7 @@ import org.janusgraph.diskstorage.keycolumnvalue.scan.ScanJobFuture;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.ScanMetrics;
 import org.janusgraph.diskstorage.keycolumnvalue.scan.StandardScanner;
 import org.janusgraph.diskstorage.log.Log;
+import org.janusgraph.diskstorage.util.StaticArrayBuffer;
 import org.janusgraph.graphdb.database.IndexSerializer;
 import org.janusgraph.graphdb.database.StandardJanusGraph;
 import org.janusgraph.graphdb.database.cache.SchemaCache;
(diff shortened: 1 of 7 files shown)


janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -545,6 +549,23 @@ public String printIndexes() {
         return this.printIndexes(true);
     }
 
+    @Override
+    public Object getVertexId(String hexString) throws DecoderException {
+        return this.graph.getIDManager().getKeyID(StaticArrayBuffer.of(Hex.decodeHex(hexString)));
+    }
```
porunov, line 555: Usually I was using `Long.decode` to decode to decode hex (standard Java) and then used `BufferUtil.getLongBuffer` to create buffer. I.e.:
```
graph.getIDManager().getKeyID(org.janusgraph.diskstorage.util.BufferUtil.getLongBuffer(Long.decode("0x3800000066ff3400")))
```

That said, I'm good with both approaches.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
```diff
@@ -545,6 +549,23 @@ public String printIndexes() {
         return this.printIndexes(true);
     }
 
+    @Override
+    public Object getVertexId(String hexString) throws DecoderException {
```
porunov, line 553: I believe if we are adding method to convert hexString to vertexId then the reversal method would also be useful.
I.e. a method to return hexString from vertexId:
```
graph.getIDManager().getKey(vertexId);
```

That said, we can add it in follow up PRs if you want.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
ntisseyre: i tried and it fails for me in the test `java.lang.NumberFormatException: For input string: "e800000000000018"`. i don't think it is encoded as Long

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
ntisseyre: I have added the method

ntisseyre:
> Thank you @ntisseyre ! I just added small two nitpicks which are not necessary needed to be resolved right now, but we can handle them in the follow up PRs if you want (as you prefer). I noticed your commit message is `Initial commit`. Just ensure to change the commit message when it's ready to be merged. Otherwise - LGTM.

Thanks for the review @porunov ! I have fixed the commit message

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/management/ManagementSystem.java
porunov: > i tried and it fails for me in the test `java.lang.NumberFormatException: For input string: "e800000000000018"`. i don't think it is encoded as Long

Ah, you right! This will only work for the Long vertex ids, but not for string vertex IDs. NVM then, you solution is better as it will covers both cases.

porunov (approved):
Thank you! LGTM.

PR #4767, opened 2025-02-11
Update `docker-compose-cql-es.yml`

tae898:
This PR is related to https://github.com/JanusGraph/janusgraph/issues/4763

It works fine on my local machine but if there are more tests needed, I'll do it.


porunov:

Issue #4860, opened 2025-10-03
List property ordering - utilize mixed index


pm-osc:
**Describe the feature:**
Extend JanusGraph to support ordering of list properties via mixed index backend. Based on my research it seems that both ElasticSearch and Solr is capable of properly ordering indexed documents also based on a multi-value properties so from mixed index backend perspective this feature is possible.

***How ordering via mixed index backend works for single cardinality property?***
When it comes to ordering a query result by a single cardinality property, JanusGraph utilizes the mixed index backend in case mixed index is defined for the given property. This is how the profile of the query looks like for the "name1" single cardinality String property.

<img width="783" height="244" alt="Image" src="https://github.com/user-attachments/assets/bb1f6443-8605-4186-9e8d-18f0ad5a4d8c" />

***How ordering via mixed index backend does NOT work for list cardinality property?***
At the same time if a "name2" property with list cardinality is used for ordering (mixed index is in place similar to the "name1" property), JanusGraph does not use any index while perform ordering (mind the OrderGlobalStep line at the end).

<img width="773" height="260" alt="Image" src="https://github.com/user-attachments/assets/d9f45df6-1f3c-48d1-b9a1-3e462b3adcb3" />

**Describe a specific use case for the feature:**
I just simply would like to efficiently order list cardinality properties by utilizing ElasticSearch mixed index backend.

Sample groovy file is provided that describes creating the graph, test data and query to execute.

[ordert_testgraph.txt](https://github.com/user-attachments/files/22682111/ordert_testgraph.txt)

porunov:
Was fixed in #4861 
Thank you @pm-osc for the contribution!

PR #4861, opened 2025-10-04
#4860 allow ordering list properties via mixed index backend

pm-osc:
Related feature request is #4860.

### For all changes:
- [ yes ] Is there an issue associated with this PR? Is it referenced in the commit message?
- [ yes ] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve?
- [ yes ] Has your PR been rebased against the latest commit within the target branch (typically `master`)?
- [ yes ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ yes ] Have you written and/or updated unit tests to verify your changes?
- [ not applicable ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)?
- [ not applicable ] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository?
- [ not applicable ] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository?

- [ not applicable ] Have you ensured that format looks appropriate for the output in which it is rendered?

### Implementation notes

Ordering properties with list cardinality via mixed index backend works as expected with Solr and Elasticsearch. For Lucene such ordering does not work, since in janusgraph-lucene/src/main/java/org/janusgraph/diskstorage/lucene/LuceneIndex.java line 516
`boolean isPossibleSortIndex = ki.getCardinality() == Cardinality.SINGLE;` line explicitly sets that only single cardinality fields are possible to sort and the following error would show up on ordering `unexpected docvalues type NONE for field 'XXXX' (expected=SORTED). Re-index with correct docvalues type`.
If this `isPossibleSortIndex` restriction is removed (e.g. by setting isPossibleSortIndex into true), a lot of tests start failing.

The solution provided in this PR is not a very elegant one. The challenge is that considering a defined order is implemented in
janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/optimize/step/HasStepFolder.java's foldInOrder() method.
At this early stage, there is still no information about the fact that against which mixed index backend the query will be executed.
To be on the safe side, order is added to the `orderBy` data structure for List cardinality property only if at least a single mixed index backend is configured and none of the configured mixed index backends is Lucene. This logic is implemented in the added `isIndexBackendsSupportsListPropertyOrdering()` method in the HasStepFolder.java file.


porunov (approved):
Thank you @pm-osc !
I never had a use-case when you would use non-single cardinality property for sorting, so don't know where it's applicable, but I assume it may be useful for someone. LGTM.

Issue #4842, opened 2025-08-28
support for Cassandra 5.0.x as a storage backend.


gitcode4shyam:
Please add support for Cassandra 5.0 as a storage backend. The new version is GA and the Java driver 4.19+ is compatible.

porunov:
Fixed in #4897

PR #4897, opened 2026-06-08
Add Cassandra 5.0 storage backend support and integration tests

Ranexia04:
Addresses #4842.

## Summary

Adds Cassandra 5.0 support to the CQL backend: a real bug fix surfaced by running the existing test suite against Cassandra 5.0, plus integration-test coverage so 5.0 is exercised the same way 3.11 / 4.0 are.

### Bug fix
When compression is disabled (`storage.cql.compression = false`), `CQLKeyColumnValueStore` used the driver's `withNoCompression()`, which emits the legacy `compression = {'sstable_compression': ''}` form. Cassandra 5.0 rejects this with `InvalidConfigurationInQueryException: Missing sub-option 'class' for the 'compression' option`. On Cassandra 5+ we now disable compression via `{'enabled': false}`, which is accepted by all supported Cassandra versions (3.x/4.x behavior is unchanged). This mirrors the existing version-aware handling already present for the *enabled*-compression options.

### Test coverage
- `cassandra5-byteordered`, `cassandra5-murmur`, `cassandra5-murmur-ssl`, `cassandra5-murmur-client-auth` Maven profiles (mirroring the `cassandra4-*` ones), pinned to `5.0.8`.
- `cassandra5-*.yaml` test configs written in the modern Cassandra 5.0 parameter format (unit-suffixed names, e.g. `read_request_timeout: 5000ms`) so the server boots with no deprecation warnings.
- `JanusGraphCassandraContainer.getConfigPrefix()` now resolves `cassandra5` for 5.x.
- CI matrix entries in `ci-backend-cql.yml` (diskstorage / graphdb / core / hadoop / ssl / client-auth, Java 11 test client).
- Docs note that 3.11 / 4.0 / 5.0 are integration-tested.

## Verification

All `cassandra5-*` CQL jobs pass on CI. The compression fix was also reproduced and confirmed against a live `cassandra:5.0` container: the legacy form reproduces the exact error, while `{'enabled': false}` succeeds and reads back as `{'enabled': 'false'}` (which `testDisableCFCompressor` accepts).

> Note on CI cost: this adds ~10 jobs to the always-on `tests` matrix. Happy to narrow the matrix (e.g. drop the byteordered or ssl/client-auth variants for 5.0) if that's preferred.

---

### For all changes:
- [x] Is there an issue associated with this PR? Is it referenced in the commit message? (#4842)
- [x] Does your PR body contain #xyz where xyz is the issue number you are trying to resolve? (#4842)

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under ASF 2.0? (no new dependencies)
- [x] If applicable, have you updated the LICENSE.txt file? (not applicable)
- [x] If applicable, have you updated the NOTICE.txt file? (not applicable)

### DIFF
--- a/janusgraph-cql/pom.xml
+++ b/janusgraph-cql/pom.xml
@@ -20,8 +20,9 @@
         <test.jvm.opts>-Xms256m -Xmx1280m -ea -XX:+HeapDumpOnOutOfMemoryError ${test.extra.jvm.opts}</test.jvm.opts>
         <test.excluded.groups>MEMORY_TESTS,PERFORMANCE_TESTS,BRITTLE_TESTS</test.excluded.groups>
         <top.level.basedir>${basedir}/..</top.level.basedir>
-        <test.docker.version.cassandra3>${cassandra.version}</test.docker.version.cassandra3>
-        <test.docker.version.cassandra4>${cassandra-dist.version}</test.docker.version.cassandra4>
+        <test.docker.version.cassandra3>3.11.10</test.docker.version.cassandra3>
+        <test.docker.version.cassandra4>4.0.6</test.docker.version.cassandra4>
+        <test.docker.version.cassandra5>5.0.8</test.docker.version.cassandra5>
         <cassandra.docker.useDefaultConfigFromImage>false</cassandra.docker.useDefaultConfigFromImage>
     </properties>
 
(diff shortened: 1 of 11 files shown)


janusgraph-cql/src/main/java/org/janusgraph/diskstorage/cql/CQLKeyColumnValueStore.java
```diff
@@ -337,7 +337,9 @@ private static CreateTableWithOptions compressionOptions(final CreateTableWithOp
                                                              final Configuration configuration,
                                                              final int cassandraMajorVersion) {
         if (!configuration.get(CF_COMPRESSION)) {
-            // No compression
+            // Cassandra 5+ rejects {'sstable_compression': ''} — use {'enabled': false} instead.
```
Ranexia04, line 340: CassandraContainer.withNoCompression() generates compression = {'sstable_compression': ''}, which Cassandra 5.0 rejects with "Missing sub-option 'class' for the 'compression' option". The {'enabled': false} form is accepted by Cassandra 3.x, 4.x, and 5.x, so this branch is only needed for the sstable_compression → enabled rename rather than an incompatibility in the disable-compression semantics themselves.

Ranexia04:
@porunov safe to merge?

janusgraph-cql/pom.xml
```diff
@@ -22,6 +22,7 @@
         <top.level.basedir>${basedir}/..</top.level.basedir>
         <test.docker.version.cassandra3>${cassandra.version}</test.docker.version.cassandra3>
         <test.docker.version.cassandra4>${cassandra-dist.version}</test.docker.version.cassandra4>
+        <test.docker.version.cassandra5>5.0.8</test.docker.version.cassandra5>
```
porunov: So, in main `pom.xml` we have:
```
<cassandra.version>3.11.10</cassandra.version>
<!-- When this version updated please also update sha256 -->
<!-- https://downloads.apache.org/cassandra/4.0.6/apache-cassandra-4.0.6-bin.tar.gz.sha256 -->
<cassandra-dist.version>4.0.6</cassandra-dist.version>
<cassandra-dist.version.sha256>86d14a8e158e4e8554388b1aab598efcb879ddb09c480edecabb3c350b58fe6b</cassandra-dist.version.sha256>
<cassandra-driver.version>4.19.1</cassandra-driver.version>
<scylla-driver.version>4.19.0.1</scylla-driver.version>
<scylladb.version>6.2.0</scylladb.version>
```

1. Let's update `cassandra-driver.version` to `4.19.3` and `scylla-driver` to `4.19.0.9`.
2. `cassandra.version` seems a bit misleading. I think it's safe to me removed and instead we could put `3.11.10` directly in `test.docker.version.cassandra3`.
3. I'm also thinking it would make sense to update Cassandra dist version (distribution packed Cassandra) to 5 as well. you can find Cassandra distribution [here](https://downloads.apache.org/cassandra/5.0.8/). Just update sha256 here and the link will be automatically be updated [here](https://github.com/JanusGraph/janusgraph/blob/3ed2758a034f22fcd6d6cdea1a40100d559fa73a/janusgraph-dist/pom.xml#L225-L232).

Also, I think it would make sense to get a note about Cassandra 5 support in [changelog under 1.2.0 version](https://github.com/JanusGraph/janusgraph/blob/master/docs/changelog.md#upgrade-instructions) and we should also specify new tested version in the changelog under [version compatibility matrix](https://github.com/JanusGraph/janusgraph/blob/master/docs/changelog.md#version-compatibility-matrix).

porunov (changes requested):
Thanks @Ranexia04 ! Looks good! See a small commit below.

porunov (approved):
LGTM! I started tests on the branch. Thanks @Ranexia04 !

porunov:
@Ranexia04 I reverted distribution packing to use Cassandra 4 instead of Cassandra 5 because Cassandra 5 requires Java >= 11, but JanusGraph still preserves compatibility with Java 8. 
Later when we update min required Java version in JanusGraph we can pack newer ES and newer Cassandra into the distribution.

Changes are in this force push: https://github.com/JanusGraph/janusgraph/compare/e231a69c1e8476223e9eb6ca7f52c407effcc041..ce5425a60d3412ef7eec9abf262a437caf8498fa

Issue #1341, opened 2018-12-07
Multiple ThreadLocal leaks during tomcat stop/undeploy


mad:
Configuration with berkelydb

When I stop tomcat application, I see this log messages:

```
07-Dec-2018 14:33:44.430 SEVERE [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [ROOT] created a ThreadLocal with key of type [java.lang.ThreadLocal.SuppliedThreadLocal] (value [java.lang.ThreadLocal$SuppliedThreadLocal@141a82c7]) and a value of type [org.janusgraph.graphdb.transaction.StandardJanusGraphTx] (value [standardjanusgraphtx[0x513ee6be]]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
07-Dec-2018 14:33:44.430 SEVERE [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [ROOT] created a ThreadLocal with key of type [org.apache.tinkerpop.gremlin.structure.util.AbstractThreadLocalTransaction$1] (value [org.apache.tinkerpop.gremlin.structure.util.AbstractThreadLocalTransaction$1@7d949cf7]) and a value of type [org.apache.tinkerpop.gremlin.structure.Transaction$READ_WRITE_BEHAVIOR$1] (value [AUTO]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
07-Dec-2018 14:33:44.431 SEVERE [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [ROOT] created a ThreadLocal with key of type [org.apache.tinkerpop.gremlin.structure.util.AbstractThreadLocalTransaction$2] (value [org.apache.tinkerpop.gremlin.structure.util.AbstractThreadLocalTransaction$2@7f153569]) and a value of type [org.apache.tinkerpop.gremlin.structure.Transaction$CLOSE_BEHAVIOR$2] (value [ROLLBACK]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
[...]

PR #4900, opened 2026-06-24
Fix NPE in GraphTransaction.doClose() on concurrent graph close (#4899)

### LINKED ISSUE #4899: NullPointerException in JanusGraphBlueprintsGraph$GraphTransaction.doClose() when graph is closed concurrently with an in-flight transaction
### Description

There is a race between `JanusGraphBlueprintsGraph.close()` and `GraphTransaction.doClose()`. `close()` sets the shared `txs` ThreadLocal field to `null`, but `doClose()` calls `txs.remove()` without a null check. If a graph is closed on one thread while another thread is finishing a `commit()`/`rollback()`, the committing thread throws an NPE in its `finally` cleanup block — *after* the storage commit has already been attempted, masking the real outcome.

This commonly surfaces during application shutdown / container termination, where in-flight Gremlin requests overlap with graph close.

### Relevant code (`janusgraph-core`, `JanusGraphBlueprintsGraph`)

```java
public synchronized void close() {
    txs.remove();
    txs = null;          // shared field nulled
}

protected void doClose() {
    super.doClose();
    transactionListeners.remove();
    txs.remove();        // NPE if close() already ran on another thread
}
```

Note that `getAutoStartTx()` and `GraphTransaction.isOpen()` **already** guard against `txs == null`, but `doClose()` does not — the null-handling is inconsistent.

### Impact

The NPE happens in a `finally` block after `super.commit()`, so the transaction may have actually committed. Callers see an NPE instead of the true commit result, making the outcome ambiguous.

### Suggested fix

Null-check the `txs` reference in `doClose()` (snapshot to a local first). This preserves the `txs = null` behavior from #1653 (which intentionally drops the strong reference to allow ThreadLocal cleanup, see #1341) while making cleanup null-safe:

```java
protected void doClose() {
    super.doClose();
    transactionListeners.remove();
    ThreadLocal<JanusGraphBlueprintsTransaction> localTxs = txs;
    if (localTxs != null) localTxs.remove();
}
```

Related: #1341, #1653.

### Version

Observed against the `JanusGraphBlueprintsGraph` cleanup logic as currently on `master`.

batrived:
Fixes #4899.

## Problem

There is a race between `JanusGraphBlueprintsGraph.close()` and `GraphTransaction.doClose()`. `close()` sets the shared `txs` ThreadLocal field to `null`, but `doClose()` called `txs.remove()` without a null check. If a graph is closed on one thread while another thread is finishing a `commit()`/`rollback()`, the committing thread throws an NPE in its `finally` cleanup block — *after* the storage commit has already been attempted, masking the real outcome.

This commonly surfaces during application shutdown / container termination, where in-flight Gremlin requests overlap with graph close. Note that `getAutoStartTx()` and `GraphTransaction.isOpen()` already guard against `txs == null`, but `doClose()` did not — the null-handling was inconsistent.

## Fix

Snapshot the shared `txs` reference to a local and null-check it before calling `remove()`:

```java
protected void doClose() {
    super.doClose();
    transactionListeners.remove();
    ThreadLocal<JanusGraphBlueprintsTransaction> localTxs = txs;
    if (localTxs != null) localTxs.remove();
}
```

This preserves the `txs = null` behavior from #1653 (which intentionally drops the strong reference to allow ThreadLocal cleanup, see #1341) while making cleanup null-safe.

## Testing

Added a regression test `ThreadLocalTxLeakTest#doCloseMustBeNullSafeWhenTxsClearedConcurrently` that simulates a concurrent `close()` nulling the field and asserts `doClose()` does not throw. Verified it **fails with an NPE** without the fix and **passes** with it. All tests in the class pass.

---

### For all changes:

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)? (N/A — no new dependencies)
- [x] If applicable, have you updated the LICENSE.txt file? (N/A)
- [x] If applicable, have you updated the NOTICE.txt file? (N/A)

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphBlueprintsGraph.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/tinkerpop/JanusGraphBlueprintsGraph.java
@@ -344,9 +344,13 @@ public abstract class JanusGraphBlueprintsGraph implements JanusGraph {
 
         @Override
         protected void doClose() {
+            // Snapshot the shared txs reference first: close() may null it out concurrently on another
+            // thread. Capturing it before the calls below keeps the race window minimal so this thread
+            // still removes its own ThreadLocal entry rather than relying on delayed GC-based cleanup.
+            ThreadLocal<JanusGraphBlueprintsTransaction> localTxs = txs;
             super.doClose();
             transactionListeners.remove();
-            txs.remove();
+            if (localTxs != null) localTxs.remove();
         }
 
         @Override
--- a/janusgraph-test/src/test/java/org/janusgraph/core/ThreadLocalTxLeakTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/core/ThreadLocalTxLeakTest.java
@@ -108,6 +108,56 @@ class ThreadLocalTxLeakTest {
         assertThreadLocalClosed();
     }
 
+    /**
+     * Reproduces https://github.com/JanusGraph/janusgraph/issues/4899: when the graph is closed on another
+     * thread (nulling the shared {@code txs} field) while a transaction is being closed, {@code doClose()}
+     * used to throw a {@link NullPointerException} in its cleanup block, masking the real commit/rollback
+     * outcome. After the fix, {@code doClose()} snapshots the {@code txs} reference and tolerates a null,
+     * so the cleanup completes without throwing.
+     */
+    @Test
+    void doCloseMustBeNullSafeWhenTxsClearedConcurrently() throws Exception {
+        CompletableFuture.runAsync(() -> {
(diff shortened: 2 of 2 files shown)


porunov (approved):
LGTM. Thank you @batrived !

batrived:
@porunov thank you for your time and reviewing it. Since I don't have merge permissions, could you please merge it when you get a chance? Thanks!

porunov:
> @porunov thank you for your time and reviewing it. Since I don't have merge permissions, could you please merge it when you get a chance? Thanks!

Based on the standard RTC rules we either need a second approval or a week for lazy consensus: https://docs.janusgraph.org/development/#review-then-commit-rtc

PR #4914, opened 2026-07-02
Fix NPE in StandardJanusGraphTx.getVertices() on concurrent graph close (#4907)

### LINKED ISSUE #4907: NullPointerException in StandardJanusGraphTx.getVertices when the transaction is closed concurrently
## Description

`StandardJanusGraphTx.getVertices(Object...)` builds its result list from `vertexCache.get(...)` and then calls `result.removeIf(JanusGraphElement::isRemoved)` ([master line 561](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java#L561)). If the transaction is released concurrently while the query is still executing — e.g. `graph.close()` runs (instance shutdown) while a Gremlin traversal is mid-iteration on another thread — `releaseTransaction()` swaps the transaction's `vertexCache` to an `EmptyVertexCache`, whose `get(...)` returns `null`. That `null` is added to the result list and then dereferenced by the `removeIf(JanusGraphElement::isRemoved)` predicate (an unbound method reference, i.e. `v -> v.isRemoved()`), throwing a `NullPointerException` instead of a clean "transaction has been closed" error.

The singular `getVertex(Object)` already tolerates this (`return (null == v || v.isRemoved()) ? null : v;`, [line 513](https://github.com/JanusGraph/janusgraph/blob/master/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java#L513)); only the plural `getVertices(Object...)` does not.

Both branches of `getVertices` are affected (the `null` originates in the shared `vertexCache` once the tx is released), but it is most easily hit on traversals using `checkExternalVertexExistence(false)` (e.g. `g.with("checkExternalVertexExistence", false).V(ids)...`): that branch adds every `vertexCache.get(...)` result unconditionally, whereas the existence-checking branch runs an existence multiquery first that would otherwise fail with a clean closed-graph error.

## Stack trace

```
java.lang.NullPointerException
	at java.base/java.util.ArrayList.removeIf(ArrayList.java:1702)
	at java.base/java.util.ArrayList.removeIf(ArrayList.java:1690)
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.getVertices(StandardJanusGraphTx.java:...)
	at org.janusgraph.graphdb.tinkerpop.JanusGraphBlueprintsTransaction.vertices(JanusGraphBlueprintsTransaction.java:154)
	at org.janusgraph.graphdb.tinkerpop.JanusGraphBlueprintsGraph.vertices(JanusGraphBlueprintsGraph.java:148)
	at org.janusgraph.graphdb.tinkerpop.optimize.strategy.JanusGraphStepStrategy.lambda$apply$0(JanusGraphStepStrategy.java:76)
	at org.apache.tinkerpop.gremlin.process.traversal.step.map.GraphStep.processNextStart(GraphStep.java:158)
	... gremlin-server traversal iteration ...
```

## How to reproduce

1. Open a graph, add and commit a vertex, note its id.
2. Start a transaction with `checkExternalVertexExistence(false)`.
3. While a `g.V(id)...` traversal is iterating on that transaction, have another thread close the graph (`StandardJanusGraph.close()` rolls back open transactions and swaps their `vertexCache` to `EmptyVertexCache`).
4. `getVertices` adds the `null` from `EmptyVertexCache.get(...)` and NPEs at `result.removeIf(...)`.

Deterministic equivalent (unit test): open a graph, start a `checkExternalVertexExistence(false)` transaction, replace its `vertexCache` field with `EmptyVertexCache.getInstance()`, then call `getVertices(id)` → NPE.

This is realistic under Gremlin Server: when an instance is shut down (e.g. autoscaling / rolling restarts) with requests still in flight, `StandardJanusGraph.closeInternal()` rolls back all open transactions (it explicitly assumes no concurrent activity during shutdown), racing any in-flight `getVertices`.

## Expected behavior

`getVertices` should not throw `NullPointerException` when the transaction is released concurrently — it should tolerate the null cache result, consistent with `getVertex(Object)`. The closed graph/transaction then surfaces as a clean error on the next operation (or the terminating commit) rather than an NPE from `getVertices`.

## Suggested fix

Make `getVertices` tolerate a `null` cache result by using the same predicate `getVertex` already uses:

```java
// before
result.removeIf(JanusGraphElement::isRemoved);
// after
result.removeIf(v -> v == null || v.isRemoved());
```

This is minimal (the `vertexCache.get(...)` add-sites are unchanged) and keeps `getVertices` consistent with `getVertex`. A `null` from the cache here means the transaction was released concurrently; tolerating it prevents the NPE, and the closed state still surfaces cleanly downstream. Happy to open a PR.

## Versions

- JanusGraph: 1.0.0 (bug also present on `master`)

batrived:
## Summary

Fixes #4907.

`StandardJanusGraphTx.getVertices(Object...)` builds its result by calling `vertexCache.contains(id)` and then `vertexCache.get(id, ...)`. If the transaction is released concurrently — e.g. `graph.close()` on another thread — between those two calls, `releaseTransaction()` swaps `vertexCache` to an `EmptyVertexCache` whose `get()` returns `null`, after `getVertices()` has already passed `verifyOpen()`. That `null` was added to the result list and then dereferenced by:

```java
result.removeIf(JanusGraphElement::isRemoved);
```

throwing a `NullPointerException` in `ArrayList.removeIf`.

The singular `getVertex()` method already guards against this (`return (null == v || v.isRemoved()) ? null : v;`). This change aligns `getVertices()` with that behavior:

```java
result.removeIf(v -> v == null || v.isRemoved());
```

## Test

Adds `StandardJanusGraphTxGetVerticesTest`, a regression test that swaps in a vertex cache reproducing the `contains()`/`get()` race window (`contains()` returns `true`, `get()` returns `null`) and asserts `getVertices()` no longer throws and filters out the null entry. The test fails with the exact NPE (`ArrayList.removeIf` → `StandardJanusGraphTx.getVertices`) before the fix and passes after it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTx.java
@@ -562,8 +562,10 @@ public class StandardJanusGraphTx extends JanusGraphBlueprintsTransaction implem
                 }
             }
         }
-        //Filter out potentially removed vertices
-        result.removeIf(JanusGraphElement::isRemoved);
+        //Filter out potentially removed vertices. A concurrent transaction release (e.g. graph.close() on
+        //another thread) can swap the vertexCache to an EmptyVertexCache whose get() returns null after this
+        //method has already passed verifyOpen(), so guard against null entries the same way getVertex() does.
+        result.removeIf(v -> v == null || v.isRemoved());
         return result;
     }
 
--- a/janusgraph-test/src/test/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTxGetVerticesTest.java
+++ b/janusgraph-test/src/test/java/org/janusgraph/graphdb/transaction/StandardJanusGraphTxGetVerticesTest.java
@@ -1,0 +1,111 @@
+// Copyright 2026 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package org.janusgraph.graphdb.transaction;
+
+import org.janusgraph.core.JanusGraph;
+import org.janusgraph.core.JanusGraphFactory;
+import org.janusgraph.core.JanusGraphVertex;
+import org.janusgraph.graphdb.internal.InternalVertex;
+import org.janusgraph.graphdb.transaction.vertexcache.VertexCache;
+import org.janusgraph.util.datastructures.Retriever;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.Field;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+class StandardJanusGraphTxGetVerticesTest {
+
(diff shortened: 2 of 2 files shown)


porunov (approved):
LGTM. Thank you @batrived !

PR #4924, opened 2026-08-02
docs: clarify supported regular expression syntax, scope, and anchor behavior

### LINKED ISSUE #1377: Document supported regular expressions
The docs currently only mention that regular expressions can be used with the predicates `textContainsRegex` and `textRegex`, but not what kind of regular expressions are supported. This leads to confusion when users try things that are not supported like anchors for example, e.g., this doesn't work:
```groovy
g.E().has('reason',textContainsRegex('^loves')).values('reason')
```
Without having looked in the source code, I assume that the limitations are based on the index backends. So, we could simply point at their documentation for this matter. The Elasticsearch docs document the supported operators and also mention that the Lucene RegEx engine is used.[\[1\]](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-regexp-query.html#regexp-syntax)

basisworks:
Fixes #1377.

### What changed

Adds a dedicated section `Regular Expression Syntax and Behavior` to `docs/index-backend/text-search.md`:

1. **Engine & Syntax**: Documents that JanusGraph's indexing backends (Elasticsearch, Lucene, Solr) process regex queries using Lucene's regular expression engine, with direct links to the official Elasticsearch & Lucene regex syntax specs.
2. **Matching Scope**: Clarifies the core difference between full-text regex (`textContainsRegex`) and string regex (`textRegex`):
   - `textContainsRegex`: Evaluates against **individual tokens**. Anchors (`^` and `$`) match individual token start/end, which explains why queries like `textContainsRegex('^loves')` act as token prefix matches rather than matching the start of the whole property string.
   - `textRegex`: Evaluates against the **entire un-tokenized string** (`Mapping.STRING`). Note that Lucene implicitly anchors regex queries to full terms.

Docs update only — no code changes.


porunov (approved):
LGTM. Thank you @basisworks !
Merging with CTR

porunov:
@basisworks before we merge it, could you please sign your commit according to EastCLA? Thanks!
https://github.com/JanusGraph/janusgraph/pull/4924#issuecomment-5157718744

Issue #4873, opened 2025-10-29
Support CDC mode for Mixed Index mutations


porunov:
Mixed Indexes are part of an external system and that's why stale indexes may appear in the system during failures. This can sometimes lead to [permanent stale indexes](https://docs.janusgraph.org/advanced-topics/stale-index/) and requires either in-flight index fixing when the index is found or proper transaction recoveries to align data between different storage systems (for example Cassandra and ElasticSearch).
This proposal it to allow using built-in change data capture (CDC) when it's available for mixed index mutations.
In such case, this will guarantee eventual consistency between mixed indexes and actual graph data.

Cassandra natively supports CDC ([doc](https://cassandra.apache.org/doc/latest/cassandra/managing/operating/cdc.html)) and there is also available [Debezium Connector for Cassandra](https://debezium.io/documentation/reference/stable/connectors/cassandra.html).

To support this we should change JanusGraph Mixed index mutation mode to skip any mutations during original transaction and instead relay on the Debezium Connector to process Mixed Index mutations later. Alternatively, we could do both (mutate mixed index during original transaction and then trigger mutations again during CDC processing - just to ensure consistency in case of failures).

Related work on transaction recovery process #4872

PR #4937, opened 2026-08-11
Cover the element ids accepted by IndexSerializer.removeElement (#4928)

### LINKED ISSUE #4928: IndexSerializer.removeElement rejects String element ids, breaking mixed-index recovery for custom vertex id types
- Version: `master` (`ac0eb23`)
- Storage Backend: any
- Mixed Index Backend: any
- Expected Behavior: mixed-index transaction recovery should work on graphs configured with `graph.allow-custom-vid-types`.
- Current Behavior: `IndexSerializer.removeElement` rejects String element ids, so recovering a *deleted* element throws `IllegalArgumentException` and aborts the restore.

### Details

https://github.com/JanusGraph/janusgraph/blob/ac0eb2392ddad3d96ab8cde2a5a9a123dbc5d839/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java#L334-L338

```java
public void removeElement(Object elementId, MixedIndexType index, Map<String,Map<String,List<IndexEntry>>> documentsPerStore) {
    Preconditions.checkArgument((index.getElement()==ElementCategory.VERTEX && elementId instanceof Long) ||
        (index.getElement().isRelation() && elementId instanceof RelationIdentifier),"Invalid element id [%s] provided for index: %s",elementId,index);
    getDocuments(documentsPerStore,index).put(element2String(elementId),new ArrayList<>());
}
```

The precondition allows only `Long` for vertices. But the very next line calls `element2String`, which explicitly supports String ids:

https://github.com/JanusGraph/janusgraph/blob/ac0eb2392ddad3d96ab8cde2a5a9a123dbc5d839/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/util/IndexRecordUtil.java#L100-L107

```java
public static String element2String(Object elementId) {
    Preconditions.checkArgument(elementId instanceof Long || elementId instanceof RelationIdentifier || elementId instanceof String);
    ...
}
```

So the two methods disagree about what a valid element id is, and the stricter one runs first.

The only caller is the transaction-log recovery path:

https://github.com/JanusGraph/janusgraph/blob/ac0eb2392ddad3d96ab8cde2a5a9a123dbc5d839/janusgraph-core/src/main/java/org/janusgraph/graphdb/log/StandardTransactionLogProcessor.java#L246-L255

```java
JanusGraphElement element = restore.retrieve(tx);
if (element!=null) {
    graph.getIndexSerializer().reindexElement(element,index,restoredDocs);
} else { //Element is deleted
    graph.getIndexSerializer().removeElement(restore.elementId,index,restoredDocs);
}
```

So on a graph using custom String vertex ids, recovering a transaction whose element was deleted throws instead of removing the stale index document — meaning the recovery mechanism that exists to repair index divergence cannot complete.

### Steps to Reproduce

1. Configure a graph with `graph.allow-custom-vid-types=true` and `tx.log-tx=true`, with a mixed index on vertices, using String vertex ids.
2. Start transaction recovery (`JanusGraphFactory.startTransactionRecovery`).
3. Cause a transaction that deletes an indexed vertex to fail on the index backend.
4. Observe: recovery throws `IllegalArgumentException: Invalid element id [...] provided for index` and the stale document is never removed.

### Suggested Fix

Align the precondition with `element2String`:

```java
Preconditions.checkArgument(
    (index.getElement()==ElementCategory.VERTEX && (elementId instanceof Long || elementId instanceof String)) ||
    (index.getElement().isRelation() && elementId instanceof RelationIdentifier),
    "Invalid element id [%s] provided for index: %s", elementId, index);
```

batrived:
Fixes #4928.

`IndexSerializer.removeElement` rejected every vertex id which is not a `Long`, but the very next line calls `element2String`, which accepts a `Long`, a `RelationIdentifier` **or** a `String`. The two methods disagreed about what a valid element id is, and the stricter of them ran first.

A graph configured with `graph.allow-custom-vid-types` assigns String vertex ids. The only caller is `StandardTransactionLogProcessor`, which calls `removeElement` when the element of a restored transaction no longer exists. So recovering a transaction that deleted an indexed vertex threw `IllegalArgumentException`, the stale index document was never removed, and the recovery mechanism that exists to repair a diverged index could not complete.

This applies the fix suggested in the issue: accept the same vertex ids `element2String` accepts. A relation index still requires a `RelationIdentifier`.

### Testing

`IndexSerializerRemoveElementTest` covers a String vertex id, a Long vertex id, a relation id, and the ids that must still be rejected. I confirmed the String case reproduces the reported failure without the fix:

```
IllegalArgument Invalid element id [customVertexId] provided for index: Mock for MixedIndexType
```

Each test also asserts the resulting document id round-trips back to the original element id through `string2ElementId`, so the document that gets removed is the right one, and that the entry list is empty, which is how a document removal is expressed.

-----

### For all changes:

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)? — no new dependencies
- [x] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository? — not applicable
- [x] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository? — not applicable

### DIFF
--- a/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java
+++ b/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/IndexSerializer.java
@@ -332,7 +332,10 @@ public class IndexSerializer {
     }
 
     public void removeElement(Object elementId, MixedIndexType index, Map<String,Map<String,List<IndexEntry>>> documentsPerStore) {
-        Preconditions.checkArgument((index.getElement()==ElementCategory.VERTEX && elementId instanceof Long) ||
+        //A String id is valid for a vertex when the graph allows custom vertex id types, and element2String below
+        //encodes it, so accept the same ids that method does
+        Preconditions.checkArgument((index.getElement()==ElementCategory.VERTEX
+                && (elementId instanceof Long || elementId instanceof String)) ||
             (index.getElement().isRelation() && elementId instanceof RelationIdentifier),"Invalid element id [%s] provided for index: %s",elementId,index);
         getDocuments(documentsPerStore,index).put(element2String(elementId),new ArrayList<>());
     }
--- a/janusgraph-core/src/test/java/org/janusgraph/graphdb/database/IndexSerializerRemoveElementTest.java
+++ b/janusgraph-core/src/test/java/org/janusgraph/graphdb/database/IndexSerializerRemoveElementTest.java
@@ -1,0 +1,95 @@
+// Copyright 2026 JanusGraph Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.janusgraph.graphdb.database;
+
+import org.janusgraph.diskstorage.indexing.IndexEntry;
+import org.janusgraph.graphdb.database.util.IndexRecordUtil;
+import org.janusgraph.graphdb.internal.ElementCategory;
+import org.janusgraph.graphdb.relations.RelationIdentifier;
+import org.janusgraph.graphdb.types.MixedIndexType;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+//removeElement is called by the transaction log recovery of StandardTransactionLogProcessor when the element of a
+//restored transaction no longer exists, to remove its stale mixed index document. Rejecting a valid element id aborts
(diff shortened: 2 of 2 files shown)


porunov:
Reviewed, rebased on current `master` and force-pushed as a single commit. Copilot generated no comments on this one.

### The production fix already landed

This is the one thing to be aware of before merging. The issue is real and your analysis of it was correct, but the precondition has since been widened to accept a String vertex id on `master`, in cbd6fc2e5 (`feat: CDC-based mixed index synchronization (#4873)`, merged 2026-08-17 — three days after this PR was opened), which is why the branch went to conflicting. The conflict is exactly the two versions of that one line.

I resolved it in favour of `master`'s line and kept your comment above it, so what the PR now contributes is:

- `IndexSerializerRemoveElementTest`, and
- the comment recording *why* a String id is valid there, which the line on `master` does not carry.

I retitled the PR and rewrote the commit message to match, so the history does not claim a fix that another commit made. Nothing else about the branch changed.

### Why I kept it rather than closing it

The test is worth having on its own. I checked it is not vacuous: narrowing the precondition back to `elementId instanceof Long` makes `shouldRemoveVertexWithCustomStringId` fail with exactly the reported error —

```
java.lang.IllegalArgumentException: Invalid element id [customVertexId] provided for index: Mock for MixedIndexType
```

— so it pins down behaviour that is currently untested and that regressed once already. The round-trip assertion through `string2ElementId` is a nice touch; it covers more than the precondition.

### Verification

`IndexSerializerRemoveElementTest` 4/4 and the `janusgraph-core` suite green on the rebased branch. The one red check on the previous revision was `tests (hbase, hbase2-diskstorage-java8, 8)`, which no longer exists now that `master` has dropped Java 8.

porunov (approved):
LGTM

PR #4953, opened 2026-09-21
Avoid redundant backend reads in schema cache fallback and edge-by-id lookup

### LINKED ISSUE #4952: Schema cache fallback does not cache empty results; edge-by-id lookup reads the label's schema vertex every transaction
- Version: `master` (`9c2c304`); present in all released 1.x versions
- Storage Backend: any (observed on a DynamoDB-backed deployment; reproduced with BerkeleyJE)
- Mixed Index Backend: n/a
- Expected Behavior: schema lookups are served from `StandardSchemaCache` after the first read, including lookups that legitimately return no entries, and resolving an edge by id does not read the edge label's schema vertex row on every transaction.
- Current Behavior: two request-path code paths read the schema vertex row from the storage backend on every call.

### 1. `StandardSchemaCache` does not cache empty results once it falls back to the bounded cache

`StandardSchemaCache.getSchemaRelations` keeps schema relations in an unbounded map until it exceeds `maxCachedRelations` (30,000 by default), then permanently switches to a bounded Guava cache. The primary map caches empty results unconditionally, but the fallback path only caches non-empty ones:

https://github.com/JanusGraph/janusgraph/blob/9c2c304/janusgraph-core/src/main/java/org/janusgraph/graphdb/database/cache/StandardSchemaCache.java#L151-L157

```java
entries = retriever.retrieveSchemaRelations(schemaId, type, dir);
if (!entries.isEmpty()) { //only cache if type exists
    schemaRelationsBackup.put(typePlusRelation, entries);
}
```

Empty results are common and legitimate: `RelationTypeVertex.getRelationIndexes` returns nothing for any edge label without a vertex-centric index. Once a graph with a large schema has crossed the cap, every traversal step and every `addEdge` on such a label issues a `getSlice` against the label's schema vertex row. On our deployment this produced roughly 4,000 reads per second per instance against a single row and hot-partition throttling on the backend.

Caching empty results in fallback mode is safe for the same reason it is safe in the primary map: adding edges of a label is a data write and never touches the label's schema vertex, and the management operations that do change a label's definition edges (`buildEdgeIndex`, `setConsistency`, ...) already broadcast `expireSchemaElement`, which evicts from both the primary map and `schemaRelationsBackup`.

### 2. `RelationIdentifierUtils.findRelation` resolves the edge label with `tx.getVertex()`

https://github.com/JanusGraph/janusgraph/blob/9c2c304/janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java#L47

```java
JanusGraphVertex typeVertex = tx.getVertex(rId.getTypeId());
```

`tx.getVertex` goes through the external vertex retriever, which with the default `query.check-external-vertex-existence=true` issues a vertex-existence `getSlice` against the label's schema vertex row. The result lives only in the per-transaction vertex cache, so `StandardSchemaCache` never sees it and every `g.E(id)` in a fresh transaction costs an extra backend read of the same schema row. `findEdgeRelations` a few lines below already resolves the type through `StandardJanusGraphTx.getExistingRelationType`, which is served from the schema cache.

### Steps to Reproduce

1. Open a graph, create two vertices and an edge, commit.
2. In a new transaction call `tx.getEdges(edgeId)` and count `edgestore.getSlice` calls with `MetricInstrumentedStore`: 2 calls are made (one for the edge, one for the label's schema vertex) where 1 is expected.
3. For the cache issue: construct `StandardSchemaCache` with a small `maxCachedRelations`, drive it past the cap, then call `getSchemaRelations` repeatedly for a schema id whose retriever returns `EntryList.EMPTY_LIST`. Each call hits the retriever.

### Suggested Fix

- Remove the emptiness guard in the fallback path so `schemaRelationsBackup` stores empty results like the primary map.
- In `findRelation`, validate the type id with `IDManager.isRelationTypeId`, resolve it with `getExistingRelationType`, and for committed non-system types check existence through the schema cache (`getSchemaRelations(typeId, BaseKey.SchemaDefinitionProperty, OUT)`) instead of `tx.getVertex`.

I have a PR with both fixes and tests (`StandardSchemaCacheTest`, `JanusGraphOperationCountingTest.testEdgeByIdDoesNotReadTypeVertex`) ready to open against this issue.

xiaoying-code:
Fixes #4952

Two request-path code paths read an edge label's schema vertex row from the storage backend on every call instead of being served by `StandardSchemaCache`. Both are fixed here; each is a small, self-contained hunk.

### 1. Cache empty results in `StandardSchemaCache` fallback mode

Once the primary schema-relations map exceeds `maxCachedRelations`, `StandardSchemaCache` permanently switches to a bounded backup cache. The backup path only cached non-empty results:

```java
if (!entries.isEmpty()) { //only cache if type exists
    schemaRelationsBackup.put(typePlusRelation, entries);
}
```

The primary map caches empty results unconditionally. In fallback mode every lookup that legitimately returns nothing, in particular `RelationTypeVertex.getRelationIndexes` for an edge label with no vertex-centric index, became a store read on every traversal step and every `addEdge`.

This PR removes the emptiness guard so the backup cache stores empty results like the primary map does.

**Why caching an empty result is safe.** The cached entry answers "which definition edges hang off this *schema* vertex". Adding edges of a label is a data write: it touches the rows of the two user vertices, never the label's schema vertex, so a cached empty answer stays correct as edges are inserted. The only operations that change a label's definition edges are management operations such as `mgmt.buildEdgeIndex(...)` or `mgmt.setConsistency(...)`. Those commit through `ManagementSystem`, which broadcasts `expireSchemaElement(schemaId)` to every instance via the management log, and `expireSchemaElement` already invalidates both the primary map and `schemaRelationsBackup`. A brand-new label cannot produce a stale empty entry either: its schema vertex does not exist until the management commit that creates it, name lookups for it return `null` and are never cached, and the same commit expires the element afterwards. Fallback mode therefore gets exactly the semantics every instance already has before the map overflows.

### 2. Resolve the edge label through the schema cache in `RelationIdentifierUtils.findRelation`

`findRelation` resolved the edge label with `tx.getVertex(typeId)`. That goes through the external vertex retriever, which with the default `query.check-external-vertex-existence=true` issues a `vertexExistenceQuery` getSlice against the label's schema vertex row in every transaction. The result only lives in the per-transaction vertex cache, so `StandardSchemaCache` never sees it and every `g.E(id)` in a fresh transaction costs an extra backend read of the same row.

The fix resolves the type the way `findEdgeRelations` already does a few lines later: validate the id with `IDManager.isRelationTypeId`, build the handle with `getExistingRelationType(typeId)`, and for committed non-system types check that the id has a definition via `getSchemaRelations(typeId, SchemaDefinitionProperty, OUT)`. A type created in the same uncommitted transaction is found in the transaction's vertex cache and skips the storage check, as before. On a cold cache this costs one read of the schema row per instance, which is then cached like every other schema access. Behaviour is otherwise unchanged: an identifier whose type has no schema vertex still resolves to no edge (`null`), and one whose type id is not a relation type id still throws `IllegalArgumentException`.

The remaining `tx.getVertex`/`getVertices` callers that can receive a schema id (`ManagementSystem.getSchemaElement`, the index-status update trigger, `StandardTransactionLogProcessor` recovery) are management or recovery paths and are left unchanged.

### Tests

- New `StandardSchemaCacheTest` (janusgraph-core): baseline behaviour of the primary map; drives the cache past its cap and checks an empty relation looked up 5 times causes 1 store read (fails with 5 on the unpatched code); `expireSchemaElement` still evicts fallback entries.
- New `JanusGraphOperationCountingTest.testEdgeByIdDoesNotReadTypeVertex` (runs for BerkeleyJE, CQL and HBase): an edge-by-id lookup in a fresh transaction performs exactly one edgestore getSlice and none for the label's schema vertex; a non-existent edge id of the same label behaves the same; a well-formed label id with no schema vertex behind it yields no edge without reading the type row; an edge of a label created in the same transaction is resolvable by id before commit with no store read. Fails on the unpatched code with 2 slices per lookup.

Run locally on Java 17: `StandardSchemaCacheTest` and the full `BerkeleyOperationCountingTest` class pass; both new tests fail without the fix.

-----

### For all changes:

### For code changes:
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)? (no new dependencies)
- [x] If applicable, have you updated the LICENSE.txt file, including the main LICENSE.txt file in the root of this repository? (not applicable)
- [x] If applicable, have you updated the NOTICE.txt file, including the main NOTICE.txt file found in the root of this repository? (not applicable)

- [x] Have you ensured that format looks appropriate for the output in which it is rendered? (not applicable)


porunov:
Thanks @xiaoying-code for contribution. Could you please sign CLA? https://github.com/JanusGraph/janusgraph/pull/4953#issuecomment-5755807886

xiaoying-code:
> Thanks @xiaoying-code for contribution. Could you please sign CLA? [#4953 (comment)](https://github.com/JanusGraph/janusgraph/pull/4953#issuecomment-5755807886)

Thanks! I am going through the process and have sent the request to the CLA Manager.

xiaoying-code:
@porunov the CLA is completed, can you please review the code if you get a chance?

porunov (commented):
Thanks for the careful write-up, @xiaoying-code. I checked both claims and ran your tests against current master (the touched files haven't changed since your base, `9c2c304`), plus some extra scenarios.

**Both problems are real.**
- `StandardSchemaCacheTest.emptyRelationsAreCachedAfterFallingBackToBoundedCache` on master: `expected: <1> but was: <5>`.
- `BerkeleyOperationCountingTest.testEdgeByIdDoesNotReadTypeVertex` on master: `expected: <1> but was: <2>`. `tx.getVertex(typeId)` goes through the external-vertex retriever, and its existence check is remembered only in the per-transaction vertex cache.

A small correction on the impact of (1): schema vertices memoize what they read (`JanusGraphSchemaVertex.getRelated` / `getDefinition`), so in fallback mode it's one read per *transaction* per schema element and lookup, not per traversal step. With short transactions that still adds up the way you describe.

**However, the `findRelation` change introduces a correctness regression, and the safety argument for the cache change doesn't hold.** Details are in the inline comments; in short:

1. **A lookup before the type is visible poisons that type on the instance.** The new existence check reads the same cache entry as `JanusGraphSchemaVertex.getDefinition()`, and an empty answer is cached. Newly created types are never evicted, so the type keeps an empty definition until restart. With assertions off, `v.edges(OUT, label)` silently returns nothing, and `addEdge` and `g.V(v).outE()` throw `NullPointerException`.
2. **Definition edges also change outside `ManagementSystem`, with no eviction.** Constraints auto-created under `schema.constraints=true` are the example. With the `StandardSchemaCache` change, fallback mode repeats a staleness bug that master only has in primary mode.
3. **Malformed ids now throw.** A type id that isn't a relation-type id and names no existing vertex used to resolve to "no edge"; it now throws `IllegalArgumentException`.
4. **The test can't see what it claims to rule out.** Schema-cache reads happen in their own transaction under the `schema` metrics group.

**Suggested direction**
- Never cache an empty `SchemaDefinitionProperty` (or `SchemaCategory`) result, in either map. Every existing schema vertex has a definition and a category, so an empty answer means "not visible (yet)", not "exists but empty". That makes the existence check in `findRelation` safe: known types come from the cache, and unknown ids read storage each time, as master's per-transaction check does. It also stops junk ids from filling the unbounded primary map. On this PR, 1,000 junk-type `getEdges` calls add 1,000 permanent entries; on master they add none. About 30,000 entries flip the cache into fallback mode for good.
- Keep caching empty `SchemaDefinitionEdge` results (your case 1), but only once ordinary transactions that commit definition edges evict what they touched, so fallback mode doesn't regress (see the inline comment).
- For type ids that aren't relation-type ids, keep master's path (`tx.getVertex` + `instanceof`). It's rare and keeps behaviour identical.
- Add a regression test for a lookup that races the type's creation (sketch inline).

**On Copilot's two comments**
- Cumulative counts across transactions are this class's convention (`testKCVSAccess1` goes 2 → 4 → 7 → 11 without resets), so I wouldn't insist on resets. The metrics-group gap is the real test issue.
- `getExistingRelationType` doesn't return `null` for a missing *user* type. By default it builds an unchecked `EdgeLabelVertex` / `PropertyKeyVertex` stub, and the schema-cache check two lines below handles the missing case. `null` only comes back for an unknown system type id, so the existing comment is accurate.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
```diff
@@ -44,13 +46,20 @@ protected static JanusGraphRelation findRelation(RelationIdentifier rId, JanusGr
...
+        // vertex-existence read against the schema vertex row for every transaction.
+        StandardJanusGraphTx stx = (StandardJanusGraphTx) tx;
+        long typeId = rId.getTypeId();
+        if (!stx.getIdInspector().isRelationTypeId(typeId))
             throw new IllegalArgumentException("Invalid RelationIdentifier: typeID does not reference a type");
+        RelationType typeVertex = stx.getExistingRelationType(typeId);
+        if (typeVertex == null) return null; // unknown system type id
+        // A type created in this transaction has no schema row yet; only check storage for committed types.
+        if (!IDManager.isSystemRelationTypeId(typeId) && !typeVertex.isNew() &&
+            stx.getGraph().getSchemaCache().getSchemaRelations(typeId, BaseKey.SchemaDefinitionProperty, Direction.OUT).isEmpty())
```
porunov: This check reads the same cache entry that `JanusGraphSchemaVertex.getDefinition()` reads for a loaded type, `(typeId, SchemaDefinitionProperty, OUT)`. When the result is empty, it's cached: the primary map caches empties unconditionally, and with the change in `StandardSchemaCache` the fallback does too. Nothing evicts that entry when the type is created later. `ManagementSystem` only adds non-new schema vertices to `updatedTypes`, `makeEdgeLabel` / `makePropertyKey` just delegate to the transaction, and auto-created types don't go through `ManagementSystem` at all. From then on, the type has an empty definition on this instance.

Deterministic single-instance repro, in-memory backend, default configuration (passes on master):

```java
JanusGraphTransaction creator = graph.newTransaction();
EdgeLabel knows = creator.makeEdgeLabel("knows").make();           // schema id assigned here
JanusGraphTransaction reader = graph.newTransaction();
reader.getEdges(new RelationIdentifier(a, knows.longId(), 1L, b)); // caches an empty definition
reader.rollback();
creator.getVertex(a).addEdge("knows", creator.getVertex(b));
creator.commit();

JanusGraphTransaction tx = graph.newTransaction();
tx.getEdgeLabel("knows").multiplicity();       // -ea: AssertionError in getDefinition(); otherwise null
tx.getVertex(a).edges(Direction.OUT, "knows"); // assertions off: silently empty
tx.traversal().V(a).outE().count().next();     // assertions off: NullPointerException
```

In a cluster, the same thing happens when instance B creates a label and an edge, and a client passes that edge id to instance A before B's write is visible to A's reads. With eventually-consistent reads (e.g. Cassandra at consistency level `ONE`) that window is real. It also covers the "stale or foreign identifier" case this line's comment mentions. An edge id from another graph whose schema was created in the same order (so its type ids line up) is looked up before this graph creates that label. I reproduced that too, in both primary and fallback mode.

On master none of this can happen, because the existence check is remembered only in the per-transaction vertex cache. With empty `SchemaDefinitionProperty` answers not cached, as suggested in the review summary, this line works as intended.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/cache/StandardSchemaCache.java
```diff
@@ -151,9 +151,8 @@ public EntryList getSchemaRelations(final long schemaId, final BaseRelationType
             entries = schemaRelationsBackup.getIfPresent(typePlusRelation);
             if (entries==null) {
                 entries = retriever.retrieveSchemaRelations(schemaId, type, dir);
-                if (!entries.isEmpty()) { //only cache if type exists
-                    schemaRelationsBackup.put(typePlusRelation, entries);
-                }
+                // Empty results are legitimate (e.g. a type without indexes) and must be cached like the primary map does
+                schemaRelationsBackup.put(typePlusRelation, entries);
```
porunov: Parity with the primary map sounds right, but the reasoning in the description doesn't hold on two counts:

- *"The only operations that change a label's definition edges are management operations."* Ordinary transactions write definition edges too. Property-key and connection constraints auto-created under `schema.constraints=true` with the default schema maker (`StandardJanusGraphTx.addProperties` / `addConnection` → `addSchemaEdge`) are one example, and anyone can call `tx.addConnection(...)` directly. Those commits never call `expireSchemaElement`, so an empty entry cached before them is stale afterwards. Master already gets this wrong in primary mode: three transactions each adding a `person -knows-> person` edge create three copies of the connection constraint where one is expected, because each reads the cached empty list. That's a pre-existing bug, not something you introduced. Master's fallback mode creates one, because it doesn't cache the empty result. With this change it creates three there too.
- *"The same commit expires the element afterwards."* New schema vertices are never expired (see the `isNew()` exclusions around `updatedTypes` in `ManagementSystem`). A brand-new type is protected only as long as nobody asked about its id before it existed, and the `findRelation` change breaks that assumption (see the comment there).

I'd keep this change for `SchemaDefinitionEdge` results only, and pair it with a fix that evicts what ordinary transactions touch, or land it after that fix. Empty `SchemaDefinitionProperty` / `SchemaCategory` results shouldn't be cached in either map, because every existing schema vertex has a definition and a category.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
```diff
@@ -44,13 +46,20 @@ protected static JanusGraphRelation findRelation(RelationIdentifier rId, JanusGr
...
-        JanusGraphVertex typeVertex = tx.getVertex(rId.getTypeId());
-        if (typeVertex == null) return null;
-
-        if (!(typeVertex instanceof RelationType))
+        // Resolve the type through the schema cache instead of tx.getVertex(), which would issue a
+        // vertex-existence read against the schema vertex row for every transaction.
+        StandardJanusGraphTx stx = (StandardJanusGraphTx) tx;
+        long typeId = rId.getTypeId();
+        if (!stx.getIdInspector().isRelationTypeId(typeId))
             throw new IllegalArgumentException("Invalid RelationIdentifier: typeID does not reference a type");
```
porunov: This changes behaviour for malformed ids. On master, a type id that isn't a relation-type id resolves to "no edge" unless it names an existing vertex. For example, the id of a removed vertex, or of one that never existed, makes `tx.getEdges(id)` return nothing. On this PR it always throws `IllegalArgumentException`. I'd keep master's path for this case (`tx.getVertex(typeId)` + the `instanceof RelationType` check). It isn't on the hot path, so the extra read doesn't matter there, and the PR stays a pure performance change.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
```diff
@@ -298,6 +301,43 @@ public void testSettingProperty() {
...
+        assertFalse(tx.getEdges(missingEdge).iterator().hasNext());
+        verifyStoreMetrics(EDGESTORE_NAME, ImmutableMap.of(M_GET_SLICE, 2L));
+        tx.commit();
+
+        // a well-formed label id with no schema vertex behind it resolves to "no edge", still without a read of the type row
+        tx = graph.buildTransaction().groupName(metricsPrefix).start();
+        long missingTypeId = IDManager.getSchemaId(IDManager.VertexIDType.UserEdgeLabel, 1L << 40);
+        RelationIdentifier missingType = new RelationIdentifier(eid.getOutVertexId(), missingTypeId, eid.getRelationId(), eid.getInVertexId());
+        assertFalse(tx.getEdges(missingType).iterator().hasNext());
+        verifyStoreMetrics(EDGESTORE_NAME, ImmutableMap.of(M_GET_SLICE, 2L));
```
porunov: The schema cache loads through its own transaction under the `schema` metrics group (`StandardJanusGraph.typeCacheRetrieval` → `groupName(METRICS_SCHEMA_PREFIX_DEFAULT)`). So `verifyStoreMetrics(EDGESTORE_NAME, …)` on this test's group can't see a read of the type row. Here the PR *does* read that row, once, and then caches the empty answer; that's the entry that goes stale in the `findRelation` comment. To pin down "no read of the type row", count the `schema` group as well, or use `verifyTypeCacheMetrics(...)` from this class. A regression test for a lookup that races the type's creation would cover the correctness side.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
xiaoying-code: Confirmed, and thanks for the deterministic repro. The root cause was exactly as you describe: this check shared the `(typeId, SchemaDefinitionProperty, OUT)` entry with `getDefinition()`, and an empty answer was cached with nothing to evict it.

Fixed in the latest revision as suggested: `StandardSchemaCache` never caches an empty `SchemaName`, `SchemaCategory` or `SchemaDefinitionProperty` result, in either map, so a type that is not visible yet is re-read on the next lookup instead of being remembered as missing.

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/cache/StandardSchemaCache.java
xiaoying-code: Agreed on both counts; the description's safety argument was wrong and is rewritten.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
xiaoying-code: Done. Ids that are not positive relation-type ids take master's path again (`tx.getVertex` + `instanceof RelationType`), so behaviour for malformed ids is unchanged. A negative type id is covered in the new race test.

janusgraph-backend-testutils/src/main/java/org/janusgraph/graphdb/JanusGraphOperationCountingTest.java
xiaoying-code: Fixed. The test now warms the cache first, then asserts on the `schema` group as well

xiaoying-code:
> Thanks for the careful write-up, @xiaoying-code. I checked both claims and ran your tests against current master (the touched files haven't changed since your base, `9c2c304`), plus some extra scenarios.
> 
> **Both problems are real.**
> 
>     * `StandardSchemaCacheTest.emptyRelationsAreCachedAfterFallingBackToBoundedCache` on master: `expected: <1> but was: <5>`.
> 
>     * `BerkeleyOperationCountingTest.testEdgeByIdDoesNotReadTypeVertex` on master: `expected: <1> but was: <2>`. `tx.getVertex(typeId)` goes through the external-vertex retriever, and its existence check is remembered only in the per-transaction vertex cache.
> 
> 
> A small correction on the impact of (1): schema vertices memoize what they read (`JanusGraphSchemaVertex.getRelated` / `getDefinition`), so in fallback mode it's one read per _transaction_ per schema element and lookup, not per traversal step. With short transactions that still adds up the way you describe.
> 
> **However, the `findRelation` change introduces a correctness regression, and the safety argument for the cache change doesn't hold.** Details are in the inline comments; in short:
> 
>     1. **A lookup before the type is visible poisons that type on the instance.** The new existence check reads the same cache entry as `JanusGraphSchemaVertex.getDefinition()`, and an empty answer is cached. Newly created types are never evicted, so the type keeps an empty definition until restart. With assertions off, `v.edges(OUT, label)` silently returns nothing, and `addEdge` and `g.V(v).outE()` throw `NullPointerException`.
> 
>     2. **Definition edges also change outside `ManagementSystem`, with no eviction.** Constraints auto-created under `schema.constraints=true` are the example. With the `StandardSchemaCache` change, fallback mode repeats a staleness bug that master only has in primary mode.
> 
>     3. **Malformed ids now throw.** A type id that isn't a relation-type id and names no existing vertex used to resolve to "no edge"; it now throws `IllegalArgumentException`.
> 
>     4. **The test can't see what it claims to rule out.** Schema-cache reads happen in their own transaction under the `schema` metrics group.
> 
> 
> **Suggested direction**
> 
>     * Never cache an empty `SchemaDefinitionProperty` (or `SchemaCategory`) result, in either map. Every existing schema vertex has a definition and a category, so an empty answer means "not visible (yet)", not "exists but empty". That makes the existence check in `findRelation` safe: known types come from the cache, and unknown ids read storage each time, as master's per-transaction check does. It also stops junk ids from filling the unbounded primary map. On this PR, 1,000 junk-type `getEdges` calls add 1,000 permanent entries; on master they add none. About 30,000 entries flip the cache into fallback mode for good.
> 
>     * Keep caching empty `SchemaDefinitionEdge` results (your case 1), but only once ordinary transactions that commit definition edges evict what they touched, so fallback mode doesn't regress (see the inline comment).
> 
>     * For type ids that aren't relation-type ids, keep master's path (`tx.getVertex` + `instanceof`). It's rare and keeps behaviour identical.
> 
>     * Add a regression test for a lookup that races the type's creation (sketch inline).
> 
> 
> **On Copilot's two comments**
> 
>     * Cumulative counts across transactions are this class's convention (`testKCVSAccess1` goes 2 → 4 → 7 → 11 without resets), so I wouldn't insist on resets. The metrics-group gap is the real test issue.
> 
>     * `getExistingRelationType` doesn't return `null` for a missing _user_ type. By default it builds an unchecked `EdgeLabelVertex` / `PropertyKeyVertex` stub, and the schema-cache check two lines below handles the missing case. `null` only comes back for an unknown system type id, so the existing comment is accurate.

Thanks @porunov, all four points check out. Reworked as suggested and force-pushed as a single commit; description updated.
Copilot's comments left as they are, per your assessment. Verified locally on Java 17 with StandardSchemaCacheTest, the full BerkeleyOperationCountingTest and InMemoryGraphTest classes; CI workflows are waiting for approval.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
```diff
@@ -44,13 +46,32 @@ protected static JanusGraphRelation findRelation(RelationIdentifier rId, JanusGr
...
+            // (no edge if nothing exists under that id, IllegalArgumentException if a non-type vertex does).
+            JanusGraphVertex vertex = tx.getVertex(typeId);
+            if (vertex == null) return null;
+            if (!(vertex instanceof RelationType))
+                throw new IllegalArgumentException("Invalid RelationIdentifier: typeID does not reference a type");
+            typeVertex = (RelationType) vertex;
+        } else if (IDManager.isSystemRelationTypeId(typeId) || stx.isVertexCached(typeId)) {
+            // System types, and types created or already loaded in this transaction, need no storage access.
+            typeVertex = stx.getExistingRelationType(typeId);
+            if (typeVertex == null) return null; // unknown system type id
```
porunov: `isVertexCached` is also true for an id the transaction already found *not* to exist. `tx.getVertex(id)` returns `null` for such an id, but the vertex cache keeps the removed stub its existence check produced. This branch then takes that stub as the type, and building the query calls `name()` on it:

```
java.lang.IllegalStateException: Vertex with id 70368744177685 was removed.
	at org.janusgraph.graphdb.vertices.AbstractVertex.verifyAccess(AbstractVertex.java:91)
	at org.janusgraph.graphdb.vertices.AbstractVertex.query(AbstractVertex.java:137)
	at org.janusgraph.graphdb.types.vertices.JanusGraphSchemaVertex.name(JanusGraphSchemaVertex.java:71)
	at org.janusgraph.graphdb.query.vertex.BaseVertexCentricQueryBuilder.types(BaseVertexCentricQueryBuilder.java:160)
	at org.janusgraph.graphdb.relations.RelationIdentifierUtils.findEdgeRelations(RelationIdentifierUtils.java:107)
	...
	at org.janusgraph.graphdb.transaction.StandardJanusGraphTx.getEdges(StandardJanusGraphTx.java:1048)
```

Repro on in-memory with the default configuration; master finds no edge:

```java
long missingTypeId = IDManager.getSchemaId(IDManager.VertexIDType.UserEdgeLabel, 1L << 40);
JanusGraphTransaction tx = graph.newTransaction();
assertNull(tx.getVertex(missingTypeId));                              // null, but a removed stub stays cached
tx.getEdges(new RelationIdentifier(outId, missingTypeId, 1L, inId)); // IllegalStateException
```

A property key id fails the same way through `findProperty`. Master returns nothing here because `tx.getVertex` filters removed vertices. Treating a removed stub as "no such type" fixes both:

```java
typeVertex = stx.getExistingRelationType(typeId);
if (typeVertex == null || typeVertex.isRemoved()) return null;
```

Your `testEdgeByIdDoesNotReadTypeVertex` already ends with the setup: one more `assertFalse(tx.getEdges(missingType).iterator().hasNext())` after its `assertNull(tx.getVertex(missingTypeId))` would pin this.

porunov (commented):
Thanks for the quick rework, @xiaoying-code, this is a big improvement. I re-ran everything from my previous review against `cfce3b9f0` on in-memory and BerkeleyJE, and all four points are resolved:

- **No more poisoning.** A lookup that races the label's creation, and a foreign edge id looked up before its label exists (primary and fallback mode), now leave the label intact: `multiplicity()` is right, reads find the edge and `addEdge` works.
- **Junk ids no longer grow the cache:** 1,000 junk-type lookups add no entries (the first revision added 1,000).
- **Malformed ids** behave as on master (the id of a removed vertex as the type id gives no edge).
- **Auto-created constraints:** three transactions create one connection constraint, in primary and fallback mode.
- The tests that should fail on master do: 2 of 5 in `StandardSchemaCacheTest`, `testEdgeByIdDoesNotReadTypeVertex` (2 slices) and `testAutoCreatedConnectionConstraintIsSeenByLaterTransactions` (9 constraints on master). The full `BerkeleyGraphTest` (218), `InMemoryGraphTest` (213) and `BerkeleyOperationCountingTest` (11) pass here, and so does checkstyle.

One new problem remains in the branch for types the transaction already holds, see the inline comment. The fix is one line; with it, my review tests and your new tests all pass.

Two small notes, neither blocking:
- The `isCacheable` javadoc says every existing schema vertex has a name. Type-modifier vertices have none. That's harmless, since nothing reads a modifier's name, but the sentence could say "a category and definition properties".
- The new eviction clears the graph-level cache only. A transaction that is open during the commit keeps the schema vertices it already loaded, so it can still add one more copy of a constraint (`ManagementSystem` also calls `tx.expireSchemaElement` on open transactions). That's bounded and fine for this PR, as is the cross-instance follow-up you mention.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
xiaoying-code: Fixed, thanks! Applied the `isRemoved()` check as suggested and added the extra `getEdges` lookup after the `assertNull` to `testEdgeByIdDoesNotReadTypeVertex`.

xiaoying-code:
> Thanks for the quick rework, @xiaoying-code, this is a big improvement. I re-ran everything from my previous review against `cfce3b9f0` on in-memory and BerkeleyJE, and all four points are resolved:
> 
>     * **No more poisoning.** A lookup that races the label's creation, and a foreign edge id looked up before its label exists (primary and fallback mode), now leave the label intact: `multiplicity()` is right, reads find the edge and `addEdge` works.
> 
>     * **Junk ids no longer grow the cache:** 1,000 junk-type lookups add no entries (the first revision added 1,000).
> 
>     * **Malformed ids** behave as on master (the id of a removed vertex as the type id gives no edge).
> 
>     * **Auto-created constraints:** three transactions create one connection constraint, in primary and fallback mode.
> 
>     * The tests that should fail on master do: 2 of 5 in `StandardSchemaCacheTest`, `testEdgeByIdDoesNotReadTypeVertex` (2 slices) and `testAutoCreatedConnectionConstraintIsSeenByLaterTransactions` (9 constraints on master). The full `BerkeleyGraphTest` (218), `InMemoryGraphTest` (213) and `BerkeleyOperationCountingTest` (11) pass here, and so does checkstyle.
> 
> 
> One new problem remains in the branch for types the transaction already holds, see the inline comment. The fix is one line; with it, my review tests and your new tests all pass.
> 
> Two small notes, neither blocking:
> 
>     * The `isCacheable` javadoc says every existing schema vertex has a name. Type-modifier vertices have none. That's harmless, since nothing reads a modifier's name, but the sentence could say "a category and definition properties".
> 
>     * The new eviction clears the graph-level cache only. A transaction that is open during the commit keeps the schema vertices it already loaded, so it can still add one more copy of a constraint (`ManagementSystem` also calls `tx.expireSchemaElement` on open transactions). That's bounded and fine for this PR, as is the cross-instance follow-up you mention.

Thanks @porunov for re-running everything. Pushed `126d1309d` with the `isRemoved()` check and the extra lookup in `testEdgeByIdDoesNotReadTypeVertex`. Javadoc reworded to "a category and definition properties". Left the open-transaction expiry out, per your note.

Locally on Java 17: `StandardSchemaCacheTest`, full `BerkeleyOperationCountingTest` and `InMemoryGraphTest` pass, checkstyle clean. CI workflows are waiting for approval again.

porunov:
Thanks, @xiaoying-code. I verified `126d1309d`. The `isRemoved()` check fixes the same-transaction case for both edges and properties, and your extra lookup pins it without a read in either metrics group. I merged the branch into current `master` (it merges cleanly with #4973's changes to `StandardJanusGraph`) and ran the full `BerkeleyGraphTest` (218), `InMemoryGraphTest` (213) and `BerkeleyOperationCountingTest` (11), `StandardSchemaCacheTest`, my review scenarios from both rounds and checkstyle. Everything passes.

Copilot's latest review raised three points. I checked each against the code and against `master`. One is a regression this PR introduces and should be fixed here; the other two are real but already exist on `master`, so they can be follow-up PRs.

**Fix in this PR: out-of-range type ids now throw**

`isRelationTypeId` only checks an id's low type bits. A malformed `RelationIdentifier` whose type id has those bits but lies far beyond the schema id range therefore reaches `getSchemaRelations`, and its range check (`schemaId <= Long.MAX_VALUE >>> 1`) throws. On `master` the same lookup goes through `tx.getVertex` and finds no edge. Repro, in-memory with the default configuration:

```java
long labelId = eid.getTypeId();                               // an existing edge label
long huge = ((Long.MAX_VALUE >>> 7) << 7) | (labelId & 127);  // same type bits, out of range
tx.getEdges(new RelationIdentifier(eid.getOutVertexId(), huge, eid.getRelationId(), eid.getInVertexId()));
// master: no edge; this PR: IllegalArgumentException
```

Sending ids outside the schema id range down the original path fixes it. Every valid type id passes this check:

```java
if (typeId <= 0 || !stx.getIdInspector().isRelationTypeId(typeId)
    || (typeId >>> IDManager.MAX_PADDING_BITWIDTH) >= IDManager.getSchemaCountBound()) {
```

I tried it on `126d1309d`: the lookup finds no edge again, and your tests and my review scenarios still pass. I'd avoid `IDManager.stripEntireRelationTypePadding` for this, since its own precondition can throw on other malformed ids. A regression test next to the negative-id case in `testEdgeByIdLookupBeforeLabelIsVisibleDoesNotPoisonSchemaCache` would pin it.

**Follow-up PR: other open transactions keep their memoized schema vertices**

Real, but not introduced here. A transaction that loaded a label's connections before another transaction commits a constraint keeps its memoized list and can add one more copy of that constraint. The per-transaction memo in `JanusGraphSchemaVertex` behaves the same way on `master`, in both cache modes, and this PR doesn't make it worse. Expiring these ids in every open transaction, as `ManagementSystem.commit` does, fits a follow-up together with the cross-instance eviction you mentioned.

**Follow-up PR: a lookup can republish a stale result after an eviction**

Also real, but it's the cache-aside race `StandardSchemaCache` already has on `master`. A lookup reads storage before a schema commit, the commit's `expireSchemaElement` runs, and the lookup then stores the old result, which stays until the next eviction. On `master` that already applies to every entry of the primary map and to non-empty entries of the fallback cache. This PR extends it to empty definition-edge results in fallback mode, and the window is about one storage read. The proper fix, such as a per-schema generation counter checked before publishing, or locking, covers both maps. It deserves its own issue and PR rather than being folded into this one.

On CI: the one failure so far, `CQLSolrTest.testRecurringIndexReplay` (`expected: <4> but was: <1>`), fails the same way on `master`'s own latest Solr run and on earlier `master` commits, so it isn't caused by this PR.

Once the range fix is in, this looks good to me.

janusgraph-core/src/main/java/org/janusgraph/graphdb/relations/RelationIdentifierUtils.java
```diff
@@ -44,13 +46,33 @@ protected static JanusGraphRelation findRelation(RelationIdentifier rId, JanusGr
...
 
-        JanusGraphVertex typeVertex = tx.getVertex(rId.getTypeId());
-        if (typeVertex == null) return null;
-
-        if (!(typeVertex instanceof RelationType))
-            throw new IllegalArgumentException("Invalid RelationIdentifier: typeID does not reference a type");
+        StandardJanusGraphTx stx = (StandardJanusGraphTx) tx;
+        long typeId = rId.getTypeId();
+        RelationType typeVertex;
+        if (typeId <= 0 || !stx.getIdInspector().isRelationTypeId(typeId)) {
```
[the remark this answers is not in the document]
porunov: Confirmed. On `master` this identifier resolves to no edge; here `getSchemaRelations` throws `IllegalArgumentException` from its range check. It's a regression, so I've asked for it to be fixed in this PR with a range check that sends such ids down the original path ([details](https://github.com/JanusGraph/janusgraph/pull/4953#issuecomment-5816604567)).

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/StandardJanusGraph.java
```diff
@@ -1240,7 +1243,30 @@ public void commit(final Collection<InternalRelation> addedRelations,
                 log.error("Could not roll-back transaction ["+transactionId+"] after failure due to exception",e2);
             }
             throw e;
+        } finally {
+            //4. Definition edges written outside of ManagementSystem (e.g. constraints auto-created under
+            //   schema.constraints=true) are not broadcast as cache evictions. Expire the local schema cache
+            //   entries of the schema vertices they touch so this instance re-reads them.
+            for (Long schemaId : schemaElementsWithChangedDefinitionEdges) {
+                schemaCache.expireSchemaElement(schemaId);
+            }
```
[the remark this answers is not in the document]
porunov: Valid, but not introduced by this PR. The per-transaction memo in `JanusGraphSchemaVertex` behaves the same on `master` in both cache modes, and the effect is one extra copy of a constraint per transaction that was open at the time. Deferred to a follow-up PR together with cross-instance eviction ([details](https://github.com/JanusGraph/janusgraph/pull/4953#issuecomment-5816604567)).

janusgraph-core/src/main/java/org/janusgraph/graphdb/database/cache/StandardSchemaCache.java
```diff
@@ -151,7 +164,7 @@ public EntryList getSchemaRelations(final long schemaId, final BaseRelationType
             entries = schemaRelationsBackup.getIfPresent(typePlusRelation);
             if (entries==null) {
                 entries = retriever.retrieveSchemaRelations(schemaId, type, dir);
-                if (!entries.isEmpty()) { //only cache if type exists
+                if (isCacheable(type, entries)) {
                     schemaRelationsBackup.put(typePlusRelation, entries);
```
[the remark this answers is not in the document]
porunov: Valid, but it's the cache-aside race `StandardSchemaCache` already has on `master`, for every primary-map entry and for non-empty fallback entries. This PR only extends it to empty definition-edge results in fallback mode. A proper fix (a generation counter or locking) covers both maps, so it belongs in its own issue and PR ([details](https://github.com/JanusGraph/janusgraph/pull/4953#issuecomment-5816604567)).

xiaoying-code:
Thanks @porunov. Pushed `ac1624b7c` with the range check as suggested and a regression test next to the negative-id case in `testEdgeByIdLookupBeforeLabelIsVisibleDoesNotPoisonSchemaCache`. Agreed on treating the other two as follow-ups.

porunov:
Thanks, @xiaoying-code, verified `ac1624b7c`. The range check sends out-of-range type ids down the original path, so `getEdges` finds no edge again instead of throwing, and the new assertion in `testEdgeByIdLookupBeforeLabelIsVisibleDoesNotPoisonSchemaCache` pins it. On the branch merged into current `master` I ran the full `BerkeleyGraphTest` (218), `InMemoryGraphTest` (213) and `BerkeleyOperationCountingTest` (11), `StandardSchemaCacheTest`, my review scenarios (including the out-of-range repro) and checkstyle: all pass.

On the previous head's CI everything passed except two jobs whose failures aren't caused by this PR. `testRecurringIndexReplay` in the Solr job fails the same way on `master`. In `tests (berkeleyje, 25)`, one interrupted thread invalidated the shared BerkeleyJE environment (`THREAD_INTERRUPTED`), and the rest of `BerkeleyGraphTest` then couldn't open it; other PRs have hit the same failure.

This looks good to me.

porunov (approved):
LGTM. **High** Copilot's comments are valid, but pre-existing - not introduced in this PR. So, fixing that as followups makes sense I believe.

